PR #4, opened 2022-06-08
Refactorization

echarlaix:


README.md
```diff
@@ -69,24 +69,28 @@ quantization_config = IncQuantizationConfig.from_pretrained(config_path, config_
 pruning_config = IncPruningConfig.from_pretrained(config_path, config_file_name="prune.yml")
 
 # Instantiate our IncQuantizer using the desired configuration
-inc_quantizer = IncQuantizer(model, quantization_config, eval_func=eval_func)
-quantizer = inc_quantizer.fit()
+quantizer = IncQuantizer(quantization_config, eval_func=eval_func)
```
hshen14, line 72: What's the position between optimum-intel and optimum? Will be optimum-intel be merged back to optimum? As we discussed, we are looking for some ways to deliver the best user experience, so transformer.training extension will be the best fit here; otherwise, uses have to introduce IncTrainer, and modify the original Trainer usage. This may not be quite related with this PR, so we may discuss more in a separate thread.

echarlaix: It's not planned for optimum-intel to be merged in optimum, nothing will change concerning the behavior of the intel package (people will still be able to install it via pip install optimum[intel] and use it as previously). Concerning `Trainer` vs `IncTrainer`, we can have this discussion during our next meeting as it is an important point.

PR #7, opened 2022-06-21
One shot optimization

echarlaix:
In this PR we add the possibility to do one shot optimization, allowing to combine quantization aware training with pruning. We should wait for a release of neural-compressor to have access to updates on the [_pre_hook_for_qat](https://github.com/intel/neural-compressor/blob/d9d51cf56a611b9d1d6e4ce47f0a04fe0f44f005/neural_compressor/adaptor/pytorch.py#L2615) method before merging.

### DIFF
--- a/examples/config/prune.yml
+++ b/examples/config/prune.yml
@@ -17,7 +17,7 @@ version: 1.0
 
 model:
   name: bert_prune
-  framework: pytorch
+  framework: pytorch_fx
 
 pruning:
   approach:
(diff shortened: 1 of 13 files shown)


echarlaix:
Hi @hshen14, currently the combination of pruning and [qat](https://github.com/intel/neural-compressor/blob/aac0a0ec860d6d875467a8b7fb119ec18713fd48/neural_compressor/experimental/component.py#L134) does not work. This comes from the model tracing and is present in [v1.12](https://github.com/intel/neural-compressor/blob/v1.12/neural_compressor/adaptor/pytorch.py#L2625) but should be solved in the next release thanks to [this](https://github.com/intel/neural-compressor/blob/master/neural_compressor/adaptor/pytorch.py#L2640) addition.

PR #9, opened 2022-07-08
Adapt changes inc release 1.13

echarlaix:


tests/test_inc.py
```diff
@@ -274,7 +274,7 @@ def eval_func(model):
             sparsity = optimizer.get_sparsity()
 
             # Verification final sparsity is equal to the targeted sparsity
-            self.assertGreaterEqual(round(sparsity), target_sparsity * 100)
+            self.assertGreaterEqual(round(sparsity), 0.5)
```
echarlaix, line 277: Why should we replace with `target_sparsity * 100` to `0.5`  @PenghuiCheng ? Is it because sparsity is never guaranteed to be the exactly `target_sparsity` ?

optimum/intel/neural_compressor/trainer.py
```diff
@@ -553,17 +553,18 @@ def compute_loss(self, model, inputs, return_outputs=False):
...
-                    inputs, teacher_model=self.agent.teacher_model._model
-                )
-
-            teacher_logits = self._get_logits(teacher_outputs)
+            teacher_logits = inputs.pop("teacher_logits", None)
+            if hasattr(self.agent, "on_after_compute_loss"):
+                if teacher_logits is None:
+                    teacher_outputs = self.agent.criterion.teacher_model_forward(inputs)
+                    teacher_logits = self._get_logits(teacher_outputs)
+            elif hasattr(self.agent, "on_post_forward"):
```
echarlaix, line 564: @PenghuiCheng Currently in [main](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/neural_compressor/trainer.py#L562) we use the `teacher_model_forward` method to compute the teacher outputs, but I agree that it makes totally sense to use `on_post_forward` to stay compatible for `neural_compressor` v1.12 or under

optimum/intel/neural_compressor/trainer.py
```diff
@@ -553,17 +553,18 @@ def compute_loss(self, model, inputs, return_outputs=False):
...
+                # Compute teacher model outputs
+                self.agent.on_post_forward(inputs, teacher_output=teacher_logits)
+                teacher_logits = self._get_logits(self.agent.criterion.teacher_outputs)
```
echarlaix, line 567:
```suggestion
            # Compute teacher model outputs
            if teacher_logits is None:
                if hasattr(self.agent, "on_after_compute_loss"):
                    teacher_outputs = self.agent.criterion.teacher_model_forward(inputs)
                    teacher_logits = self._get_logits(teacher_outputs)
                elif hasattr(self.agent, "on_post_forward"):
                    self.agent.on_post_forward(inputs, teacher_output=teacher_logits)
                    teacher_logits = self._get_logits(self.agent.criterion.teacher_outputs)
                else:
                    raise ValueError("Unable to compute the teacher outputs")
```

optimum/intel/neural_compressor/trainer.py
```diff
@@ -553,17 +553,18 @@ def compute_loss(self, model, inputs, return_outputs=False):
...
-            teacher_outputs = inputs.pop("teacher_outputs", None)
-            # Compute teacher model outputs
-            if teacher_outputs is None:
-                teacher_outputs = self.agent.criterion.teacher_model_forward(
-                    inputs, teacher_model=self.agent.teacher_model._model
-                )
-
-            teacher_logits = self._get_logits(teacher_outputs)
+            teacher_logits = inputs.pop("teacher_logits", None)
+            if hasattr(self.agent, "on_after_compute_loss"):
```
echarlaix, line 560: Is it equivalent to `hasattr(self.agent.criterion, "teacher_model_forward")` ?

tests/test_inc.py
PenghuiCheng: because， the target  sparsity in yaml is for every operator sparsity, not for whole model sparsity, so we can the sparsity of op is all greater than 0.02 in log, but there are many weights, like embedding op, didn't be pruned.  And the function of optimizer.get_sparsity() is get the model sparsity. the model sparsity is 0.84, so I set it to 0.5.
the log is as below:
2022-08-05 17:30:41 [INFO]                                                  Name         Shape NNZ (dense)  NNZ (sparse)  Sparsity(%)   Std      Mean  Abs-M
ean
0   distilbert.embeddings.word_embeddings.module.w...  [30522, 768]    23440896             0         0.00  **_0.05_** -3.83e-02      0.05
1   distilbert.embeddings.position_embeddings.modu...    [512, 768]      393216             0         0.00  **_0.02**_ -4.01e-05      0.01
2   distilbert.transformer.layer.0.attention.q_lin...    [768, 768]      589824             0         0.00  _**0.04**_  5.97e-05      0.03
3   distilbert.transformer.layer.0.attention.k_lin...    [768, 768]      589824             0         0.00  _**0.04**_  9.21e-06      0.03

optimum/intel/neural_compressor/trainer.py
PenghuiCheng: on_after_compute_loss is not equivalent to teacher_model_forward, in future , we will use on_after_compute_loss callback to compute distillation loss. the usage is like below:
student_outputs = student_model(input)
student_loss = user_criterion(student_outputs, labels)
total_loss = agent.on_after_compute_loss(input, student_outputs, student_loss, tearch_outputs)
here, teacher_outputs is optional.
The callback of on_after_compute_loss is a new API in 1.13. The purpose of this is to reuse the user's criterion.
But in on_after_compute_loss, it didn't handle a tuple of student_outputs and teacher_outputs. so we didn't use the on_after_compute_loss directly, in the future I will commit another commit to use on_after_compute_loss directly.

tests/test_inc.py
echarlaix: Thanks for the clarification

optimum/intel/neural_compressor/trainer.py
PenghuiCheng: The on_after_compute_loss API usage is as above, so here is only a transitional change. in the future, we only use total_loss == agent.on_after_compute_loss(None, student_outputs, student_loss, teacher_outputs) directly.

optimum/intel/neural_compressor/trainer.py
echarlaix: My question was more about does `hasattr(self.agent, "on_after_compute_loss")` is equivalent to `hasattr(self.agent.criterion, "teacher_model_forward")` as `self.agent.criterion.teacher_model_forward` is called after this condition. Also thanks for your explanation, does it mean that the `teacher_model_forward` method will be deprecated ? I find this method very useful as it gives us more flexibility : and compute the loss ourselves with the trainer `compute_distillation_loss` method

PenghuiCheng (approved):

optimum/intel/neural_compressor/trainer.py
PenghuiCheng: The teacher_model_forward will exist in the old version and the new version in neural_compressor,  we can use it anyway. Yes, if you want to use the trainer compute_distillation_loss method, you can use hasattr(self.agent.criterion, "teacher_model_forward") condition, but it is only in the new version that will return outputs. So you need to judge that is a new version neural_compressor first.

optimum/intel/neural_compressor/trainer.py
PenghuiCheng: If teacher_logits is None, agent.on_post_forward will compute the tearcher_outputs, but if teacher_logits is not None, agent.on_post_forward didn't recompute teacher_outputs, just only assign to self.agent.criterion.teacher_outputs

PenghuiCheng: Oh, yes, I got your point now, your suggestion is OK.

PR #11, opened 2022-07-13
Remove 'INC' since 'INC' is not an Intel formal approved trademark

PenghuiCheng:
# What does this PR do?
This PR will remove "INC" from documents, since "INC" is not an intel formal approved trademark.


hshen14 (approved):
LGTM

echarlaix (approved):

PR #19, opened 2022-07-28
Add test

echarlaix:


xin3he:
Hi Ella, agent.pre_epoch_begin was called twice so I removed one to avoid loading failures.

echarlaix:
Good catch, thanks a lot @xin3he

PR #20, opened 2022-08-02
fix variable name

jiqing-feng:
The variable name of distillation should be distiller, otherwise the distiller parameter in optimizer would be a none parameter.

@echarlaix


echarlaix:
Good catch @jiqing-feng, thanks !

PR #24, opened 2022-08-05
fix the crash in distillation if loss_weights set to be [1, 0] in yam…

sywangyi:
…l config

# What does this PR do?

Fixes # (issue)
fix the crash in distillation if loss_weights set to be [1, 0] in yaml config

## Before submitting

### DIFF
--- a/optimum/intel/neural_compressor/trainer.py
+++ b/optimum/intel/neural_compressor/trainer.py
@@ -562,12 +562,12 @@ class IncTrainer(Trainer):
                 teacher_outputs = self.agent.criterion.teacher_model_forward(
                     inputs, teacher_model=self.agent.teacher_model._model
                 )
-
-            teacher_logits = self._get_logits(teacher_outputs)
-            distillation_loss = self.compute_distillation_loss(logits, teacher_logits)
-            loss *= self.agent.criterion.loss_weights[0]
-            loss += distillation_loss * self.agent.criterion.loss_weights[1]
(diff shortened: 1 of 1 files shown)


sywangyi:
@jiqing-feng @echarlaix please notice this issue

echarlaix:
Good catch @sywangyi, thanks for fixing this

optimum/intel/neural_compressor/trainer.py
```diff
@@ -562,12 +562,12 @@ def compute_loss(self, model, inputs, return_outputs=False):
                 teacher_outputs = self.agent.criterion.teacher_model_forward(
                     inputs, teacher_model=self.agent.teacher_model._model
                 )
-
-            teacher_logits = self._get_logits(teacher_outputs)
-            distillation_loss = self.compute_distillation_loss(logits, teacher_logits)
-            loss *= self.agent.criterion.loss_weights[0]
-            loss += distillation_loss * self.agent.criterion.loss_weights[1]
-            loss /= self.agent.criterion.loss_weights[0] + self.agent.criterion.loss_weights[1]
+            if teacher_outputs is not None:
```
echarlaix, line 565: Why not adding additional constraint before applying distillation instead such as ` self.agent.criterion.loss_weights[1] > 0`.

echarlaix: Also `self.agent.criterion.loss_weights[1] + self.agent.criterion.loss_weights[0] > 0`  should be checked

PR #26, opened 2022-08-08
fix the crash when the KL loss weight is set to 0

sywangyi:
loss is incorrectly calculated if there's teacher logit output

# What does this PR do?

Fixes # (issue)
fix the crash when the KL loss weight is set to 0
loss is incorrectly calculated if there's teacher logit output

## Before submitting


sywangyi:
@echarlaix @jiqing-feng please notice the issue

PR #27, opened 2022-08-10
fix the import error of cached_path, hf_bucket_url

sywangyi:
from transformers.file_utils import cached_path, hf_bucket_url

to align with the following PR merged in transformer
https://github.com/huggingface/transformers/pull/18497

# What does this PR do?

Fixes # (issue)
 fix the import error of cached_path, hf_bucket_url
 to align with the following PR merged in transformer
 https://github.com/huggingface/transformers/pull/18497

## Before submitting

### DIFF
--- a/optimum/intel/neural_compressor/configuration.py
+++ b/optimum/intel/neural_compressor/configuration.py
@@ -15,11 +15,13 @@
 import logging
 import os
 from functools import reduce
+from pathlib import Path
 from typing import Any, Optional, Union
 
-from transformers.file_utils import cached_path, hf_bucket_url
+from transformers.utils import TRANSFORMERS_CACHE, is_offline_mode
 
 import yaml
+from huggingface_hub import hf_hub_download
 from neural_compressor.conf.config import Conf, Distillation_Conf, Pruning_Conf, Quantization_Conf
 from optimum.intel.neural_compressor.utils import CONFIG_NAME
 
@@ -97,37 +99,46 @@ class IncConfig:
         revision = kwargs.get("revision", None)
 
         config_file_name = config_file_name if config_file_name is not None else CONFIG_NAME
+
         if os.path.isdir(config_name_or_path):
             config_file = os.path.join(config_name_or_path, config_file_name)
         elif os.path.isfile(config_name_or_path):
             config_file = config_name_or_path
         else:
-            config_file = hf_bucket_url(config_name_or_path, filename=config_file_name, revision=revision)
-
-        try:
-            resolved_config_file = cached_path(
-                config_file,
-                cache_dir=cache_dir,
-                force_download=force_download,
-                resume_download=resume_download,
-            )
-        except EnvironmentError as err:
-            logger.error(err)
-            msg = (
-                f"Can't load config for '{config_name_or_path}'. Make sure that:\n\n"
-                f"-'{config_name_or_path}' is a correct model identifier listed on 'https://huggingface.co/models'\n\n"
-                f"-or '{config_name_or_path}' is a correct path to a directory containing a {config_file_name} file\n\n"
-            )
-
-            if revision is not None:
-                msg += (
-                    f"- or '{revision}' is a valid git identifier (branch name, a tag name, or a commit id) that "
-                    f"exists for this model name as listed on its model page on 'https://huggingface.co/models'\n\n"
+            local_files_only = False
+            if is_offline_mode():
+                logger.info("Offline mode: forcing local_files_only=True")
+                local_files_only = True
+            if cache_dir is None:
+                cache_dir = TRANSFORMERS_CACHE
+            if isinstance(cache_dir, Path):
+                cache_dir = str(cache_dir)
+            try:
+                config_file = hf_hub_download(
+                    repo_id=config_name_or_path,
+                    filename=config_file_name,
+                    revision=revision,
+                    cache_dir=cache_dir,
+                    force_download=force_download,
+                    resume_download=resume_download,
+                    local_files_only=local_files_only,
+                )
+            except EnvironmentError as err:
+                logger.error(err)
+                msg = (
+                    f"Can't load config for '{config_name_or_path}'. Make sure that:\n\n"
+                    f"-'{config_name_or_path}' is a correct model identifier listed on 'https://huggingface.co/models'\n\n"
+                    f"-or '{config_name_or_path}' is a correct path to a directory containing a {config_file_name} file\n\n"
                 )
 
-            raise EnvironmentError(msg)
+                if revision is not None:
+                    msg += (
+                        f"- or '{revision}' is a valid git identifier (branch name, a tag name, or a commit id) that "
+                        f"exists for this model name as listed on its model page on 'https://huggingface.co/models'\n\n"
+                    )
 
-        config = cls(resolved_config_file)
+                raise EnvironmentError(msg)
+        config = cls(config_file)
 
         return config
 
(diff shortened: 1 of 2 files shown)


sywangyi:
@jiqing-feng @echarlaix @PenghuiCheng  please aware of the issue

optimum/intel/neural_compressor/configuration.py
```diff
@@ -17,7 +17,7 @@
 from functools import reduce
 from typing import Any, Optional, Union
 
-from transformers.file_utils import cached_path, hf_bucket_url
+from transformers.utils import cached_file
```
PenghuiCheng, line 20: We should be compatible with the old version, is it OK?

PenghuiCheng (approved):

optimum/intel/neural_compressor/configuration.py
```diff
@@ -17,7 +17,15 @@
...
-from transformers.file_utils import cached_path, hf_bucket_url
+
+try:
+    from transformers.utils import cached_file
+
+    _use_cached_file = True
+except ImportError:
+    from transformers.file_utils import cached_path, hf_bucket_url
+
+    _use_cached_file = False
```
echarlaix, line 28:
```suggestion
from huggingface_hub import hf_hub_download
```

optimum/intel/neural_compressor/configuration.py
```diff
@@ -97,20 +105,29 @@ def from_pretrained(cls, config_name_or_path: str, config_file_name: Optional[st
...
-        else:
-            config_file = hf_bucket_url(config_name_or_path, filename=config_file_name, revision=revision)
-
+        if _use_cached_file == False:
+            if os.path.isdir(config_name_or_path):
+                config_file = os.path.join(config_name_or_path, config_file_name)
+            elif os.path.isfile(config_name_or_path):
+                config_file = config_name_or_path
+            else:
+                config_file = hf_bucket_url(config_name_or_path, filename=config_file_name, revision=revision)
```
echarlaix, line 114: I suggest using `hf_hub_download` as follows :
```
config_file_name = config_file_name if config_file_name is not None else CONFIG_NAME
if os.path.isdir(config_name_or_path):
    config_file = os.path.join(config_name_or_path, config_file_name)
elif os.path.isfile(config_name_or_path):
    config_file = config_name_or_path
else:
    try:
        config_file = hf_hub_download(
            repo_id=config_name_or_path,
            filename=config_file_name,
            revision=revision,
            cache_dir=cache_dir,
        )
    except EnvironmentError as err:
        logger.error(err)
        msg = (
            f"Can't load config for '{config_name_or_path}'. Make sure that:\n\n"
            f"-'{config_name_or_path}' is a correct model identifier listed on 'https://huggingface.co/models'\n\n"
            f"-or '{config_name_or_path}' is a correct path to a directory containing a {config_file_name} file\n\n"
        )

        if revision is not None:
            msg += (
                f"- or '{revision}' is a valid git identifier (branch name, a tag name, or a commit id) that "
                f"exists for this model name as listed on its model page on 'https://huggingface.co/models'\n\n"
            )

        raise EnvironmentError(msg)

config = cls(config_file)
```

optimum/intel/neural_compressor/quantization.py
```diff
@@ -271,15 +281,19 @@ def from_pretrained(
...
-                state_dict_path = model_name_or_path
-            else:
-                state_dict_path = hf_bucket_url(model_name_or_path, filename=q_model_name, revision=revision)
+            if _use_cached_file == False:
+                if os.path.isdir(model_name_or_path):
+                    state_dict_path = os.path.join(model_name_or_path, q_model_name)
+                elif os.path.isfile(model_name_or_path):
+                    state_dict_path = model_name_or_path
+                else:
+                    state_dict_path = hf_bucket_url(model_name_or_path, filename=q_model_name, revision=revision)
```
echarlaix, line 290: I suggest using `hf_hub_download` as follows :
```
  if os.path.isdir(model_name_or_path):
      state_dict_path = os.path.join(model_name_or_path, q_model_name)
  elif os.path.isfile(model_name_or_path):
      state_dict_path = model_name_or_path
  else:
      try:
          state_dict_path = hf_hub_download(model_name_or_path, filename=q_model_name, revision=revision)
      except EnvironmentError as err:
          logger.error(err)
          msg = (
              f"Can't load config for '{model_name_or_path}'. Make sure that:\n\n - '{model_name_or_path}' is a "
              f"correct model identifier listed on 'https://huggingface.co/models'\n\n - or "
              f"'{model_name_or_path}' is a correct path to a directory containing a {q_model_name} file\n\n"
          )

          if revision is not None:
              msg += (
                  f"- or '{revision}' is a valid git identifier (branch name, a tag name, or a commit id)  "
                  f"thatexists for this model name as listed on its model page on "
                  f"'https://huggingface.co/models'\n\n"
              )

          raise EnvironmentError(msg)
```

echarlaix (commented):
Thanks for the fix @sywangyi !

optimum/intel/neural_compressor/quantization.py
sywangyi: hi， @echarlaix ， I see  cached_file called hf_hub_download as well. meanwhile, in cached_file, there's a judgement of offline mode . in some case, we need to use offline mode.   if we use hf_hub_download, should we move the judgement of offline mode here?

echarlaix: I agree with you, that would be great

sywangyi: so, my suggestion is to keep it as what this PR did, since both old cached_path and new cached_file API contain the logic of how to dispose offline mode.

echarlaix: My suggestion would be to add the `local_files_only` variable instead. Currently in the PR, the behavior is not strictly the same depending on the version of transformers (`config_name_or_path` should not be the path of the file but only the directory containing it when using `cached_file`) also the addition of exception to handle the different imports might add additional difficulties in the future to maintain or modify the code. What do you think ?

sywangyi: Hi, echarlaix, that means we need to port nearly all the logic in function "cached_file" here, including offline mode check, fetch local cache logic, maybe some function need to be added in __init__.py in transformer/utils like try_to_load_from_cache

echarlaix: In our case can't we just add the offline mode detection (wtih the `local_files_only` variable) ?

sywangyi: Hi, Echarlaix, there are three cases here
1) the config file is in local. like current distillation.cfg in example . need to add logic if it's local file here.
2) the config is in huggingface.co/models, not downloaded yet. need to use hf_hub_download to download to cache
3) there's a download cache in local. the config is there. no need to use hf_hub_download if it's offline mode, just try_to_load_from_cache. the function is very useful when we test in no network environment.

I thinks we need to support all the scenario here as transformers.

echarlaix: Hi @sywangyi, for me all those 3 cases would be supported, can you go into more details about which one would not be handled correctly ? Also when loading a config from the cache `hf_hub_download` will be used (and not `try_to_load_from_cache`) as `_commit_hash` will be `None`

sywangyi: hi, echarlaix, if __commit_hash will be None in the case, nothing concern from me. just update the patch according to your suggestion and add cache, offline support

echarlaix (approved):
Thanks for the fix @sywangyi ! 🔥

PR #28, opened 2022-08-18
support reloading for int8 models

xin3he:
# What does this PR do?

In this PR we enhance the example to support reloading for int8 models. users and reload quantized model from `output_dir` or `model_name` on [model hub](https://huggingface.co/models?other=Intel%C2%AE%20Neural%20Compressor).

## Before submitting

### DIFF
--- a/examples/neural_compressor/test_examples.py
+++ b/examples/neural_compressor/test_examples.py
@@ -86,6 +86,24 @@ class TestExamples(unittest.TestCase):
                 results = get_results(tmp_dir)
                 self.assertGreaterEqual(results["eval_accuracy"], 0.70)
 
+            test_args = f"""
+                run_glue.py
+                --model_name_or_path {tmp_dir}
+                --task_name sst2
+                --metric eval_accuracy
+                --do_eval
+                --per_device_eval_batch_size 1
+                --max_eval_samples 50
+                --load_int8 \
+                --verify_loading
+                --output_dir {tmp_dir}
+                """.split()
+
+            with patch.object(sys, "argv", test_args):
+                run_glue.main()
+                results = get_results(tmp_dir)
+                self.assertGreaterEqual(results["eval_accuracy"], 0.70)
+
     def test_run_qa(self):
         quantization_approach = "static"
 
--- a/examples/neural_compressor/text-classification/README.md
+++ b/examples/neural_compressor/text-classification/README.md
@@ -36,11 +36,24 @@ python run_glue.py \
     --verify_loading \
     --output_dir /tmp/sst2_output
 ```
-The following example reloads the quantized DistilBERT and verifies the accuracy on sst-2 task. [Int8 models](https://huggingface.co/models?other=Intel%C2%AE%20Neural%20Compressor) on model hub are also available.
+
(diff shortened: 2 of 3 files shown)


PenghuiCheng (approved):

examples/text-classification/run_glue.py
```diff
@@ -571,162 +587,169 @@ def take_train_steps(model, trainer, resume_from_checkpoint, last_checkpoint):
...
+        if optim_args.load_int8:
+            loaded_model = model
+        else:
+            loaded_model = IncQuantizedModelForSequenceClassification.from_pretrained(training_args.output_dir)
         loaded_model.eval()
+
         result_loaded_model = take_eval_steps(loaded_model, trainer, metric_name)
 
-        if result_loaded_model != result_optimized_model:
+        if not optim_args.load_int8 and result_loaded_model != result_optimized_model:
```
echarlaix, line 752: Not sure about the case where `optim_args.load_int8` is set to `True` as we don't have the original model's accuracy to compare it with

echarlaix (commented):
Great addition @xin3he, thanks !

examples/text-classification/README.md
```diff
@@ -36,6 +36,17 @@ python run_glue.py \
     --verify_loading \
     --output_dir /tmp/sst2_output
 ```
+The following example reloads the quantized DistilBERT and verifies the accuracy on sst-2 task. [Int8 models](https://huggingface.co/models?other=Intel%C2%AE%20Neural%20Compressor) on model hub are also available.
+
+```bash
+python run_glue.py \
+    --model_name_or_path /tmp/sst2_output \
```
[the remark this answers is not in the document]
xin3he: Yes, but my concern is keeping the series with the tuning command. I will add one more intro. about loading int8 models from the [Intel](https://huggingface.co/Intel) org

examples/text-classification/run_glue.py
```diff
@@ -571,162 +587,169 @@ def take_train_steps(model, trainer, resume_from_checkpoint, last_checkpoint):
     def train_func(model):
         return take_train_steps(model, trainer, resume_from_checkpoint, last_checkpoint)
 
-    quantizer = None
-    pruner = None
-    distiller = None
+    if optim_args.apply_quantization or optim_args.apply_pruning or optim_args.apply_distillation:
 
-    if not optim_args.apply_quantization and not optim_args.apply_pruning and not optim_args.apply_distillation:
```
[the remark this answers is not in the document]
xin3he: Good suggestion, thanks

examples/text-classification/run_glue.py
xin3he: Yes, I also have this concern. One possible method is checking the module class whether there is 'quantize' string in its name.

PenghuiCheng (approved):

examples/text-classification/run_glue.py
echarlaix: We could create a separate function verifying the model's quantization by comparing what layer is being quantized with the configuration describing the quantization process, but it might be a little bit heavy

hshen14:
@xin3he  Please rebase the code. We will need to support reloading INT8 models.

PenghuiCheng (approved):

PR #40, opened 2022-09-12
Fix documentation

echarlaix:
Closes https://github.com/huggingface/optimum/issues/379


lewtun (approved):
Thanks for the quick fix @echarlaix ! I checked the build and everything LGTM

PR #41, opened 2022-09-12
Add OVModel static shapes

echarlaix:
In this PR we add the static shapes option, we also enable model caching and create the inference request during the first inference (to avoid unuseful repetitive model compilation)

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -36,7 +36,16 @@ core = Core()
 
 logger = logging.getLogger(__name__)
 
-_SUPPORTED_DEVICES = {"CPU", "GPU"}
+_SUPPORTED_DEVICES = {
+    "CPU",
+    "GPU",
+    "AUTO",
+    "AUTO:CPU,GPU",
+    "AUTO:GPU,CPU",
+    "MULTI",
+    "MULTI:CPU,GPU",
+    "MULTI:GPU,CPU",
+}
 
 
 @add_start_docstrings(
@@ -54,7 +63,7 @@ class OVBaseModel(OptimizedModel):
         self._device = kwargs.get("device", "CPU")
         self.is_dynamic = kwargs.get("dynamic_shapes", True)
         self.ov_config = {"PERFORMANCE_HINT": "LATENCY"}
-        if self._device == "GPU" and self.is_dynamic:
+        if "GPU" in self._device and self.is_dynamic:
             raise ValueError(
                 "Support of dynamic shapes for GPU devices is not yet available. Set `dynamic_shapes` to `False` to continue."
             )
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -68,7 +68,9 @@ class OVBaseModel(OptimizedModel):
                 "Support of dynamic shapes for GPU devices is not yet available. Set `dynamic_shapes` to `False` to continue."
             )
         if self.is_dynamic:
-            model = self._reshape(model, -1, -1)
+            height = -1 if self.export_feature == "image-classification" else None
+            width = -1 if self.export_feature == "image-classification" else None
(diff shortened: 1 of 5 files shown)


optimum/intel/openvino/modeling_base.py
```diff
@@ -229,6 +236,20 @@ def _create_infer_request(self, model):
...
+            shapes[inputs][1] = sequence_length
+        model.reshape(shapes)
+        return model
+
+    def reshape(self, batch_size: int, sequence_length: int):
+        self.is_dynamic = True if batch_size == -1 and sequence_length == -1 else False
+        self.model = self._reshape(self.model, batch_size, sequence_length)
+        self.request = self._create_infer_request(self.model)
+
     def _ensure_supported_device(self, device: str = None):
```
helena-intel, line 253: Now that both CPU and GPU are supported, we should also support AUTO, MULTI and HETERO. See https://docs.openvino.ai/latest/openvino_docs_OV_UG_Working_with_devices.html 
Device names can then be for example MULTI:CPU,GPU, or just AUTO (which will automatically choose the best device), or AUTO:CPU,GPU (for example to specify that AUTO is allowed to choose from CPU and GPU, but not VPU even if it is available on the system) or AUTO:GPU,CPU, etc.

optimum/intel/openvino/modeling_base.py
```diff
@@ -229,6 +236,20 @@ def _create_infer_request(self, model):
...
+    def _reshape(self, model: openvino.runtime.Model, batch_size: int, sequence_length: int):
+        shapes = {}
+        for inputs in model.inputs:
+            shapes[inputs] = inputs.get_partial_shape()
+            shapes[inputs][0] = batch_size
+            shapes[inputs][1] = sequence_length
+        model.reshape(shapes)
+        return model
+
+    def reshape(self, batch_size: int, sequence_length: int):
```
AlexKoff88, line 248: The current implementation works the way that the user needs to reshape the model every time exactly the shape required for the input. So that if I reshape the model to [1,32] and pass the input with [1,17], I got an error. My assumption was that it should be possible to define the static shape (max sequence length) and if the user passes the input which is less than the shape it will be padded and the output will be truncated.

AlexKoff88: I also tried a w/a proposed by @helena-intel ```tokenizer = AutoTokenizer.from_pretrained(model_checkpoint, max_length=32, padding="max_length")``` but it does not work.

echarlaix: Can you try :
```python
inputs = tokenizer(text, return_tensors="pt", max_length=32, padding="max_length", truncation=True)
```

optimum/intel/openvino/modeling_base.py
echarlaix: Good point, I will add these to the list of the supported devices

optimum/intel/openvino/modeling_base_seq2seq.py
```diff
@@ -309,5 +319,41 @@ def _from_transformers(
...
 
+    def _reshape(self, model: openvino.runtime.Model, batch_size: int, sequence_length: int, is_decoder=True):
+        shapes = {}
+        for inputs in model.inputs:
+            shapes[inputs] = inputs.get_partial_shape()
+            shapes[inputs][0] = batch_size if not is_decoder else -1
+            if inputs.get_any_name().startswith("past_key_values"):
+                shapes[inputs][2] = -1
+            elif is_decoder and not inputs.get_any_name().startswith("encoder"):
+                shapes[inputs][1] = -1
```
AlexKoff88, line 330: I think we should print a warning message here: "WARNING: Decoder part of the model does not support static shapes and kept dynamic."

echarlaix: I have added it !

AlexKoff88 (approved):

PR #49, opened 2022-09-23
Add option to use cache fo seq2seq models

echarlaix:
In this PR we enable the `use_cache` option for `OVSeq2SeqModel`'s where an extra decoder with pre-computed key/values as one of its inputs is exported, allowing the usage of the pre-computed key/values hidden-states


helena-intel (approved):
Thanks Ella!

PR #50, opened 2022-09-23
Add NNCF post-training static quantization

echarlaix:
Introduces `OVQuantizer` a class handling post-training quantization process of NNCF

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -19,10 +19,10 @@ from pathlib import Path
 from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import torch
+import transformers
 from datasets import Dataset, load_dataset
 from torch.onnx import export as onnx_export
 from torch.utils.data import DataLoader, RandomSampler
-import transformers
 from transformers import PreTrainedModel, default_data_collator
 from transformers.onnx import FeaturesManager, OnnxConfig
 
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -22,6 +22,7 @@ import torch
 from datasets import Dataset, load_dataset
 from torch.onnx import export as onnx_export
 from torch.utils.data import DataLoader, RandomSampler
+import transformers
 from transformers import PreTrainedModel, default_data_collator
 from transformers.onnx import FeaturesManager, OnnxConfig
 
@@ -53,11 +54,11 @@ class OVQuantizer(OptimumQuantizer):
     Handle the NNCF quantization process.
     """
 
-    def __init__(self, model: PreTrainedModel, **kwargs):
+    def __init__(self, model: transformers.PreTrainedModel, **kwargs):
         """
         Args:
-            model (`PreTrainedModel`):
-                The model to quantize.
+            model (`transformers.PreTrainedModel`):
+                The [PreTrainedModel](https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel) to quantize.
             seed (`int`, defaults to 42):
                 The random seed to use when shuffling the calibration dataset.
         """
@@ -111,11 +112,7 @@ class OVQuantizer(OptimumQuantizer):
         )
         controller.prepare_for_export()
 
-        self.feature = HfApi().model_info(self.model.config._name_or_path).pipeline_tag or self.feature
-        if self.feature in ["sentiment-analysis", "text-classification", "zero-shot-classification"]:
-            self.feature = "sequence-classification"
-        elif self.feature in ["feature-extraction", "fill-mask"]:
-            self.feature = "default"
+        self._set_feature()
 
         self.model.config.save_pretrained(save_directory)
         model_type = self.model.config.model_type.replace("_", "-")
@@ -154,6 +151,16 @@ class OVQuantizer(OptimumQuantizer):
                 opset_version=10,
             )
             model.enable_dynamic_graph_building()
+
+    def _set_feature(self):
+        if self.feature is None:
+            self.feature = HfApi().model_info(self.model.config._name_or_path).pipeline_tag
+            if self.feature in ["sentiment-analysis", "text-classification", "zero-shot-classification"]:
+                self.feature = "sequence-classification"
+            elif self.feature in ["feature-extraction", "fill-mask"]:
+                self.feature = "default"
+            elif self.feature is None:
+                raise ValueError("The feature could not be extracted and needs to be specified for the ONNX export.")
 
     def get_calibration_dataset(
         self,
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -117,6 +117,7 @@ class OVQuantizer(OptimumQuantizer):
         elif self.feature in ["feature-extraction", "fill-mask"]:
             self.feature = "default"
 
+        self.model.config.save_pretrained(save_directory)
         model_type = self.model.config.model_type.replace("_", "-")
         onnx_config_cls = FeaturesManager._SUPPORTED_MODEL_TYPE[model_type][self.feature]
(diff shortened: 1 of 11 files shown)


AlexKoff88 (approved):

optimum/intel/openvino/quantization.py
```diff
@@ -0,0 +1,223 @@
...
+        Args:
+            dataset_name (`str`):
+                The dataset repository name on the Hugging Face Hub or path to a local directory containing data files
+                to load to use for the calibration step.
+            num_samples (`int`, defaults to 100):
+                The maximum number of samples composing the calibration dataset.
+            dataset_config_name (`str`, *optional*):
+                The name of the dataset configuration.
+            dataset_split (`str`, *optional*):
+                Which split of the dataset to use to perform the calibration step.
```
helena-intel, line 180: Is there a default? If a dataset has train/val splits and I don't specify a split, will the entire dataset be used, or will I get an error?

setup.py
```diff
@@ -30,7 +30,7 @@
 
 EXTRAS_REQUIRE = {
     "neural-compressor": "neural-compressor",
-    "openvino": "openvino>=2022.2.0",
+    "openvino": ["openvino>=2022.2.0", "nncf"],
```
helena-intel, line 33: Can we have nncf as an additional extra, instead of included in OpenVINO? I would like to keep the option of only using OpenVINO Runtime, with as few as possible dependencies. This is essential for using OpenVINO for deployment.
You would then do `pip install optimum-intel[openvino,nncf]` to install both OpenVINO and NNCF with optimum-intel.

tests/openvino/test_quantization.py
```diff
@@ -0,0 +1,63 @@
...
+from parameterized import parameterized
+
+
+class OVQuantizerTest(unittest.TestCase):
+    SUPPORTED_ARCHITECTURES_WITH_EXPECTED_QUANTIZED_MATMULS = (
+        ("distilbert-base-uncased-finetuned-sst-2-english", 82),
+    )
+
+    @parameterized.expand(SUPPORTED_ARCHITECTURES_WITH_EXPECTED_QUANTIZED_MATMULS)
+    def test_static_quantization(self, model_name, expected_fake_quantize):
```
helena-intel, line 33: Could we have a test that verifies that outputs are as expected? I would like to have at least 1 sanity check that accuracy is not impacted on an update.

helena-intel (commented):
Thanks @echarlaix ! I tested this and it worked well. I have some comments, none of them are critical. If you want to merge now and address some comments in a future PR that would be fine

setup.py
AlexKoff88: nncf is really a small dependency compared to transformers which are required as far as I understand. @helena-intel, do you mean that we can have only "openvino" as a dependency once we import the model from Hub? or we need optimum and transformers anyway?

optimum/intel/openvino/quantization.py
```diff
@@ -0,0 +1,223 @@
...
+class OVQuantizer(OptimumQuantizer):
+    """
+    Handle the NNCF quantization process.
+    """
+
+    def __init__(self, model: PreTrainedModel, **kwargs):
+        """
+        Args:
+            model (`PreTrainedModel`):
+                The model to quantize.
```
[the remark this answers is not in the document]
echarlaix: Ok I added additional explanation in a comment

optimum/intel/openvino/quantization.py
```diff
@@ -0,0 +1,223 @@
...
+        # TODO : Create model
+        return cls(model, **kwargs)
+
+    def quantize(
+        self,
+        quantization_config: NNCFConfig,
+        calibration_dataset: Dataset,
+        save_directory: Union[str, Path],
+        file_name: Optional[str] = None,
+        batch_size: int = 8,
```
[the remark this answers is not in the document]
echarlaix: It looks reasonable to me, 8 is also the [default](https://github.com/huggingface/transformers/blob/v4.22.2/src/transformers/training_args.py#L160) value for both the training and evaluation batch size of the `TrainingArguments`, and in case of larger models, users will have the possibility to reduce it. But we can modify it if you prefer this option, I don't have a strong opinion on this

setup.py
helena-intel: nncf is small, but it has dependencies which still add up. Dependencies can also have version conflicts with dependencies of other tools you need, so it is very useful to have an option to install optimum-intel[openvino] with the minimum number of dependencies. We discussed in the OpenVINO Runtime integration that in the future we could also remove PyTorch from the base install requirements, to make the footprint smaller.

These are current nncf dependencies (on Python 3.8/Linux): `addict, jsonschema, jstyleson, matplotlib, natsort, networkx, ninja, pandas, pillow, pydot, pymoo, pyparsing, scikit-learn, scipy, texttable, tqdm, wheel`

optimum/intel/openvino/quantization.py
```diff
@@ -0,0 +1,223 @@
...
+        preprocess_function: Optional[Callable] = None,
+        preprocess_batch: bool = True,
+        use_auth_token: bool = False,
+    ) -> Dataset:
+        """
+        Create the calibration `datasets.Dataset` to use for the post-training static quantization calibration step.
+
+        Args:
+            dataset_name (`str`):
+                The dataset repository name on the Hugging Face Hub or path to a local directory containing data files
```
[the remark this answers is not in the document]
echarlaix: I added some additional description

setup.py
AlexKoff88: From what I see now in the mandatory requirements for `pip install optimum[openvino]` we have:
```
INSTALL_REQUIRES = [
    "optimum",
    "transformers>=4.17.0",
    "datasets",
    "torch",
    "sentencepiece",
    "scipy",
]
```
`"transformers"` lib, in its turn, has a lot of additional requirements. So nncf requirements do not look like a burden in this case.

optimum/intel/openvino/quantization.py
echarlaix: No in this case there will be an error, and because in many cases there is a train/eval split it makes sense to add the training set as default, I will add this modification

tests/openvino/test_quantization.py
echarlaix: Sure, I will add the test

setup.py
helena-intel: In total nncf installs 33 packages, transformers 16. Unlike transformers, NNCF has things like matplotlib that someone who just wants to use transformer models will never use. Every additional package risks causing issues. We see this with some regularity with openvino-dev dependencies that get conflicts with eachother. If at some point matplotlib decides to only support numpy version X and transformers requires numpy version Y it will be impossible to install optimum-intel[openvino]. It is nice if the base package is as small as possible.

optimum/intel/openvino/quantization.py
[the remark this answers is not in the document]
helena-intel: Thanks, we can keep it for now and test memory use, and if needed change it later. I often have to manually set a much smaller batch size when training models - but I suppose a higher end system is expected there.

helena-intel (approved):

setup.py
AlexKoff88: BTW, I counted 19 dependencies including torch in the setup.py. But I agree that this can be too much.

echarlaix: I added nncf as an additional extra as proposed by @helena-intel

optimum/intel/openvino/quantization.py
```diff
@@ -0,0 +1,242 @@
...
+#  limitations under the License.
+
+import inspect
+import io
+import logging
+from itertools import chain
+from pathlib import Path
+from typing import Callable, Dict, List, Optional, Tuple, Union
+
+import torch
```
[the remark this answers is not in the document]
echarlaix: I added the corresponding operators patching

AlexKoff88:
Let's merge.

PR #52, opened 2022-09-26
[Draft]: Added working export for NNCF model

AlexKoff88:
- Added working export (w/ some hardcode)
- Added default quantization config for reference

### DIFF
--- a/optimum/intel/openvino/__init__.py
+++ b/optimum/intel/openvino/__init__.py
@@ -22,3 +22,5 @@ from .modeling import (
 )
 from .modeling_seq2seq import OVModelForSeq2SeqLM
 from .utils import OV_DECODER_NAME, OV_DECODER_WITH_PAST_NAME, OV_ENCODER_NAME, OV_XML_FILE_NAME
+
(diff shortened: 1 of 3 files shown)


optimum/intel/openvino/__init__.py
```diff
@@ -22,3 +22,5 @@
 )
 from .modeling_seq2seq import OVModelForSeq2SeqLM
 from .utils import OV_DECODER_NAME, OV_DECODER_WITH_PAST_NAME, OV_ENCODER_NAME, OV_XML_FILE_NAME
+
+from .nncf_config import DEFAULT_QUANTIZATION_CONFIG
```
helena-intel, line 26: @echarlaix @AlexKoff88 Can we sync on the import location? With this PR DEFAULT_QUANTIZATION_CONFIG is imported `from optimum.intel.openvino` but OVQuantizer is imported `from optimum.intel.openvino.quantization`. They should have the same root. Since optimum.intel.openvino is already three levels, I'm happy with everything under that, but open to other ideas.

AlexKoff88: Sure

PR #55, opened 2022-09-27
Replace INC QAT and pruning example to dynamic quantization

echarlaix:
Replace INC QAT and pruning example to dynamic quantization in quickstart.
Closes https://github.com/huggingface/hub-docs/issues/328


hshen14 (approved):
LGTM

PR #58, opened 2022-09-27
Improve openvino example

echarlaix:


README.md
```diff
@@ -112,4 +112,21 @@ tokenizer = AutoTokenizer.from_pretrained(model_id)
...
+
+By default, OVModels support dynamic shapes, enabling inputs of every shapes.
+To decrease latency, static shapes can be enabled by giving the desired inputs shapes.
+
+```python
+# Fix the batch size to 1 and the sequence length to 9
+model.reshape(1, 9)
+```
+
+Currently, OpenVINO GPU only supports static shapes. FP16 precision can also be enabled in order to further decrease latency.
```
helena-intel, line 125:
```suggestion
Currently, OpenVINO only supports static shapes for inference on GPU. FP16 precision can also be enabled in order to further decrease latency.
```
or "OpenVINO GPU inference"?

PR #67, opened 2022-10-06
Add OVTrainer enabling NNCF QAT

echarlaix:
In this PR we introduce an `OVTrainer` enabling NNCF quantization aware training, given the appropriate `OVConfig` configuration.

### DIFF
--- a/optimum/intel/openvino/trainer.py
+++ b/optimum/intel/openvino/trainer.py
@@ -540,6 +540,8 @@ class OVTrainer(Trainer):
 
     def _onnx_export(self, model: NNCFNetwork, config: OnnxConfig, f: Union[str, io.BytesIO]):
         model_inputs = config.generate_dummy_inputs(self.tokenizer, framework=TensorType.PYTORCH)
+        device = model.device
+        model_inputs = dict((k, v.to(device)) for k, v in model_inputs.items())
 
         with torch.no_grad():
             model.eval()
--- a/optimum/intel/openvino/trainer.py
+++ b/optimum/intel/openvino/trainer.py
@@ -69,6 +69,9 @@ from .utils import OV_XML_FILE_NAME
 if is_apex_available():
     from apex import amp
 
+if is_sagemaker_mp_enabled():
+    import smdistributed.modelparallel.torch as smp
+
 core = Core()
 
 logger = logging.get_logger(__name__)
--- a/optimum/intel/openvino/trainer.py
+++ b/optimum/intel/openvino/trainer.py
@@ -23,6 +23,7 @@ from itertools import chain
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
 import torch
+import torch.distributed as dist
 from packaging import version
 from torch.onnx import export as onnx_export
 from torch.utils.data import DataLoader, Dataset, RandomSampler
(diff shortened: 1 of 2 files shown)


optimum/intel/openvino/trainer.py
```diff
@@ -0,0 +1,558 @@
...
+        data_collator: Optional[DataCollator] = None,
+        train_dataset: Optional[Dataset] = None,
+        eval_dataset: Optional[Dataset] = None,
+        tokenizer: Optional[PreTrainedTokenizerBase] = None,
+        model_init: Callable[[], PreTrainedModel] = None,
+        compute_metrics: Optional[Callable[[EvalPrediction], Dict]] = None,
+        callbacks: Optional[List[TrainerCallback]] = None,
+        optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
+        preprocess_logits_for_metrics: Callable[[torch.Tensor, torch.Tensor], torch.Tensor] = None,
+        ov_config: Optional[OVConfig] = None,
```
AlexKoff88, line 82: I would call it `nncf_config`.

echarlaix: The configuration could contain information more general to OpenVINO (related to the export for example), so I would be in favor of not constraining it to nncf, what do you think ?

AlexKoff88: this was addressed in the previous PR that is merged already.

optimum/intel/openvino/trainer.py
```diff
@@ -0,0 +1,558 @@
...
+
+            # Export the compressed model to the ONNX format
+            output_path = os.path.join(output_dir, OV_XML_FILE_NAME)
+            self.compression_controller.prepare_for_export()
+            model_type = self.model.config.model_type.replace("_", "-")
+            onnx_config_cls = FeaturesManager._SUPPORTED_MODEL_TYPE[model_type][self.feature]
+            onnx_config = onnx_config_cls(self.model.config)
+            use_external_data_format = onnx_config.use_external_data_format(self.model.num_parameters())
+            f = io.BytesIO() if not use_external_data_format else output_path.replace(".xml", ".onnx")
+            self._onnx_export(self.model, onnx_config, f)
```
AlexKoff88, line 524: This is an error-prone code. Please save the current device, move the model to CPU, export, and move to the initial device.

optimum/intel/openvino/trainer.py
```diff
@@ -0,0 +1,558 @@
...
+            callbacks,
+            optimizers,
+            preprocess_logits_for_metrics,
+        )
+
+        self.ov_config = ov_config
+        self.feature = feature
+        self.compression_controller = None
+
+        if self.ov_config is not None and self.args.do_train:
```
AlexKoff88, line 104: Let's do the same for `ov_config` in case it is `None` as we do it in `OVQuantizer`.

optimum/intel/openvino/trainer.py
```diff
@@ -0,0 +1,558 @@
...
+
+        self._memory_tracker.stop_and_update_metrics(metrics)
+
+        self.log(metrics)
+
+        self.control = self.callback_handler.on_train_end(args, self.state, self.control)
+
+        return TrainOutput(self.state.global_step, train_loss, metrics)
+
+    def _save(self, output_dir: Optional[str] = None, state_dict=None):
```
helena-intel, line 482: Is this the expected way to save the quantized IR model, or is this called by a public method in the base class? It does not work for me. I did `trainer.save_model()` and `trainer._save()` after training, but there is no IR file in output_dir. It seems that self.compression_controller is None when I call this.

optimum/intel/openvino/trainer.py
echarlaix: I added a modification to move the `model_inputs` to the correct device to fix the ONNX export

optimum/intel/openvino/trainer.py
echarlaix: For the `OVQuantizer`, it make sense to me that the user will always want to apply static quantization when using it (so the `OVConfig` can be kept optional). However, for the `OVTrainer`, a user could be interested in training a model with and without QAT in order to compare both, so I am in favor of keeping QAT optional (enabling it when an `OVConfig` is given).

optimum/intel/openvino/trainer.py
echarlaix: To save the model you can use `trainer.save_model()` (which will call `trainer._save()`). Did you give an `ov_config` when instantiating the `OVTrainer` ? (if no `ov_config` is given, `self.compression_controller` will default to `None` and no QAT will be applied)

optimum/intel/openvino/trainer.py
AlexKoff88: OV specific is only in QAT, without it it does not make sense to replace `Trainer` with `OVTrainer`. So, you can use `Trainer` and then use `OVModelFor...`

optimum/intel/openvino/trainer.py
helena-intel: Thanks, it is intuitive that save_model() will also save the INT8 model and I'm glad that that's how it works! The issue was indeed solved by adding an OVConfig.

optimum/intel/openvino/trainer.py
echarlaix: What if we integrate additional compression techniques in the future such as pruning or distillation, QAT shouldn't be the default option in my opinion. Also having a config specifying the quantization process would keep consistency with `optimum.intel.neural_compressor.IncTrainer`.

AlexKoff88: Ok, let's go with the explicit definition of the config file.

AlexKoff88 (approved):

helena-intel (approved):

PR #69, opened 2022-10-08
add distributed training guide on quickstart

jiqing-feng:
@echarlaix @sywangyi 
Hi, this is a supplement to quickstart.mdx that teaches users how to use distributed training.


sywangyi:
@ftian1

docs/source/quickstart.mdx
```diff
@@ -85,4 +85,70 @@ tokenizer = AutoTokenizer.from_pretrained(model_id)
 pipe_cls = pipeline("text-classification", model=model, tokenizer=tokenizer)
 text = "He's a dreadful magician."
 outputs = pipe_cls(text)
+```
+
+# Distributed training
+
+Distributed training is implemented by intel oneCCL and MPI.
```
ftian1, line 92: will we support GLOO, NCCL and other communication libraries?

yao-matrix: We'll support GLOO, as it's a default CCL in PyTorch. And we'll not validate NCCL here, since we treat optimum-intel mainly on Intel's stacks.

@jiqing-feng  Suggest to wording as below(just an example):
CPU distributed distillation supports both Gloo and oneCCL as collective communication backend. And for performance seek, Intel recommend to use oneCCL backend.
We will show how to use oneCCL backend-ed distributed distillation as below steps.

docs/source/quickstart.mdx
```diff
@@ -85,4 +85,96 @@ tokenizer = AutoTokenizer.from_pretrained(model_id)
 pipe_cls = pipeline("text-classification", model=model, tokenizer=tokenizer)
 text = "He's a dreadful magician."
 outputs = pipe_cls(text)
+```
+
+# Distributed training
+
+When distillation on a single CPU is too slow, we can use multiple CPUs. This guide focuses on PyTorch-based DDP enabling distributed CPU distillation efficiently.
```
echarlaix, line 92: I would remove the mention to distillation (as it makes it seems like this option is specific to the distillation case)

echarlaix (commented):
Thanks for adding this to the documentation ! I think it would be great for distributed training to have its own [section](https://github.com/huggingface/optimum-intel/blob/main/docs/source/_toctree.yml#L5), what do you think?

echarlaix:
Thanks for adding a new section ! You will also need to add it to `_toctree.yml` https://github.com/huggingface/optimum-intel/blob/main/docs/source/_toctree.yml#L5 for the documentation to be generated

jiqing-feng:
@echarlaix Hi Ella. Thanks for your comments and I have fixed all of them. Could you please help me to review these changes? Thanks!

echarlaix (approved):
Thanks a lot for iterating on it ! Could you just add it as a subsection of Optimum Intel before we merge ? (as done [here](https://github.com/huggingface/optimum-intel/blob/add-ov-doc/docs/source/_toctree.yml#L9))

PR #74, opened 2022-10-12
Add default onnx opset depending on openvino version

echarlaix:


optimum/intel/openvino/utils.py
```diff
@@ -22,3 +22,6 @@
 ONNX_ENCODER_NAME = "encoder_model.onnx"
 ONNX_DECODER_NAME = "decoder_model.onnx"
 ONNX_DECODER_WITH_PAST_NAME = "decoder_with_past_model.onnx"
+
+MAX_ONNX_OPSET_2022_2_0 = 10
+MAX_ONNX_OPSET = 16
```
AlexKoff88, line 27: MAX_ONNX_OPSET also depends on the PyTorch version being used. PyTorch 1.12 claims opset 13 support: https://github.com/pytorch/pytorch/releases/tag/v1.12.0. I would suggest changing this to 13 as well because this is what OpenVINO claims as well.

AlexKoff88 (approved):

helena-intel (approved):
Thanks Ella! I verified that with this change a roberta model that fails with opset 10 works with OpenVINO 2022.3.

PR #78, opened 2022-10-17
Add openvino QAT question answering and image classification

echarlaix:


examples/openvino/image-classification/run_image_classification.py
```diff
@@ -0,0 +1,394 @@
...
+    Resize,
+    ToTensor,
+)
+from transformers import (
+    MODEL_FOR_IMAGE_CLASSIFICATION_MAPPING,
+    AutoConfig,
+    AutoFeatureExtractor,
+    AutoModelForImageClassification,
+    HfArgumentParser,
+    Trainer,
```
AlexKoff88, line 45: Probably no need to import `Trainer` here.

examples/openvino/question-answering/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Note that if your dataset contains samples with no possible answers (like SQuAD version 2), you need to pass along the flag `--version_2_with_negative`.
+
+### Fine-tuning BERT on SQuAD1.0
+
+This example code fine-tunes BERT on the SQuAD1.0 dataset while applying quantization aware training (QAT).
+
+```bash
+python run_qa.py \
+  --model_name_or_path bert-base-uncased \
```
AlexKoff88, line 34: I don't think this is the right model to use in this example because it is not trained for SQuAD. We encourage users to go for QAT only after they have a floating-point baseline trained for the downstream task (and PTQ should be prior to QAT). I would suggest to change this to any model fine-tuned on SQuAD.

examples/openvino/image-classification/run_image_classification.py
echarlaix: Good catch !

examples/openvino/question-answering/README.md
echarlaix: It makes sense to change it to a fine-tuned models to show that it can be done in a short training time. But I think that depending on the quantization process, there is definitely an interest in quantizing earlier in fine-tuning (in the case where some sensitive modules are quantized for example), this is not our case so I think it's a good idea to have fine-tuned models in the examples!

AlexKoff88 (approved):

examples/openvino/question-answering/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+# Question answering
+
+This folder contains [`run_qa.py`](https://github.com/huggingface/optimum/blob/main/examples/openvino/question-answering/run_qa.py), a script to fine-tune a 🤗 Transformers model on a question answering dataset while applying quantization aware training (QAT). QAT can be easily applied by replacing the Transformers [`Trainer`](https://huggingface.co/docs/transformers/main/en/main_classes/trainer#trainer) with the Optimum [`OVTrainer`].
```
helena-intel, line 18: Could you add that trainer_qa.py includes a QuestionAnsweringTrainer specifically for OVTrainer?

examples/openvino/question-answering/README.md
```diff
@@ -0,0 +1,44 @@
...
+uses special features of those tokenizers. You can check if your favorite model has a fast tokenizer in
+[this table](https://huggingface.co/transformers/index.html#supported-frameworks).
+
+Note that if your dataset contains samples with no possible answers (like SQuAD version 2), you need to pass along the flag `--version_2_with_negative`.
+
+### Fine-tuning BERT on SQuAD1.0
+
+Here we show how to apply quantization aware training (QAT) on a fine-tuned DistilBERT on the SQuAD1.0 dataset.
+
+```bash
```
helena-intel, line 32: Can we mention `--max_train_samples` and `--max_eval_samples` here as a way to speed up the process? This is very useful for verifying functionality too. Even on my system with GPU, SQuAD takes a long time to train one epoch.

And maybe a link to Trainer documentation https://huggingface.co/docs/transformers/main_classes/trainer ?

examples/openvino/question-answering/run_qa.py
```diff
@@ -0,0 +1,687 @@
...
+from transformers.utils.versions import require_version
+
+import evaluate
+from optimum.intel.openvino import OVConfig
+from trainer_qa import QuestionAnsweringTrainer
+from utils_qa import postprocess_qa_predictions
+
+
+# Will error if the minimal version of Transformers is not installed. Remove at your own risks.
+check_min_version("4.23.0")
```
helena-intel, line 54: If examples need 4.23, can we have that in the main optimum-intel dependency requirements too? Does this example really need 4.23?

helena-intel (approved):
Thanks for adding this! I tried the QA example with several models and it worked well. 
I have a few comments, could also be addressed after merge.

examples/openvino/question-answering/README.md
echarlaix: It's a good point as an epoch on the entire dataset is not required , I have set `max_train_samples` to 1024 (which gave me good final F1 score).

examples/openvino/question-answering/run_qa.py
echarlaix: This example should work with `transformers` 4.17.0 as well, it's coming from transformers v4.23.0 and has thus be validated for this version. It's more for information purposes and can be removed by the user, do you think it could be an issue for users ?

PR #96, opened 2022-11-07
Add example_inputs for FX trace if PyTorch > 1.13

PenghuiCheng:
# What does this PR do?

This PR is for adding example_input for FX trace, since PyTorch 1.13 and above version, prepare_fx has a example_inputs parameter.

## Before submitting

### DIFF
--- a/examples/neural_compressor/language-modeling/run_clm.py
+++ b/examples/neural_compressor/language-modeling/run_clm.py
@@ -613,9 +613,7 @@ def main():
 
         if quant_approach == IncQuantizationMode.STATIC:
             calib_dataloader = trainer.get_train_dataloader()
-            quantizer = IncQuantizer(
-                q8_config, eval_func=eval_func, calib_dataloader=calib_dataloader
-            )
+            quantizer = IncQuantizer(q8_config, eval_func=eval_func, calib_dataloader=calib_dataloader)
         else:
             example_inputs = next(iter(trainer.get_train_dataloader()))
             example_inputs = [example_inputs[key] for key in example_inputs]
--- a/examples/neural_compressor/language-modeling/run_clm.py
+++ b/examples/neural_compressor/language-modeling/run_clm.py
@@ -611,10 +611,17 @@ def main():
 
             q8_config.set_config("model.framework", "pytorch_fx")
 
-        calib_dataloader = trainer.get_train_dataloader() if quant_approach != IncQuantizationMode.DYNAMIC else None
-        quantizer = IncQuantizer(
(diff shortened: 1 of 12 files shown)


echarlaix (approved):

examples/neural_compressor/language-modeling/run_clm.py
```diff
@@ -611,10 +611,15 @@ def train_func(model):
...
-        )
+        if quant_approach == IncQuantizationMode.STATIC:
+            calib_dataloader = trainer.get_train_dataloader()
+            quantizer = IncQuantizer(q8_config, eval_func=eval_func, calib_dataloader=calib_dataloader)
+        else:
+            example_inputs = next(iter(trainer.get_train_dataloader()))
+            example_inputs = [example_inputs[key] for key in example_inputs]
+            quantizer = IncQuantizer(
+                q8_config, eval_func=eval_func, train_func=train_func, example_inputs=example_inputs
+            )
```
echarlaix, line 622: Why giving an `example_inputs` in the case of dynamic quantization ? 
I would suggest :
```suggestion
        example_inputs = None
        if quant_approach == IncQuantizationMode.AWARE_TRAINING:
            example_inputs = next(iter(trainer.get_train_dataloader()))
            example_inputs = [example_inputs[key] for key in example_inputs]
        calib_dataloader = trainer.get_train_dataloader() if quant_approach == IncQuantizationMode.STATIC else None
        quantizer = IncQuantizer(
            q8_config,
            eval_func=eval_func,
            calib_dataloader=calib_dataloader,
            example_inputs=example_inputs,
        )
```
Also, what is the interest of having multiple inputs in `example_inputs` instead of using only one ?

optimum/intel/neural_compressor/quantization.py
```diff
@@ -97,6 +100,12 @@ def __init__(
         self.calib_func = calib_func
         if calib_dataloader is not None:
             calib_dataloader = IncDataLoader.from_pytorch_dataloader(calib_dataloader)
+        elif example_inputs is not None:
+            if is_torch_less_than_1_13:
+                raise EnvironmentError("PyTorch must be 1.13 or above.")
```
echarlaix, line 105: Shouldn't we keep the `calib_dataloader` for post-training static quantization as it was previously required and add the `example_inputs` only for QAT ? I think it's not worth it to add breaking changes just before the big refactorization we are working on. Also the user would need to preprocess the dataset according to the model's input which can be tricky

echarlaix (changes requested):
I think we should discuss whether it makes sense to add `example_inputs` before the refactorization

PR #99, opened 2022-11-07
Add dump of ONNX file on demand using config

AlexKoff88:


optimum/intel/openvino/quantization.py
```diff
@@ -169,6 +173,7 @@ def _onnx_export(model: NNCFNetwork, config: OnnxConfig, model_inputs: Dict, f:
             )
         max_onnx_opset = min(config.default_onnx_opset, MAX_ONNX_OPSET)
         opset = max_onnx_opset if _openvino_version > version.Version("2022.2.0") else MAX_ONNX_OPSET_2022_2_0
+        opset = opset if not ov_config.dump_onnx else max(opset, MIN_ONNX_QDQ_OPSET)
```
echarlaix, line 176: If the opset is set to `MIN_ONNX_QDQ_OPSET` instead of `MAX_ONNX_OPSET_2022_2_0` wouldn't that cause any issue for the resutling OpenVINO model ?

optimum/intel/openvino/configuration.py
```diff
@@ -44,14 +44,13 @@ class OVConfig(BaseConfig):
...
-        self,
-        compression: Optional[Dict] = None,
-        input_info: Optional[List] = None,
-        **kwargs,
+        self, compression: Optional[Dict] = None, input_info: Optional[List] = None, dump_onnx: bool = False, **kwargs
     ):
         super().__init__()
         self.compression = compression or DEFAULT_QUANTIZATION_CONFIG
         self.input_info = input_info
+        self.dump_onnx = dump_onnx
```
echarlaix, line 52: I would rename it to 
```suggestion
        self.save_onnx_model = save_onnx_model
```

optimum/intel/openvino/quantization.py
AlexKoff88: It could but the user should not use save_onnx_model option in this case. This is because per-channel QuantizeLinear/DequantizeLinear is supported starting from  opset13.

echarlaix: The problem should be solved in the next openvino release where the min ONNX opset will be 13 right ? In the meantime we can either add a warning to explain it or even remove the export to the openvino format in the case where `save_onnx_model` is set to `True`.

AlexKoff88: I am ok with the warning but a new OV version is coming soon (pre-release is available) so let's keep export to OV as well.

echarlaix: We can add a logger warning explaning the potential risks encountered if `MIN_ONNX_QDQ_OPSET` is strictly higher than `MAX_ONNX_OPSET` / `MAX_ONNX_OPSET_2022_2_0` (depending on the `openvino` version).

AlexKoff88: To tell the truth, the current statement that `MAX_ONNX_OPSET_2022_2_0=10` is not fully correct and it is caused by the fact that some models may not work while OpenVINO 2022.2 supporst ONNX opset higher than 13. This is the local issue that will be fixed in 22.3 soon.
I updated the warnings. Please check.

optimum/intel/openvino/quantization.py
```diff
@@ -160,15 +163,26 @@ def _save_pretrained(model: openvino.runtime.Model, output_path: str):
...
+                    logger.warning(
+                        f"The minimal ONNX opset for the given model architecture is {config.default_onnx_opset}. Currently, "
+                        f"some models may not work with the installed version of OpenVINO. You can update OpenVINO "
+                        f"to 2022.3.* version or use ONNX opset version {MAX_ONNX_OPSET_2022_2_0} to resolve the issue."
+                    )
+                else:
+                    logger.warning(
+                        f"The minimal ONNX opset for QDQ format export is {MIN_ONNX_QDQ_OPSET}. Currently, some models"
+                        f"may not work with the installed version of OpenVINO in this opset. You can update OpenVINO "
+                        f"to 2022.3.* version or do not use \"save_onnx_model\" option."
```
echarlaix, line 181: I would replace with 
```suggestion
                        f"to 2022.3.* version or set `save_onnx_model` to `False`."
```

optimum/intel/openvino/trainer.py
```diff
@@ -559,15 +561,24 @@ def _set_feature(self):
...
+                    logger.warning(
+                        f"The minimal ONNX opset for the given model architecture is {config.default_onnx_opset}. Currently, "
+                        f"some models may not work with the installed version of OpenVINO. You can update OpenVINO "
+                        f"to 2022.3.* version or use ONNX opset version {MAX_ONNX_OPSET_2022_2_0} to resolve the issue."
+                    )
+                else:
+                    logger.warning(
+                        f"The minimal ONNX opset for QDQ format export is {MIN_ONNX_QDQ_OPSET}. Currently, some models"
+                        f"may not work with the installed version of OpenVINO in this opset. You can update OpenVINO "
+                        f"to 2022.3.* version or do not use \"save_onnx_model\" option."
```
echarlaix, line 577: I would replace with 
```suggestion
                        f"to 2022.3.* version or set `save_onnx_model` to `False`."
```

echarlaix (approved):
LGTM, thanks for adding this !

PR #102, opened 2022-11-10
Move PyTorch install to [extras]

helena-intel:
This PR removes PyTorch from the default installed packages, and moves it to the extras "neural_compressor" and "nncf". The reason for this change is that PyTorch is a pretty large framework which from the latest release by default also installs several CUDA related packages. This takes up more than 3GB of space, and takes quite a long time to download if you're on a slowish internet connection. These packages are unnecessary for users who just want to do inference on an OpenVINO model and it is useful to have a more minimal package for these users. 

The GitHub Actions tests passed in my fork, but they may not catch everything so I made this a draft PR for now so we can discuss/review if there are edge-cases that need to be resolved.

Also made a minor change to the test_openvino.yml file to allow running the test manually without having to make a PR.

### DIFF
--- a/.github/workflows/test_openvino.yml
+++ b/.github/workflows/test_openvino.yml
@@ -3,10 +3,12 @@
 name: OpenVINO - Test
 
 on:
+  workflow_dispatch:
   push:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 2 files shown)


hshen14:
PyTorch is a hard dependency for transformers. I don't think it makes sense to move to extra.

AlexKoff88 (approved):

echarlaix:
You're right @helena-intel, torch is currently a required dependency for optimum, which we preview to set as a soft dependency in the future.

helena-intel:
Thanks Ella! Happy to hear that this will be addressed in the future.

PR #108, opened 2022-11-17
Add model compilation when instantiating an OVModel

echarlaix:


helena-intel:
Can there be an option to not compile the model on instantiating? For some models (e.g. large encoder-decoder models) compiling can be pretty slow at the moment, and it is unnecessary to compile it to the device if you want to reshape the model, or move it to GPU, or just save the IR model with `.save_pretrained()` for example.

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -143,22 +143,29 @@ def __init__(
         self.device = torch.device("cpu")
         self.main_input_name = "input_ids"
         self.decoder_with_past = None
+        disable_compilation = kwargs.get("disable_compilation", False)
```
AlexKoff88, line 146: `compile=False` would be preferable to me.

AlexKoff88:
I agree with the changes but need to showcase this in the documentation as well.

AlexKoff88 (approved):

echarlaix:
> I agree with the changes but need to showcase this in the documentation as well.

An explanation of the changes was already added in the documentation, do you have something specific in mind that you would like us to add ?

AlexKoff88:
> > I agree with the changes but need to showcase this in the documentation as well.
> 
> An explanation of the changes was already added in the documentation, do you have something specific in mind that you would like us to add ?

Sorry, I missed .mdx file. Is it possible to highlight it as a Note? Alternatively, it can be a code snippet with an example of model preparation flow where `model.compile()` is placed at the end.

echarlaix:
> > > I agree with the changes but need to showcase this in the documentation as well.
> > 
> > 
> > An explanation of the changes was already added in the documentation, do you have something specific in mind that you would like us to add ?
> 
> Sorry, I missed .mdx file. Is it possible to highlight it as a Note? Alternatively, it can be a code snippet with an example of model preparation flow where `model.compile()` is placed at the end.

Sure, I just added an example in the documentation

AlexKoff88:
> > > > I agree with the changes but need to showcase this in the documentation as well.
> > > 
> > > 
> > > An explanation of the changes was already added in the documentation, do you have something specific in mind that you would like us to add ?
> > 
> > 
> > Sorry, I missed .mdx file. Is it possible to highlight it as a Note? Alternatively, it can be a code snippet with an example of model preparation flow where `model.compile()` is placed at the end.
> 
> Sure, I just added an example in the documentation

Thanks, @echarlaix. Looks good.

helena-intel (approved):

PR #111, opened 2022-11-21
Fix PR doc generation

regisss:
# What does this PR do?

This PR fixes the issue where the generated PR doc cannot be pushed to the remote server. See https://github.com/huggingface/optimum/pull/495.

## Before submitting


echarlaix (approved):
Thanks a lot for taking care of this @regisss !!

PR #118, opened 2022-11-24
Refactorization intel neural compressor

echarlaix:
In this PR, we adapt the `INCQuantizer` and `INCTrainer` classes after the refactorization on the quantization API from `neural-compressor`, simplifying its usage.

Should be merged after `neural-compressor` next release.


PenghuiCheng (approved):

README.md
```diff
@@ -28,6 +28,45 @@ pip install git+https://github.com/huggingface/optimum-intel.git
...
+from neural_compressor.config import PostTrainingQuantConfig
+from optimum.intel.neural_compressor import INCQuantizer
+
+model_name = "distilbert-base-cased-distilled-squad"
+model = AutoModelForQuestionAnswering.from_pretrained(model_name)
+# The directory where the quantized model will be saved
+save_dir = "quantized_model"
+# Load the quantization configuration detailing the quantization we wish to apply
+quantization_config = PostTrainingQuantConfig(approach="dynamic")
+quantizer = INCQuantizer.from_pretrained(model)
```
ftian1, line 49: Ella, do you have any future plan to have a AutoQuantizer liker Transformers to automatically dispatch to right quantizer?

README.md
```diff
@@ -28,6 +28,45 @@ pip install git+https://github.com/huggingface/optimum-intel.git
...
+quantizer.quantize(quantization_config=quantization_config, save_directory=save_dir, save_onnx_model=True)
+```
+
+To load a quantized model hosted locally or on the 🤗 hub, you can do as follows :
+```python
+from optimum.intel.neural_compressor import INCQuantizedModelForSequenceClassification
+from optimum.onnxruntime import ORTModelForSequenceClassification
+
+# Load the ONNX model hosted locally
+onnx_model = ORTModelForSequenceClassification.from_pretrained(save_dir)
```
ftian1, line 60: 1. should the ORTModelForSequenceClassification be ORT**Quantized**ModelForSequenceClassification
2. is asking user to explicitly invoke ORTModelXxx or INCQuantizedModelXxx a good way to load model? shall we make it transparent for user?

optimum/intel/neural_compressor/trainer.py
```diff
@@ -59,40 +60,93 @@
...
+        data_collator: Optional[DataCollator] = None,
+        train_dataset: Optional[Dataset] = None,
+        eval_dataset: Optional[Dataset] = None,
+        tokenizer: Optional[PreTrainedTokenizerBase] = None,
+        model_init: Callable[[], PreTrainedModel] = None,
+        compute_metrics: Optional[Callable[[EvalPrediction], Dict]] = None,
+        callbacks: Optional[List[TrainerCallback]] = None,
+        optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
+        preprocess_logits_for_metrics: Callable[[torch.Tensor, torch.Tensor], torch.Tensor] = None,
+        quantization_config: Optional[_BaseQuantizationConfig] = None,
```
ftian1, line 102: shall we list these configs here as three input parameters or a single parameter which is a list of configs? whether the latter has better extensibility?

optimum/intel/neural_compressor/trainer.py
```diff
@@ -59,40 +60,93 @@
...
+            preprocess_logits_for_metrics,
+        )
+
+        inc_config = []
+        self.task = task
+        self.quantization_config = quantization_config
+        self.pruning_config = pruning_config
+        self.distillation_config = distillation_config
+
+        self.compression_controller = None
```
ftian1, line 129: INC used `compression_manager` class to store compression related info. looks compression_controller is the concept of NNCF. suggest to change the variable name to compression_manager to avoid confusing.

optimum/intel/neural_compressor/quantization.py
```diff
@@ -35,101 +40,292 @@
...
-from huggingface_hub import hf_hub_download
+from huggingface_hub import HfApi, hf_hub_download
 from neural_compressor.adaptor.pytorch import PyTorch_FXAdaptor, _cfg_to_qconfig, _propagate_qconfig, get_torch_version
 from neural_compressor.adaptor.torch_utils.util import get_embedding_contiguous
-from neural_compressor.conf.config import Quantization_Conf
-from neural_compressor.experimental import Quantization
-from neural_compressor.utils.pytorch import _load_int8_orchestration
+from neural_compressor.model.torch_model import PyTorchModel
+from neural_compressor.quantization import fit
+from neural_compressor.utils.pytorch import _load_int8_orchestration, load
```
ftian1, line 58: shall we remove _load_int8_orchestration? it's an internal func and shouldn't use because it may be changed in the future in INC.

optimum/intel/neural_compressor/quantization.py
```diff
@@ -35,101 +40,292 @@
...
+
+    def quantize(
+        self,
+        save_directory: Union[str, Path],
+        quantization_config: "PostTrainingQuantConfig",
+        calibration_dataset: Dataset = None,
+        file_name: Optional[str] = None,
+        batch_size: int = 8,
+        data_collator: Optional[DataCollator] = None,
+        remove_unused_columns: bool = True,
```
ftian1, line 144: sorry, I didn't get the point of this parameter. could you elaborate more?

optimum/intel/neural_compressor/quantization.py
```diff
@@ -35,101 +40,292 @@
...
-                    )
-                self.quantization._calib_dataloader = self.calib_dataloader
+    @classmethod
+    def from_pretrained(cls, model: PreTrainedModel, **kwargs):
+        # TODO : Create model
+        return cls(model, **kwargs)
+
+    def quantize(
+        self,
+        save_directory: Union[str, Path],
```
ftian1, line 138: can we adjust the order of input parameters? such as put quant_config and calib_dataset as the first two parameters as they are important, organize data_collator/batch-size following w/ calibration_dataset, and merge save_directory and file_name to be one?(not sure the latter is feasible).

README.md
echarlaix: `ORTModelForSequenceClassification` allows to load any ONNX model (quantized or not) so we shouldn't rename it to `ORTQuantizedModelForSequenceClassification` in my opinion. Something that we can think about is change `INCQuantizedModelXxx` to `INCModelXxx` to reduce the friction (even though we lose an important information, as `INCQuantizedModelXxx`  should be used for quantized model only)

echarlaix: I don't think we should make it transparent as users would need `ORTModelForXxx` to load ONNX model and run inference with ONNX Runtime while `INCModelForXxx` is to load a PyTorch model quantized with INC, so both are for different case and requires different dependencies

optimum/intel/neural_compressor/trainer.py
echarlaix: IMO three parameters make more sense right now as it's more understandable for users but happy to discuss about different possibilities ! Why do you think we should make something more extensible ? Will we have more configuration to pass in the future ? I think ideally we would have one parameter combining all the details related to all the different compression techniques, and create the adapted INC configuration directly in the `INCTrainer`, what do you think ? In case we plan to do this, I think it would be better to add it in a future PR as this one might be already pretty big

optimum/intel/neural_compressor/trainer.py
echarlaix: Sure I'll change it !

README.md
echarlaix: What do you mean by an `AutoQuantizer` dispatching to the right quantizer ?

optimum/intel/neural_compressor/quantization.py
echarlaix: Good catch ! it's not used, I'll remove this import

optimum/intel/neural_compressor/quantization.py
echarlaix: Sure I can reorder `quant_config` and `save_directory`, `calibration_dataset` needs to be after those two though cause it's optional (not needed for dynamic quantization). `file_name`  was added to leave the possibility to overwrite the default file name `"pytorch_model.bin"`, but I'll remove it as it can be confusing

optimum/intel/neural_compressor/quantization.py
echarlaix: It gives the possibility to remove or not the columns of the dataset unused by the model forward method

PR #122, opened 2022-11-29
Enable openvino export and inference for stable diffusion models

echarlaix:


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -0,0 +1,229 @@
...
+            "`optimum.intel.openvino.OVModelForStableDiffusion` is experimental and might change in the future."
+        )
+        self.model = model
+        self._device = kwargs.get("device", "CPU")
+        self.ov_config = {"PERFORMANCE_HINT": "LATENCY"}
+        self.request = None
+
+        # Reshape the U-NET model
+        if "encoder_hidden_states" in [inputs.get_any_name() for inputs in self.model.inputs]:
+            self.model = self._reshape(self.model)
```
anton-l, line 48: Would it help to have this operation in diffusers during export?

echarlaix: Yes I'll check if by modifying the ONNX export, we can make this reshaping step optional

PR #125, opened 2022-12-01
Add IPEX inference_mode contextmanager to enable optimization on Intel platform.

mfuntowicz:
This PR introduces a new contextmanager in the exact same spirit of the one from PyTorch: `torch.inference_mode()` to enable IPEx optimization on Intel platform.

Usage:

```
from optimum.intel import inference_mode
with inference_mode(model=<pipeline|model>, dtype=torch.float32) as model:
    model(...)
```


setup.py
```diff
@@ -18,6 +18,7 @@
     "torch",
     "sentencepiece",
     "scipy",
+    "intel_extension_for_pytorch"
```
echarlaix, line 21: Should we move it to `EXTRAS_REQUIRE` to keep the openvino / neural-compressor subpackages independent from IPEX ?

optimum/intel/ipex/inference.py
```diff
@@ -0,0 +1,78 @@
...
+        self._dtype = dtype
+        self._graph_mode = False  # Let's keep for future use when it doesn't hang anymore
+        self._use_onednn = use_onednn
+        self._original = None
+
+    def __enter__(self):
+
+        with torch.inference_mode():
+            with ipex.verbose(self._verbose):
+                ipex.enable_onednn_fusion(True)
```
echarlaix, line 50: Should oneDNN fusion be enabled when `self._use_onednn` is `True` ?

optimum/intel/ipex/inference.py
```diff
@@ -0,0 +1,78 @@
...
+
+class inference_mode:
+    __slots__ = ("_model", "_dtype", "_graph_mode", "_use_onednn", "_verbose", "_original")
+
+    def __init__(
+        self,
+        model: Union[nn.Module, Pipeline],
+        dtype: torch.dtype = torch.float32,
+        use_onednn: bool = True,
+        verbose: bool = False
```
echarlaix, line 37: It could be nice to add a docstring to give more information on these parameters

echarlaix (approved):
LGTM, thanks for adding this @mfuntowicz 🔥🔥

optimum/intel/ipex/inference.py
mfuntowicz: Let's always use oneDNN otherwise we dont get the performance boost.

setup.py
mfuntowicz: Done ✔️

optimum/intel/ipex/inference.py
mfuntowicz: Done ✔️

PR #128, opened 2022-12-04
question-answering quantization notebook with NNCF

helena-intel:
# What does this PR do?

Add a Jupyter notebook that shows how to do post-training quantization with NNCF on a QA model. By default csarron/bert-base-uncased-squad-v1 is used, I also tested with a distilbert and roberta model pretrained on SQuAD.


notebooks/openvino/question_answering_quantization.ipynb
```diff
@@ -0,0 +1,945 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "eaed3927-e315-46d3-8889-df3f3bbcbf6b",
+   "metadata": {},
+   "source": [
+    "# Quantize a Hugging Face Question-Answering Model with OpenVINO\n",
+    "\n",
+    "This notebook shows how to quantize a question answering model with OpenVINO's Neural Network Compression Framework (NNCF). \n",
```
AlexKoff88, line 10:
```suggestion
    "This notebook shows how to quantize a question answering model with OpenVINO's [Neural Network Compression Framework](https://github.com/openvinotoolkit/nncf) (NNCF). \n",
```

AlexKoff88:
It looks awesome really. There can be still questions about the representativeness of such a comparison but I think we can omit them for now.

notebooks/openvino/question_answering_quantization.ipynb
```diff
@@ -0,0 +1,945 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "58df445d-af43-4ba1-8195-7d8f00b8f82f",
+   "metadata": {},
+   "source": [
+    "### Model Size\n",
+    "\n",
+    "Quantization reduces the size of the model by up to four times. We save the FP32 PyTorch model and define a function to show the model size for the PyTorch and OpenVINO models."
```
echarlaix, line 805: Do you know why the resulting model is only 2.39x smaller ? I would remove the "up to four times" sentence because of this

notebooks/openvino/question_answering_quantization.ipynb
```diff
@@ -0,0 +1,945 @@
...
+      "INT8 size decrease: 2.39x\n"
+     ]
+    }
+   ],
+   "source": [
+    "def get_model_size(model_folder, framework):\n",
+    "    \"\"\"\n",
+    "    Return OpenVINO or PyTorch model size in Mb. model_folder should point to a\n",
+    "    path that contains pytorch_model.bin for a PyTorch model, and openvino_model.xml/.bin\n",
+    "    for an OpenVINO model. framework should be \"pytorch\" or \"openvino\".\n",
```
echarlaix, line 838:
```suggestion
    "    Return OpenVINO or PyTorch model size in Mb. \n",
    "    Arguments: \n",
    "        model_folder: \n",
    "            Directory containing a pytorch_model.bin for a PyTorch model, and an openvino_model.xml/.bin for an OpenVINO model. \n",
    "        framework: \n",
    "            Define whether the model is a PyTorch or an OpenVINO model. \n"
```

echarlaix (commented):
LGTM, thanks for adding this @helena-intel !

notebooks/openvino/question_answering_quantization.ipynb
```diff
@@ -0,0 +1,945 @@
...
+     "iopub.status.idle": "2022-12-04T14:40:25.150427Z",
+     "shell.execute_reply": "2022-12-04T14:40:25.150085Z",
+     "shell.execute_reply.started": "2022-12-04T14:40:25.147933Z"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "MODEL_ID = \"csarron/bert-base-uncased-squad-v1\"\n",
+    "DATASET_NAME = \"squad\"\n",
+    "VERSION_2_WITH_NEGATIVE = False\n",
```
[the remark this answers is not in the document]
helena-intel: I used the capital letters to indicate that these are "constants", to make it obvious later in the notebook that these are settings set in the beginning. It makes it much harder to accidentally use the same name in a function and then inadvertently use the global variable instead of an intended local one. 
But if this is against a style guide, I'm happy to change.

notebooks/openvino/question_answering_quantization.ipynb
helena-intel: I was also surprised to see it. It is because by default some values are added to ignored_scopes. I was unsure how to explain this, suggestions welcome, also @AlexKoff88

notebooks/openvino/question_answering_quantization.ipynb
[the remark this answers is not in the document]
echarlaix: Usually we keep lower case in our notebooks, I think capital letters might make things a little less readable but I don't have a strong opinion on this, so I'll leave this decision to you

PR #131, opened 2022-12-09
[OV] Seq2Seq pipeline improvements

jiwaszki:
This PR add potential performance improvements to Seq2Seq OpenVINO pipelines.

### DIFF
--- a/optimum/intel/openvino/modeling_base_seq2seq.py
+++ b/optimum/intel/openvino/modeling_base_seq2seq.py
@@ -66,7 +66,7 @@ class OVBaseModelForSeq2SeqLM(OVBaseModel):
         self.model_save_dir = kwargs.get("model_save_dir")
         self._device = kwargs.get("device", "CPU")
         self.is_dynamic = kwargs.get("dynamic_shapes", True)
-        self.ov_config = {"PERFORMANCE_HINT": "LATENCY"}
+        self.ov_config = {}
         if "GPU" in self._device:
             raise ValueError("Support of dynamic shapes for GPU devices is not yet available.")
         if self.is_dynamic:
--- a/optimum/intel/openvino/modeling_seq2seq.py
+++ b/optimum/intel/openvino/modeling_seq2seq.py
@@ -330,12 +330,10 @@ class OVEncoder:
             inputs["attention_mask"] = Tensor(attention_mask.numpy(), shared_memory=True)
 
         # Run inference
-        self.request.start_async()
+        self.request.start_async(inputs)
         self.request.wait()
 
-        last_hidden_state = torch.from_numpy(
-                                self.request.get_tensor("last_hidden_state").data
-                            ).to(self.device)
+        last_hidden_state = torch.from_numpy(self.request.get_tensor("last_hidden_state").data).to(self.device)
 
         return BaseModelOutput(last_hidden_state=last_hidden_state)
 
@@ -393,7 +391,11 @@ class OVDecoder:
 
         if past_key_values is not None:
             # Flatten the past_key_values
-            past_key_values = [Tensor(past_key_value.numpy(), shared_memory=True) for pkv_per_layer in past_key_values for past_key_value in pkv_per_layer]
+            past_key_values = [
+                Tensor(past_key_value.numpy(), shared_memory=True)
+                for pkv_per_layer in past_key_values
+                for past_key_value in pkv_per_layer
+            ]
             # Add the past_key_values to the decoder inputs
             for input_name, past_key_value in zip(self.key_value_input_names, past_key_values):
                 inputs[input_name] = past_key_value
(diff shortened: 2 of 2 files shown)


helena-intel (approved):
Thanks @jiwaszki ! This improves performance by quite a lot on large-ish sequences! On a quick test on Ice Lake I saw about 30% speedup with a sequence length of 84. from I verified on a subset of  iwslt2017 and opus100 that the bleu score is not impacted by this change.

optimum/intel/openvino/modeling_base_seq2seq.py
```diff
@@ -66,7 +66,7 @@ def __init__(
         self.model_save_dir = kwargs.get("model_save_dir")
         self._device = kwargs.get("device", "CPU")
         self.is_dynamic = kwargs.get("dynamic_shapes", True)
-        self.ov_config = {"PERFORMANCE_HINT": "LATENCY"}
```
echarlaix, line 69: Why is it more efficient to remove the `"PERFORMANCE_HINT"` key ?

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -375,25 +386,40 @@ def forward(
...
+        # Add the encoder_hidden_states inputs when needed
+        if "encoder_hidden_states" in self.input_names:
+            encoder_hidden_states = _contiguous_helper(encoder_hidden_states.numpy())
+            inputs["encoder_hidden_states"] = Tensor(encoder_hidden_states, shared_memory=True)
 
         # Run inference
-        outputs = self.request.infer(inputs)
-        outputs = {key.get_any_name(): value for key, value in outputs.items()}
+        self.request.start_async(inputs)
+        self.request.wait()
```
echarlaix, line 418: When should it be used ? Also should we add the same modifications for other `OVModels` ?

echarlaix (approved):
LGTM, thanks a lot for adding this @jiwaszki !

optimum/intel/openvino/modeling_base_seq2seq.py
helena-intel: > Why is it more efficient to remove the `"PERFORMANCE_HINT"` key ?

Jan is on holiday so I'll answer for now: great question! It is not expected, but this showed improved performance. We suspect that the latency performance hint was optimized for single models, not encoder/decoder models. We will look into this more.

jiwaszki: @echarlaix exactly as @helena-intel stated. It will be consulted with OV teams, as this case is very interesting one.

PS While I am on holidays, I am still receiving notifications from open source -- I will answer occasionally:)

optimum/intel/openvino/modeling_seq2seq.py
jiwaszki: @echarlaix It really depends on the application itself. First I can explain why I prefer `async_infer`:

- It fully releases GIL (synchronous mode is currently doing it partially). Thus other applications on Python threads can capture GIL and perform their work in the meantime.
-  Specific case of `OVEncoder` is showing that only one tensor has to retrieved from `InferRequest`. This approach creates less overhead comparing to simple `infer()` call, as synchronous function is creating whole dictionary as returned value, which results in few extra memory accesses by dispatching logic.
- It is more scalable solution. Actually, if there were some parts of the code that can be done without interfering with model or inputs, I would also put them between `async_infer()` and `wait()` calls. This will boost pipeline even more.

However there are downsides as well:

- Concurrency and data races which may appear. Especially when combined with "zero-copy" shared memory approach that was added in PR.
- GIL itself. As long as the model execution is relatively long, releasing GIL could be beneficial. If said runtime is shorter than releasing and reacquiring of the lock, then it makes no sense to release it. It will be just a waste of time.
- Asynchronous code is simply more complex. It requires extra effort to write and maintain. I wouldn't recommend this approach for "proof of concept" pipelines.

> When should it be used? Also should we add the same modifications for other OVModels?

As I mentioned above, there are a lot of factors. General rule of thumb - if data safety can be ensured in the pipeline, go with asynchronous mode with shared memory optimisations. However, it really depends on the application itself.

If you want I can assist you while optimising OpenVINO pipelines. I will be very happy to gather your feedback and further improve OpenVINO Python API according to it.

I hope I didn't make this comment too long! 😄

optimum/intel/openvino/modeling_base_seq2seq.py
echarlaix: Thank you for the explanation !

optimum/intel/openvino/modeling_seq2seq.py
echarlaix: Thanks a lot for the detailed explanation @jiwaszki !! That would be great !

PR #132, opened 2022-12-12
Catch when IPEx raises `RuntimeError` not being able to optimize a mo…

mfuntowicz:
This was making the Inference API hanging forever.


echarlaix (approved):
LGTM !

PR #135, opened 2022-12-15
Adapt OVModel for optimum recent changes

echarlaix:


tests/openvino/test_modeling.py
```diff
@@ -66,6 +70,21 @@
 SEED = 42
 
 
+class ORTModelIntegrationTest(unittest.TestCase):
```
AlexKoff88, line 73: why the test is called `ORTModelIntegrationTest`?

echarlaix: Good catch, I'll rename it !

AlexKoff88 (approved):

PR #136, opened 2022-12-16
Change the ground truth to original picture

PenghuiCheng:
# What does this PR do?

This PR will change the ground truth picture to the original one that the user used in examples maybe.

## Before submitting


hshen14 (commented):
Good catch. LGTM

PR #138, opened 2022-12-21
Update examples and fixed pruning error

PenghuiCheng:
# What does this PR do?
This PR will adapt to new API of neural_compressor, and Fixed pruning sparsity issue.

## Before submitting

### DIFF
--- a/examples/neural_compressor/language-modeling/run_clm.py
+++ b/examples/neural_compressor/language-modeling/run_clm.py
@@ -157,17 +157,19 @@ class OptimizationArguments:
         default=0.1,
         metadata={"help": "Targeted sparsity when pruning the model."},
     )
-    start_epoch: int = field(
+    start_step: int = field(
         default=0,
-        metadata={"help": "Epoch for which the pruning process will start."},
+        metadata={"help": "step for which the pruning process will start."},
     )
-    end_epoch: Optional[int] = field(
+    end_step: Optional[int] = field(
         default=None,
-        metadata={"help": "Epoch for which the pruning process will end."},
+        metadata={"help": "step for which the pruning process will end."},
     )
     pruning_approach: str = field(
-        default="basic_magnitude",
-        metadata={"help": "Pruning approach. Supported approach is basic_magnitude."},
+        default="snip_momentum",
+        metadata={
+            "help": "Pruning approach. Supported approach is 'magnitude'," "'gradient', 'snip_momentum', and 'snip'."
+        },
     )
     apply_distillation: bool = field(
         default=False,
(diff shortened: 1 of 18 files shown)


examples/neural_compressor/language-modeling/run_clm.py
```diff
@@ -591,16 +593,22 @@ def compute_metrics(eval_preds):
 
     if optim_args.apply_pruning:
 
-        if optim_args.end_epoch is None:
-            end_epoch = training_args.num_train_epochs
+        if optim_args.end_step is None:
+            end_step = training_args.num_train_epochs * (
```
echarlaix, line 597: Shouldn't we include the number of `gradient_accumulation_steps` ? Also I think the number of epoch might be easier for users to set, what do you think ?

tests/neural_compressor/test_optimization.py
```diff
@@ -229,7 +237,7 @@ def compute_metrics(p: EvalPrediction):
                 transformers_outputs = transformers_model(**tokens)
             self.assertTrue(torch.allclose(onnx_outputs.logits, transformers_outputs.logits, atol=1e-4))
             sparsity = trainer.get_model_sparsity()
-            self.assertGreaterEqual(sparsity, target_sparsity * 100)
+            self.assertGreaterEqual(sparsity, target_sparsity * 100 / 2)
```
echarlaix, line 240: Where does the  `/ 2` comes from ?

echarlaix (commented):
Thanks for adding this @PenghuiCheng !

examples/neural_compressor/language-modeling/run_clm.py
```diff
@@ -157,17 +157,19 @@ class OptimizationArguments:
...
         default=None,
-        metadata={"help": "Epoch for which the pruning process will end."},
+        metadata={"help": "step for which the pruning process will end."},
     )
     pruning_approach: str = field(
-        default="basic_magnitude",
-        metadata={"help": "Pruning approach. Supported approach is basic_magnitude."},
+        default="snip_momentum",
+        metadata={
+            "help": "Pruning approach. Supported approach is 'magnitude'," "'gradient', 'snip_momentum', and 'snip'."
```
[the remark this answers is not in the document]
PenghuiCheng: Added the pruning method introduction in README and optimization_inc.mdx

examples/neural_compressor/language-modeling/run_clm.py
```diff
@@ -591,16 +593,22 @@ def compute_metrics(eval_preds):
...
+                optim_args.end_step,
+                training_args.num_train_epochs * (len(train_dataset) // training_args.per_device_train_batch_size),
+            )
 
         pruning_config = WeightPruningConfig(
-            start_epoch=optim_args.start_epoch,
-            end_epoch=end_epoch,
+            start_step=optim_args.start_step,
+            end_step=end_step,
+            excluded_op_names=["classifier"],
```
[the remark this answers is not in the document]
PenghuiCheng: Yes, excluded_op_names is for improving accuracy in some cases. I think we can remove it in examples. Users can ensure which op to be excluded.

examples/neural_compressor/language-modeling/run_clm.py
PenghuiCheng: Pruning occurs after the every optimizer.step() is called. So I think needn't include gradient_accumulation_steps. Our think the step can slightly adjust sparsity.

tests/neural_compressor/test_optimization.py
PenghuiCheng: "get_model_sparsity()" return the sparsity of the whole model, in the model, no all ops will be pruned, there are many op unpruned, and target_sparsity is only indicates that the sparsity for the op that can do prune. In this case, if we set the target sparsity to 0.9,  the model sparsity is about 0.56, So I divide the result by 2.

examples/neural_compressor/language-modeling/run_clm.py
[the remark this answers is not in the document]
echarlaix: Yes sure we can remove it!

tests/neural_compressor/test_optimization.py
PenghuiCheng: @echarlaix, I add the explanation of pruning in the README and docs/source/optimization_inc.mdx. 
Thanks!

examples/neural_compressor/language-modeling/run_clm.py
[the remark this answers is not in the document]
echarlaix: Thank you !

tests/neural_compressor/test_optimization.py
echarlaix: Thank you @PenghuiCheng

echarlaix (approved):

PR #139, opened 2022-12-21
FP16 hotfix

AlexKoff88:
A hotfix for FP16 that provides the correct results when running models with dynamic shapes and FP16 precision on GPU.


echarlaix (approved):
LGTM, thanks for fixing it !

PR #141, opened 2022-12-23
Modify openvino required version

echarlaix:


fxmarty (approved):
LGTM sorry for the breaking change! We should be more careful.

Edit: maybe should not be merged actually https://huggingface.slack.com/archives/C02P0559X9S/p1671819829307639?thread_ts=1671816524.896739&cid=C02P0559X9S

PR #146, opened 2022-12-27
[INC] Neural Coder enables API Design Adaptor for Optimum-Intel APIs

kaiyaointel:
# What does this PR do?

This PR enables Neural Coder API Design Adaptor in Optimum-Intel's source code by adding the file ```optimum-intel/optimum/intel/neural_compressor/neural_coder_adaptor.py``` and adds a Python launcher to Optimum-Intel for INC optimizations ```optimum.intel.launcher```.

When user uses this launcher ```optimum.intel.launcher``` to call INC's  Neural Coder to auto-enable Optimum-Intel's quantization optimizations, Neural Coder will by default use the API design in this file (by referring the constants in ```NeuralCoderAdaptor``` class) to insert the API code lines and transform the code.

Typical usage: ```python -m optimum.intel.launcher run_glue.py``` to run ```run_glue.py``` with Optimum-Intel automatically enabled in it (in this case, the code change is explicitly invisible to users, which is a no-code solution)

When Optimum-Intel quantization API design changes, Optimum-Intel developers or INC developers can update this file accordingly to adapt the latest API design change.

Update guidance:
- General
Please simply modify the constants in ```NeuralCoderAdaptor``` class to reflect the latest API design change by rearranging the code lines in the multi-line string.
- Special aliases
  - EVAL_FUNC_LINES: keep this **unchanged** - this is an alias referring to the evaluate function lines auto-detected by Neural Coder algorithm
  - OPTIMUM_QUANT_CONFIG: keep this **unchanged** - this is an alias referring to the quantization config specified by Neural Coder's user in Neural Coder's front-end interface if they wish to use customized config file for quantization
  - MODEL_NAME: keep this **unchanged** - this is an alias referring to the model name auto-detected by Neural Coder algorithm
## Before submitting


kaiyaointel:
updated API design strings with latest API design

hshen14:
@echarlaix , please review this PR. It's better to include this in Optimum-Intel v1.6 release.

kaiyaointel:
@hshen14 I tried to fix the code format, you may trigger CI again

kaiyaointel:
@hshen14 fixed isort, pls trigger again, thanks

hshen14:
@echarlaix  please take a review and see whether it's ok for a merge.

optimum/intel/launcher.py
```diff
@@ -0,0 +1,14 @@
+from neural_coder.launcher import Launcher
+
+from .neural_compressor.neural_coder_adaptor import NeuralCoderAdaptor
```
echarlaix, line 3: This file should be moved to the `neural_compressor` directory as it's related to INC.
Also the command line `python -m optimum.intel.neural_compressor.launcher run_glue.py` is better for clarity I think, as the optimizations are related to INC

optimum/intel/neural_compressor/neural_coder_adaptor.py
```diff
@@ -0,0 +1,26 @@
+class NeuralCoderAdaptor:
```
echarlaix, line 1: It would be great to add a test as well to verify no modifications in the future is breaking anything

optimum/intel/neural_compressor/neural_coder_adaptor.py
```diff
@@ -0,0 +1,26 @@
...
+from optimum.intel.neural_compressor import INCQuantizer
+quantization_config = PostTrainingQuantConfig(approach="dynamic")
+quantizer = INCQuantizer.from_pretrained(MODEL_NAME)
+quantizer.quantize(quantization_config=quantization_config, save_directory="quantized_model", save_onnx_model=False)
+MODEL_NAME = quantizer._quantized_model
+MODEL_NAME.eval()
+"""
+
+    default_quant_static = f"""\
+def eval_func(model):
```
echarlaix, line 17: Why do we need an `eval_func` in our case as it doesn't seem to be used ?

optimum/intel/neural_compressor/neural_coder_adaptor.py
```diff
@@ -0,0 +1,26 @@
...
+    EVAL_FUNC_LINES
+from neural_compressor.config import PostTrainingQuantConfig
+from optimum.intel.neural_compressor import INCQuantizer
+quantization_config = PostTrainingQuantConfig(approach="static")
+quantizer = INCQuantizer.from_pretrained(MODEL_NAME)
+quantizer.quantize(quantization_config=quantization_config, calibration_dataset=eval_dataset, save_directory="quantized_model", save_onnx_model=False)
+MODEL_NAME = quantizer._quantized_model
+MODEL_NAME.eval()
```
echarlaix, line 25: Why is it necessary ?

echarlaix (commented):
Left a couple of comments, thanks for this addition !

optimum/intel/neural_compressor/neural_coder_adaptor.py
kaiyaointel: Thanks for raising this. We seem to have missed referencing ```eval_func``` in the API. @PenghuiCheng Do we use ```eval_func``` like this in new Optimum API? ```INCQuantizer.from_pretrained(MODEL_NAME, eval_fn=eval_func)```, or do we use it some other way? Thanks

optimum/intel/neural_compressor/neural_coder_adaptor.py
kaiyaointel: Are you referring to line 25, or?

optimum/intel/launcher.py
hshen14: NeuralCoder as a launcher provides the capability to expand to the other use cases e.g., IPEX, OV. So that's why we intend to make this under intel. Does it make sense?

kaiyaointel: We have been discussing this: the reason we currently put ```launcher.py``` at this location (under ```intel```) is because Neural Coder in the future might enable more Intel optimizations for Optimum-Intel like IPEX etc, which means it is not limited to INC. Also we're thinking if the command line becomes ```optimum.intel.neural_compressor.launcher``` it seems a bit long and less user-friendly. @hshen14 Any additional comments?

optimum/intel/neural_compressor/neural_coder_adaptor.py
kaiyaointel: Thanks for raising this. There's only two string constants in this Class and will not have more items in the future. Would you advice on how to add a test for only string constants? Thanks!

PenghuiCheng:
@kaiyaointel ， could we add an example for neural coder usage？

optimum/intel/neural_compressor/neural_coder_adaptor.py
kaiyaointel: Discussed with Haihao. We will add a test for this.

optimum/intel/neural_compressor/neural_coder_adaptor.py
PenghuiCheng: @kaiyaointel， there are two usages： 1. user only provides eval_dataloader and metric then it will quantize with built-in eval_func, 2. user can provide an eval_func to INCQuantizer to tune the quantized model.

kaiyaointel: OK, thanks, I will modify accordingly

PenghuiCheng: @echarlaix , I found that there is no built-in eval_func in INCQuantizer, so all examples didn't been tuned by  neural_compressor,  Why INCQuantizer didn't provide a built-in eval_func ?

optimum/intel/neural_compressor/neural_coder_adaptor.py
kaiyaointel: Test added. @echarlaix

optimum/intel/launcher.py
echarlaix: Because of the `neural_coder` dependency, I think it would make more sense to first add it to `neural_compressor` and to move it in the future in case of needs, what do you think ?

optimum/intel/neural_compressor/neural_coder_adaptor.py
echarlaix: Thanks @kaiyaointel !

optimum/intel/neural_compressor/neural_coder_adaptor.py
echarlaix: Sure, I can include the `eval_fn` in the `neural_compressor` examples.

optimum/intel/neural_compressor/neural_coder_adaptor.py
echarlaix: Yes !

tests/neural_compressor/test_neural_coder_adaptor.py
```diff
@@ -0,0 +1,11 @@
+import unittest
+
+from optimum.intel.neural_compressor.neural_coder_adaptor import NeuralCoderAdaptor
+
+
+class NeuralCoderAdaptorTest(unittest.TestCase):
+    def test_string_type(self):
```
echarlaix, line 7: Should we add a test to verify that model are correctly quantized using the launcher ?

optimum/intel/neural_compressor/neural_coder_adaptor.py
kaiyaointel: This line converts model to eval (evaluation) mode for PyTorch inference, we added this so that it's clear that the model will be in eval mode for the rest part of the code. Because quantized mode can only be used in evaluation mode not training mode (as far as I know)

tests/neural_compressor/test_neural_coder_adaptor.py
kaiyaointel: We can add a test on Launcher in some way. It's a bit hard to structure the UT for Launcher... Let me think about it!

optimum/intel/neural_compressor/neural_coder_adaptor.py
echarlaix: Ok this was added in case the model is used after the quantization step, thanks for the explanation !

echarlaix (approved):

optimum/intel/launcher.py
kaiyaointel: hi @echarlaix I have made the change per our discussion

kaiyaointel:
hi @echarlaix I have made the change per our discussion by relocating launcher.py's folder place, please merge the PR if it looks okay to you now.

PR #150, opened 2022-12-31
Enable OVTrainer with Joint Pruning, Quantization and Distillation

vuiseng9:
This PR extends OVTrainer with additional compression algorithms provided by NNCF such as pruning, self-distillation on top of the quantization support in the current OVTrainer. Distillation with larger (smaller) teacher is also enabled. OVTrainer also provides end-to-end OpenVINO IR optimization, including structured pruning. With that, users can achieve acceleration jointly from structured pruning and quantization and accuracy retention with distillation. Tests and examples of respective compression and joint optimization have been added.

## Before submitting


examples/openvino/audio-classification/README.md
```diff
@@ -0,0 +1,104 @@
...
+    --evaluation_strategy epoch \
+    --save_strategy epoch \
+    --load_best_model_at_end True \
+    --metric_for_best_model accuracy \
+    --save_total_limit 3 \
+    --seed 0 \
+    --push_to_hub
+```
+
+On a single V100 GPU (16GB), this script should run in ~14 minutes and yield accuracy of **98.26%**.
```
AlexKoff88, line 62: It would also be great to have similar information about time and resources for JPQD run.

examples/openvino/audio-classification/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+👀 See the results here: [anton-l/wav2vec2-base-ft-keyword-spotting](https://huggingface.co/anton-l/wav2vec2-base-ft-keyword-spotting)
+
+> If your model classification head dimensions do not fit the number of labels in the dataset, you can specify `--ignore_mismatched_sizes` to adapt it.
+
+### Joint Pruning, Quantization and Distillation (JPQD) of Wav2Vec2 for Keyword Spotting
+`OVTrainer` also provides an advanced optimization worflow through the NNCF when Transformer model can be structurally pruned along with 8-bit quantization and distillation. Following is an example to jointly prune, quantize, distill a Wav2Vec2 model tuned for keywork spotting task. Do take note of additional NNCF config `--nncf_compression_config`.
```
AlexKoff88, line 69: Can we add the info about what to expect after JPQD here? the similar to what the SQuAD sample has.

examples/openvino/image-classification/configs/swin-base-movement-sparsity.json
```diff
@@ -0,0 +1,15 @@
+{
+    "algorithm": "movement_sparsity",
+    "params": {
+        "warmup_start_epoch": 2,
+        "warmup_end_epoch": 5,
+        "importance_regularization_factor": 0.030,
+        "enable_structured_masking": true
+    },
+    "sparse_structure_by_scopes": [
```
AlexKoff88, line 9: Why does this only about pruning (not pruning + quantization)? From these two methods, quantization is the simplest one and delivers better performance. Most likely, user will choose quantization from two alternatives.  Stacking it with pruning brings additional benefits.

AlexKoff88:
Overall it looks great. But we need to wait for a new release of NNCF that is coming soon. 

Please also install isort-5.11.4 and black-22.12.0 and run `make style` to fix the code quality issues.

optimum/intel/openvino/trainer.py
```diff
@@ -536,18 +649,72 @@ def _save(self, output_dir: Optional[str] = None, state_dict=None):
...
+                onnx_config = self.onnx_config
+
+            if self._is_pruning_controller_exists():
+                # Note:
+                # OpenVINO provides automated structured pruning of sparse structure in IR.
+                # However it requires static axes, current export utilizes nncf exporter
+                # which generates static-shaped IR.
+
+                f = os.path.join(output_dir, "model.onnx")
+                self.compression_controller.export_model(
```
AlexKoff88, line 665: Can we return dynamic shapes after the pruning? It can be easily done using `model.reshape([-1,-1])` after calling `convert_model()`.

examples/openvino/audio-classification/README.md
```diff
@@ -0,0 +1,104 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Audio classification examples
+
+The following examples showcase how to fine-tune `Wav2Vec2` for audio classification using PyTorch.
```
helena-intel, line 19: While it is nice to show that NNCF finetuning is just adding a few commands to the PyTorch example, it is a bit confusing to me to show how to finetune with PyTorch in the OpenVINO example. Could we refer with a link to the PyTorch example and in the OpenVINO examples just show OpenVINO/NNCF code?

optimum/intel/openvino/configuration.py
```diff
@@ -68,10 +68,16 @@ def add_input_info(self, model_inputs: Dict):
...
 
-    def _enable_standard_onnx_export_option(self):
-        if isinstance(self.compression, dict):
+    def _set_standard_onnx_export_option(self):
+        # This method depends on self.save_onnx_model.
+        # save_onnx_model is defaulted to false so that the final model output is
+        # in OpenVINO IR to realize performance benefit in OpenVINO runtime.
+        # True value of save_onnx_model will save a model in onnx format.
+        # Quantized performance of onnx model in onnxruntime will require OpenVINO backend.
+        # Otherwise, default onnxruntime backend gives a full-precision performance instead of low-precision's.
```
helena-intel, line 77: I am concerned that referring people to ORT here will cause some confusion. This is a private method for people who need it, but the default supported method to run models created with optimum[openvino] is with OpenVINO Runtime.

optimum/intel/openvino/configuration.py
```diff
@@ -54,7 +54,7 @@ def __init__(
         self.compression = compression or DEFAULT_QUANTIZATION_CONFIG
         self.input_info = input_info
         self.save_onnx_model = save_onnx_model
-        self._enable_standard_onnx_export_option()
+        self._set_standard_onnx_export_option()
```
helena-intel, line 57: This method is currently in use by users. Is renaming necessary?

optimum/intel/openvino/trainer.py
```diff
@@ -346,6 +372,9 @@ def _inner_training_loop(
                 else args.max_steps * args.gradient_accumulation_steps
             )
             self.control = self.callback_handler.on_epoch_begin(args, self.state, self.control)
+            if self.compression_controller is not None:
+                self.compression_controller.scheduler.epoch_step()
+                print(self.compression_controller.statistics().to_str())
```
helena-intel, line 377: Could this be done with `logger.info()` instead?

setup.py
```diff
@@ -30,8 +30,10 @@
 
 EXTRAS_REQUIRE = {
     "neural-compressor": ["neural-compressor>=2.0.0", "onnx", "onnxruntime"],
-    "openvino": ["openvino>=2022.3.0", "transformers>=4.20.0", "onnx"],
-    "nncf": ["nncf"],
+    "openvino": ["openvino-dev>=2022.3.0", "transformers>=4.20.0", "onnx"],
```
helena-intel, line 33: Why the change from openvino->openvino-dev for the openvino extra?

helena-intel (commented):
Thanks for the PR @vuiseng9 ! It is really great to have this available in optimum[openvino]. I will try the examples when I have access to a computer with a GPU. It would be great if the models from the examples are to the Hugging Face hub as well.

It would be great to have this feature documented in the documentation, at https://huggingface.co/docs/optimum/intel/optimization_ov (source: docs/source/optimization_ov.mdx).

examples/openvino/image-classification/README.md
```diff
@@ -39,4 +39,31 @@ python run_image_classification.py \
     --save_total_limit 3 \
     --seed 1337 \
     --output_dir /tmp/beans_outputs/
+```
+
+### Structured Pruning on Swin for food101
+Following is an example to prune Swin-base model structurally while fine-tuning for food101. More on how to configure the pruning algorithm, see NNCF documentation [here](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/experimental/torch/sparsity/movement/MovementSparsity.md).
```
helena-intel, line 45: Would it be possible to add how long this took to train on one GPU, and what the resulting accuracy is, like in the other JPQD examples?

optimum/intel/openvino/trainer.py
vuiseng9: Good catch )

setup.py
vuiseng9: IR pruning requires model optimizer which is bundled in openvino-dev. See [here](https://github.com/vuiseng9/optimum-intel/blob/pr-nncf-jpqd/optimum/intel/openvino/trainer.py#L75).

examples/openvino/image-classification/README.md
vuiseng9: Yes, it's WIP.

optimum/intel/openvino/configuration.py
vuiseng9: I felt that it is a misnomer. The method sets "export_to_onnx_standard_ops" in the compression config and the value depends on instance initialization argument "save_onnx_model" which can be true or false, it is only enabled when it is true.

examples/openvino/audio-classification/README.md
vuiseng9: WIP

examples/openvino/audio-classification/README.md
vuiseng9: Added.

examples/openvino/image-classification/configs/swin-base-movement-sparsity.json
vuiseng9: WIP

optimum/intel/openvino/trainer.py
vuiseng9: Collectively decided to retain current implementation for better performance.

examples/openvino/audio-classification/README.md
vuiseng9: @helena-intel I am not following, could you elaborate what you are looking for? what do you mean by OpenVINO example?

setup.py
vuiseng9: @helena-intel convert_model internally uses serialize, we did attempt to use serialize earlier but it would turn out to be a heavy routine in the OVTrainer to prune the IR and the routine will be a duplicate of convert_model. 

Regarding extra, do you mean we place openvino-dev as part of NNCF keypair? `"nncf": ["nncf", "openvino-dev>=2022.3.0],`

optimum/intel/openvino/trainer.py
helena-intel: That is unexpected. All models in optimum-intel are dynamic shapes by default, even though there is better performance for static shapes. Users can easily reshape their model to static shapes for optimal performance (with a possible loss of accuracy resulted from truncating data). Is it just "better performance" or is JQPD not recommended at all for dynamic shapes? If this works with static shapes only it should be very clearly documented.

examples/openvino/audio-classification/README.md
helena-intel: The other examples show basic QAT with NNCF first, and then JPQD. For example: "Here we show how to apply quantization aware training (QAT) on a fine-tuned Vision Transformer (ViT) on the beans dataset (to classify the disease type of bean leaves)". This example says  "how to fine-tune `Wav2Vec2` for audio classification using PyTorch.". For consistency and clarity, it is nice if this example also mentions NNCF/QAT for the first example, similar to the other examples.

In general for the examples, as a user I would like to know when to use the first example and when the second. If JQPD is better, should I just start with that? Is there any benefit still for just quantization?

We should preferably not just list examples as "how" to do things but also give a bit of a context about why to choose a particular method.

setup.py
helena-intel: I hoped it was just  `apply_pruning_transformation` and other transformations from openvino._offline_transformations and then serialize. Is it more than that? We could see if it is useful to add a generic utils function for that to optimum-intel.

openvino-dev includes quite a few other dependencies which adds some risk of things going wrong and additional unnecessary downloads for users so I prefer to avoid it if it is just for saving a few lines of code. 
If we do need openvino-dev, then yes, I meant it could be part of the "nncf" extra in setup.py, because this is only needed when NNCF is also needed.

optimum/intel/openvino/configuration.py
AlexKoff88: Both names look meaningful. Moreover, this is not a public API, I would recommend not changing the initial name since there are concerns.

vuiseng9: Reverted

examples/openvino/audio-classification/README.md
vuiseng9: see [audio classification readme.](https://github.com/vuiseng9/optimum-intel/blob/pr-nncf-jpqd/examples/openvino/audio-classification/README.md)

examples/openvino/audio-classification/README.md
vuiseng9: see [here](https://github.com/vuiseng9/optimum-intel/blob/pr-nncf-jpqd/examples/openvino/question-answering/README.md#joint-pruning-quantization-and-distillation-jpqd-for-bert-on-squad10).

optimum/intel/openvino/trainer.py
vuiseng9: Keep generated IR to static shape until MO/OV runtime is patched.

setup.py
vuiseng9: see [here](https://github.com/vuiseng9/optimum-intel/blob/pr-nncf-jpqd/setup.py#L33-L37)

examples/openvino/image-classification/configs/swin-base-movement-sparsity.json
vuiseng9: Added [here](https://github.com/vuiseng9/optimum-intel/blob/pr-nncf-jpqd/examples/openvino/question-answering/README.md#joint-pruning-quantization-and-distillation-jpqd-for-bert-on-squad10).

examples/openvino/image-classification/README.md
vuiseng9: see [here](https://github.com/vuiseng9/optimum-intel/tree/pr-nncf-jpqd/examples/openvino/image-classification#joint-pruning-quantization-and-distillation-on-swin-for-food101)

optimum/intel/openvino/trainer.py
```diff
@@ -78,6 +92,7 @@
 
 set_log_level(logging.ERROR)
 logger = logging.get_logger(__name__)
+logger.setLevel(logging.INFO)
```
helena-intel, line 95: With loglevel INFO  "Statistics of the sparsified model" printed almost 3000 lines when I tested this. That drowns out all other information in the terminal.

Could we handle that differently? For example, send that to a logfile?

optimum/intel/openvino/training_args.py
```diff
@@ -0,0 +1,29 @@
+from dataclasses import dataclass, field
+
+from transformers import TrainingArguments
+
+
+@dataclass
+class OVTrainingArguments(TrainingArguments):
```
helena-intel, line 7: @vuiseng9  can you explain why some parameters are part of the configuration json/dict, and others part of the OVTrainingArguments? It is not immediately obvious to me why for example warmup_start_epoch is an item in the config, but distillation_weight and temperature are OVTrainingArguments.

examples/openvino/question-answering/README.md
```diff
@@ -44,4 +44,32 @@ python run_qa.py \
...
+    --learning_rate 3e-5 \
+    --num_train_epochs 8 \
+    --per_device_eval_batch_size 128 \
+    --per_device_train_batch_size 16 \
+    --max_seq_length 384 \
+    --doc_stride 128 \
+    --logging_steps 1 \
+    --evaluation_strategy steps \
+    --eval_steps 250 \
+    --save_steps 500 \
```
helena-intel, line 71: Do we need to save every 500 steps by default? This saved about 180GB of checkpoints on my disk. Not everyone has a very large harddisk, it's annoying when you set a demo to train overnight and wake up to a completely full harddisk and training that crashed half way through because of that.

docs/source/optimization_ov.mdx
```diff
@@ -112,6 +112,76 @@ metrics = trainer.evaluate()
 trainer.save_model()
 ```
 
+## Distillation
+Distillation is an effective method to improve model performance by learning from the knowledge of, usually, larger model. `OVTrainer` also provides the capability to perform distillation during training. Users are required to load the teacher and feed the loaded model to `OVTrainer` as follows. Subclassed from `TrainingArguments`, `OVTrainingArguments` is provided to ease users to specify distillation hyperparameters (such as teacher model path, distillation weightage and temperature), as well as to expose the hyperparameters as commandline arguments to the training script.
+```diff
```
helena-intel, line 117: The documentation build shows some formatting issues. I would try to add a blank line before this. See https://moon-ci-docs.huggingface.co/docs/optimum-intel/pr_150/en/optimization_ov

vuiseng9: thanks for the suggestion! next iteration here https://github.com/huggingface/optimum-intel/pull/150/commits/c42936c5796db560491b913e80eecd6aef38c625

optimum/intel/openvino/training_args.py
```diff
@@ -0,0 +1,29 @@
...
+    nncf_compression_config: str = field(
+        default=None, metadata={"help": "NNCF configuration .json file for compression-enabled training"}
+    )
+    teacher_model_or_path: str = field(
+        default=None, metadata={"help": "Path to pretrained model or model identifier from huggingface.co/models"}
+    )
+    distillation_weight: float = field(
+        default=0.5, metadata={"help": "weightage of distillation loss, value between 0.0 to 1.0"}
+    )
+    distillation_temperature: float = field(default=2.0, metadata={"help": "temperature of distillation."})
```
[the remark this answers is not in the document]
vuiseng9: @helena-intel Optimal distillation temperature and weight will be on case-by-case basis, heuristics are required but these values are good value to start with. Having default values also ease the users as HF transformers require many input arguments.

optimum/intel/openvino/training_args.py
vuiseng9: @helena-intel The json configuration specifies hyperparameters for NNCF algorithms. NNCF native distillation is a self distillation which can be limiting. OVTrainer is extended with distillation that allows larger models as teacher. OVTrainingArguments is provided as a companion to OVTrainer to standardize adaptation of existing script/task for distillation and nncf-supported algorithms. For hyperparameters of NNCF algorithms, it would be challenging to expose through OVTrainingArguments as users can mix and match NNCF algorithms.

examples/openvino/question-answering/README.md
vuiseng9: Agree, I found it annoying too when I started to use HF's transformers. Added checkpoint limit in https://github.com/huggingface/optimum-intel/pull/150/commits/3e7ff7879d15eb5a02f4a099fa9a9e0e196abcc9 .

optimum/intel/openvino/trainer.py
vuiseng9: @helena-intel  see https://github.com/huggingface/optimum-intel/pull/150/commits/9e2f4ee7385dcb98256fcac6fe1445729a39874c .

optimum/intel/openvino/training_args.py
helena-intel: I'm more wondering the other way around: it is not obvious to me that distillation_weight and temperate are not "hyperparameters for NNCF algorithms", so I'm wondering why they aren't in the config but have to be specified separately in the trainer.

optimum/intel/openvino/trainer.py
```diff
@@ -100,6 +119,7 @@ def __init__(
         preprocess_logits_for_metrics: Callable[[torch.Tensor, torch.Tensor], torch.Tensor] = None,
         ov_config: Optional[OVConfig] = None,
         feature: Optional[str] = None,
+        onnx_config: Optional[OnnxConfig] = None,
```
echarlaix, line 122: In which case the user needs to use its own `onnx_config` ? I would remove this parameter in order to not let the possibility to have different onnx_config that we do not support during export

optimum/intel/openvino/trainer.py
```diff
@@ -536,18 +667,77 @@ def _save(self, output_dir: Optional[str] = None, state_dict=None):
...
+                # Note:
+                # OpenVINO provides automated structured pruning of sparse structure in IR.
+                # However it requires static axes, current export utilizes nncf exporter
+                # which generates static-shaped IR.
 
-            # Load and save the compressed model
-            model = core.read_model(f) if use_external_data_format else core.read_model(f.getvalue(), b"")
-            compress_quantize_weights_transformation(model)
-            openvino.runtime.serialize(model, output_path, output_path.replace(".xml", ".bin"))
+                f = os.path.join(output_dir, "model.onnx")
```
echarlaix, line 682: [`ONNX_WEIGHTS_NAME`](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/utils.py#L21) could be used instead of "model.onnx" (should have been done before)

optimum/intel/openvino/trainer.py
```diff
@@ -536,18 +667,77 @@ def _save(self, output_dir: Optional[str] = None, state_dict=None):
...
+                self.compression_controller.export_model(
+                    f, input_names=list(onnx_config.inputs.keys()), output_names=list(onnx_config.outputs.keys())
+                )
+                self._generate_openvino_ir(f)
```
echarlaix, line 686: For pruned models, would it be possible to also leave the possibility to save the ONNX model (to leave the possiblity to run inference using ONNX Runtime) ?

tests/openvino/test_quantization.py
```diff
@@ -162,6 +195,176 @@ def test_aware_training_quantization(self, model_name, expected_fake_quantize, e
...
+            trainer.train()
+            trainer.save_model()
+            self.assertIn("distillation_loss", trainer.compression_metrics)
+            self.assertIn("compression_loss", trainer.compression_metrics)
+
+            state_dict = torch.load(Path(tmp_dir, WEIGHTS_NAME), map_location="cpu")
+            num_binary_masks = sum(key.endswith("_binary_mask") for key in state_dict)
+            self.assertEqual(expected_binary_masks, num_binary_masks)
+
+            ovmodel = OVModelForSequenceClassification.from_pretrained(tmp_dir)
```
echarlaix, line 363: It could be interesting to also compare the outputs between the `trainer.model` and the `ovmodel` to verify the loading, as done [here](https://github.com/huggingface/optimum-intel/blob/main/tests/neural_compressor/test_optimization.py#L256) (I can do it in a following PR as well)

examples/openvino/audio-classification/configs/wav2vec2-base-jpqd.json
```diff
@@ -0,0 +1,56 @@
...
+            "warmup_start_epoch": 3,
+            "warmup_end_epoch": 8,
+            "importance_regularization_factor": 0.100,
+            "enable_structured_masking": true
+        },
+        "sparse_structure_by_scopes": [
+            {
+                "mode": "block",
+                "sparse_factors": [32, 32],
+                "target_scopes": "{re}.*Wav2Vec2Attention.*"
```
echarlaix, line 14: Why not adapt each config depending on the architecture instead of adding multiple configuration ?

docs/source/optimization_ov.mdx
```diff
@@ -112,6 +112,81 @@ metrics = trainer.evaluate()
...
+        "warmup_end_epoch":    4,
+        "importance_regularization_factor":  0.01,
+        "enable_structured_masking":  True
+    },
+    "sparse_structure_by_scopes": [
+        {"mode":  "block",   "sparse_factors": [32, 32], "target_scopes": "{re}.*BertAttention.*"},
+        {"mode":  "per_dim", "axis":  0,                 "target_scopes": "{re}.*BertIntermediate.*"},
+        {"mode":  "per_dim", "axis":  1,                 "target_scopes": "{re}.*BertOutput.*"},
+    ],
+    "ignored_scopes": ["{re}.*NNCFEmbedding", "{re}.*qa_outputs*", "{re}.*LayerNorm.*"]
```
echarlaix, line 161: In this example, the task is text classification so I would remove the part of the configuration adapted for a QA task

echarlaix (commented):
Thanks a lot for this great addition @vuiseng9  🔥

optimum/intel/openvino/trainer.py
vuiseng9: fixed in https://github.com/huggingface/optimum-intel/pull/150/commits/76991dcb79da98ea520a644e42f840a05c790ae4

docs/source/optimization_ov.mdx
vuiseng9: Good catch, patched for text classification in https://github.com/huggingface/optimum-intel/pull/150/commits/e93045ec3bf6058943a2f29ee7f9d7a48d35a451

docs/source/optimization_ov.mdx
```diff
@@ -112,6 +112,81 @@ metrics = trainer.evaluate()
 trainer.save_model()
 ```
 
+## Distillation
```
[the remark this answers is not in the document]
AlexKoff88: I agree that having a separate section that describes "Distillation" looks cumbersome, especially, if to note that we don't recommend distillation as a single method for model optimization. On the contrary, it is an auxiliary method that is used along with other optimization techniques. What we recommend is PTQ, QAT, and JPQD (distillation is a part of it).
I would recommend moving the description of the distillation to "Advanced Compression with NNCF" section.

examples/openvino/audio-classification/README.md
AlexKoff88: @helena-intel, do you mean that we should have a description of how to quantize the model in every example prior to deeper optimization methods such as JPQD? It makes sense to me if we have such consistency across all the samples but it requires some extra work to create a config and verify that it works. 
@vuiseng9, is it possible for you to check if QAT (with no pruning and distillation) works and add examples of the commands in every document?

examples/openvino/audio-classification/configs/wav2vec2-base-jpqd.json
AlexKoff88: @echarlaix, for sure, we are aiming at handling different model architectures in an automated way in NNCF without a need to specify such complicated configs on the user's side with "target_scopes", "ignored_scopes", etc. We have an idea of how we can do this and planning some POC. But for now, this is the only way how to make pruning work properly. If you have any idea how to make this more elegant feel free to propose. We can discuss as well.

tests/openvino/test_quantization.py
AlexKoff88: @vuiseng9, please address this

optimum/intel/openvino/training_args.py
```diff
@@ -0,0 +1,29 @@
...
+from transformers import TrainingArguments
+
+
+@dataclass
+class OVTrainingArguments(TrainingArguments):
+    """
+    Arguments pertaining to OpenVINO/NNCF-enabled training flow
+    """
+
+    nncf_compression_config: str = field(
```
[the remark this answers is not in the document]
AlexKoff88: This is related to the previous two comments. Since distillation is always use-case and topology specific it makes sense not to introduce it in the common OVTrainingArguments but rather create a specific Trainer object for the method, i.e. `JPQDQuestionAnsweringTrainer` and have all the distillation-specific parameters there only, it could be additional `OptimizationArguments` as proposed by @echarlaix. 
It is also possible to have a base `JPQDTrainer` and inherit from it every time.
@helena-intel, @echarlaix, @vuiseng9, what do you think?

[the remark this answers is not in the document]
helena-intel: I like a `JPQDTrainer` and/or `JPQDQuestionAnsweringTrainer`. For "regular" QA we already have a specific trainer too, because question answering requires some specific data processing, so to me a `JPQDTrainer` would be in line with that.

I do really like the option to add a `nncf_compression_config` in the trainer command line though, and I would like to consider adding that to the OVTrainer as an optional field (and also adapt the existing examples). The default config is designed to work well most of the time, but there can always be cases where you want to experiment with the config, or want to make other tradeoffs for speed/accuracy/model size and users should have that option.

docs/source/optimization_ov.mdx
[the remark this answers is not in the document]
echarlaix: Shouldn't the three sections "While training optimization", "Distillation" and "Advanced Compression with NNCF" be merged into one in your opinion @AlexKoff88 ? (describing clearly all the available compression techniques available at training time)

examples/openvino/audio-classification/configs/wav2vec2-base-jpqd.json
echarlaix: That's great news ! One thing that can be done is to set the layer patterns for each model architectures, and for the `ignored_scopes` this could be solved by using the task information (already required by the `OVTrainer`)

optimum/intel/openvino/training_args.py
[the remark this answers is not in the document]
echarlaix: Not sure to understand why we would need a `JPQDTrainer`, in my opinion it would be easier from a user perspective to have one unique `OVTrainer` which can be used to combine any of the compression techniques it enables

examples/openvino/text-classification/configs/bert-base-jpqd.json
```diff
@@ -0,0 +1,47 @@
...
+        "params": {
+            "warmup_start_epoch": 1,
+            "warmup_end_epoch": 2,
+            "importance_regularization_factor": 0.05,
+            "enable_structured_masking": true,
+        },
+        "sparse_structure_by_scopes": [
+            {"mode": "block", "sparse_factors": [32, 32], "target_scopes": "{re}.*BertAttention.*"},
+            {"mode": "per_dim", "axis": 0, "target_scopes": "{re}.*BertIntermediate.*"},
+            {"mode": "per_dim", "axis": 1, "target_scopes": "{re}.*BertOutput.*"},
```
echarlaix, line 13: Could it be possible to replace it with : 
```suggestion
            {"mode": "block", "sparse_factors": [32, 32], "target_scopes": "{re}.*Attention.*"},
            {"mode": "per_dim", "axis": 0, "target_scopes": "{re}.*Intermediate.*"},
            {"mode": "per_dim", "axis": 1, "target_scopes": "{re}.*Output.*"},
```
to have a more generic config (adapted to more architectures such as roberta for example)

docs/source/optimization_ov.mdx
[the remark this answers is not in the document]
AlexKoff88: I think so. I would also call it like:
- "Training-time optimization":
  - "QAT"
  - "JPQD" (distillation is described inside)

examples/openvino/audio-classification/configs/wav2vec2-base-jpqd.json
AlexKoff88: > set the layer patterns for each model architecture

@echarlaix, how it can be accomplished?

optimum/intel/openvino/training_args.py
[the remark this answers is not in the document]
echarlaix: In my opinion we should have one class (`OVTrainer`), allowing users to either apply quantization, pruning, knowledge distillation or any combination of these techniques. I don't see the gain of having a specific class for JPQD but I might be missing an important point

examples/openvino/audio-classification/configs/wav2vec2-base-jpqd.json
echarlaix: We could modify the `OVConfig` directly in the `OVTrainer` where we would have the model architecture information. Here is an [example](https://github.com/huggingface/nn_pruning/blob/main/nn_pruning/model_structure.py#L104) of some patterns defined for couple of architectures (bert, bart and t5) we did in `nn-pruning`. What is the solution you're currently investigating ? And would it be integrated to NNCF directly ?

optimum/intel/openvino/training_args.py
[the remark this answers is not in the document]
AlexKoff88: @echarlaix, I see your point. But, in my opinion, this is not a practical approach to let the user try various combinations of optimization methods while we know the best one. For example, using pruning without quantization does not make sense because the latter has certain advantages, namely, you can quickly quantize the model, it does not hurt accuracy a lot, and you can get a decent speedup after quantization, it does not require training HW in the case of PTQ. So that the workflow here should be quantized first and then use pruning on top of quantization if you still are not satisfied with the performance.
Let's discuss this problem.

optimum/intel/openvino/trainer.py
AlexKoff88: @vuiseng9, can you please clarify this?

examples/openvino/text-classification/README.md
```diff
@@ -0,0 +1,80 @@
...
+    --num_train_epochs 3 \
+    --logging_steps 10 \
+    --evaluation_strategy steps \
+    --eval_steps 250 \
+    --save_strategy epoch \
+    --fp16 \
+    --seed 42
+```
+
+On a single V100 GPU, this script should run in ~40 minutes and yield accuracy of **92.9%**.
```
helena-intel, line 47: I trained for 1 epoch (on an A10G) and got 92.55% accuracy (in 22 minutes). That's super close to 5-epoch accuracy. IMO it would be good to mention that as an option, This is an example, it's useful to know that you can have a good model quickly.

examples/openvino/audio-classification/README.md
vuiseng9: @AlexKoff88 Yes, we did that. That was how we found out that the default QAT doesn't work for wav2vec2 and provided a tailored config and relative performance improvement from JPQD in the documentation was compared against the QAT.

In all the support tasks, there are quantization and JPQD samples.

tests/openvino/test_quantization.py
vuiseng9: @AlexKoff88 the torch model here could have a different input shape to the IR model. We faced this issue and temporarily using mock tokens to check if IR model can be inferred. Once the dynamic shape IR issue is resolved, we can compare the `trainer.model` and `ovmodel` output.

optimum/intel/openvino/trainer.py
vuiseng9: The original implementation of [onnx_export](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/trainer.py#L562) requires an `OnnxConfig` object as input and corresponding `OnnxConfig` for a given task/architecture can be extracted from transformers' `FeatureManager` ([see](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/trainer.py#L539)). For some reasons, `Wav2vec2` has no entry in the `FeaturesManager`. In order to comply to design, we expose this to the user. We also created one for `Wav2vec2` in this pull request. 
@echarlaix Any reason why `wav2vec2` has not entry in `wav2vec2`?

echarlaix: The ONNX export of wav2vec2 models was recently added to [`optimum`](https://github.com/huggingface/optimum/blob/v1.6.3/optimum/exporters/onnx/model_configs.py#L802), I opened https://github.com/huggingface/optimum-intel/pull/173 to integrate it to the `OVQuantizer` and wanted to wait the merging of https://github.com/huggingface/optimum-intel/pull/150 before adding it to the `OVTrainer` but will integrate it to https://github.com/huggingface/optimum-intel/pull/150 as it's now needed on your end

echarlaix: https://github.com/huggingface/optimum-intel/pull/173 was merged, feel free to rebase !

examples/openvino/audio-classification/configs/wav2vec2-base-jpqd.json
AlexKoff88: @echarlaix, thanks for the link, it's very helpful. Regarding your question, right, it is supposed to be a part of NNCF which is responsible for graph analysis. Once we are ready with POC we can share the details. But the idea is to get rid of specifying model-related patterns in the config.

examples/openvino/text-classification/configs/bert-base-jpqd.json
AlexKoff88: @vuiseng9, can you please comment here?

optimum/intel/openvino/training_args.py
```diff
@@ -0,0 +1,29 @@
...
+        default=None, metadata={"help": "NNCF configuration .json file for compression-enabled training"}
+    )
+    teacher_model_or_path: str = field(
```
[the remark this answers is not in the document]
AlexKoff88: I agree, @vuiseng9 can you please address this comment?

examples/openvino/question-answering/run_qa.py
```diff
@@ -45,7 +45,9 @@
 from transformers.utils.versions import require_version
 
 import evaluate
-from optimum.intel.openvino import OVConfig
+import jstyleson as json
```
helena-intel, line 48: Is this extra requirement necessary? It seems it's only to load the non-standard json, and shouldn't nncf's safe_open do that already?

examples/openvino/audio-classification/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+Speech recognition models that have been pretrained in unsupervised fashion on audio data alone,
+*e.g.* [Wav2Vec2](https://huggingface.co/transformers/main/model_doc/wav2vec2.html),
+[HuBERT](https://huggingface.co/transformers/main/model_doc/hubert.html),
+[XLSR-Wav2Vec2](https://huggingface.co/transformers/main/model_doc/xlsr_wav2vec2.html), have shown to require only
+very little annotated data to yield good performance on speech classification datasets.
+
```
helena-intel, line 26: Can there be a note that "on Linux, you need to install libsndfile using your distribution’s package manager, for example sudo apt install libsndfile1" (from the soundfile pypi page)?

examples/openvino/question-answering/run_qa.py
vuiseng9: @helena-intel `safe_open` is only responsible for opening the file. The reason of using non-standard json is that, as a user myself, I find that I occasionally put some comments in the json, it is a flexibility during experiments. It was inspired by nncf out-of-the-box vision examples which use the lib `jstyleson`.

optimum/intel/openvino/trainer.py
vuiseng9: Rebased )

helena-intel:
@echarlaix the minimum optimum version in setup.py is now set to 1.6.1. I got an error when running the audio example with that version. Should it be updated to 1.6.4?

optimum/intel/openvino/trainer.py
AlexKoff88: @echarlaix, this call saves onnx model: ` exporter.export_model(f)`

docs/source/optimization_ov.mdx
```diff
@@ -112,18 +118,104 @@ metrics = trainer.evaluate()
 trainer.save_model()
 ```
 
+
+### Joint Pruning, Quantization and Distillation (JPQD)
+
+Other than quantization, compression methods like pruning and distillation are common in further improving the task performance and efficiency. Structured pruning slims a model for lower computational demands while distillation leverages knowledge of a teacher, usually, larger model to improve model prediction. Combining these methods with quantization can result in optimized model with significant efficiency improvement while enjoying good task accuracy retention. In `optimum.openvino`, `OVTrainer` provides the capability to jointly prune, quantize and distill a model during training. Following is an example on how to perform the optimization on BERT-base for the sst-2 task.
```
AlexKoff88, line 124: I think we should put a disclaimer here about what kind of model architectures are supported for now.

vuiseng9: Agree, added in this commit https://github.com/huggingface/optimum-intel/pull/150/commits/18dba544fa0a0b42270de282307b86e6b6c010a1

vuiseng9: for easy reference/readibility, go to https://moon-ci-docs.huggingface.co/docs/optimum-intel/pr_150/en/optimization_ov after next workflow completes

optimum/intel/openvino/training_args.py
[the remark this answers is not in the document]
vuiseng9: Thanks for the feedback, the current design of `OVTrainingArguments` only extends `TrainingArguments` with distillation hyperparameters (weightage and temperature) since users are required to handle teacher loading and feeding it as an argument ti `OVTrainer`. Similarly, `nncf_compression_config` is removed as user is expected to set `OVConfig` to `OVTrainer`.

AlexKoff88 (approved):

helena-intel (approved):
Thanks @vuiseng9 ! This is a great addition. Thanks for your patience in figuring out some of the issues we ran into with the examples and for the fix of the issue with reshaping BERT models. Really happy we have this in optimum-intel now.

docs/source/optimization_ov.mdx
```diff
@@ -112,18 +118,98 @@ metrics = trainer.evaluate()
...
 
-After applying quantization on our model, we can then easily load it with our [`OVModelForXxx`](https://huggingface.co/docs/optimum/intel_inference#optimum-inference-with-openvino) classes and perform inference with OpenVINO Runtime using the Transformers [pipelines](https://huggingface.co/docs/transformers/main/en/main_classes/pipelines).
+After applying quantization on our model, we can then easily load it with our `OVModelFor<Task>` classes and perform inference with OpenVINO Runtime using the Transformers [pipelines](https://huggingface.co/docs/transformers/main/en/main_classes/pipelines).
 
 ```python
 from transformers import pipeline
 from optimum.intel.openvino import OVModelForSequenceClassification
 
-ov_model = OVModelForSequenceClassification.from_pretrained(save_dir)
+ov_model = OVModelForSequenceClassification.from_pretrained(model_dir)
```
echarlaix, line 206: I would keep `save_dir` for users to load the quantized model obtained through by either PTQ or QAT, or use a valid model id such as  `"helenai/distilbert-base-uncased-finetuned-sst-2-english-ov-int8"`, as currently `model_dir` is not defined anywhere

examples/openvino/question-answering/README.md
```diff
@@ -44,4 +44,33 @@ python run_qa.py \
   --max_seq_length 384 \
   --doc_stride 128 \
   --output_dir /tmp/outputs_squad/
-```
\ No newline at end of file
+```
+
+### Joint Pruning, Quantization and Distillation (JPQD) for BERT on SQuAD1.0
+`OVTrainer` also provides an advanced optimization workflow through the NNCF when Transformer model can be structurally pruned along with 8-bit quantization and distillation. Below is an example which demonstrates how to jointly prune, quantize BERT-base for SQuAD 1.0 using NNCF config `--nncf_compression_config` and distill from BERT-large teacher. This example closely resembles the movement sparsification work of [Lagunas et al., 2021, Block Pruning For Faster Transformers](https://arxiv.org/pdf/2109.04838.pdf). This example takes about 12 hours with a single V100 GPU and the output model attains about 40% structured sparsity (denominated by linear layers only). The joint pruned and quantized OpenVINO IR provides >60% more throughput over quantize-only model on AWS EC2 instance (c6i.32xlarge). Attributed to large teacher distillation, the sparse-quantized model has F1 improved to >89%.
```
echarlaix, line 50: I would prefer to add a link towards an existing benchmark, removing this part in the meantime.
```suggestion
`OVTrainer` also provides an advanced optimization workflow through the NNCF when Transformer model can be structurally pruned along with 8-bit quantization and distillation. Below is an example which demonstrates how to jointly prune, quantize BERT-base for SQuAD 1.0 using NNCF config `--nncf_compression_config` and distill from BERT-large teacher. This example closely resembles the movement sparsification work of [Lagunas et al., 2021, Block Pruning For Faster Transformers](https://arxiv.org/pdf/2109.04838.pdf). This example takes about 12 hours with a single V100 GPU and the output model attains about 40% structured sparsity (denominated by linear layers only). 
```
Also concerning the  40% structured sparsity,I think we should make it clearer that it's on the model without taking into account the embeddings

examples/openvino/text-classification/README.md
```diff
@@ -0,0 +1,80 @@
...
+    --logging_steps 10 \
+    --evaluation_strategy steps \
+    --eval_steps 250 \
+    --save_strategy epoch \
+    --save_total_limit 3 \
+    --fp16 \
+    --seed 42
+```
+
+On a single V100 GPU, this script should run in ~1.8 hours, and yield accuracy of **92.2%** with ~40% pruned weights in the Transformer blocks. on On AWS EC2 instance (c6i.32xlarge), the sparse-quantized IR with OpenVINO runtime provides 2.5X throughput of model optimized with quantization alone.
```
echarlaix, line 80: Same comment, I would prefer a link towards an existing benchmark

setup.py
```diff
@@ -26,8 +26,11 @@
 
 EXTRAS_REQUIRE = {
     "neural-compressor": ["neural-compressor>=2.0.0", "onnx", "onnxruntime"],
-    "openvino": ["openvino>=2022.3.0", "onnx", "onnxruntime"],
-    "nncf": ["nncf"],
+    "openvino": ["openvino>=2022.3.0", "transformers>=4.20.0", "onnx", "onnxruntime"],
```
echarlaix, line 29:
```suggestion
    "openvino": ["openvino>=2022.3.0", "onnx", "onnxruntime"],
```

echarlaix (approved):
Awesome work @vuiseng9, thanks a lot for your contribution ! 
Could you add your addition to the documentation in a an other PR in order for us to merge it after `optimum-intel` next release ?

setup.py
vuiseng9: revised in [489022f](https://github.com/huggingface/optimum-intel/pull/150/commits/489022fcf259866a8b9214f385f98ccef58b4beb)

optimum/intel/openvino/trainer.py
vuiseng9: Updates: IR generated by OVTrainer is dynamic now. It is functional for prediction and the model input can be reshaped to static/dynamic shape. Tests on IR for reshape-ability and its equivalence to torch output have been added [6a93a79](https://github.com/huggingface/optimum-intel/pull/150/commits/6a93a79325656fcc3307b3d8759630238d058b87) and [9d30755](https://github.com/huggingface/optimum-intel/pull/150/commits/9d30755d0121e434d258142ade0e4ca71febbcb7)

optimum/intel/openvino/trainer.py
vuiseng9: JPQD flow exporter now reuse the main branch export flow.

tests/openvino/test_quantization.py
vuiseng9: @echarlaix @AlexKoff88 since we have dynamic IR now, we also cover this test now. See https://github.com/huggingface/optimum-intel/pull/150/commits/6a93a79325656fcc3307b3d8759630238d058b87 .

docs/source/optimization_ov.mdx
vuiseng9: @echarlaix Thanks for the suggestion, revised per your suggestion [16da776](https://github.com/huggingface/optimum-intel/pull/150/commits/16da7763b1a725f0389165395ca7cdd5e0b207d8)

examples/openvino/question-answering/README.md
vuiseng9: We do mention it "about 40% structured sparsity _(denominated by linear layers only)_"

vuiseng9:
@echarlaix thank you for the reviews, we appreciate your attention! We've attended to most of your comments. The only open is regarding the performance description in the readme of the examples. As of now, we don't have public links to performance numbers. But we do think having a description about efficiency improvement is important because it informs the users about the value of a costlier pipeline than quantization alone. Hope you are okay with this. We can revisit the performance links once they become available to public. Could you kindly approve and merge?

PR #160, opened 2023-01-04
Fix openvino compatibility with transformers 

echarlaix:


AlexKoff88 (approved):

helena-intel:
Thanks Ella! This fixed the issue I had with transformers installed from source. Should we specify the minimum optimum version as 1.6.* in setup.py? I had an issue with optimum 1.5 now, but 1.6 worked fine.

PR #162, opened 2023-01-04
Add small tests for optimum/transformer update testing

helena-intel:
Add basic functionality check for optimum[openvino] integration with released and source versions of optimum and transformers. This will allow us to find issues with updated dependencies earlier. 
The test only checks if no exceptions occur on loading, saving and doing inference (with pipeline) on the most common model types. 
The test is configured to run daily. With initial testing I saw some differences depending on Python version, so the test checks with both Python 3.7 and 3.10. 
The test is excluded from tests that are run on PR (except for PRs that modify the test itself) because this test overlaps with them.

TODO (in later PR): Add ViT model type.

### DIFF
--- a/.github/workflows/test_openvino_basic.yml
+++ b/.github/workflows/test_openvino_basic.yml
@@ -22,6 +22,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # Testing lower and upper bound of supported Python versions
+        # This also ensures that the test fails if dependencies break for Python 3.7
         python-version: ["3.7", "3.10"]
         transformers: ['transformers', 'git+https://github.com/huggingface/transformers.git']
         optimum: ['optimum', 'git+https://github.com/huggingface/optimum.git']
--- a/.github/workflows/test_openvino_basic.yml
+++ b/.github/workflows/test_openvino_basic.yml
@@ -4,6 +4,14 @@ on:
   workflow_dispatch:
   schedule:
     - cron:  '41 1 * * *'  # run every day at 1:41
+  push:
+    paths:
+    - 'tests/openvino/test_modeling_basic.py'
+    - '.github/workflows/test_openvino_basic.yml'
+  pull_request:
+    paths:
+    - 'tests/openvino/test_modeling_basic.py'
+    - '.github/workflows/test_openvino_basic.yml'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
@@ -26,7 +34,7 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
-    
+
     - name: Install dependencies
       run: |
         # Install openvino manually to prevent dependency conflicts when .[openvino] pins
--- a/.github/workflows/test_openvino_basic.yml
+++ b/.github/workflows/test_openvino_basic.yml
@@ -1,0 +1,41 @@
+name: OpenVINO - Basic Test
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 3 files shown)


.github/workflows/test_openvino_basic.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.7", "3.10"]
+        transformers: ['transformers', 'git+https://github.com/huggingface/transformers.git']
+        optimum: ['optimum', 'git+https://github.com/huggingface/optimum.git']
+
+    runs-on: ubuntu-20.04
```
AlexKoff88, line 29: Python 3.8 comes by default with ubuntu-20.04. I wonder why we are going to check `3.7` and `3.10`?

helena-intel: I find it useful to have lower and upper bound tests. The main tests use 3.8 and 3.9, so having 3.7 and 3.10 here would be complimentary to that (we could probably also upgrade the main test to use Python 3.10, since that's the default in Ubuntu 22.04 and now supported by OpenVINO). It is unlikely that tests would fail in 3.8 but work fine in both 3.7 and 3.10. I can add 3.8 here, I didn't include it because it seemed a bit over-enthusiastic to test 3 python versions and I would like to test 3.7 too. One reason for the test to fail is that dependencies of dependencies suddenly drop support for lower Python versions in their new releases, and having the lowest supported version in the automatic tests alerts us to that.

AlexKoff88: Ok, now it makes sense. Placing a comment about that could be helpful.

AlexKoff88 (approved):

.github/workflows/test_openvino_basic.yml
helena-intel: Thanks, I added a comment.

tests/openvino/test_modeling_basic.py
```diff
@@ -0,0 +1,85 @@
...
+        model_class = eval(model_class_str)
+        model = model_class.from_pretrained(model_id, from_transformers=True)
+        model.save_pretrained(f"{model_id}_ov")
+        model = model_class.from_pretrained(f"{model_id}_ov")
+
+        input_text = ["hello world"]
+        if model_class_str == "OVModelForQuestionAnswering":
+            input_text *= 2
+        elif model_class_str == "OVModelForMaskedLM":
+            input_text[0] = f"{input_text} {tokenizer.mask_token}"
```
echarlaix, line 58:
```suggestion
            input_text[0] = f"{input_text[0]} {tokenizer.mask_token}"
```

tests/openvino/test_modeling_basic.py
```diff
@@ -0,0 +1,85 @@
...
+    "hf-internal-testing/tiny-random-t5": "OVModelForSeq2SeqLM",
+    "hf-internal-testing/tiny-random-bart": "OVModelForSeq2SeqLM",
+}
+
+TASKS = {
+    "OVModelForMaskedLM": "fill-mask",
+    "OVModelForSequenceClassification": "text-classification",
+    "OVModelForQuestionAnswering": "question-answering",
+    "OVModelForCausalLM": "text-generation",
+    "OVModelForSeq2SeqLM": "text-generation",
```
echarlaix, line 37:
```suggestion
    "OVModelForSeq2SeqLM": "text2text-generation",
```

echarlaix (approved):
LGTM, thanks a lot for this addition @helena-intel

PR #163, opened 2023-01-06
add jit mode for ipex to improve inference performance for pipeline

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/intel/ipex/inference.py
+++ b/optimum/intel/ipex/inference.py
@@ -48,9 +48,15 @@ class _ModelFallbackWrapper:
     """,
 )
 class inference_mode:
-    __slots__ = ("_model", "_dtype", "_graph_mode", "_verbose", "_original")
+    __slots__ = ("_model", "_dtype", "_graph_mode", "_verbose", "_original", "_jit")
 
-    def __init__(self, model: Union[nn.Module, Pipeline], dtype: torch.dtype = torch.float32, verbose: bool = False):
+    def __init__(
+        self,
+        model: Union[nn.Module, Pipeline],
+        dtype: torch.dtype = torch.float32,
+        verbose: bool = False,
+        jit: bool = False,
+    ):
         """
         Args:
             model (`torch.nn.Module` or `transformers.Pipeline`):
@@ -71,6 +77,7 @@ class inference_mode:
         self._dtype = dtype
         self._graph_mode = False  # Let's keep for future use when it doesn't hang anymore
         self._original = None
+        self._jit = jit
 
     def __enter__(self):
         if self._model.framework == "pt":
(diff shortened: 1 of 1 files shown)


sywangyi:
@echarlaix @yao-matrix @jianan-gu please help review

sywangyi:
@mfuntowicz

sywangyi:
@juliensimon

sywangyi:
@kding1

juliensimon:
@mfuntowicz @echarlaix Can you please take a look? This seems to have a nice impact on inference performance, and if so, I'd like to use it for the upcoming inference post. Thank you.

optimum/intel/ipex/inference.py
```diff
@@ -87,7 +90,19 @@ def __enter__(self):
...
 
                             # Enable automatic mixed precision (AMP) if we are going to target `bfloat16`
-                            with torch.cpu.amp.autocast(enabled=(self._dtype == torch.bfloat16)):
+                            with torch.cpu.amp.autocast(enabled=(self._dtype == torch.bfloat16)), torch.no_grad():
+                                if self._model.tokenizer is not None:
+                                    try:
+                                        jit_inputs = []
+                                        dummy_input = self._model.tokenizer("")
+                                        for key in dummy_input:
+                                            jit_inputs.append(torch.ones((1, len(dummy_input[key])), dtype=torch.long))
```
echarlaix, line 99: Dummy inputs can also be generated using the optimum.exporters appropriated [configuration](https://github.com/huggingface/optimum/blob/v1.6.1/optimum/exporters/onnx/__main__.py#L77-#L78) 
```
dummy_inputs = config.generate_dummy_inputs(framework="pt")
```
It will be extended in the future but would would require us to add `onnx` as a new dependecy until then

echarlaix (commented):
Thanks for this addition @sywangyi ! I'll let @mfuntowicz approve this PR before merging

mfuntowicz:
The JIT property seems to enable some additional speedup for some models but I'm not sure it would be compatible with all our usages.

For instance, when working with text generation, the optional nature of `past_keys_values` would introduce some burden to make it right.

Still, I think it's an interesting feature, would it be possible to move the ability to JIT a model to the `inference_mode` constructor to let the user decide? Also, jitting might not be the default path for PyTorch users along with introducing some "compilation" time for every new shape provided.

If we include this as a customization of `inference_mode` I would be ok to merge 👍🏻

mfuntowicz (approved):
LGTM! 👍🏻

juliensimon:
Thanks, I'll give it a go.

juliensimon:
Happy to confirm a spectacular improvement. Latency is down another 20-25%. Well done.

PR #164, opened 2023-01-09
Temporarily disable CACHE_DIR on CPU

helena-intel:
There is an issue with OpenVINO 2022.3 that results in a RuntimeError when loading some INT8 models from cache on CPU (I encountered it with BERT QA models). While that issue is being resolved, this PR temporarily disables CACHE_DIR when the device is not GPU. Caching is mostly useful for GPU (since model loading takes longer on GPU), so disabling it for CPU should not significantly impact performance. 

I also tried to implement CACHE_DIR with `core.set_property(device_name="GPU")`, which would be cleaner and more flexible, but I ran into different issues with that. Currently `core` is also set as a global object and if we're going to use `set_property` it would be best to make `core` a property of the `OVBaseModel`. I think that's not necessary, the workaround in this PR works and it's expected to be temporary. Open to other opinions/suggestions.

I deleted `cache_dir.mkdir(parents=True, exist_ok=True)` because OpenVINO creates the cache dir if it doesn't exist (with parents if necessary).

I added a QA quantization tests that would have caught this issue.

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -82,13 +82,11 @@ class OVBaseModel(OptimizedModel):
         self.ov_config = {"PERFORMANCE_HINT": "LATENCY"}
         self.preprocessors = kwargs.get("preprocessors", [])
         enable_compilation = kwargs.get("compile", True)
-        cache_dir = Path(self.model_save_dir).joinpath("model_cache")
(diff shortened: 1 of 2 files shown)


AlexKoff88 (approved):

echarlaix (approved):
Thank you for the fix @helena-intel

PR #165, opened 2023-01-11
Option to load with CPU with torch

bconsolvo:
When I go to use the `from_pretrained` function, it automatically tries to load it with CUDA. But CUDA is not available on a CPU. I set up an option so that you can load the `state_dict` with CPU only.

### DIFF
--- a/optimum/intel/neural_compressor/quantization.py
+++ b/optimum/intel/neural_compressor/quantization.py
@@ -414,6 +414,7 @@ class INCModel:
         model_name_or_path: str,
         inc_config: Union[IncOptimizedConfig, str] = None,
         q_model_name: Optional[str] = None,
+        cpu_flag: bool = False,
         **kwargs
     ) -> torch.nn.Module:
         """
@@ -535,7 +536,11 @@ class INCModel:
             raise ValueError("INC IPEX is currently not supported")
 
         # Load the state dictionary of the model to verify whether the model is quantized or not
-        state_dict = torch.load(state_dict_path)
+        if not cpu_flag: 
+            state_dict = torch.load(state_dict_path)
+        elif cpu_flag:
+            state_dict = torch.load(state_dict_path,map_location='cpu')
+        
         if "best_configure" not in state_dict:
             return model
 


optimum/intel/neural_compressor/quantization.py
```diff
@@ -535,7 +536,11 @@ def from_pretrained(
             raise ValueError("INC IPEX is currently not supported")
 
         # Load the state dictionary of the model to verify whether the model is quantized or not
-        state_dict = torch.load(state_dict_path)
+        if not cpu_flag: 
+            state_dict = torch.load(state_dict_path)
+        elif cpu_flag:
+            state_dict = torch.load(state_dict_path,map_location='cpu')
+
```
echarlaix, line 543: What do you think about always loading the model state dict on the cpu by default ? Would be most likely the most common inference usage + could avoid potential out of memory error (as [in](https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L2062) or [in](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L415) and discussed [in](https://discuss.pytorch.org/t/out-of-memory-error-when-resume-training-even-though-my-gpu-is-empty/30757) and [in](https://discuss.pytorch.org/t/cuda-error-out-of-memory-when-load-models/38011))

echarlaix (commented):
Thanks for the fix @bconsolvo !! Also before we merge, can you run `make style` to to fix the code quality issues ?

optimum/intel/neural_compressor/quantization.py
bconsolvo: Yes, that makes sense to me, to load with CPU by default, and to have the option to load with GPU.

bconsolvo: In most cases, I would say using GPU by default would make more sense, but in the case of Optimum Intel, probably most are using CPU by default at this point.

bconsolvo: Yes, agreed. By default, it tries to load it on the GPU, which doesn't make sense.

optimum/intel/neural_compressor/quantization.py
```diff
@@ -535,7 +536,11 @@ def from_pretrained(
             raise ValueError("INC IPEX is currently not supported")
 
         # Load the state dictionary of the model to verify whether the model is quantized or not
-        state_dict = torch.load(state_dict_path)
+        if not cpu_flag: 
+            state_dict = torch.load(state_dict_path)
+        elif cpu_flag:
+            state_dict = torch.load(state_dict_path,map_location='cpu')
```
echarlaix, line 542: I think this would makes sense then :
```suggestion
        state_dict = torch.load(state_dict_path, map_location="cpu")

```

optimum/intel/neural_compressor/quantization.py
echarlaix: Yes always loading and keeping the model on the CPU makes sense then

echarlaix:
Added in https://github.com/huggingface/optimum-intel/pull/249 to integrate it to `optimum-intel` next release, thanks a lot for your contibution @bconsolvo !

PR #168, opened 2023-01-18
Make batch size and sequence length dynamic for causal lm

echarlaix:
In this PR we make the batch size and sequence length dimension dynamic for causal language model

### DIFF
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -473,6 +473,35 @@ class OVModelForCausalLM(OVModel, GenerationMixin):
             # Ensure attention mask is on the same device as the input IDs
             return torch.ones(inputs.shape[:2], dtype=torch.long, device=inputs.device)
 
+    def _reshape(
+        self,
+        model: openvino.runtime.Model,
+        batch_size: int,
+        sequence_length: int,
+        height: int = None,
+        width: int = None,
+    ):
+        if batch_size != -1:
+            logger.warning(
+                f"The batch size was set to {batch_size} during reshape which is not supported for causal language model."
+                "This value is set to -1."
+            )
+            batch_size = -1
+        if sequence_length != -1:
+            logger.warning(
+                f"The sequence length was set to {sequence_length} during reshape, this dimension needs to be dynamic for causal language model."
+                "This value is set to -1."
+            )
+            sequence_length = -1
+
+        return super()._reshape(
+            model=model,
+            batch_size=batch_size,
+            sequence_length=sequence_length,
+            height=height,
+            width=width,
+        )
+
 
 @add_start_docstrings(
     """
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -481,17 +481,9 @@ class OVModelForCausalLM(OVModel, GenerationMixin):
         height: int = None,
         width: int = None,
     ):
-        if batch_size != -1:
-            logger.warning(
-                f"The batch size was set to {batch_size} during reshape which is not supported for causal language model."
-                "This value is set to -1."
-            )
+        if batch_size != -1 or sequence_length != -1:
+            logger.warning("Static shapes are not supported for causal language model.")
             batch_size = -1
-        if sequence_length != -1:
-            logger.warning(
-                f"The sequence length was set to {sequence_length} during reshape, this dimension needs to be dynamic for causal language model."
-                "This value is set to -1."
-            )
             sequence_length = -1
 
         return super()._reshape(


optimum/intel/openvino/modeling.py
```diff
@@ -473,6 +473,35 @@ def _prepare_attention_mask_for_generation(
...
+        self,
+        model: openvino.runtime.Model,
+        batch_size: int,
+        sequence_length: int,
+        height: int = None,
+        width: int = None,
+    ):
+        if batch_size != -1:
+            logger.warning(
+                f"The batch size was set to {batch_size} during reshape which is not supported for causal language model."
```
helena-intel, line 486: Since both dynamic batch and dynamic sequence length are not supported, I'm wondering if it would be clearer to show one message when a user tries to reshape the model to static shapes, that causal language models need dynamic shapes and that reshaping is not supported. What do you think?

echarlaix: I updated the warning message

helena-intel (approved):

Vipitis:
so is Causal modeling (as in text generation) impossible on GPU? since GPU doesn't support dynamic shapes.

helena-intel:
@Vipitis dynamic shapes on GPU will be supported in the next OpenVINO release.

Vipitis:
that is great to hear. Do you have any eta for how soon that will be available, or is there already a working branch I could try? I am working on an evaluation task and need to run some models. Would love to do that locally on my arc card.

helena-intel:
Hi @Vipitis sorry for the delay. The next OpenVINO release is planned for end of May. We also expect to have a prerelease next week (but please note that both dates are not set in stone). If you want to try this earlier, I recommend building OpenVINO from source. See https://github.com/openvinotoolkit/openvino/blob/master/docs/dev/build_linux.md for Linux. 

In short: install system dependencies with the install_dependencies_sh script. Install Python dependencies. For Python, this should work on x86 Linux: 

```
pip install --upgrade pip
pip install --upgrade numpy cython patchelf wheel setuptools
```

If the default Python on your system is >= 3.7, <= 3.10 you can then do, from the directory where you cloned OpenVINO:
```
mkdir build && cd build
cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_PYTHON=ON -DENABLE_WHEEL=ON  ..
make --jobs=$(nproc --all)
```

Your build folder will then contain a `wheels` folder with openvino and openvino-dev wheels which you can install with `pip install *whl` from that folder.

If you have a different default Python or are on another platform, see the documentation: https://github.com/openvinotoolkit/openvino/blob/master/docs/dev/build.md

Please let me know if you run into issues, or open an issue in the OpenVINO repo.

Vipitis:
Thanks for the help @helena-intel - I was able to get the newest version easily via steps here: https://huggingface.co/docs/optimum/installation

However, it doesn't seem useable to me at the moment. My hope was to it with `evaluate`, but found I that is is incredibly slow with a `transformers.pipeline`. Even after calling compile(), first inference is really slow - then subsequent inference is faster(still seems slower than CPU, but I haven't tested that thoroughly, But if you generate more tokens for example - first inference is slow again. There is also a reproducibility issue once you move the model to GPU.

helena-intel:
Hi @Vipitis , that means you'll have the latest version of Optimum, but to get the most recent build of OpenVINO you will need to compile from source with the steps I mentioned. Or you can wait for the next pre-release. Even though dynamic shapes on GPU do not raise errors anymore in the current OpenVINO version, this is preview-support and there are known issues. I have not yet tested CausalLM models with dynamic shapes on GPU yet, but in general, support for dynamic shapes with NLP models on GPU has much improved over the past months. 

Dynamic shapes on GPU leverages OpenVINO's caching feature. This is automatically enabled in optimum-intel. Caching is done for each input shape. Depending on the model and data, you may see improved inference after the cache has been filled (so for example after you have run inference over the dataset already). 
To leverage this caching, you should save the model to disk and not use the `export=True` method to convert the model on-the-fly (Do `model.save_pretrained("local_model_path")` and then load the model with `OVModelForCausalLM.from_pretrained("local_model_path")`

There's also caching of past_key_values for CausalLM models. Currently OVModelForCausalLM does not support that yet, but that will be supported very soon. See this PR: https://github.com/huggingface/optimum-intel/pull/272 That should also speed up inference.

Could you share what model you use for testing, so we can also test it from our side?

Vipitis:
today I had `openvino==2023.0.0.dev20230217, optimum==1.7.4.dev0, optimum-intel==1.7.3`. 
My experimental setup as shown in the screenshot, includes execute times, running all the cells in order exactly once.
Code_lCYxV6AcYI
Also shows the difference in output when the model is moved to GPU.

If I let the model autoregressively generate all the way to it's context size limit, it should have cached all the possible input shapes in between, right? Really unsure about saving models as my goal is to implement intel-gpu inference for an [EvaluationSuite](https://huggingface.co/spaces/Vipitis/ShaderEval) I am working on. Also `evaluate` has a bunch of development going on as well

helena-intel:
Hi @Vipitis ! We have a new OpenVINO [prerelease](https://pypi.org/project/openvino/2023.0.0.dev20230407/). This fixes the accuracy issue with dynamic shapes on GPU. Unfortunately there is an issue with GPT-2 on GPU and it results in an error on inference with default settings.  On GPU, inference is in FP16 by default, which causes the issue for this model. This will be fixed, and for now you can work around the error by forcing inference in FP32 by adding an `ov_config={"INFERENCE_PRECISION_HINT": "f32"}` argument to `OVModelForCausalLM.from_pretrained()`. This will be slower than FP16 inference.

Your use case is definitely expected to work! But at the moment, dynamic shapes are not officially supported yet on GPU, and there are still known issues, as you noticed. 

As for caching, when setting `export=True` a temporary directory is used, which is removed after the script finishes. So for GPU dynamic shapes, it is recommended to either save the model to disk, or load an OpenVINO model from the Hugging Face hub. Both methods will create a cache on disk that will be re-used every time you run the script. I uploaded [gpt2](https://huggingface.co/helenai/gpt2-ov) and [bloom-560m](https://huggingface.co/helenai/bigscience-bloom-560m-ov) to the hub. You can use these, or upload your own models. I'm also happy to upload other models. Loading an OpenVINO model from directly (from the hub or a local directory) is quite a bit faster too.

The PR to add caching for CausalLM models was merged. It speeds up inference quite a bit. If you haven't tried it yet: `python -m pip install git+https://github.com/huggingface/optimum-intel.git`. It is still not fast on GPU though (iGPU is much slower than CPU for me). We will look into this.

Are you primarily interested in CausalLM models? For most models I do see faster inference on iGPU than CPU (on my 11th gen Core laptop).

Vipitis:
he @helena-intel I have already had my time to play around with the prerelease. I didn't expect there be a specific issue to GPT2, I will try some other models to see if they work - perhaps especially some that aren't completely based on GPT2. It's a project around code intelligence, so languge models that are trained on code and can generate. Right now I am looking at just CausalLM, but there is also the Fill-in-the-middle approach which could be really interesting. One of the models I would like to try is [Santacoder](https://huggingface.co/bigcode/santacoder) as well as various fine tuned versions of it.

I do now believe that my CPU is causing the massive slowdown I am experiencing. It seems to be stuck at 100% usage while running inference on GPU (and all the associated caching), it's a Skylake i5 and is well overdue for an upgrade. I do have a TGL laptop as well and could compare CPU, iGPU (and the nvidia dGPU in it) to see if there is a major issue on my home system. But I would really love to use the A750 with all it's vram and 250W to chew through my 10k examples in the eval set.

PR #185, opened 2023-02-06
Standardize OpenVINO inference and quantization with INC and ORT 

echarlaix:


AlexKoff88:
I am ok with the changes, but just want to note that this breaks the existing API. If to consider that we already have users of the PTQ workflow, these changes will break their scripts.

AlexKoff88 (approved):

optimum/intel/openvino/quantization.py
```diff
@@ -220,18 +229,20 @@ def _onnx_export(
...
+            self.task = HfApi().model_info(self.model.config._name_or_path).pipeline_tag
+            if self.task in ["sentiment-analysis", "text-classification", "zero-shot-classification"]:
+                self.task = "sequence-classification"
+            elif self.task in ["feature-extraction", "fill-mask"]:
+                self.task = "default"
+            elif self.task == "text-generation":
+                self.task = "causal-lm"
+            elif self.task is None:
+                raise ValueError(
+                    "The task defining the model topology could not be extracted and needs to be specified for the ONNX export."
```
helena-intel, line 243:
```suggestion
                    "The task defining the model topology could not be extracted and needs to be specified for the ONNX export. For example: `OVQuantizer.from_pretrained(model, task='text-classification')`"
```

No preferences on specific phrasing, but I think it would be useful to give an example of how to specify this task.

helena-intel:
> I am ok with the changes, but just want to note that this breaks the existing API. If to consider that we already have users of the PTQ workflow, these changes will break their scripts.

I was happy to see deprecation warnings. I hope they can be kept for a while to alert users to this change.

echarlaix:
> > I am ok with the changes, but just want to note that this breaks the existing API. If to consider that we already have users of the PTQ workflow, these changes will break their scripts.
> 
> I was happy to see deprecation warnings. I hope they can be kept for a while to alert users to this change.

We can keep the deprecation warnings until at least v1.7 to give time to users to adapt their code. We will also only integrate this modifications in the documentation after the next `optimum-intel` release to avoid any confusion

helena-intel:
> [Temporarily add onnxruntime dependancy required by optimum exporters]

Is this really necessary?

echarlaix:
> > [Temporarily add onnxruntime dependancy required by optimum exporters]
> 
> Is this really necessary?

This is resulting from a recent change done in `optimum.exporters.onnx` (https://github.com/huggingface/optimum/pull/731 : https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/base.py#L308) which verifies and fixes the dynamic axis during the ONNX export (some axis are being set as dynamic during the export when they are static). We plan to make it optional in the future

PR #187, opened 2023-02-07
Fix past key values usage

fxmarty:
transformers 4.26.0 broke the usage of past key and values in Optimum:

https://github.com/huggingface/transformers/blob/v4.25.1/src/transformers/models/t5/modeling_t5.py#L1703
vs
https://github.com/huggingface/transformers/blob/v4.26.0/src/transformers/models/t5/modeling_t5.py#L1715

This PR fixes the issue and adds the corresponding test.

Related: https://github.com/huggingface/optimum/pull/756 https://github.com/huggingface/optimum/issues/754

### DIFF
--- a/optimum/intel/openvino/modeling_seq2seq.py
+++ b/optimum/intel/openvino/modeling_seq2seq.py
@@ -225,7 +225,7 @@ class OVModelForSeq2SeqLM(OVBaseModelForSeq2SeqLM, GenerationMixin):
     def prepare_inputs_for_generation(
         self,
         input_ids,
-        past=None,
+        past_key_values=None,
         attention_mask=None,
         head_mask=None,
         decoder_head_mask=None,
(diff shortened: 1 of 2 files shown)


optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -236,7 +236,7 @@ def prepare_inputs_for_generation(
     ) -> Dict:
         return {
             "decoder_input_ids": input_ids,
-            "past_key_values": past,
+            "past_key_values": past_key_values,
```
echarlaix, line 239: To support `transformers<4.26`, can you add something like:
```suggestion
            "past_key_values": past_key_values or kwargs.get("past", None),
```

echarlaix (commented):
Thanks a lot for the fix @fxmarty !

echarlaix (approved):

PR #195, opened 2023-02-21
Add stable diffusion OpenVINO pipeline

echarlaix:
In this PR we add a stable diffusion pipeline enabling OpenVINO Runtime. To export the model to the OpenVINO IR `export` must be set to `True`. Needs https://github.com/huggingface/optimum/pull/786 to be merged.

```python
from optimum.intel.openvino import OVStableDiffusionPipeline

model_id = "runwayml/stable-diffusion-v1-5"
stable_diffusion = OVStableDiffusionPipeline.from_pretrained(model_id, export=True)
prompt = "sailing ship in storm by Leonardo da Vinci"
images = stable_diffusion(prompt).images
```
Closes https://github.com/huggingface/diffusers/issues/1760 and https://github.com/huggingface/diffusers/issues/2202

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -34,11 +34,6 @@ from optimum.exporters import TasksManager
 from optimum.exporters.onnx import export_models, get_stable_diffusion_models_for_export
 from optimum.pipelines.diffusers.pipeline_stable_diffusion import StableDiffusionPipelineMixin
 from optimum.utils import (
-    DIFFUSION_MODEL_TEXT_ENCODER_SUBFOLDER,
-    DIFFUSION_MODEL_UNET_SUBFOLDER,
-    DIFFUSION_MODEL_VAE_DECODER_SUBFOLDER,
-)
-from optimum.utils.constant import (
     CONFIG_NAME,
     DIFFUSION_MODEL_TEXT_ENCODER_SUBFOLDER,
     DIFFUSION_MODEL_UNET_SUBFOLDER,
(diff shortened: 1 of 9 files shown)


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -0,0 +1,462 @@
...
+
+        self._create_inference_request()
+
+        inputs = {
+            "sample": sample,
+            "timestep": timestep,
+            "encoder_hidden_states": encoder_hidden_states,
+        }
+
+        outputs = self.request.infer(inputs)
```
AlexKoff88, line 449: This was accelerated in the latest version of OpenVINO via non-copy use of passed numpy data. Hope, it will result in some speedup when moving from openvino==2022.3 to openvino=2023.0.*.

AlexKoff88 (approved):

helena-intel:
Thanks @echarlaix it is amazing to be able to run stable diffusion on my laptop CPU with only a few lines of code! 

I have some small questions:
- Could you clarify the `export` parameter? It seems similar to what `from_transformers` does for OVModels and I'm wondering if there is a difference and/or if it would be possible to use `from_transformers` here too for consistency.
- Thanks for explicitly adding easy-to-use static shapes support! Do you think we could consider removing some duplication so that if a model has been reshaped to static shapes with `stable_diffusion.reshape( batch_size=batch_size, height=height, width=width, num_images_per_prompt=num_images_per_prompt)`  and you call `stable_diffusion(prompt)` the pipeline will pick up the shapes from the model? When you explicitly reshape the model, it seems you have to use these shapes in the pipeline? It would be a slight deviation from other stable diffusion pipelines, but it would make code with static shapes easier to read/write, and of course explicitly setting height, width etc in the pipeline would still work too for users who are familiar with that.

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -0,0 +1,462 @@
...
+            inputs.get_any_name(): OV_TO_NP_TYPE[inputs.get_element_type().get_type_name()]
+            for inputs in self.model.inputs
+        }
+        self.ov_config = ov_config or self.parent_model.ov_config
+        self.request = None
+
+    def _create_inference_request(self):
+        if self.request is None:
+            logger.info("Compiling the encoder and creating the inference request ...")
+            compiled_model = core.compile_model(self.model, self.device, self.ov_config)
```
helena-intel, line 418: This does not yet include model caching, which should improve model compiling time on GPU a lot. See modeling_base.py: https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_base.py#L304

echarlaix:
> Could you clarify the export parameter? It seems similar to what from_transformers does for OVModels and I'm wondering if there is a difference and/or if it would be possible to use from_transformers here too for consistency.

My suggestion is to rename the `from_transformers` parameter to `export` which I think should be clearer (also `from_transformers` could be confusing when exporting `diffusers` models) for all `OVModels` (while keeping the `from_transformers` support until `optimum-intelv1.8`)

> Thanks for explicitly adding easy-to-use static shapes support! Do you think we could consider removing some duplication so that if a model has been reshaped to static shapes with `stable_diffusion.reshape( batch_size=batch_size, height=height, width=width, num_images_per_prompt=num_images_per_prompt)`  and you call `stable_diffusion(prompt)` the pipeline will pick up the shapes from the model? When you explicitly reshape the model, it seems you have to use these shapes in the pipeline? It would be a slight deviation from other stable diffusion pipelines, but it would make code with static shapes easier to read/write, and of course explicitly setting height, width etc in the pipeline would still work too for users who are familiar with that.

I think modifying the pipeline behavior is perilous as the user wouldn't necessarily know about it (even in the case where we add a warning). Something that we can do is raise an error when the shapes are not matching and showcase example of static reshaping in the documentation what do you think ?

optimum/intel/openvino/modeling_diffusion.py
echarlaix: Added in [`09be7f9`](https://github.com/huggingface/optimum-intel/pull/195/commits/09be7f97c5c38ff4927f59e8e1e68b30f34c8523)

helena-intel:
> My suggestion is to rename the `from_transformers` parameter to `export` which I think should be clearer (also `from_transformers` could be confusing when exporting `diffusers` models) for all `OVModels` (while keeping the `from_transformers` support until `optimum-intelv1.8`)

That sounds good!

> I think modifying the pipeline behavior is perilous as the user wouldn't necessarily know about it (even in the case where we add a warning). Something that we can do is raise an error when the shapes are not matching and showcase example of static reshaping in the documentation what do you think ?

I think it can also be the other way around. When I reshaped the model I was surprised the inference didn't work out of the box. I guess I'm not seeing why you would explicitly reshape the model, but then want to do inference on different shapes? And since static shapes speed up inference, I think it doesn't make much sense not to to use them - but your simple 4 line example is so much friendlier than the 10 lines you now need with static shapes (https://huggingface.co/helenai/runwayml-stable-diffusion-v1-5-ov-fp32)

Not a dealbreaker though, and we can also discuss this later.

helena-intel (approved):

PR #198, opened 2023-02-24
Added explicit cast to numpy to avoid issues with pre-release

AlexKoff88:
This PR fixed issue of incompatibility with the new version of OpenVINO. InferRequest.infer() method does not accept torch.Tensor. numpy.ndarray is required.

Generative models were fixed in the past.

### DIFF
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -157,13 +157,13 @@ class OVModelForSequenceClassification(OVModel):
         self.compile()
 
         inputs = {
-            "input_ids": input_ids,
-            "attention_mask": attention_mask,
+            "input_ids": input_ids.numpy(),
+            "attention_mask": attention_mask.numpy(),
         }
 
         # Add the token_type_ids when needed
         if "token_type_ids" in self.input_names:
-            inputs["token_type_ids"] = token_type_ids
+            inputs["token_type_ids"] = token_type_ids.numpy()
 
         # Run inference
         outputs = self.request.infer(inputs)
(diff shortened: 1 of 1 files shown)


helena-intel:
Thanks for the quick fix @AlexKoff88 ! I hope this can be fixed from the OpenVINO side before release because we just added official support for using numpy-compatible tensors in 2022.

optimum/intel/openvino/modeling.py
```diff
@@ -157,13 +157,13 @@ def forward(
         self.compile()
 
         inputs = {
-            "input_ids": input_ids,
-            "attention_mask": attention_mask,
+            "input_ids": input_ids.numpy(),
```
helena-intel, line 160: It would be safer/more flexible to cast to numpy with `np.array(input_ids)` (and other tensors) because then inference will also work if the input data is already in Numpy format. It is not obvious that you would need to use PyTorch tensors for OpenVINO inference.

echarlaix (approved):

optimum/intel/openvino/modeling.py
AlexKoff88: I agree. Changed it for generative models as well.

setup.py
```diff
@@ -26,10 +26,10 @@
 
 EXTRAS_REQUIRE = {
     "neural-compressor": ["neural-compressor>=2.0.0", "onnx", "onnxruntime"],
-    "openvino": ["openvino>=2022.3.0", "onnx", "onnxruntime"],
+    "openvino": ["openvino==2023.0.0.dev20230217", "onnx", "onnxruntime"],
```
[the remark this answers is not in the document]
AlexKoff88: Last time I tried `>=` for prerelease it didn't work. Let's try again)

PR #199, opened 2023-02-24
Update tests in checking IR outputs with torch outputs

yujiepan-work:
This PR sets a more reasonable value (0.1%) for relative tolerance, when checking if OpenVINO IR outputs equal to torch model outputs.

However, OpenVINO IR is facing output issue with a customized quantization config.  See https://github.com/openvinotoolkit/nncf/issues/1498 .  
This may sometimes cause failed tests. For now we mark to skip tests with custom quantization config.  Once we upgrade to use a higher OpenVINO release, we will delete the skipping.


yujiepan-work:
run test

PR #201, opened 2023-02-24
Enable numpy arrays for openvino model inputs

echarlaix:


helena-intel (approved):
Thank you for making this more robust! We discussed some time ago to make the PyTorch dependency optional in the future and then checking `isinstance(input_ids, torch.Tensor)` would fail, while `np.array(tensor)` works when tensor is either a numpy array or a torch tensor. But that's something we can address then.

AlexKoff88 (approved):

PR #209, opened 2023-03-02
Enable overflow_fix in JPQD examples

helena-intel:
Enable overflow_fix for JPQD examples to restore accuracy on computers without VNNI.
With overflow_fix disabled, accuracy on non-VNNI machines is terrible, model outputs are almost random. 
Enabling overflow_fix has no disadvantages: accuracy is now as expected on both VNNI and non-VNNI machines.

Also removed some comma's from the config file to make it valid JSON.

cc @vuiseng9


AlexKoff88 (approved):

AlexKoff88:
@vuiseng9

echarlaix (approved):
LGTM, thanks @helena-intel !

PR #215, opened 2023-03-04
Enable overflow_fix in ovtrainer tests on machines without VNNI support

yujiepan-work:
This PR addresses the tests on checking torch outputs and IR outputs:

(1) sets a reasonable relative error tolerance value (rtol)  as 0.01% when checking IR and torch outputs;

(2) when doing `test_training` on a machine without VNNI support (e.g., this CI system), will first modify the test config with "overflow_fix" enabled, and then do the following checking. Otherwise the CI computer cannot pass the tests with rtol=0.01%, see https://github.com/openvinotoolkit/nncf/issues/1498#issuecomment-1446052227 

(3) to detect if VNNI is supported, we added a new test requirement: [py-cpuinfo](https://github.com/workhorsy/py-cpuinfo), which is MIT license.

### DIFF
--- a/tests/openvino/test_training.py
+++ b/tests/openvino/test_training.py
@@ -367,7 +367,7 @@ class OVTrainerTrainingTest(unittest.TestCase):
                     torch.allclose(
                         torch.softmax(ovmodel_logits, dim=-1),
                         torch.softmax(torch_logits, dim=-1),
-                        rtol=0.2,
+                        rtol=0.01,
                     )
                 )
 
--- a/tests/openvino/test_training.py
+++ b/tests/openvino/test_training.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import re
 import tempfile
 import unittest
 from copy import deepcopy
@@ -100,9 +101,8 @@ def initialize_movement_sparsifier_parameters_by_sparsity(
                 operand.bias_importance.copy_(bias_init_tensor)
 
 
-def is_avx512_vnni_supported() -> bool:
-    flags = [flag.lower() for flag in cpuinfo.get_cpu_info()["flags"]]
-    return "avx512_vnni" in flags or "avx512vnni" in flags or "avx_vnni" in flags
+def is_avx_vnni_supported() -> bool:
+    return any(re.search("avx.*vnni", flag.lower()) is not None for flag in cpuinfo.get_cpu_info()["flags"])
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


yujiepan-work:
@AlexKoff88   Hi Alexander, this PR about overflow fix on no-vnni-support machines, is ready for review. @vuiseng9

yujiepan-work:
@helena-intel  Hi Helena, this PR is ready for review, it is about checking IR and torch outputs on non-vnni-support machines.

AlexKoff88 (approved):

tests/openvino/test_training.py
```diff
@@ -98,6 +100,11 @@ def initialize_movement_sparsifier_parameters_by_sparsity(
                 operand.bias_importance.copy_(bias_init_tensor)
 
 
+def is_avx512_vnni_supported() -> bool:
+    flags = [flag.lower() for flag in cpuinfo.get_cpu_info()["flags"]]
+    return "avx512_vnni" in flags or "avx512vnni" in flags
```
helena-intel, line 105:
```suggestion
    return "avx512_vnni" in flags or "avx512vnni" in flags or "avx_vnni" in flags
```

Alder Lake and Raptor Lake do not support avx512 but do have VNNI. When I checked on a model with overflow_fix enabled accuracy was fine there.

helena-intel (commented):
Thanks @yujiepan-work ! We discussed setting the default value for overflow_fix to enable for QAT, but we should do more testing. It's good to have this work for tests now. I proposed a small change to support Alder Lake and Raptor Lake.

tests/openvino/test_training.py
vuiseng9: @helena-intel that is a good observation. Client CPUs provide different VNNI setup. I think the keyword should be just "vnni".

yujiepan-work: i have added the "avx_vnni" flag. Since this is intended for tests, i think "avx512_vnni", avx512vnni, avx_vnni should be able to cover

yujiepan-work: Changed to `re.search("avx.*vnni", flag) is not None` to capture possible flags, e.g., avx512_vnni, avx2_vnni, avx_vnni, avx_vnni_int8, x86_avx_vnni

vuiseng9:
@echarlaix I've seen that some of our PRs are merged. Thank you! Just a friendly reminder for this PR. It is ready. Could you kindly review and merged if there is no open? Thanks.

echarlaix (approved):
Thanks a lot for adding this @yujiepan-work

PR #220, opened 2023-03-07
Update documentation

echarlaix:


AlexKoff88 (approved):

docs/source/index.mdx
```diff
@@ -21,30 +21,17 @@ limitations under the License.
...
-```
-
-or to install from source including dependencies:
-
-```bash
-python -m pip install "optimum-intel[extras]"@git+https://github.com/huggingface/optimum-intel.git
-```
-
-where `extras` can be one or more of `neural-compressor`, `openvino`, `nncf`.
+[OpenVINO](https://docs.openvino.ai/latest/index.html) is an open-source toolkit that enables high performance inference capabilities on a variety of Intel [processors](https://docs.openvino.ai/latest/openvino_docs_OV_UG_supported_plugins_Supported_Devices.html). It is supplied with a set of tools to optimize your models with compression techniques such as quantization, pruning and knowledge distillation. Optimum Intel provides a simple interface to optimize your Transformers and Diffusers models, convert them to the OpenVINO Intermediate Representation (IR) format and run inference using OpenVINO Runtime.
```
helena-intel, line 24: I like having, CPUs, GPUs and DL accelerators spelled out. Could we keep that?

docs/source/installation.mdx
```diff
@@ -0,0 +1,41 @@
...
+-->
+
+# Installation
+
+To install the latest release of 🤗 Optimum Intel with the corresponding required dependencies, you can do respectively:
+
+| Accelerator                                                                                                            | Installation                                        |
+|:-----------------------------------------------------------------------------------------------------------------------|:----------------------------------------------------|
+| [Intel Neural Compressor (INC)](https://www.intel.com/content/www/us/en/developer/tools/oneapi/neural-compressor.html) | `python -m pip install "optimum[neural-compressor]"`|
+| [Intel OpenVINO](https://docs.openvino.ai/latest/index.html)                                                           | `python -m pip install "optimum[openvino,nncf]"`    |
```
helena-intel, line 24: It would be useful to mention that to do inference on stable diffusion models, you need to do `pip install optimum-intel[diffusers]` or a note to manually install diffusers if  we do not need a specific version

docs/source/index.mdx
echarlaix: Will add it back !

docs/source/installation.mdx
echarlaix: I will add it in the documentation under the "Stable Diffusion" section

PR #226, opened 2023-03-08
Update OpenVINO notebook requirements

helena-intel:
The notebook used to require source versions of optimum-intel and NNCF, but not anymore!
I also added a cell to the notebook to install the requirements directly in the notebook, to make it possible to run in cloud environments like colab.

### DIFF
--- a/notebooks/openvino/requirements.txt
+++ b/notebooks/openvino/requirements.txt
@@ -1,6 +1,4 @@
-git+https://github.com/openvinotoolkit/nncf.git@develop
-git+https://github.com/huggingface/optimum-intel.git
-openvino>=2022.3.0.dev20221125
+optimum-intel[openvino, nncf]
 datasets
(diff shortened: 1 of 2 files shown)


notebooks/openvino/requirements.txt
```diff
@@ -1,6 +1,4 @@
-git+https://github.com/openvinotoolkit/nncf.git@develop
```
AlexKoff88, line 1: @vuiseng9, was asking about the ability to install NNCF from the source. This is required for audio models as far as I understood. @vuiseng9, can you comment please?

AlexKoff88: Never mind, this comment related to the samples only.

AlexKoff88 (approved):

echarlaix (approved):
Thanks for adding it @helena-intel

PR #227, opened 2023-03-10
Trace model by mock data so it can run generation task.

jiqing-feng:
@sywangyi 
Trace model by mock data so it can run generation task.


optimum/intel/ipex/inference.py
```diff
@@ -19,6 +19,54 @@
 if is_ipex_available():
     import intel_extension_for_pytorch as ipex
 
+def sparse_model_config(model_config):
+    embedding_size = None
+    if hasattr(model_config, 'hidden_size'):
```
echarlaix, line 24: You can use the [`NormalizedConfigManager`](https://github.com/huggingface/optimum/blob/main/optimum/utils/normalized_config.py#L151) to get the the model number of attention heads, hidden size and number of layer. See [here](https://github.com/huggingface/optimum/blob/main/optimum/onnxruntime/optimization.py#L62) and [here ](https://github.com/huggingface/optimum/blob/main/optimum/onnxruntime/optimization.py#L169-#L170),

optimum/intel/ipex/inference.py
```diff
@@ -19,6 +19,54 @@
...
+    elif hasattr(model_config, 'n_embd'):
+        embedding_size = model_config.n_embd
+
+    num_head = None
+    if hasattr(model_config, 'num_attention_heads'):
+        num_head = model_config.num_attention_heads
+    elif hasattr(model_config, 'n_head'):
+        num_head = model_config.n_head
+
+    if embedding_size is None or num_head is None or num_head == 0:
```
echarlaix, line 37:
```suggestion
    if embedding_size is None or num_head is None or num_head is None:

```

optimum/intel/ipex/inference.py
```diff
@@ -19,6 +19,54 @@
...
+        raise ValueError("Check the model config")
+
+    num_embedding_size_per_head = int(embedding_size/num_head)
```
echarlaix, line 40:
```suggestion
    num_embedding_size_per_head = embedding_size // num_head

```

optimum/intel/ipex/inference.py
```diff
@@ -19,6 +19,54 @@
...
+                                            torch.zeros(input_length, num_head, 1, num_embedding_size_per_head).to(dtype)])]) * num_block_layers)
+    dummy_input['attention_mask'] = torch.cat([torch.zeros(dummy_input['attention_mask'].shape[0], 1).to(dummy_input['attention_mask'].dtype), \
+                                                dummy_input['attention_mask']], -1)
+    if model_pipeline.task == 'text-generation':
+        jit_inputs = tuple([dummy_input['input_ids'], past_key_values, dummy_input['attention_mask']])
+    elif model_pipeline.task == 'text2text-generation':
+        # need to do ...
+        jit_inputs = None
+    else:
+        jit_inputs = tuple([dummy_input['input_ids'], dummy_input['attention_mask']])
```
echarlaix, line 67: Can you replace the single by double quotes everywhere for uniformity ? 
```suggestion
        jit_inputs = tuple([dummy_input["input_ids"], dummy_input["attention_mask"]])

```

jiqing-feng:
@echarlaix Hi, Ella. Thanks for your comments! I have fixed all the issues and applied [NormalizedConfigManager](https://github.com/huggingface/optimum/blob/main/optimum/utils/normalized_config.py#L151) in my code. Would you please help to review it? Thanks!

optimum/intel/ipex/inference.py
```diff
@@ -20,6 +22,51 @@
...
+        )
+    else:
+        past_key_values = tuple(
+            (
+                torch.zeros(input_length, num_attention_heads, 1, num_embedding_size_per_head).to(dtype),
+                torch.zeros(input_length, num_attention_heads, 1, num_embedding_size_per_head).to(dtype),
+            )
+            for _ in range(num_block_layers)
+        )
+    dummy_input["attention_mask"] = torch.cat(
```
echarlaix, line 52: Why is this modification necessary ?

optimum/intel/ipex/inference.py
```diff
@@ -20,6 +22,51 @@
...
+    if model_pipeline.model.config == "bloom":
+        past_key_values = tuple(
+            (
+                torch.zeros(int(num_attention_heads * input_length), num_embedding_size_per_head, 1).to(dtype),
+                torch.zeros(int(num_attention_heads * input_length), 1, num_embedding_size_per_head).to(dtype),
+            )
+            for _ in range(num_block_layers)
+        )
+    else:
+        past_key_values = tuple(
```
echarlaix, line 45: A more robust way to generate the `past_key_values`  could be to use the [`DummyPastKeyValuesGenerator`](https://github.com/huggingface/optimum/blob/v1.7.1/optimum/utils/input_generators.py#L373), it will handle specific cases such as for [bloom](https://github.com/huggingface/optimum/blob/v1.7.1/optimum/exporters/onnx/model_configs.py#L195).

optimum/intel/ipex/inference.py
```diff
@@ -20,6 +22,51 @@
...
+        )
+    dummy_input["attention_mask"] = torch.cat(
+        [
+            torch.zeros(dummy_input["attention_mask"].shape[0], 1).to(dummy_input["attention_mask"].dtype),
+            dummy_input["attention_mask"],
+        ],
+        -1,
+    )
+    if model_pipeline.task == "text-generation" and model_pipeline.model.config.use_cache:
+        jit_inputs = (dummy_input["input_ids"], past_key_values, dummy_input["attention_mask"])
```
echarlaix, line 60: The parameter order can very depending on the modeling definition, a more robust way to create the dummy inputs would be to generate them using a configuration such as done for the optimum ONNX [export](https://github.com/huggingface/optimum/blob/v1.7.1/optimum/exporters/onnx/convert.py#L208).

optimum/intel/ipex/run_text-generation.py
echarlaix, line 1: Would it be possible to move this test in the tests directory ? And to add a comparison between the outputs using `inference_mode` from `torch` and `optimum`  (`ipex`)

optimum/intel/ipex/inference.py
jiqing-feng: We generated the past_key_values so the input sequence length increases, so the attention mask needs to increase its length to match the input length.

optimum/intel/ipex/inference.py
jiqing-feng: Thanks for your advice, but for now it does not have a specific process for bloom models. Maybe I can submit another PR for changing the `DummyPastKeyValuesGenerator ` to adapt bloom models. Alternatively, I can use the `DummyPastKeyValuesGenerator ` to generate common data and reshape it for bloom.

optimum/intel/ipex/inference.py
jiqing-feng: It seems great, I can have a try. BTW, do you have any examples of how to use it? Thanks!

jiqing-feng:
@echarlaix Hi, Ella. Thanks for your comments. I have fixed all of them. Would you please help me to review it? Thanks. BTW, there are 2 failed checks and it seems not related to my changes.

optimum/intel/ipex/inference.py
```diff
@@ -20,7 +26,25 @@
...
+def prepare_jit_inputs(model):
+    generation_config = GenerationConfig(config=model.config, task="text-generation")
+    input_config = {
+        "input_ids": {"batch_size": 1, "sequence_length": 1},
+        "past_key_values": {"batch_size": 1, "sequence_length": 1},
+        "attention_mask": {"batch_size": 1, "sequence_length": 2},
+    }
+
+    jit_inputs = generation_config.generate_dummy_inputs(framework="pt", **input_config)
+    jit_inputs["past_key_values"] = tuple(jit_inputs["past_key_values"])
```
echarlaix, line 38: In my previous comment I was suggesting to use directly the onnx config allowing to have the inputs directly generated for a lot of architectures, was this suggestion preferred in order to  not have to add a dependency to onnx ? (which is not necessary in this case I agree but this would makes things much easier IMO)

```python
from optimum.exporters.tasks import TasksManager

onnx_config_class = TasksManager.get_exporter_config_constructor(
    exporter="onnx",
    model=model,
    task=task,
)
dummy_inputs = onnx_config_class.generate_dummy_inputs(framework="pt")
```

with task being the task being defined in https://huggingface.co/docs/optimum/main/en/exporters/task_manager#pytorch which can be inferred such as :
```python
task = TasksManager.infer_task_from_model(model_id)
```

examples/pipeline/run_text-generation.py
```diff
@@ -0,0 +1,54 @@
...
+model = model.eval().to(torch.device("cpu"))
+model = model.to(memory_format=torch.channels_last)
+
+# 32 tokens
+input_seq = "DeepSpeed is a machine learning framework for deep neural networks and deep reinforcement learning. It is written in C++ and is available for Linux, Mac OS X,"
+
+print("Input sequence is: ")
+print(input_seq)
+
+generate_kwargs = dict(max_new_tokens=32, do_sample=False, num_beams=4, num_beam_groups=1)
```
echarlaix, line 31: I don't think an example would be relevant here, but would rather have it in the documentation instead what do you think ? Also it would make sense to add a test as well !

echarlaix:
> @echarlaix Hi, Ella. Thanks for your comments. I have fixed all of them. Would you please help me to review it? Thanks. BTW, there are 2 failed checks and it seems not related to my codes.

Thanks for iterating on it ! Will take a look at the failing tests

optimum/intel/ipex/inference.py
jiqing-feng: The onnx config class may not meet my demand since I need ('input_ids', 'past_key_values', 'attention_masks') in the inputs. BTW, I running the previous code and got the error: `AttributeError: 'functools.partial' object has no attribute 'generate_dummy_inputs'`

jiqing-feng: @echarlaix Hi, Ella. I have tried `OnnxConfig` but it doesn't work. When I run the previous codes, I got the error: `AttributeError: 'functools.partial' object has no attribute 'generate_dummy_inputs'`. When I run `OnnxConfig(model.config, task)`, I got this error: `TypeError: Can't instantiate abstract class OnnxConfig with abstract method inputs`.

echarlaix: Apologies @jiqing-feng, you also need to create an instance of OnnxConfig, here is the updated code snippet :

```python
from optimum.exporters.tasks import TasksManager

onnx_config_class = TasksManager.get_exporter_config_constructor(
    exporter="onnx",
    model=model,
    task=task,
)
onnx_config = onnx_config_class(model.config)
dummy_inputs = onnx_config.generate_dummy_inputs(framework="pt")
```
Note : if `task="causal-lm"`, the `past_key_values` will not be generated. For that, you need to have  `task="causal-lm-with-past"` instead.

Also, you can also get the task from pipelines using the `task` attribute, when using it you will need to use a mapping such as [`_TASK_ALIASES`](https://github.com/huggingface/optimum-intel/blob/v1.7.3/optimum/intel/openvino/__init__.py#L46) (which can be moved to `optimum.intel.utils` as it can be useful for all subpackages).

optimum/intel/ipex/inference.py
```diff
@@ -1,11 +1,22 @@
...
+from transformers.modeling_outputs import CausalLMOutputWithPast, Seq2SeqLMOutput
 from transformers.pipelines import Pipeline
-from transformers.utils import is_ipex_available
+from transformers.utils import ModelOutput, is_ipex_available
+
+from optimum.exporters.onnx import OnnxConfig
+from optimum.exporters.tasks import TasksManager
+from optimum.utils import DummyPastKeyValuesGenerator, NormalizedConfigManager, NormalizedTextConfig
+
+from ..openvino import _TASK_ALIASES
```
echarlaix, line 18: This could cause issue as openvino is not a hard dependecy, you can rebase and use :

```suggestion
from ..utils.constant import _TASK_ALIASES
```
(moved in https://github.com/huggingface/optimum-intel/pull/266)

optimum/intel/ipex/inference.py
```diff
@@ -1,11 +1,22 @@
...
+from .dummy_inputs import GenerationConfig
```
echarlaix, line 19:
```suggestion

```

optimum/intel/ipex/inference.py
```diff
@@ -20,7 +31,48 @@
...
+        param_regex = re.compile(rf"{param}(\.\d*)?")
+        to_insert = []
+        for name, dynamic_axes in inputs.items():
+            if re.match(param_regex, name):
+                to_insert.append((name, dynamic_axes))
+        # TODO: figure out a smart way of re-ordering potential nested structures.
+        to_insert = sorted(to_insert, key=lambda t: t[0])
+        for name, dynamic_axes in to_insert:
+            ordered_inputs[name] = dynamic_axes
+    return ordered_inputs
```
echarlaix, line 51: Wouldn't this work :    
```suggestion
    return tuple(inputs[key] for key in sig.parameters if inputs.get(key, None) is not None)
```

optimum/intel/ipex/inference.py
```diff
@@ -20,7 +31,48 @@
...
+        task=task,
+    )
+    onnx_config = onnx_config_class(model.config)
+    dummy_inputs = onnx_config.generate_dummy_inputs(framework="pt")
+    dummy_inputs = ordered_inputs(dummy_inputs, model)
+
+    inputs = []
+    for k, v in dummy_inputs.items():
+        if v is not None and not isinstance(v, bool):
+            inputs.append(v)
```
echarlaix, line 70: why is this needed ? if yes can this be integrated directly in `ordered_inputs` ?

optimum/intel/ipex/inference.py
```diff
@@ -29,15 +81,45 @@ def __init__(self, optimized, default):
 
     def __call__(self, *args, **kwargs):
         try:
-            return self._optimized(*args, **kwargs)
+            kwargs.pop("position_ids", None)
```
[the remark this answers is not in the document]
jiqing-feng: It's because inputs of some models like GPT2 have `position_id` which is not contained in the jit traced model. GPT2 can use this optimization by removing `position_ids`. I can remove `kwargs.pop("position_ids", None)` and models like GPT2 will not use the optimization.

jiqing-feng:
@echarlaix Hi Ella, thanks for your comments. The unit test has been added. I have run the unit test and it is ok. I kept `_ModelFallbackWrapper` and use another class named `_ModelGenerationWrapper` for generative models. Could you please review these changes? Thanks!

BTW, the import error in the failed check may be due to the constant being recently added and the check did not catch up with the latest codes.

echarlaix:
> BTW, the import error in the failed check may be due to the constant being recently added and the check did not catch up with the latest codes.

Could you rebase then ? 
Also I added https://github.com/huggingface/optimum-intel/pull/267 for you to integrate your test (a yaml file is currently missing so the test you added will not be triggered)

jiqing-feng:
> > BTW, the import error in the failed check may be due to the constant being recently added and the check did not catch up with the latest codes.
> 
> Could you rebase then ? Also I added #267 for you to integrate your test (a yaml file is currently missing so the test you added will not be triggered)

@echarlaix Thanks for your help, I have rebased my codes. Should I remove my test file `tests/pipeline/test_text-generation.py` and add my test case on the `tests/ipex/test_inference.py`? BTW, I want to demonstrate how to use the text-generation pipeline with ipex and jit optimization, could you please give me a director or file where I can add my demonstration.

echarlaix:
> @echarlaix Thanks for your help, I have rebased my codes. Should I remove my test file `tests/pipeline/test_text-generation.py` and add my test case on the `tests/ipex/test_inference.py`?

Yes please !

> BTW, I want to demonstrate how to use the text-generation pipeline with ipex and jit optimization, could you please give me a director or file where I can add my demonstration? Thanks!

Depending on the example you want to integrate a [notebook](https://github.com/huggingface/optimum-intel/tree/main/notebooks) or an [example](https://github.com/huggingface/optimum-intel/tree/main/examples) script could be adapted.

tests/pipeline/test_text-generation.py
```diff
@@ -0,0 +1,43 @@
...
+import torch
+from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer, pipeline
+
+from optimum.intel import inference_mode as ipex_inference_mode
+from parameterized import parameterized
+
+
+MODEL_NAMES = {
+    "bloom": "bigscience/bloom-7b1",
+    "gptj": "EleutherAI/gpt-j-6B",
```
echarlaix, line 14: For tests, please use smaller models instead :
```suggestion
    "bloom": "hf-internal-testing/tiny-random-bloom",
    "gptj": "hf-internal-testing/tiny-random-gptj",
```

tests/pipeline/test_text-generation.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    def test_pipeline_inference(self, model_arch):
+        model_id = MODEL_NAMES[model_arch]
+        kwargs = dict(torch_dtype=torch.bfloat16, use_cache=True, low_cpu_mem_usage=True, return_dict=False)
+        kwargs["torch_dtype"] = torch.bfloat16
+        kwargs["use_cache"] = True
+        kwargs["low_cpu_mem_usage"] = True
+        kwargs["return_dict"] = False
```
echarlaix, line 31:
```suggestion
        kwargs = {
            "torch_dtype" : torch.bfloat16,
            "use_cache" : True,
            "low_cpu_mem_usage" : True,
            "return_dict": False,
        }
```

tests/pipeline/test_text-generation.py
```diff
@@ -0,0 +1,43 @@
...
+        model.eval()
+        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        inputs = "DeepSpeed is a machine learning framework for deep neural networks and deep reinforcement learning. It is written in C++ and is available for Linux, Mac OS X,"
+        generate_kwargs = {"max_new_tokens": 32, "do_sample": False, "num_beams": 4, "num_beam_groups": 1}
+        text_generator = pipeline("text-generation", model=model, tokenizer=tokenizer)
+        with torch.inference_mode():
+            output = text_generator(inputs, **generate_kwargs)
+        with ipex_inference_mode(text_generator, dtype=torch.bfloat16, verbose=False, jit=True) as ipex_text_generator:
+            output_ipex = ipex_text_generator(inputs, **generate_kwargs)
+        self.assertTrue(output[0]["generated_text"] == output_ipex[0]["generated_text"])
```
echarlaix, line 43:
```suggestion
        self.assertEqual(output[0]["generated_text"], output_ipex[0]["generated_text"])
```

optimum/intel/ipex/inference.py
```diff
@@ -20,6 +30,37 @@
...
+        sig = inspect.signature(model.call)
+
+    return tuple(inputs[key] for key in sig.parameters if inputs.get(key, None) is not None)
+
+
+def prepare_jit_inputs(model: PreTrainedModel, task: str):
+    """
+    Prepare tuple inputs for jit trace model
+    """
+    task = _TASK_ALIASES[task]
```
echarlaix, line 49: The model is traced once, but shouldn't it be done twice when `use_cache` is set to `True` ? If not, what happens for the model first generation (when no `past_key_values` provided), in this case is the original model used instead ?

jiqing-feng: Yes, the original model will be used when predicting the first token when `use_cache=True`. There is no need to save two traced models because it barely helps with latency and needs extra memory.

echarlaix: Could you share the speed up you obtained for the different cases ? (if you have them). Also what are the speed up observed in comparison to generation without IPEX ? This could be intereting to have in the tests as well (to verify that we always get the expected speed up) what do you think ?

echarlaix: cc @mfuntowicz

optimum/intel/ipex/inference.py
```diff
@@ -99,18 +189,19 @@ def __enter__(self):
...
+                                        torch._C._jit_set_profiling_mode(False)
                                         model = torch.jit.trace(model, jit_inputs, strict=False)
                                         model = torch.jit.freeze(model)
                                         model(*jit_inputs)
                                         model(*jit_inputs)
                                     except Exception as e:
                                         logger.warning(f"failed to use PyTorch jit mode due to: {e}.")
                                 # Patching model with the new one
-                                self._model.model = _ModelFallbackWrapper(model, self._original)
+                                if "generation" in self._model.task:
```
echarlaix, line 201: If "text2text-generation" task is also concerned, we should add some tests as well, what do you think ?

jiqing-feng: I have changed it. We only consider text-generation for now, but the text2text-generation is on the way.

jiqing-feng:
Hi @echarlaix , here is the speed-up data, I only tested it on two models, bloom-7b1 and gptj-6b.
node:
image
envs:
intel-extension-for-pytorch 2.1.0+git3a39c02
torch                       2.1.0a0+gitf08dd7d
performance:
image


BTW. I think we can keep tests only support to validate the outputs because performance may depend on nodes and environments. It is better to have the performance test in a notebook, I can submit this notebook when I finish the text2text-generation task.

jiqing-feng:
Hi, Ella. Thanks for your comments. Here is the speed-up data.
node:
image
envs:
torch 2.1.0a0+gitf08dd7d
ipex 2.1.0+git3a39c02
performance:
image


BTW, I think we can keep tests only to support validating the outputs since performance depends on different nodes and environments. I can submit a notebook to run the performance and collect the speed-up data after this PR is merged.

yao-matrix:
@echarlaix, any more actions need us to do to merge this PR? Thx

optimum/intel/ipex/inference.py
echarlaix: Thanks !

echarlaix:
waiting for @mfuntowicz review

mfuntowicz:
Thanks a lot @jiqing-feng for working on this.

Everything sounds ok on my side, I'll let the final word to @echarlaix for merging.

What speedup we can expect for `float32`? Asking because we still don't have access to SPR nodes for production use cases and `bfloat16` is still not the go-to on CPU.

I'd be in favour of defaulting to `float32` and not `bfloat16` for now and see when SPR is more generally available for us to add capabilities to detect CPU ISA and adjust `float32` -> `bfloat16` accordingly.

wdyt @jiqing-feng ? @echarlaix ?

jiqing-feng:
> Thanks a lot @jiqing-feng for working on this.
> 
> Everything sounds ok on my side, I'll let the final word to @echarlaix for merging.
> 
> What speedup we can expect for `float32`? Asking because we still don't have access to SPR nodes for production use cases and `bfloat16` is still not the go-to on CPU.
> 
> I'd be in favour of defaulting to `float32` and not `bfloat16` for now and see when SPR is more generally available for us to add capabilities to detect CPU ISA and adjust `float32` -> `bfloat16` accordingly.
> 
> wdyt @jiqing-feng ? @echarlaix ?

Sure! It is reasonable. Please let me know if there is anything I can do.

optimum/intel/ipex/inference.py
```diff
@@ -99,18 +185,18 @@ def __enter__(self):
...
+                                        jit_inputs = prepare_jit_inputs(self._model.model, self._model.task)
                                         model = torch.jit.trace(model, jit_inputs, strict=False)
                                         model = torch.jit.freeze(model)
                                         model(*jit_inputs)
                                         model(*jit_inputs)
                                     except Exception as e:
                                         logger.warning(f"failed to use PyTorch jit mode due to: {e}.")
                                 # Patching model with the new one
-                                self._model.model = _ModelFallbackWrapper(model, self._original)
+                                if "text-generation" in self._model.task:
```
echarlaix, line 196: Because we want to exclude the `"text2text-generation"` task, shouldn't it be :
```suggestion
                                if self._model.task == "text-generation":

```

optimum/intel/ipex/inference.py
```diff
@@ -40,6 +77,55 @@ def __getattr__(self, item):
...
 
 
+class _ModelGenerationWrapper(GenerationMixin):
+    __slots__ = ("_optimized", "_default")
+
+    def __init__(self, optimized, default):
+        self._optimized = optimized
+        self._default = default
+
+    def __call__(self, *args, **kwargs):
```
echarlaix, line 87: We should replace it with https://github.com/huggingface/optimum-intel/pull/283 before merging no ?

PR #231, opened 2023-03-12
Add JPQD evaluation notebook

helena-intel:
Add JPQD evaluation notebook. Since JPQD QA takes about 12 hours to train, it doesn't make sense to do it in a notebook (if the browser crashes or the computer goes to sleep, training would stop). So I just refer to the example and use the notebook to evaluate the model.

This makes the notebook similar to the PTQ QA notebook. I thought about removing duplication but I think duplication in examples is not so bad, at least for now. It's nice that examples are standalone. 

Since JPQD starts from a plain bert-base-uncased model I finetuned a bert-base-uncased model following the transformers run_qa.py example to compare performance.

Instead of making this a JPQD specific notebook, it could make more sense to make it a generic QA INT8 evaluation notebook, but on the other hand, it's an example, people can surely change it for similar purposes, and it's nice to promote JPQD.

TODO: the intro text at the top needs to explain a bit more about JPQD.

Colab link: https://colab.research.google.com/github/helena-intel/optimum-intel/blob/jpqd-notebook/notebooks/openvino/question_answering_quantization_jpqd.ipynb (performance is probably bad on Colab because there is no AVX512/VNNI).

@vuiseng9

### DIFF
--- a/notebooks/openvino/question_answering_quantization_jpqd.ipynb
+++ b/notebooks/openvino/question_answering_quantization_jpqd.ipynb
@@ -1,0 +1,892 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "eaed3927-e315-46d3-8889-df3f3bbcbf6b",
+   "metadata": {},
+   "source": [
+    "# Joint Pruning Quantization and Distillation with OpenVINO and NNCF\n",
+    "\n",
+    "With quantization, we reduce the precision of the model's weights and activations from floating point (FP32) to integer (INT8). This results in a smaller model with faster inference times with OpenVINO Runtime. \n",
+    "\n",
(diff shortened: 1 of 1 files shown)


AlexKoff88:
I think that it could be more useful if we can show the performance and accuracy trade-offs for three models:
- Original Transformer model (fp32)
- Quantized model (PTQ/QAT)
- Pruned and quantized (JPQD, distillation is an auxiliary method here)

ljaljushkin:
@yujiepan-work and @vuiseng9 implemented very nice lightweight tests for JPQD training. 
9 epochs take just a few seconds on a single card. I'd reuse them for this notebook. 
https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/sparsity/movement/test_training.py#L237

if we need a very good accuracy/performance results, there are longer tests to consider:
https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/sparsity/movement/test_training.py#L318
If I am not mistaken, it takes minutes. Probably, @yujiepan-work could say the exact time.

PR #210, opened 2023-03-02
Add JPQD examples of audio classification and image classification

yujiepan-work:
This PR adds examples of OpenVINO OVTrainer:

(1) QAT and JPQD examples in audio-classification.
(2) JPQD example in image-classification.
(3) renames "teacher_model_or_path" to "teacher_model_name_or_path" in the example scripts. Should be more appropriate

PR #236, opened 2023-03-15
Fix OVTrainer IR saving for structurally pruned model

yujiepan-work:
## Background
To save the structurally pruned IR model, the current workaround is:  
dynamic shape onnx -> dynamic shape IR ->  static shape IR -> pruned static shape IR -> pruned dynamic shape IR.

Swin model will get errors at "dynamic shape IR ->  static shape IR " step because swin cannot accept arbitrary static shape. To fix this, this PR proposes to use the input info in `ovconfig` to get a proper static shape.

## Changes
- "dynamic shape IR ->  static shape IR " step in generating pruned IR will now use `ovconfig`'s input info as static shape.
- If the above step failed, then IR will not be saved, so that users can realize that something wrong happens.
- Add new tests to check ovtrainer can save swin / wav2vec models with structured pruning configs. 
- The new tests are in `test_trainer.py`.  Currently, the file contains only the ovtrainer tests on **text-classification**.  This PR extends new cases on image-classification and audio-classification because we want to check swin and wav2vec2 models.  To better organize these 3 cases, we abstract a new **base class** to aggregate common codes, and create the test classes specific for each case. Though the codes are highly refactored, existing checks on text-classification are 100%  preserved.
- Add a new test requirement `librosa` for audio task tests, see [license](https://github.com/librosa/librosa/blob/main/LICENSE.md)

### DIFF
--- a/optimum/intel/openvino/trainer.py
+++ b/optimum/intel/openvino/trainer.py
@@ -719,8 +719,10 @@ class OVTrainer(Trainer):
                     if not save_as_external_data:
                         onnx_path.write_bytes(f.getvalue())
                     logger.warning(
-                        f"Error encountered during IR pruning: {err}. {onnx_path} is dumped for debug. Run continues."
+                        f"Error encountered during OpenVINO IR pruning: {err}. {onnx_path} is dumped for debug. "
+                        "Run continues without saving OpenVINO IR files."
                     )
+                    return
             else:
                 if self._get_compression_controller_by_cls(QuantizationController) is not None:
                     compress_quantize_weights_transformation(ov_model)
@@ -749,9 +751,10 @@ class OVTrainer(Trainer):
 
     def _reshape_ir(self, ov_model: openvino.runtime.Model, static_shape: bool) -> openvino.runtime.Model:
         new_input_cfg = dict()
+        input_name_vs_shape = {item["keyword"]: item["sample_size"] for item in self.ov_config.input_info}
         for input_ in ov_model.inputs:
             if static_shape is True:
-                new_input_cfg[input_.any_name] = PartialShape(list(range(1, len(input_.partial_shape) + 1)))
+                new_input_cfg[input_.any_name] = PartialShape(input_name_vs_shape[input_.any_name])
             else:
                 new_input_cfg[input_.any_name] = PartialShape([-1] * len(input_.partial_shape))
         ov_model.reshape(new_input_cfg)
--- a/optimum/intel/openvino/trainer.py
+++ b/optimum/intel/openvino/trainer.py
@@ -754,7 +754,9 @@ class OVTrainer(Trainer):
         input_name_vs_shape = {item["keyword"]: item["sample_size"] for item in self.ov_config.input_info}
         for input_ in ov_model.inputs:
             if static_shape is True:
-                new_input_cfg[input_.any_name] = PartialShape(input_name_vs_shape[input_.any_name])
+                new_input_cfg[input_.any_name] = PartialShape(
+                    [1, *input_name_vs_shape[input_.any_name][1:]]
+                )  # use batch size of 1 for static shape IR
             else:
                 new_input_cfg[input_.any_name] = PartialShape([-1] * len(input_.partial_shape))
         ov_model.reshape(new_input_cfg)
--- a/optimum/intel/openvino/trainer.py
+++ b/optimum/intel/openvino/trainer.py
@@ -755,7 +755,7 @@ class OVTrainer(Trainer):
         for input_ in ov_model.inputs:
             if static_shape is True:
                 new_input_cfg[input_.any_name] = PartialShape(
-                    [1, *input_name_vs_shape[input_.any_name][1:]]
+                    [1] + input_name_vs_shape[input_.any_name][1:]
                 )  # use batch size of 1 for static shape IR
             else:
                 new_input_cfg[input_.any_name] = PartialShape([-1] * len(input_.partial_shape))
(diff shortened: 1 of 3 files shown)


yujiepan-work:
Hi, @AlexKoff88 @helena-intel, this PR fixes the IR saving for swin model when it is structurally pruned by JPQD. Could you help to look at it?

AlexKoff88 (approved):

vuiseng9:
This PR should be merged before https://github.com/huggingface/optimum-intel/pull/210

optimum/intel/openvino/trainer.py
```diff
@@ -719,8 +719,10 @@ def _save(self, output_dir: Optional[str] = None, state_dict=None):
                     if not save_as_external_data:
                         onnx_path.write_bytes(f.getvalue())
                     logger.warning(
-                        f"Error encountered during IR pruning: {err}. {onnx_path} is dumped for debug. Run continues."
+                        f"Error encountered during OpenVINO IR pruning: {err}. {onnx_path} is dumped for debugging. "
```
bosbes, line 722: I expect most users will not know what to do with the ONNX model for debugging. It's a blanket exception, it's a bit risky to only show a warning. Shouldn't we reraise the exception and then still dump the ONNX model? Then it's at least clear that something is broken, and you get the exact error, instead of just a generic warning that may be easy to miss.

optimum/intel/openvino/trainer.py
```diff
@@ -749,9 +751,12 @@ def _should_apply_pruning_transform(self) -> bool:
 
     def _reshape_ir(self, ov_model: openvino.runtime.Model, static_shape: bool) -> openvino.runtime.Model:
         new_input_cfg = dict()
+        input_name_vs_shape = {item["keyword"]: item["sample_size"] for item in self.ov_config.input_info}
```
bosbes, line 754: Is there a possibility that `self.ov_config.input_info` is not defined at this point?

optimum/intel/openvino/trainer.py
```diff
@@ -719,8 +719,10 @@ def _save(self, output_dir: Optional[str] = None, state_dict=None):
                     if not save_as_external_data:
                         onnx_path.write_bytes(f.getvalue())
                     logger.warning(
-                        f"Error encountered during IR pruning: {err}. {onnx_path} is dumped for debug. Run continues."
+                        f"Error encountered during OpenVINO IR pruning: {err}. {onnx_path} is dumped for debugging. "
+                        "Run continues without saving OpenVINO IR files."
                     )
+                    return
             else:
                 if self._get_compression_controller_by_cls(QuantizationController) is not None:
```
helena-intel, line 727: It's a blanket exception, it's a bit risky to only show a warning. Shouldn't we reraise the exception and then still dump the ONNX model? Then it's clear that something is broken, and you get the exact error

optimum/intel/openvino/trainer.py
helena-intel, line 754: Is there a possibility that self.ov_config.input_info is not defined at this point?

optimum/intel/openvino/trainer.py
yujiepan-work: > It's a blanket exception, it's a bit risky to only show a warning. Shouldn't we reraise the exception and then still dump the ONNX model? Then it's clear that something is broken, and you get the exact error

I agree with your concern. See updates: https://github.com/huggingface/optimum-intel/blob/5618c3cf45d61bab48d702a0254001395e66f182/optimum/intel/openvino/trainer.py#L717-L724

optimum/intel/openvino/trainer.py
yujiepan-work: Line176, Lind196

tests/openvino/test_training.py
```diff
@@ -116,317 +98,737 @@ class OVTrainerTestDescriptor:
...
+                for inputs in self.trainer.get_eval_dataloader(dataset):
+                    self.assertSequenceEqual(inputs["input_ids"].shape, [batch_size, seq_length])
+                    ovmodel_outputs = ovmodel(**inputs)
+                    self.assertIn("logits", ovmodel_outputs)
+                    ovmodel_logits = ovmodel_outputs.logits
+                    with torch.no_grad():
+                        torch_logits = torch_model(**inputs).logits
+                    torch.testing.assert_close(
+                        torch.softmax(ovmodel_logits, dim=-1),
+                        torch.softmax(torch_logits, dim=-1),
```
echarlaix, line 500: Could this be replaced with :
```suggestion
                        ovmodel_logits,
                        torch_logits,
```

echarlaix (approved):
It looks great, thanks a lot @yujiepan-work !

tests/openvino/test_training.py
yujiepan-work: The softmax here is intended to amplify the logits so that atol=1e-4 can pass.

I check the model logits and find that sometimes the values are very close to zero, e.g., 0.0071. In **quantization** cases where higher error is expected than fp32 models, the original logit values are easy to accidentally fail at atol=1e-4 (e.g., 0.0071 vs. 0.0073).

In the latest commit I have deleted the softmax, but I also have to increase atol to 1e-3 to pass the tests (see latest checks here， ignoring the INC part which should not our problem). Is that acceptable? I see some other tests are also using 1e-3, e.g.,
https://github.com/huggingface/optimum-intel/blob/844be9adc9eb3bea8f87c22fa9112a4fb3a31d88/tests/openvino/test_modeling.py#L646

echarlaix: A tolerance of 1e-3 sounds reasonable, thanks a lot for updating it !

PR #244, opened 2023-03-20
Add dummy objects

echarlaix:
Allow importing classes from the openvino or neural_compressor subpackages from `optimum.intel`, an error will be raised when an instance is created (specifying which dependency is missing)


AlexKoff88:
I am ok with the changes if the backward compatibility is preserved. It looks like so based on the changes.

AlexKoff88 (approved):

PenghuiCheng (approved):

echarlaix:
> I am ok with the changes if the backward compatibility is preserved. It looks like so based on the changes.

Yes backward compatibility is preserved !

PR #262, opened 2023-03-29
Remove GPU dynamic shapes restriction

helena-intel:
- Remove message about GPU dynamic shapes. Dynamic shapes on GPU work in the current OpenVINO prerelease, though there are still some known issues. We currently only show this message when a user loads a model though, not when moving a model `.to("gpu")`.
- Change comment for `.half()`. It's not necessary anymore to explicitly convert the model to  FP16 for increased performance on GPU; OpenVINO will do this automatically. It's still a useful method to save the model weights with half the size. 

We should discuss how to support setting explicit inference hints https://docs.openvino.ai/latest/openvino_docs_OV_UG_supported_plugins_CPU.html

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -85,11 +85,6 @@ class OVBaseModel(OptimizedModel):
         self.preprocessors = kwargs.get("preprocessors", [])
         enable_compilation = kwargs.get("compile", True)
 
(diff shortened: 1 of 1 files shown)


echarlaix (approved):
Thanks a lot for updating it @helena-intel, could you integrate it in the [documentation ](https://github.com/huggingface/optimum-intel/blob/v1.7.3/docs/source/inference.mdx?plain=1#L55) as well ?

PR #263, opened 2023-03-29
Add inc config

echarlaix:
In this PR we add the `INCConfig` class, which allows to summarize the different compression techniques that were applied on a model and their characteristics, as well as `neural-compressor` and `torch` versions (useful when loading a quantized model). If present on the hub or local directory where a model is being loaded, the configuration will be loaded and the torch version will be checked for quantized model

### DIFF
--- a/optimum/intel/__init__.py
+++ b/optimum/intel/__init__.py
@@ -96,6 +96,7 @@ except OptionalDependencyNotAvailable:
     ]
 else:
     _import_structure["neural_compressor"] = [
+        "INCConfig",
         "INCModelForCausalLM",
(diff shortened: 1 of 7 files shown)


hshen14:
@echarlaix That's the absolutely right direction. Thank you for proposing this PR. INC queries the framework capabilities e.g., torch 2.0 supports INT8 ops like A/B/C, and know how to do smart things like auto-tuning if some ops are not suitable for INT8.

ftian1 (approved):

PenghuiCheng (approved):

PR #265, opened 2023-03-30
Apply new Python API features from OpenVINO 2023.0 release

jiwaszki:
This PR tests features from main branch of OpenVINO including:

* Shared memory mode for inference calls
* New return type `OVDict` that is able to address elements with string values
* New `CompiledModel.__call__` with automatic request creation

This PR can be tested against https://github.com/openvinotoolkit/openvino/pull/16665

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -539,5 +539,5 @@ class OVModelVaeEncoder(OVModelPart):
         inputs = {
             "sample": sample,
         }
-        outputs = self.request.infer(inputs, shared_memory=True)
+        outputs = self.request(inputs, shared_memory=True)
         return list(outputs.values())


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -447,9 +447,8 @@ def __init__(
 
     def _create_inference_request(self):
         if self.request is None:
-            logger.info("Compiling the encoder and creating the inference request ...")
```
echarlaix, line 450: It would be great to rename the `_create_inference_request` method to `_compile` for stable diffusion models as well to keep uniformity

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -318,36 +313,29 @@ def forward(
         attention_mask: torch.LongTensor = None,
         **kwargs,
     ) -> BaseModelOutput:
-        self._create_inference_request()
+        self._compile()
 
-        # Check if inputs are c-like, if not - convert them
-        input_ids = _contiguous_helper(np.array(input_ids))
```
echarlaix, line 324: It looks much clearer now ! Could this also be integrated for `OVModelForCausalLM` from modeling_decoder.py as well ? happy to do it it in a following PR if not

echarlaix (approved):
It looks much clearer thanks ! Could you update the openvino version in the setup so that we can merge it soon ?

optimum/intel/openvino/modeling_seq2seq.py
jiwaszki: I edited it as much as I could. Hopefully it's better now!:)

I am sorry for the late response here. However, now OV is bumped in setup.py and features can be used from the official release.

optimum/intel/openvino/modeling_diffusion.py
jiwaszki: Fixed

echarlaix (approved):
LGTM, thanks for updating it !

PR #274, opened 2023-04-07
Enable past key values

echarlaix:
In this PR we enable the possibility to leverage the `pask_key_values` for causal language model, this can be done by setting `use_cache` to `True` when exporting a model. For inference, this behavior will be determined by the model itself (whether it accepts `pask_key_values` as inputs)


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -0,0 +1,347 @@
...
+            **kwargs,
+        )
+
+    def _reshape(self, model: openvino.runtime.Model, batch_size: int, sequence_length: int, is_decoder=True):
+        shapes = {}
+        for inputs in model.inputs:
+            shapes[inputs] = inputs.get_partial_shape()
+            shapes[inputs][0] = -1
+            if inputs.get_any_name().startswith("past_key_values"):
+                shapes[inputs][2] = -1
```
eaidova, line 182:
```suggestion
                                if inputs.get_any_name().endswith("value"):
                    key_for_value = model.input(inputs.get_any_name().replace(".value", ".key"))
                    if shapes[inputs][2] == key_for_value.get_partial_shape()[1]:
                        shapes[inputs][1] = -1
                        continue
                    shapes[inputs][2] = -1
```

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -0,0 +1,347 @@
...
+            }
+
+        # Create empty past_key_values for decoder_with_past first generation step
+        elif self.use_cache:
+            shape_input_ids = input_ids.shape
+            for input_name in self.key_value_input_names:
+                model_inputs = self.model.input(input_name)
+                shape = model_inputs.get_partial_shape()
+                shape[0] = shape_input_ids[0]
+                shape[2] = 0
```
eaidova, line 308: To take care about models with inverted_value_shape (e.g. bloom) 
```suggestion
                shape[2 if shape[2].is_dynamic else 1] = 0
```

echarlaix: Tried your suggestion but got an error due to the batch size defintion which should change for bloom architecture, was fixed in [`06d285`](https://github.com/huggingface/optimum-intel/pull/274/commits/06d28520091e6e2e630a83db58118bc8f3cfdda1).

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -240,32 +240,105 @@ def forward(
...
+        self, past_key_values: Tuple[Tuple[torch.Tensor]], beam_idx: torch.Tensor
     ) -> Tuple[Tuple[torch.Tensor]]:
         """
         This function is used to re-order the `past_key_values` cache if [`~PreTrainedModel.beam_search`] or
-        [`~PreTrainedModel.beam_sample`] is called. This is required to match `past_key_values` with the correct
-        beam_idx at every generation step.
+        [`~PreTrainedModel.beam_sample`] is called.
+        This is required to match `past_key_values` with the correct beam_idx at every generation step.
         """
+        if self.config.model_type == "bloom":
```
AlexKoff88, line 267: Is this the only case that has such a specific?

AlexKoff88 (approved):

eaidova (approved):

optimum/intel/openvino/modeling_decoder.py
echarlaix: Yes, bloom models needs the `past_key_values` to be reshaped before each generation steps to match the ones from the other architectures

helena-intel (approved):
Thanks @echarlaix ! This is great. One question: with this method, the cache is used if the model was exported with `use_cache=True`. If you did that and then later want to disable the cache, you have to export the model again. That's probably not what a user coming from PyTorch would expect, they are used to being able to change this parameter when loading a model.

Could we for example throw an exception if the user loads a model that includes the cache with `use_cache=False` and tell them that to disable the cache they should export the model again with `use_cache=False` (and probably vice versa)? Doesn't need to be in this PR, it would be great to have this merged soon.

echarlaix:
> Could we for example throw an exception if the user loads a model that includes the cache with `use_cache=False` and tell them that to disable the cache they should export the model again with `use_cache=False` (and probably vice versa)? Doesn't need to be in this PR, it would be great to have this merged soon.

Makes sense ! I will add it in a following PR

PR #281, opened 2023-04-13
Add CACHE_DIR back for CPU

helena-intel:
CACHE_DIR on CPU was disabled in https://github.com/huggingface/optimum-intel/pull/164
I don't see the issue anymore in the 2023.0.0.dev20230217 prerelease, which is the current minimum OpenVINO version in setup.py, so we can enable CACHE_DIR for CPU again.


echarlaix (approved):
LGTM !

PR #290, opened 2023-04-18
Update code and move modeling.py to utils folder.

PenghuiCheng:
# What does this PR do?

Move modeling.py to optimum.intel.utils folder and update the tracing code.

## Before submitting


optimum/intel/utils/modeling.py
```diff
@@ -0,0 +1,337 @@
...
+                "use_auth_token": use_auth_token,
+                "cache_dir": cache_dir,
+                "subfolder": subfolder,
+                "local_files_only": local_files_only,
+                "force_download": force_download,
+            }
+            model = TasksManager.get_model_from_task(task, model_id, **model_kwargs)
+            model_inputs = None
+        else:
+            return cls._from_transformers(
```
echarlaix, line 236: I would rather keep the `export` parameter instead of `torchscript` to keep consistency accross all the `OptimizedModel`. To make it more explicit, why not change the class name instead ? To something like `TorchScriptModelForCausalLM` ?

optimum/intel/utils/modeling.py
```diff
@@ -0,0 +1,337 @@
...
+
+            self.generation_config = GenerationConfig.from_model_config(config)
+
+    @staticmethod
+    def load_model(file_name: Union[str, Path]):
+        model = torch.jit.load(file_name)
+        torch.jit.freeze(model.eval())
+        return model
+
+    def set_traced_model(self, traced_model):
```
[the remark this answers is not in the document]
PenghuiCheng: This function is for quantization with INC, for example:
model = AutoModelForGeneration.from_pretrained(args.model_name_or_path, torchscript=True)

def eval_function(q_model):
    model.set_traced(model)
    for input in dataloader:
        out=model(**input)
    eval_metric = metric(out, label)
    return eval_metric

q_model = inc_quantizer.quantize(model, config, calib_dataloader, eval_func=eval_function)

optimum/intel/utils/modeling.py
PenghuiCheng: I have retained the export parameter, but I don't think the semantics of export match the tracing jit model very well, so I added the torchscript parameter. It's ok for me to change the class name, TorchScriptModelForCausalLM is OK.

PenghuiCheng: How about TorchScriptModelForGeneration? because not all generation model is CausalLM, like gpt2 is GPT2LMHeadModel

optimum/intel/generation/modeling.py
```diff
@@ -0,0 +1,345 @@
...
+            model_kwargs = {
+                "revision": revision,
+                "use_auth_token": use_auth_token,
+                "cache_dir": cache_dir,
+                "subfolder": "",
+                "local_files_only": local_files_only,
+                "force_download": force_download,
+            }
+
+            model = TasksManager.get_model_from_task(task, "../../ipex/text-generation/model", **model_kwargs)
```
echarlaix, line 204: Shouldn't this be removed 
```suggestion

```

optimum/intel/generation/modeling.py
```diff
@@ -0,0 +1,345 @@
...
+                model_id=model_id,
+                config=config,
+                use_auth_token=use_auth_token,
+                revision=revision,
+                force_download=force_download,
+                cache_dir=cache_dir,
+                local_files_only=local_files_only,
+                use_cache=use_cache,
+                **kwargs,
+            )
```
echarlaix, line 254: When would this case happen ? `_from_transformers` should be called when `export=True`

```suggestion
```

optimum/intel/generation/modeling.py
```diff
@@ -0,0 +1,345 @@
...
+            model_kwargs = {
+                "revision": revision,
+                "use_auth_token": use_auth_token,
+                "cache_dir": cache_dir,
+                "subfolder": subfolder,
+                "local_files_only": local_files_only,
+                "force_download": force_download,
+            }
+            model = TasksManager.get_model_from_task(task, model_id, **model_kwargs)
+            model_inputs = None
```
echarlaix, line 242: In which case would we need the original model ?

```suggestion
```

optimum/intel/generation/modeling.py
```diff
@@ -204,16 +302,13 @@ def _from_transformers(
         save_dir = TemporaryDirectory()
         save_dir_path = Path(save_dir.name)
         torch.jit.save(traced_model, save_dir_path / WEIGHTS_NAME)
+        config.torchscript = True
 
-        return cls._from_pretrained(
```
echarlaix, line 208: I think we should keep `cls._from_pretrained` in order to have all the model transformation at one place (like  the 2 iterations to convert model to int8 model for IPEX)

optimum/intel/generation/modeling.py
```diff
@@ -186,8 +277,15 @@ def _from_transformers(
             "force_download": force_download,
         }
 
-        model = TasksManager.get_model_from_task(task, model_id, **model_kwargs)
-        model.config.return_dict = False
+        if isinstance(model_id, str):
+            model = TasksManager.get_model_from_task(task, model_id, **model_kwargs)
+            model.config.return_dict = False
+        elif isinstance(model_id, torch.nn.Module):
```
echarlaix, line 283: I don't think we should leave the possibility to give all the options to give as input the model path, an `nn.Module` and a `RecursiveScriptModule` but to only give the model path. This could be confusing to users in my opinion, also this doesn't respect the format we have for Optimum `OVModel` and `ORTModel`

optimum/intel/generation/modeling.py
```diff
@@ -108,6 +105,94 @@ def forward(
 
         return CausalLMOutputWithPast(logits=outputs[0], past_key_values=outputs[1] if self.use_cache else None)
 
+    @classmethod
+    @add_start_docstrings(FROM_PRETRAINED_START_DOCSTRING)
+    def from_pretrained(
```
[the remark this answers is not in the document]
PenghuiCheng: removed

echarlaix (approved):
LGTM, thanks a lot for iterating on this @PenghuiCheng  !!

PR #292, opened 2023-04-19
[OpenVINO]: Fix import from ONNX format

AlexKoff88:


optimum/intel/openvino/modeling_base.py
```diff
@@ -111,9 +111,12 @@ def load_model(file_name: Union[str, Path]):
         """
         if isinstance(file_name, str):
             file_name = Path(file_name)
-        bin_file_name = file_name.with_suffix(".bin") if file_name.suffix == ".xml" else None
 
-        return core.read_model(file_name, bin_file_name)
+        model = core.read_model(file_name)
+        if file_name.suffix == ".onnx":
```
echarlaix, line 116: When should `apply_moc_transformations` be called?

echarlaix (approved):
LGTM, thanks for the fix @AlexKoff88

optimum/intel/openvino/modeling_base.py
echarlaix: Thanks for the explanation

PR #295, opened 2023-04-19
Add generation for bloom models

echarlaix:


optimum/intel/__init__.py
```diff
@@ -109,9 +109,6 @@
         "INCSeq2SeqTrainer",
         "INCTrainer",
     ]
-    _import_structure["generation.modeling"] = [
```
echarlaix, line 112: cc @PenghuiCheng removed temporarily until validated

PenghuiCheng (approved):

PR #300, opened 2023-04-23
Integrate text-generation pipeline from inference.py to TSModelForCausalLM

jiqing-feng:
Hi @echarlaix . I have integrated my changes from inference.py to TSModelForCausalLM, refer to: https://github.com/huggingface/optimum-intel/pull/227.

I also use the function `export_model` to enable ipex model in TSModelForCausalLM. Would like your opinion. Thanks!

### DIFF
--- a/optimum/intel/generation/modeling.py
+++ b/optimum/intel/generation/modeling.py
@@ -21,7 +21,7 @@ from typing import Optional, Tuple, Union
 
 import torch
 from huggingface_hub import hf_hub_download
-from transformers import AutoConfig, AutoModelForCausalLM, PretrainedConfig
+from transformers import AutoConfig, AutoModelForCausalLM, PretrainedConfig, PreTrainedModel
 from transformers.modeling_outputs import CausalLMOutputWithPast
 from transformers.utils import WEIGHTS_NAME
 
@@ -81,8 +81,33 @@ class TSModelForCausalLM(OptimizedModel, GenerationMixin):
         torch.jit.freeze(model.eval())
         return model
 
+    @staticmethod
+    def jit_trace(model: PreTrainedModel, task: str, config: PretrainedConfig, use_cache: bool = True):
+        model.config.return_dict = False
+        signature = inspect.signature(model.forward) if hasattr(model, "forward") else inspect.signature(model.call)
+        onnx_config_class = TasksManager.get_exporter_config_constructor(model=model, exporter="onnx", task=task)
+        onnx_config = onnx_config_class(model.config, use_past=use_cache)
+        dummy_inputs = onnx_config.generate_dummy_inputs(framework="pt")
+        model_inputs = {
+            key: dummy_inputs[key] for key in signature.parameters if dummy_inputs.get(key, None) is not None
+        }
+        if use_cache:
+            traced_model = torch.jit.trace(model, example_inputs=tuple(model_inputs.values()))
+        else:
+            traced_model = torch.jit.trace(model, example_kwarg_inputs=model_inputs)
+        traced_model = torch.jit.freeze(traced_model.eval())
+        traced_model(**model_inputs)
+        traced_model(**model_inputs)
+
+        return traced_model
+
     def _save_pretrained(self, save_directory: Union[str, Path], file_name: Optional[str] = None, **kwargs):
         torch.jit.save(self.model, os.path.join(save_directory, WEIGHTS_NAME))
+
+    @classmethod
+    def export_model(cls, model: PreTrainedModel, task: str, use_cache: bool = True):
+        traced_model = cls.jit_trace(model, task, model.config, use_cache)
+        return cls(traced_model, model.config)
 
     def forward(
         self,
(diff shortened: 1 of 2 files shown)


optimum/intel/generation/modeling.py
```diff
@@ -81,9 +81,34 @@ def load_model(file_name: Union[str, Path]):
         torch.jit.freeze(model.eval())
         return model
 
+    @staticmethod
+    def jit_trace(model: PreTrainedModel, task: str, config: PretrainedConfig, use_cache: bool = True):
```
echarlaix, line 85: I think it would make sense to create a separate function `trace` (working for all architectures not only causal LM) and to use it in `TSModelForCausalLM._from_transformers` instead of having `jit_trace` and `export_model` methods.

echarlaix (commented):
Thanks a lot for this integration @jiqing-feng

jiqing-feng:
Hi @echarlaix , I have created a separate function trace in a new file, could you please help to review it? Thanks! 

BTW, the failed check seems not related to my changes.

optimum/intel/generation/tracing.py
```diff
@@ -0,0 +1,29 @@
+import inspect
+
+import torch
+from transformers import PreTrainedModel
+
+from optimum.exporters import TasksManager
+
+
+def prepare_jit_inputs(model: PreTrainedModel, task: str, use_cache: bool = True):
```
echarlaix, line 9: Could we keep `prepare_jit_inputs` and `jit_trace` in `modeling.py` ?

optimum/intel/ipex/inference.py
```diff
@@ -188,16 +102,34 @@ def __enter__(self):
...
+                                            jit_inputs = []
+                                            dummy_input = self._model.tokenizer("")
+                                            for key in dummy_input:
+                                                jit_inputs.append(
+                                                    torch.ones((1, len(dummy_input[key])), dtype=torch.long)
+                                                )
+                                            model = torch.jit.trace(model, jit_inputs, strict=False)
+                                            model = torch.jit.freeze(model)
+                                            model(*jit_inputs)
+                                            model(*jit_inputs)
```
echarlaix, line 126: Why not :  
```suggestion
                                        model = jit_trace(
                                            model=model,
                                            task=self._model.task,
                                            use_cache=self._original.config.use_cache,
                                        )
                                        if self._model.task == "text-generation":
                                            model = TSModelForCausalLM(
                                                model=model,
                                                config=self._original.config,
                                                use_cache=self._original.config.use_cache,
                                            )

```

echarlaix (commented):
Thanks for integrating `TSModelForCausalLM` to `inference_mode` @jiqing-feng !!

jiqing-feng:
Hi @echarlaix , thanks for your advice. I have updated the code. Could you please review it? Thanks!

optimum/intel/ipex/inference.py
```diff
@@ -188,16 +101,23 @@ def __enter__(self):
...
+                                                use_cache=self._original.config.use_cache,
+                                            )
                                     except Exception as e:
+                                        self._jit = False
                                         logger.warning(f"failed to use PyTorch jit mode due to: {e}.")
                                 # Patching model with the new one
-                                if self._model.task == "text-generation":
-                                    self._model.model = _ModelGenerationWrapper(model, self._original)
+                                if self._model.task == "text-generation" and self._jit:
+                                    self._model.model = model
```
echarlaix, line 120: Shouldn't it be for all cases :
`self._model.model = _ModelFallbackWrapper(model, self._original)`

jiqing-feng: Hi @echarlaix . If I use _ModelFallbackWrapper, when I execute `model.generate`, it will go to the func  `__getattr__(self, item)`, and will return `getattr(self._default, item)` so I actually execute `self._default.generate`. I didn't use _ModelFallbackWrapper because I cannot use the generation of my optimized model.

echarlaix: Why not subclass from `_ModelFallbackWrapper` to enable it then ?

jiqing-feng:
Hi, @echarlaix . I subclass a new class from `_ModelFallbackWrapper` and use it, but `ipex.optimize()` will change the shape of bloom models so the check failed. 
BTW, could we merge it if there are no big issues? I have the following PR for enabling text2text-generation and other generation tasks. Some issues like variable naming could be fixed in the next PR.
Thanks!

optimum/intel/ipex/inference.py
PenghuiCheng: @echarlaix , I think we should enhance TSModelForCausalLM for all text-generation tasks, then for text-generation task, we don't need to use _ModelFallbackWrapper, is it OK?

jiqing-feng:
Hi @echarlaix , I have updated the test cases of [310](https://github.com/huggingface/optimum-intel/pull/310). Could you help me have a look at the failed check? I can run all test cases successfully in my node. It is probably because of the version of Pytorch.

sywangyi:
@mfuntowicz please help review, This PR utilizes the TSModelForCausalLM and fix the jit issue in llama text generation

PR #303, opened 2023-04-25
Enable export of decoder of quantized models to be exported with pkv

echarlaix:


AlexKoff88 (approved):

helena-intel (approved):
Thanks for adding this!

PR #305, opened 2023-04-27
enable loading of INC quantized stable diffusion model

echarlaix:


hshen14 (approved):
LGTM. @echarlaix , thanks for the PR.

PR #306, opened 2023-05-01
Use static shapes for SWIN model in training test

helena-intel:
There is an issue with OpenVINO 2023.0 and dynamic batch sizes for tiny-swin, which causes the training tests to fail with an error. This PR sets static shapes for the swin model in the training test to workaround this. 

The issue is fixed in https://github.com/openvinotoolkit/openvino/pull/17276 but I still see issues. Compiling this model https://huggingface.co/helenai/swin-base-food101-jpqd-ov stalls with dynamic shapes (this is also the case in 2022.3) and works fine when adding compile=False and reshaping to static shapes first. @vuiseng9 @AlexKoff88 for now we should probably look into modifying the JPQD SWIN example to return static models.

### DIFF
--- a/tests/openvino/test_training.py
+++ b/tests/openvino/test_training.py
@@ -131,7 +131,9 @@ class OVTrainerBaseTrainingTest(unittest.TestCase, ABC):
 
         # check saved ovmodel IR and output
         ovmodel = self.get_ov_model()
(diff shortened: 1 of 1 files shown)


AlexKoff88 (approved):

echarlaix (approved):

PR #308, opened 2023-05-03
Set height and width by default for static models

echarlaix:
In this PR we set the height and width of the outputs by default for statically reshaped models


AlexKoff88 (approved):

helena-intel (approved):
Thank you! This is really nice to have.

PR #310, opened 2023-05-08
fix regression of not using jit model for text-generation/text-classf…

sywangyi:
…ication/token classfication/question-answer task

also add UT to avoid such regression. UT will check if the optimized model is jit model.

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/tests/ipex/test_inference.py
+++ b/tests/ipex/test_inference.py
@@ -20,6 +20,7 @@ from parameterized import parameterized
 # TODO : add more tasks
 from transformers import (
     AutoModelForCausalLM,
+    AutoModelForQuestionAnswering,
     AutoModelForSequenceClassification,
     AutoModelForTokenClassification,
     AutoTokenizer,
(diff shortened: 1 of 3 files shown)


sywangyi:
@echarlaix @jiqing-feng please help review

optimum/intel/utils/constant.py
```diff
@@ -15,7 +15,11 @@
 _TASK_ALIASES = {
     "sequence-classification": "text-classification",
     "sentiment-analysis": "text-classification",
+    "text-classification": "text-classification",
```
echarlaix, line 18: not needed, we currently do `task = _TASK_ALIASES.get(task, task)`

tests/ipex/test_inference.py
```diff
@@ -54,27 +55,48 @@ class IPEXIntegrationTest(unittest.TestCase):
     )
 
     TEXT_GENERATION_SUPPORTED_ARCHITECTURES = (
-        "bloom",
```
echarlaix, line 57: Why are these architectures now unsupported ?

sywangyi: currently, bloom ang gpt_neox has some issue in jit, because we use tuple as jit input. I have a fix to support dict input in jit.trace, which will be released in torch 2.1. I add check in ci for jit to avoid regression like self.assertTrue(isinstance(ipex_text_generator.model._optimized, torch.jit.RecursiveScriptModule)), this two architextures will lead to fail in this assertion. will add some architecture back in 2.1

sywangyi: previously this case could pass because it's silently switch to eager models when jit.trace fails

echarlaix: I see, thanks for the fix @sywangyi !

echarlaix (approved):
Thanks for the fix @sywangyi !!

PR #311, opened 2023-05-08
Suppress torch version mismatch warning with NNCF

vshampor:
# What does this PR do?
Removes the `WARNING:nncf:NNCF provides best results with torch==1.13.1, while current torch version is 1.13.0+cu117. If you encounter issues, consider switching to torch==1.13.1` line from the logs when using `optimum-intel`. 

NNCF does this at `nncf.torch` import-time once, and it seems that NNCF is being imported in `optimum-intel` regardless of being used in such situations.


echarlaix (approved):
Thanks for the addition @vshampor ! Please run the `make style` command  to format your code so that the code quality test pass

PR #312, opened 2023-05-09
Explicitly set batch=1 for NNCF in order to avoid issue with Wave2Vec

ljaljushkin:
# What does this PR do?

Explicitly set batch size equal 1 in the NNCFConfig in order to avoid issue with Wave2Vec with the upcoming NNCF release.
With switching to NNCF v2.5.0, JPQD uses automatic structured pruning algorithm and currently it doesn't work properly with batch > 1 for Wave2Vec model. 
It works fine when batch is 1. With this PR it is set in the NNCFConfig only, it should not affect training and inference with different batch sizes, since it's only used for NNCF internals to set up compression.
The proper fix for the Wave2Vec will be introduced in the next releases of NNCF. Here's the draft (https://github.com/openvinotoolkit/nncf/pull/1784) that will be finalized later.
The fix is checked on the existing NNCF, JPQD tests.

Fixes # (issue)
Ticket 110126 in the internal NNCF tracking system


ljaljushkin:
@AlexKoff88 @yujiepan-work @vuiseng9

optimum/intel/openvino/trainer.py
```diff
@@ -168,7 +168,8 @@ def __init__(
             model_inputs = next(iter(train_dataloader))
             for label_name in self.label_names:
                 model_inputs.pop(label_name)
-            self.ov_config.add_input_info(model_inputs)
+            force_single_batch = self._should_apply_pruning_transform()
```
ljaljushkin, line 171: @AlexKoff88 @echarlaix Made the fix more local. Now it affects only pruning transform case.
Similar what is implemented for ngraph export here:
https://github.com/huggingface/optimum-intel/blob/3714c3226da20a78d2fb1008da2e3592f294b7f1/optimum/intel/openvino/trainer.py#L707-L710

optimum/intel/openvino/configuration.py
```diff
@@ -56,10 +56,10 @@ def __init__(
         self._enable_standard_onnx_export_option()
         self.optimum_version = kwargs.pop("optimum_version", None)
 
-    def add_input_info(self, model_inputs: Dict):
+    def add_input_info(self, model_inputs: Dict, force_single_batch: bool = False):
         self.input_info = [
             {
-                "sample_size": list(value.shape),
+                "sample_size": [1] + list(value.shape[1:] if force_single_batch else value.shape),
```
AlexKoff88, line 62:
```suggestion
                "sample_size": [1] + list(value.shape[1:] if force_batch_one else value.shape),
```

AlexKoff88 (approved):
I am mostly ok with the changes

optimum/intel/openvino/configuration.py
ljaljushkin: renamed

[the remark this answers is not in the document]
ljaljushkin: thank you, corrected!

ljaljushkin:
> @echarlaix Could this be merged before the upcoming dot release?

And these, please
https://github.com/huggingface/optimum-intel/pull/315
https://github.com/huggingface/optimum-intel/pull/319

ljaljushkin:
> Thanks for the fix @ljaljushkin ! Can you rebase as well as run the `make style` command to fix the tests

Done

Issue #313, opened 2023-05-10
`get_nncf_wrapped_model` method does not exist in latest nncf


yujiepan-work:
The `get_nncf_wrapped_model` method for NNCFNetwork in NNCF has been deleted in https://github.com/openvinotoolkit/nncf/pull/1584 Corresponding changes needed.

![image](https://github.com/huggingface/optimum-intel/assets/105195712/ab8ed1ca-8ea8-427d-8b1f-60e1824c82ba)

ljaljushkin:
fixed here:
https://github.com/openvinotoolkit/nncf/pull/1819/files

echarlaix:
Thanks for the update @ljaljushkin !

PR #314, opened 2023-05-10
Prepare to NNCF v2.5 release

ljaljushkin:
# What does this PR do?

This PR updates deprecate NNCF API calls for switching to the upcoming NNCF 2.5.0 release.
The `.forward` on the NNCFNetwork function will always be an NNCF-wrapped forward version, no need to call `get_nncf_wrapped_model`

The API changes were introduced in the https://github.com/openvinotoolkit/nncf/pull/1584

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -281,7 +281,7 @@ class OVQuantizer(OptimumQuantizer):
 
 
 def _onnx_export_nncf_model(model: NNCFNetwork, config: OnnxConfig, output: Union[str, io.BytesIO], opset: int = None):
-    signature = inspect.signature(model.get_nncf_wrapped_model().forward)
+    signature = inspect.signature(model.forward)
     signature = list(signature.parameters.keys())
     opset = opset or config.DEFAULT_ONNX_OPSET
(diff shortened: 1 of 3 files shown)


ljaljushkin:
@vshampor @AlexKoff88 @yujiepan-work @vuiseng9 please take a look

yujiepan-work:
This can solve https://github.com/huggingface/optimum-intel/issues/313

vshampor (approved):

helena-intel:
Thanks for this @ljaljushkin ! When is the NNCF 2.5 release planned? This should not be merged before that, because with `pip install optimum-intel[openvino,nncf]` users will get 2.4 at the moment. We should also update setup.py to update the nncf extra to "nncf>=2.5.0".

ljaljushkin:
> 

Thanks, Helena! Added nncf 2.5.0 to setup.py. AFAIK, the plan was to release at the same time as openvino 2023.0

echarlaix (approved):
LGTM, thanks for this addition @ljaljushkin ! Will merge it once nncf v2.5.0 release is out

helena-intel:
@echarlaix NNCF 2.5 is released! https://github.com/openvinotoolkit/nncf/releases/tag/v2.5.0

PR #320, opened 2023-05-24
Update JPQD config in audio-classification example

yujiepan-work:
# What does this PR do?

This PR updates the NNCF config for JPQD (joint pruning, quantization and distillation) in the example of audio-classification. The new config will quantize more layers (feature_extractor). The output model has a lower sparsity, but has higher throughput and thus should be more favored.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case). NA
- [x] Did you make sure to update the documentation with your changes? yes
- [ ] Did you write any new necessary tests? no needed

### DIFF
--- a/examples/openvino/audio-classification/README.md
+++ b/examples/openvino/audio-classification/README.md
@@ -93,4 +93,4 @@ torchrun --nproc-per-node=1 run_audio_classification.py \
     --seed 42
 ```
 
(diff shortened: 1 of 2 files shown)


yujiepan-work:
@echarlaix @helena-intel @AlexKoff88 @vuiseng9  Please take a look

yujiepan-work:
As a reference,
|            | SPR throughput (fps) | Xeon 5218 throughput |
|------------|----------------------|----------------------|
| old config | 3200 fps            | 240 fps             |  
| new config | 3600 fps            | 410 fps             |  
|            |                      |                      |

AlexKoff88 (approved):

echarlaix (approved):
Thanks @yujiepan-work ! Failing tests are unrelated so will merge now

PR #322, opened 2023-05-25
add TSModelForSeq2SeqLM for enabling jit trace on text2text-generation

jiqing-feng:
Hi @echarlaix .  
Thanks for your support. The previous [PR](https://github.com/huggingface/optimum-intel/pull/300) enables jit trace on text-generation has been merged. 

In this PR, we create a new class TSModelForSeq2SeqLM to enable jit trace on text2text-generation. I tested it on google/flan-t5-base and it will bring 10% speed-up. In this class, we traced the encoder and decoder separately and we also need another decoder called decoder_pkv if we use past_key_values. 

We use an empty tensor to mock past_key_values in text-generation as shown [there](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/generation/modeling.py#L148) because the past_key_values are only cat with other tensor, it will not impact the result if we set past_key_values as empty tensor. 

However, in text2text-generation, we cannot avoid this difference because past_key_values will change the hidden_states, like in [T5_model](https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5/modeling_t5.py#L497-L515). That's why we need 2 decoders (decoder and decoder_pkv) to handle this problem. 

Would you please help me review it? Would like your opinion. Thanks! 

BTW, the failed tests may be caused by pytorch version since jit trace with dict inputs have a bug in torch<=2.0.1, but it has been fixed by [this_PR](https://github.com/pytorch/pytorch/pull/97866). So I can successfully run all the tests locally with torch==2.1.0 and ipex==2.1.0. Do you mind upgrading the lib version in the tests?

cc @sywangyi


jiqing-feng:
Hi @echarlaix . I have fixed the failed checks. The jit_trace function will return the eager mode models if the torch version is lower than `2.1.0` in text2text-generation. Could you please help me review it? Thx!

optimum/intel/generation/modeling.py
```diff
@@ -403,3 +472,553 @@ def _convert_to_standard_cache(
...
+            self.decoder_start_token_id = kwargs.get("decoder_start_token_id", None)
+        elif config.decoder_start_token_id is not None:
+            self.decoder_start_token_id = config.decoder_start_token_id
+        elif config.bos_token_id is not None:
+            self.decoder_start_token_id = config.bos_token_id
+
+        if self.decoder_start_token_id is None:
+            logger.warning("Please assign the decoder_start_token_id, it may cause unexpected errors.")
+
+        if is_transformers_version("<=", "4.25.1"):
```
echarlaix, line 520: This can be removed as `[optimum](https://github.com/huggingface/optimum/blob/v1.8.6/setup.py#L18)` now support `transformers>=4.26.0`

optimum/intel/generation/modeling.py
```diff
@@ -403,3 +472,553 @@ def _convert_to_standard_cache(
...
+                decoder_pkv(**dummy_inputs_3)
+
+        return cls(
+            decoder=decoder,
+            encoder=encoder,
+            decoder_pkv=decoder_pkv,
+            config=config,
+            model_save_dir_1=model_save_dir_1,
+            model_save_dir_2=model_save_dir_2,
+            model_save_dir_3=model_save_dir_3,
```
echarlaix, line 703: Not used so would remove

optimum/intel/generation/modeling.py
```diff
@@ -403,3 +472,553 @@ def _convert_to_standard_cache(
...
+    export_feature = "text2text-generation"
+    main_input_name = "input_ids"
+    base_model_prefix = "torch_script_model"
+
+    def __init__(
+        self,
+        decoder,
+        encoder,
+        config: PretrainedConfig = None,
+        decoder_pkv: Optional[torch.jit.RecursiveScriptModule] = None,
```
echarlaix, line 488: we shouldn't have to export the decoder twice but support both cases (with and without cache) with only one as done in https://github.com/huggingface/optimum-intel/blob/c48f06ad3303df8f1c530db6f953c23f6e71553c/optimum/intel/generation/modeling.py#L150

tests/generation/test_modeling.py
```diff
@@ -145,3 +155,120 @@ def test_compare_with_and_without_past_key_values(self):
             f"With pkv latency: {with_pkv_timer.elapsed:.3f} ms, without pkv latency: {without_pkv_timer.elapsed:.3f} ms,"
             f" speedup: {without_pkv_timer.elapsed / with_pkv_timer.elapsed:.3f}",
         )
+
+
+class TSModelForSeq2SeqLMTest(unittest.TestCase):
+    SUPPORTED_ARCHITECTURES = ("t5",)
```
echarlaix, line 161: thnaks for adding the tests ! Could we add tests for additionnal architectures ? (like bart, mbart, ..)

tests/generation/test_modeling.py
```diff
@@ -12,22 +12,32 @@
...
-from optimum.intel.generation.modeling import TSModelForCausalLM
+from optimum.intel.generation.modeling import TSModelForCausalLM, TSModelForSeq2SeqLM
 
 
 MODEL_NAMES = {
     "bloom": "hf-internal-testing/tiny-random-BloomModel",
     "gptj": "hf-internal-testing/tiny-random-gptj",
     "gpt2": "hf-internal-testing/tiny-random-gpt2",
     "gpt_neo": "hf-internal-testing/tiny-random-GPTNeoModel",
+    "t5": "google/flan-t5-base",
```
echarlaix, line 40: We should use a tiny model so that the tests are not too computationally expensive
```suggestion
    "t5": "hf-internal-testing/tiny-random-t5",
```

PenghuiCheng:
Close this first, will try to use one decoder model for both w/o the past_key_values input.

PR #323, opened 2023-05-26
Rewrite seq2seq model class for neural_compressor

PenghuiCheng:
# What does this PR do?
This PR is based on https://github.com/huggingface/optimum-intel/pull/318. In this PR, we create a new class INCModelForSeq2SeqLM, in this class, we support loading a quantized model with neural compressor and tracing the model for inference for text2text generation. We will also support neural engine backend inference in this class in the future.
In this class, if you want jit trace the model, we will create three jit models: encoder, decoder and decoder_with_past(if with past_key_values). If without past_key_values, then only two jit models be created: encoder and decoder.

BTW, the example may be failed if using pytorch version <=2.0.1 since jit trace with dict inputs have a bug, but it has been fixed by https://github.com/pytorch/pytorch/pull/97866. So the example needs pip install nightly version torch.

## Before submitting


optimum/intel/neural_compressor/modeling_base_seq2seq.py
```diff
@@ -0,0 +1,419 @@
...
+        model_dim = None
+        model_parallel = None
+
+        decoder_with_past = None
+        if config.torchscript:
+            # Load model from a local directory
+            if os.path.isdir(model_id):
+                encoder = cls.load_model(os.path.join(model_id, encoder_file_name))
+                decoder = cls.load_model(os.path.join(model_id, decoder_file_name))
+                decoder_with_past = (
```
echarlaix, line 189: Why do we need a `decoder_with_past` ? We should keep only one decoder which would handle both cache and not

optimum/intel/neural_compressor/modeling_seq2seq.py
```diff
@@ -0,0 +1,344 @@
...
+            The PyTorch model associated to the decoder.
+        decoder_with_past (`JIT model or PyTorch model`):
+            The PyTorch model associated  to the decoder with past key values.
+        config (`transformers.PretrainedConfig`):
+            [PretrainedConfig](https://huggingface.co/docs/transformers/main_classes/configuration#transformers.PretrainedConfig)
+            is an instance of the configuration associated to the model. Initializing with a config file does
+            not load the weights associated with the model, only the configuration.
+"""
+
+ENCODER_INPUTS_DOCSTRING = r"""
```
echarlaix, line 52: This is also present for openvino, we should move it and reuse it

optimum/intel/neural_compressor/modeling_seq2seq.py
```diff
@@ -0,0 +1,344 @@
...
+            The encoder `last_hidden_state` of shape `(batch_size, encoder_sequence_length, hidden_size)`.
+        past_key_values (`tuple(tuple(torch.FloatTensor), *optional*)`
+            Contains the precomputed key and value hidden states of the attention blocks used to speed up decoding.
+            The tuple is of length `config.n_layers` with each tuple having 2 tensors of shape
+            `(batch_size, num_heads, decoder_sequence_length, embed_size_per_head)` and 2 additional tensors of shape
+            `(batch_size, num_heads, encoder_sequence_length, embed_size_per_head)`.
+"""
+
+
+TRANSLATION_EXAMPLE = r"""
```
echarlaix, line 96: This is also present for openvino, so same comment

optimum/intel/neural_compressor/modeling_base_seq2seq.py
```diff
@@ -0,0 +1,419 @@
...
+        encoder_file_name: Optional[str] = ENCODER_NAME,
+        decoder_file_name: Optional[str] = DECODER_NAME,
+        decoder_with_past_file_name: Optional[str] = DECODER_WITH_PAST_NAME,
+        local_files_only: bool = False,
+        use_cache: bool = True,
+        torch_dtype: Optional[Union[str, "torch.dtype"]] = None,
+        **kwargs,
+    ):
+        """
+        Loads a model and its configuration file from a directory or the HF Hub.
```
echarlaix, line 148: I would remove the docstring from `_from_pretrained` (which is copy pasted everywhere so we can keep them in "public" method (`from_pretrained` or  `from_transformers`) which will be accessible in the documentation, it's also true for openvino, I will take care of removing it as it's not very readable)

optimum/intel/neural_compressor/modeling_base_seq2seq.py
```diff
@@ -0,0 +1,419 @@
...
+        force_download: bool = False,
+        cache_dir: Optional[str] = None,
+        subfolder: str = "",
+        local_files_only: bool = False,
+        task: Optional[str] = None,
+        use_cache: bool = True,
+        **kwargs,
+    ):
+        """
+        Export a vanilla Transformers model into an ONNX model using `transformers.onnx.export_onnx`.
```
echarlaix, line 316: same comment

optimum/intel/neural_compressor/modeling_decoder.py
```diff
@@ -0,0 +1,299 @@
...
+if is_transformers_version("<", "4.25.0"):
+    from transformers.generation_utils import GenerationMixin
+else:
+    from transformers.generation import GenerationMixin
+
+
+logger = logging.getLogger(__name__)
+
+
+TEXT_GENERATION_EXAMPLE = r"""
```
echarlaix, line 40: same comment

PR #335, opened 2023-06-02
fix: self.config -> self.model.config else throws up error during evaluation in INCSeq2SeqTrainer.

AbhishekSalian:
# What does this PR do?

- The predictions during evaluation are converted into ids. model.config was missing so the prediction was not converted to a token because of the missing hyperparameter in the model config.


echarlaix (approved):
LGTM, thanks a lot for the fix @AbhishekSalian

Issue #327, opened 2023-05-30
OV Bloomz with `use_cache=True` generated output differ from ORT and Pytorch


sammysun0711:
Hi, I met following issue when I compared output results of ```bloomz-560m``` with OV, Pytorch and ORT on CPU.
It seems that Pytorch,ORT and OV without cache generate same outputs, while OV with cache (by default) generate strange outputs: [bloomz_with_cache_issue.log](https://github.com/huggingface/optimum-intel/files/11595122/bloomz_with_cache_issue.log)

```
Input text:  DeepSpeed is a machine learning framework.
Pytorch generated text:             DeepSpeed is a machine learning framework. It is a software tool that uses deep learning to automatically learn from large volumes of data.
ORT generated text:                 DeepSpeed is a machine learning framework. It is a software tool that uses deep learning to automatically learn from large volumes of data.
OV without cache generated text:    DeepSpeed is a machine learning framework. It is a software tool that uses deep learning to automatically learn from large volumes of data.
OV with cache generated text:       DeepSpeed is a machine learning framework. Machine... Machine... Machine... Machine... Machine... Machine... Machine... Machine... Machine... Machine...
```
Here is a code sample to reproduce this issue on latest master: 
```
from transformers import AutoModelForCausalLM, AutoTokenizer
from optimum.intel import OVModelForCausalLM
from optimum.onnxruntime import ORTModelForCausalLM

MODEL_ID = "bigscience/bloomz-560m"
tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
pt_model = AutoModelForCausalLM.from_pretrained(MODEL_ID, use_cache=True)
ort_model = ORTModelForCausalLM.from_pretrained(MODEL_ID, export=True, use_cache=True)
ov_model_with_cache = OVModelForCausalLM.from_pretrained(MODEL_ID, export=True, use_cache=True)
ov_model_without_cache = OVModelForCausalLM.from_pretrained(MODEL_ID, export=True, use_cache=False)
input_text = "DeepSpeed is a machine learning framework."
print("Input text: ", input_text)
inputs = tokenizer(input_text, return_tensors="pt")

pt_gen_tokens = pt_model.generate(**inputs, max_new_tokens=20)
pt_gen_text = tokenizer.batch_decode(pt_gen_tokens, skip_special_tokens=True)[0]

ort_gen_tokens = ort_model.generate(**inputs, max_new_tokens=20)
[...]

PR #340, opened 2023-06-08
Fix OV model for BLOOM architecture

echarlaix:
In this PR we patch the model before the ONNX / OpenVINO export for both bloom and llama architecture in order to fix the decoder mask generation.
Fix https://github.com/huggingface/optimum-intel/issues/327

cc @sammysun0711 @usstq

### DIFF
--- a/.github/workflows/test_openvino.yml
+++ b/.github/workflows/test_openvino.yml
@@ -30,7 +30,6 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install diffusers==0.16.1
         pip install .[openvino,nncf,tests]
     - name: Test with Pytest
       run: |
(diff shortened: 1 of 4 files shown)


eaidova:
@echarlaix  I believe this fix is not complete. function which you patch for llama model,  can be also find in other model types e.g. blanderbot and opt, so I assume that more models can be affected here.

echarlaix:
Hi @eaidova ! Sure let me check it right now and add a fix ! Currently, [here](https://github.com/huggingface/optimum-intel/blob/d95534d374da203ee75111e0eb51ae1bcd431553/tests/openvino/test_modeling.py#L417) are the models architectures we are officially supporting, it could make sense to disable or at least add a warning when exporting other architectures.

PR #342, opened 2023-06-09
small improvements in diffusion pipeline for openvino

eaidova:
# What does this PR do?

This PR propose small improvements for OVStableDiffusionPipeline
1. When diffusion pipeline compiled it logs message about compiling encoder 3 times, which may confuse users. Added name of model part in logs to make these logs more useful (e.g. in case if one of the models filed during compilation, we can understand which one).
2. Model with static shape does not allow to change shape during pipeline call, but user should know current height and width to understand when model requires reshaping. this PR proposes properties current_height and current_width  where user can check current pipeline status

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -339,6 +339,20 @@ class OVStableDiffusionPipeline(OVBaseModel, StableDiffusionPipelineMixin):
     def device(self) -> str:
         return self._device.lower()
 
+    @property
+    def current_height(self) -> int:
+        height = self.unet.model.inputs[0].get_partial_shape()[2]
+        if height.is_dynamic:
+            return -1
+        return height.get_lenght() * self._vae_scale_factor
+
+    @property
+    def current_width(self) -> int:
+        width = self.unet.model.inputs[0].get_partial_shape()[3]
+        if width.is_dynamic:
+            return -1
+        return width.get_lenght() * self._vae_scale_factor
+
     def _reshape_unet(
         self,
         model: openvino.runtime.Model,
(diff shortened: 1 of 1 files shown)


eaidova:
@AlexKoff88  @helena-intel  could you please take a look?

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -339,6 +339,20 @@ def to(self, device: str):
     def device(self) -> str:
         return self._device.lower()
 
+    @property
+    def current_height(self) -> int:
```
AlexKoff88, line 343: Maybe just call properties height and width but I would let @echarlaix to decide here.

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -339,6 +339,20 @@ def to(self, device: str):
...
+        height = self.unet.model.inputs[0].get_partial_shape()[2]
+        if height.is_dynamic:
+            return -1
+        return height.get_lenght() * self._vae_scale_factor
```
helena-intel, line 347:
```suggestion
        return height.get_length() * self._vae_scale_factor
```

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -339,6 +339,20 @@ def to(self, device: str):
...
+
+    @property
+    def current_width(self) -> int:
+        width = self.unet.model.inputs[0].get_partial_shape()[3]
+        if width.is_dynamic:
+            return -1
+        return width.get_lenght() * self._vae_scale_factor
```
helena-intel, line 354:
```suggestion
        return width.get_length() * self._vae_scale_factor
```

helena-intel (commented):
Thanks @eaidova ! Could you also create a test for this in https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_modeling.py ?

optimum/intel/openvino/modeling_diffusion.py
eaidova: thank you, fixed

eaidova:
> Thanks @eaidova ! Could you also create a test for this in https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_modeling.py ?

@helena-intel, thank you, added

optimum/intel/openvino/modeling_diffusion.py
eaidova: renamed

eaidova:
@AlexKoff88  @helena-intel  @echarlaix 

I observe failing on ci, because stable diffusion onnx export failed with torch 2.0.1 (aten::scaled_dot_product_attention is not registered for export. It will be added in next release as it is already exists on pytorch master and in nightly packages. 
I added registering custom operation for that as workaround, please take a look, if it is not acceptable, I can remove it

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -285,6 +286,12 @@ def _from_transformers(
         feature_extractor: Optional["CLIPFeatureExtractor"] = None,
         **kwargs,
     ):
+        import torch
+
+        if packaging.version.parse(torch.__version__) > packaging.version.parse("1.13.1") and packaging.version.parse(
+            torch.__version__
+        ) < packaging.version.parse("2.1.0"):
```
echarlaix, line 293: You can also use [`is_torch_version`](https://github.com/huggingface/optimum-intel/blob/86cdea9893493c106d3be516c7a47c9acc4c8a9e/optimum/intel/utils/import_utils.py#L176):
```suggestion

        if is_torch_version(">", "1.13.1") and is_torch_version("<=", "2.0.1"):
```

AlexKoff88:
@echarlaix, can you please take a look what is wrong with the documentation build?

echarlaix (approved):
Thanks a lot @eaidova ! 

> @echarlaix, can you please take a look what is wrong with the documentation build?

The documentation generation was fixed in https://github.com/huggingface/optimum-intel/pull/343

PR #346, opened 2023-06-12
Add support of OVModel in quantizer

AlexKoff88:
- Support OVModel in the optimum OVQuantizer as a more robust approach for PTQ

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -23,7 +23,6 @@ from huggingface_hub import hf_hub_download
 from huggingface_hub.utils import EntryNotFoundError
 from openvino._offline_transformations import apply_moc_transformations, compress_model_transformation
 from openvino.runtime import Core
-from openvino.tools import mo
 from transformers import PretrainedConfig
 from transformers.file_utils import add_start_docstrings
 
@@ -246,10 +245,7 @@ class OVBaseModel(PreTrainedModel):
             model_save_dir = Path(model_cache_path).parent
             file_name = file_names[0]
 
-        if from_onnx:
-            model = mo.convert_model(file_name)
-        else:
-            model = cls.load_model(file_name)
+        model = cls.load_model(file_name)
         return cls(model, config=config, model_save_dir=model_save_dir, **kwargs)
 
     @classmethod
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -23,6 +23,7 @@ from huggingface_hub import hf_hub_download
 from huggingface_hub.utils import EntryNotFoundError
 from openvino._offline_transformations import apply_moc_transformations, compress_model_transformation
 from openvino.runtime import Core
+from openvino.tools import mo
 from transformers import PretrainedConfig
 from transformers.file_utils import add_start_docstrings
 
@@ -187,7 +188,6 @@ class OVBaseModel(PreTrainedModel):
                     "The file names `ov_model.xml` and `ov_model.bin` will be soon deprecated."
                     "Make sure to rename your file to respectively `openvino_model.xml` and `openvino_model.bin`"
                 )
-            model = cls.load_model(file_name)
             model_save_dir = model_id
         # Download the model from the hub
         else:
@@ -227,7 +227,12 @@ class OVBaseModel(PreTrainedModel):
                     "Make sure to rename your file to respectively `openvino_model.xml` and `openvino_model.bin`"
                 )
             model_save_dir = Path(model_cache_path).parent
-            model = cls.load_model(file_names[0])
+            file_name = file_names[0]
+            
+        if from_onnx:
+            model = mo.convert_model(file_name)
+        else:
+            model = cls.load_model(file_name)
         return cls(model, config=config, model_save_dir=model_save_dir, **kwargs)
 
     @classmethod
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -228,7 +228,7 @@ class OVBaseModel(PreTrainedModel):
                 )
             model_save_dir = Path(model_cache_path).parent
             file_name = file_names[0]
-            
+
         if from_onnx:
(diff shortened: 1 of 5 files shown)


optimum/intel/openvino/quantization.py
```diff
@@ -137,11 +140,83 @@ def quantize(
...
+            calibration_dataset=calibration_dataset,
+            batch_size=batch_size,
+            remove_unused_columns=remove_unused_columns,
+            data_collator=data_collator,
+        )
+
+        quantization_dataset = nncf.Dataset(calibration_dataloader, lambda x: x)
+        quantized_model = nncf.quantize(
+            self.model.model,
+            quantization_dataset,
```
echarlaix, line 192: Is there a possibility to enable the pre-computed pkv depending on `config.use_cache` for text generation ?

echarlaix (approved):
It looks great, thanks a lot for the addition @AlexKoff88

optimum/intel/openvino/quantization.py
```diff
@@ -104,9 +106,10 @@ def quantize(
         save_directory: Union[str, Path],
         quantization_config: OVConfig = None,
         file_name: Optional[str] = None,
-        batch_size: int = 8,
+        batch_size: int = 1,
```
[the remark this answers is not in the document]
AlexKoff88: Yes, mostly for safety reasons, especially for LLMs. It should not hurt performance since it is just a matter of 300 inferences by default and batch size does help here that much.

optimum/intel/openvino/quantization.py
AlexKoff88: @echarlaix, I've added support of OVModelForCausalLM with pkv. I also need to change the way how the model export works (added `mo.convert_model()` for ONNX -> OV IR). This is a more safe way and we faced a lot of problems before when using read_model or ONNX directly.

optimum/intel/openvino/quantization.py
```diff
@@ -137,11 +141,145 @@ def quantize(
...
+        )
+
+        quantization_dataset = nncf.Dataset(calibration_dataloader, lambda x: x)
+        quantized_model = nncf.quantize(
+            self.model.model,
+            quantization_dataset,
+            model_type=nncf.ModelType.TRANSFORMER if not kwargs.get("model_type") else kwargs.get("model_type"),
+            fast_bias_correction=True
+            if not kwargs.get("fast_bias_correction")
+            else kwargs.get("fast_bias_correction"),
```
echarlaix, line 201: it looks equivalent to:
```suggestion
            fast_bias_correction=kwargs.get("fast_bias_correction", True),
```

optimum/intel/openvino/quantization.py
```diff
@@ -137,11 +141,145 @@ def quantize(
...
+        calibration_dataloader = self._get_calibration_dataloader(
+            calibration_dataset=calibration_dataset,
+            batch_size=batch_size,
+            remove_unused_columns=remove_unused_columns,
+            data_collator=data_collator,
+        )
+
+        # Prefeth past_key_values
+        self.model.compile()
+        subset_size = 300 if not kwargs.get("subset_size") else kwargs.get("subset_size")
```
echarlaix, line 229:
```suggestion
        subset_size = kwargs.get("subset_size", 300 )
```

optimum/intel/openvino/quantization.py
```diff
@@ -137,11 +141,145 @@ def quantize(
...
+            fast_bias_correction=True
+            if not kwargs.get("fast_bias_correction")
+            else kwargs.get("fast_bias_correction"),
+            subset_size=300 if not kwargs.get("subset_size") else kwargs.get("subset_size"),
+            **kwargs,
+        )
+        self.model.model = quantized_model
+        self.model.save_pretrained(save_directory)
+
+    def _quantize_ovcausallm(
```
echarlaix, line 208: thanks for adding it !

echarlaix (approved):

optimum/intel/openvino/quantization.py
```diff
@@ -137,11 +141,145 @@ def quantize(
...
+
+        quantization_dataset = nncf.Dataset(calibration_dataloader, lambda x: x)
+        quantized_model = nncf.quantize(
+            self.model.model,
+            quantization_dataset,
+            model_type=nncf.ModelType.TRANSFORMER if not kwargs.get("model_type") else kwargs.get("model_type"),
+            fast_bias_correction=True
+            if not kwargs.get("fast_bias_correction")
+            else kwargs.get("fast_bias_correction"),
+            subset_size=300 if not kwargs.get("subset_size") else kwargs.get("subset_size"),
```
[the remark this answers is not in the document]
AlexKoff88: I agree that `num_samples` is better but I am just bypassing parameters to the underlying NNCF PTQ API here. In case I rename it I will need to introduce mapping.
Regarding the warning, this is what NNCF should throw.

AlexKoff88:
@echarlaix, I am done with PR. Just to note that there will be a follow-up PR with quantization config and test update once we fix one NNCF issue that blocks quantizing AutoModels.

optimum/intel/openvino/modeling_base.py
```diff
@@ -114,11 +114,28 @@ def load_model(file_name: Union[str, Path]):
...
+            return model
+
         if isinstance(file_name, str):
             file_name = Path(file_name)
         bin_file_name = file_name.with_suffix(".bin") if file_name.suffix == ".xml" else None
 
-        return core.read_model(file_name, bin_file_name)
+        model = core.read_model(file_name, bin_file_name)
+        if file_name.suffix == ".onnx":
+            model = fix_op_names_duplicates(model)  # should be called during model conversion to IR
```
echarlaix, line 136: Should  `fix_op_names_duplicates` be called when converting an ONNX model to OpenVINO or is this needed to enable NNCF quantization on an OpenVINO model ?

optimum/intel/openvino/quantization.py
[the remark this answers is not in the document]
echarlaix: Ok perfect for the warning !

echarlaix (approved):

PR #350, opened 2023-06-13
Fix PR doc build from forks

regisss:
# What does this PR do?

The doc needs to be pushed from a workflow which is triggered within the HF org.

## Before submitting


echarlaix (approved):
LGTM, thanks a lot for the fix @regisss !!

PR #354, opened 2023-06-16
Improved performance of decoders

AlexKoff88:
Improved performance of decoders: GPT-like, bloom, etc., and Seq2seq models.
Observed a significant speedup of long sequences, e.g. Dolly-3B, 500 tokens: +30% when running on a client CPU.

### DIFF
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -310,13 +310,12 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
         # Run inference
         self.request.start_async(inputs, shared_memory=True)
         self.request.wait()
-        output = self.request.results
 
-        logits = torch.from_numpy(output["logits"]).to(self.device)
+        logits = torch.from_numpy(self.request.get_tensor("logits").data).to(self.device)
 
         if self.use_cache:
             # Tuple of length equal to : number of layer * number of past_key_value per decoder layer (2 corresponds to the self-attention layer)
-            past_key_values = tuple(output[key] for key in self.key_value_output_names)
+            past_key_values = tuple(self.request.get_tensor(key).data for key in self.key_value_output_names)
             # Tuple of tuple of length `n_layers`, with each tuple of length equal to 2 (k/v of self-attention)
             past_key_values = tuple(
                 past_key_values[i : i + self.num_pkv] for i in range(0, len(past_key_values), self.num_pkv)
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -239,10 +239,8 @@ class OVBaseDecoderModel(OVModel):
 
     def compile(self):
         if self.request is None:
-            logger.info("Compiling the model...")
-            cache_dir = Path(self.model_save_dir).joinpath("model_cache")
-            ov_config = {**self.ov_config, "CACHE_DIR": str(cache_dir)}
-            self.request = core.compile_model(self.model, self._device, ov_config).create_infer_request()
+            super().compile()
+            self.request = self.request.create_infer_request()
 
 
 @add_start_docstrings(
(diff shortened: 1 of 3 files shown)


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -301,15 +308,14 @@ def forward(
             inputs["attention_mask"] = np.array(attention_mask)
 
         # Run inference
-        outputs = self.request(inputs, shared_memory=True)
+        self.request.start_async(inputs, shared_memory=True)
```
helena-intel, line 311: Is there any danger in using shared_memory in combination with async in this case?

AlexKoff88: it is not truly the async since we wait for the results right after the call, it is more a way to access the results without copying them (this is what you get using the current synchronous API).

optimum/intel/openvino/quantization.py
```diff
@@ -234,17 +235,26 @@ def __init__(self, request):
...
             def __getattr__(self, attr):
                 if attr in self.__dict__:
                     return getattr(self, attr)
                 return getattr(self.request, attr)
 
         self.model.request = InferRequestWrapper(self.model.request)
-        for i, data in enumerate(calibration_dataloader):
-            self.model.generate(**data, max_new_tokens=10)
+        for _, data in enumerate(calibration_dataloader):
+            self.model.generate(**data, max_new_tokens=100)
```
echarlaix, line 257: Is this modification added to reduce accuracy degradation resulting from quantization? If yes, what did you observe when varying this parameter ?

optimum/intel/openvino/quantization.py
```diff
@@ -234,17 +235,26 @@ def __init__(self, request):
...
+
             def __getattr__(self, attr):
                 if attr in self.__dict__:
                     return getattr(self, attr)
                 return getattr(self.request, attr)
 
         self.model.request = InferRequestWrapper(self.model.request)
-        for i, data in enumerate(calibration_dataloader):
-            self.model.generate(**data, max_new_tokens=10)
+        for _, data in enumerate(calibration_dataloader):
```
echarlaix, line 256: Not related to the PR, but what do you think about uniformizing how quantization is applied on causal langage models depending on whether the user gives a `torch.nn.Module` or a `OVBaseDecoderModel` (the number of generation steps is currently not the same). We could also instantiate an `OVModel` in the `from_pretrained` method when the given model is a `PreTrainedModel`

echarlaix (approved):

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -237,6 +237,13 @@ def reshape(self, batch_size: int, sequence_length: int):
         logger.warning("Static shapes are not supported for causal language model.")
         return self
 
+    def compile(self):
+        if self.request is None:
+            logger.info("Compiling the model...")
+            cache_dir = Path(self.model_save_dir).joinpath("model_cache")
+            ov_config = {**self.ov_config, "CACHE_DIR": str(cache_dir)}
+            self.request = core.compile_model(self.model, self._device, ov_config).create_infer_request()
```
echarlaix, line 245: could be replaced by : 
```suggestion
            super().compile()
            self.request = self.request.create_infer_request()
```

optimum/intel/openvino/quantization.py
AlexKoff88: In the process, and will update a bit later.

optimum/intel/openvino/modeling_decoder.py
helena-intel: Ah, thanks! This is a great solution for now. I heard sync inference in OpenVINO should be improved in the next release, with an option to not return the data.

optimum/intel/openvino/quantization.py
AlexKoff88: It is hard to accomplish this with the current NNCF PTQ API implementation we have for PyTorch. I think we should deprecate PTQ for PyTorch at some point because it also introduces ambiguity for the user about what workflow to use for quantization.

AlexKoff88:
I think it is ready for merge.

helena-intel (approved):
🚀 🔥

echarlaix (approved):
LGTM, thanks a lot @AlexKoff88

PR #357, opened 2023-06-21
8-bit weight compression

AlexKoff88:
- Introduced data-free weight compression to 8 bits.
- It is lossless for LLMs and reduces the footprint significantly
- Runtime optimization will be available in the next release of OV
- Reworked quantization


optimum/intel/openvino/quantization.py
```diff
@@ -330,25 +329,86 @@ def _quantize_torchmodel(
...
         # Load and save the compressed model
-        model = core.read_model(f) if save_as_external_data else core.read_model(f.getvalue(), b"")
+        model = core.read_model(onnx_path)
         self._save_pretrained(model, output_path)
         quantization_config.save_pretrained(save_directory)
+        if not quantization_config.save_onnx_model:
+            os.remove(onnx_path)
+
+    def compress_weights(
+        self, save_directory: Union[str, Path], quantization_config: OVConfig = None, file_name: Optional[str] = None
```
helena-intel, line 357: Should the INT8_WEIGHT_COMPRESSION_CONFIG be the default quantization config? I tried gpt2 and bloom-560m without specifying a config and got very bad model outputs.

optimum/intel/openvino/quantization.py
```diff
@@ -330,25 +329,86 @@ def _quantize_torchmodel(
...
+            output=onnx_path,
+        )
 
         # Load and save the compressed model
-        model = core.read_model(f) if save_as_external_data else core.read_model(f.getvalue(), b"")
+        model = core.read_model(onnx_path)
         self._save_pretrained(model, output_path)
         quantization_config.save_pretrained(save_directory)
+        if not quantization_config.save_onnx_model:
+            os.remove(onnx_path)
```
helena-intel, line 354: It seems this removes the ONNX model, but not the model.onnx_data

helena-intel (commented):
This looks amazing! 

Model outputs with the INT8_WEIGHT_COMPRESSION_CONFIG looked good for me on gpt2 (though not the same as the fp32 model), but not with bloom-560m. Is this expected to work on all models?

The compress_weights() function shows a lot of log lines like  `INFO:nncf:Not adding activation input quantizer for operation`. But that's expected here. Would it be an option to make this a debug level logger in NNCF?

optimum/intel/openvino/quantization.py
AlexKoff88: export(...) already does most of the work, I need only read from the .onnx and remove it if it is not needed. 
There reason, why I save to the file, comes from LLMs, this helps to avoid some issues. Anyway, next month we are planning to replace it with direct export from pytorch to OV model which speed ups the whole export process significantly. The work on the OV side is almost done.

optimum/intel/openvino/quantization.py
AlexKoff88: It is the default actually in this method. Can you share the model id for which you get poor results?

optimum/intel/openvino/quantization.py
helena-intel: That's great news! But I didn't mean that the ONNX saving is problematic, it's just that for large models this also creates a model.onnx_data which is not removed. I think we could just add:
```python
try:
    os.remove(f"{onnx_path}_data")
except FileNotFoundError:
    pass
```

optimum/intel/openvino/quantization.py
AlexKoff88: This is the initial implementation that is based on the legacy NNCF API to unblock LLMs acceleration. We are working on a more unified one in the NNCF and will plug it in once it is done.

optimum/intel/openvino/quantization.py
```diff
@@ -306,6 +301,7 @@ def _quantize_torchmodel(
             data_collator=data_collator,
         )
         model_inputs = next(iter(calibration_dataloader))
+        print("model_inputs: ", model_inputs)
```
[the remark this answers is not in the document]
AlexKoff88: removed

optimum/intel/openvino/quantization.py
```diff
@@ -330,25 +328,90 @@ def _quantize_torchmodel(
...
+            quantization_config (`OVConfig`, *optional*):
+                The configuration containing the parameters related to quantization.
+            file_name (`str`, *optional*):
+                The model file name to use when saving the model. Overwrites the default file name `"model.onnx"`.
+
+        Example:
+        ```python
+        >>> from optimum.intel.openvino import OVQuantizer, OVModelForCausalLM
+        >>> from transformers import AutoModelForSequenceClassification
+        >>> model = OVModelForCausalLM.from_pretrained("databricks/dolly-v2-3b")
```
echarlaix, line 378: Shouldn't it be :
```suggestion
        >>> model = AutoModelForCausalLM.from_pretrained("databricks/dolly-v2-3b")
```
as `_quantize_torchmodel` is called in `compress_weights` ?

optimum/intel/openvino/quantization.py
```diff
@@ -330,25 +328,90 @@ def _quantize_torchmodel(
...
+                The directory where the quantized model should be saved.
+            quantization_config (`OVConfig`, *optional*):
+                The configuration containing the parameters related to quantization.
+            file_name (`str`, *optional*):
+                The model file name to use when saving the model. Overwrites the default file name `"model.onnx"`.
+
+        Example:
+        ```python
+        >>> from optimum.intel.openvino import OVQuantizer, OVModelForCausalLM
+        >>> from transformers import AutoModelForSequenceClassification
```
echarlaix, line 377:
```suggestion
        >>> from transformers import AutoModelForCausalLM
```

optimum/intel/openvino/quantization.py
```diff
@@ -330,25 +328,90 @@ def _quantize_torchmodel(
...
         self._save_pretrained(model, output_path)
         quantization_config.save_pretrained(save_directory)
+        if not quantization_config.save_onnx_model:
+            os.remove(onnx_path)
+            try:
+                os.remove(f"{onnx_path}_data")
+            except FileNotFoundError:
+                pass
+
+    def compress_weights(
```
echarlaix, line 359: I would prefer to add this in the `.quantize()` method, enabling it with an argument such as `weights_only` or `disable_activation`

echarlaix (approved):
Really nice PR, thanks a lot @AlexKoff88 !

optimum/intel/openvino/quantization.py
echarlaix: Also in the case where a user gives a config should we verify the activations `ignored_scopes` value ? Otherwise quantization could be applied on the activations when this is unpexpected no?

tests/openvino/test_quantization.py
```diff
@@ -142,6 +143,41 @@ def preprocess_function(examples, tokenizer):
...
+            transformers_model = model_cls.auto_model_class.from_pretrained(model_name)
+            tokenizer = AutoTokenizer.from_pretrained(model_name)
+            if tokenizer.pad_token is None:
+                tokenizer.pad_token = tokenizer.eos_token
+
+            quantizer = OVQuantizer.from_pretrained(transformers_model, task=task)
+            quantizer.compress_weights(save_directory=tmp_dir)
+            model = model_cls.from_pretrained(tmp_dir)
+
+            # TODO: uncomment once move to a newer version of NNCF which has some fixes
```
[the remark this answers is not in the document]
AlexKoff88: I meant fixes that are already available in the NNCF from GitHub that allows properly handling some torch operations, e.g. "baddmm"

optimum/intel/openvino/quantization.py
AlexKoff88: thanks!

tests/openvino/test_quantization.py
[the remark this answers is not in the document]
AlexKoff88: Updated the comment

optimum/intel/openvino/quantization.py
AlexKoff88: There are several works on the mixing sparsity + quantization to compress models in the post-training mode. At some point, we can extend the implementation.

AlexKoff88: > Also in the case where a user gives a config should we verify the activations `ignored_scopes` value ? Otherwise quantization could be applied on the activations when this is unpexpected no?

You are right here, for now, it is possible to hack this method using quantization_config. As I mentioned we will pug a new API in the future. I added `quantization_config` for experimenting purposes mostly, but no problem to delete it from here. What do you think?

tests/openvino/test_quantization.py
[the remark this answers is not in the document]
echarlaix: Thanks for the explanation !

optimum/intel/openvino/quantization.py
echarlaix: > It may work. The thing is that in NNCF it is supposed to be a separate API call (implementation is in progress). @echarlaix, what are the benefits of having all in one API, in your opinion?


Currently we try to unify the quantization API of the different optimum subpackages in order for users to be able to easily transition from one to another (`ORTQuantizer` -> `OVQuantizer`)
We plan to create 
Having a close behavior among the quantization API of the different optimum subpackages makes it easier for users 


We want the quantization API of the different optimum subpackages to have a behave as close as possible in order 


so that it's easy for users to transition from one subpackageto the other, 
Currently we try to unify the quantization API of the different optimum subpackages in order for users to be able to easily transition from one to another (`ORTQuantizer` -> `OVQuantizer`)

echarlaix: > There are several works on the mixing sparsity + quantization to compress models in the post-training mode. At some point, we can extend the implementation.

Super interesting ! Happy to help on the integration if needed

echarlaix: > > Also in the case where a user gives a config should we verify the activations `ignored_scopes` value ? Otherwise quantization could be applied on the activations when this is unpexpected no?
> 
> You are right here, for now, it is possible to hack this method using quantization_config. As I mentioned we will pug a new API in the future. I added `quantization_config` for experimenting purposes mostly, but no problem to delete it from here. What do you think?

I agree with the current integration, it makes sense to leave the users the possibility to give a config (as currently the case for weights and activations quantization). Something that we can do is overwritte the config by adding the default patterns when missing and add a warning, what do you think ? Also after a quick look to the repo, I can see that all the examples config are for QAT / JPQD and our PTQ documentation / examples all use the default configuration, so it's not as likely as I thought originally, happy to do it in a following PR if easier for you

helena-intel (approved):

optimum/intel/openvino/quantization.py
echarlaix: Done in https://github.com/huggingface/optimum-intel/pull/359

PR #358, opened 2023-06-25
fix bug of cli <recipe>

qgao007:
# What does this PR do?

This PR fixed the bug with CLI parameter <recipes>. <recipes> has to be defaulted to empty {} if not provided in command line.


echarlaix (approved):
Thanks for the fix @qgao007

PR #362, opened 2023-06-28
Allow user to set the CACHE_DIR

### LINKED ISSUE #352: Add an option to disable OpenVINO model caching
Add an option to disable model caching and to override the CACHE_DIR directory when loading an `OVModel` 

cc @helena-intel

peterchen-intel:
1. user can set CACHE_DIR to empty and disable the model cache
2. or user can set the CACHE_DIR to another folder

# What does this PR do?

Fixes #352 

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -332,8 +332,12 @@ class OVBaseModel(PreTrainedModel):
     def compile(self):
         if self.request is None:
             logger.info("Compiling the model...")
-            cache_dir = Path(self.model_save_dir).joinpath("model_cache")
-            ov_config = {**self.ov_config, "CACHE_DIR": str(cache_dir)}
+            ov_config = {**self.ov_config}
+            if "CACHE_DIR" not in ov_config.keys():
+                cache_dir = Path(self.model_save_dir).joinpath("model_cache")
+                ov_config["CACHE_DIR"] = str(cache_dir)
+                logger.info(f"Set CACHE_DIR to {str(cache_dir)}")
+
             self.request = core.compile_model(self.model, self._device, ov_config)
 
     def _reshape(
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -85,6 +85,11 @@ class OVBaseModel(PreTrainedModel):
         self._device = device.upper()
         self.is_dynamic = dynamic_shapes
         self.ov_config = ov_config if ov_config is not None else {"PERFORMANCE_HINT": "LATENCY"}
+
+        if "CACHE_DIR" not in self.ov_config.keys():
+            cache_dir = Path(self.model_save_dir).joinpath("model_cache")
+            self.ov_config["CACHE_DIR"] = str(cache_dir)
+            logger.info(f"Set CACHE_DIR to {str(cache_dir)}")
         self.preprocessors = kwargs.get("preprocessors", [])
         enable_compilation = kwargs.get("compile", True)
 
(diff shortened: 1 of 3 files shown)


helena-intel:
Thanks @peterchen-intel ! CACHE_DIR is also set in modeling_diffusion.py and modeling_seq2seq.py. 

We should add a docstring for this method and add this to documentation, but we can do that in a separate PR. It was not obvious to me that setting CACHE_DIR to an empty string would disable caching, but setting it to None would result in a segmentation fault.

optimum/intel/openvino/modeling_base.py
```diff
@@ -332,8 +332,12 @@ def _from_transformers(
     def compile(self):
         if self.request is None:
             logger.info("Compiling the model...")
-            cache_dir = Path(self.model_save_dir).joinpath("model_cache")
-            ov_config = {**self.ov_config, "CACHE_DIR": str(cache_dir)}
+            ov_config = {**self.ov_config}
```
helena-intel, line 335:
```suggestion
            ov_config = self.ov_config
```

echarlaix: Agreed, also what do you think about modifying `self.ov_config` directly so that we don't do this for each compilation step ?

peterchen-intel: @helena-intel  Do you agree to move this logic to __init__?

echarlaix: Would definitely make sense to me, wdyt @helena-intel ?

helena-intel: I have no preference on implementation, but IMO setting a config belongs to model compile, not model init (in OpenVINO the config is also set on compile_model, not read_model). So you should be able to do `model.compile(ov_config={"INFERENCE_PRECISION_HINT":"f32"})` and still get automatic model caching.

Vipitis:
would it be possible to set it to the huggingface cache, so exported and cached models are discovered by `from_pretrained` directly? As well as show up when looking at `huggingface-cli scan-cache` ?
Apart from that is the exported onnx model still required after caching the model, I suspect that might be device specific. The ONNX files seem to be twice the size as well as the cache. So you end up with 400% of additional storage needed.

optimum/intel/openvino/modeling_base.py
peterchen-intel: @helena-intel This change (set default CACHE_DIR in __init__) doesn't impact the further implementation (if it will happen) to overwrite any ov_config with compile() API.

helena-intel: It's not uncommon to want to change ov_config after loading the model, for example:

```python
model = OVModelFor....from_pretrained(model_id)
# do things
model.ov_config = {"INFERENCE_PRECISION_HINT": "f32"}
model.to("GPU")
model.compile()
```

For example in a gradio app where people can select different options, or in a benchmarking script where you want to iterate over several ov_config values.

With the new implementation this will have the side effect that there is no model caching anymore. Since the effect of caching is so large, and also since some people include model loading in "total inference time" (which is not good practice, but that's a separate discussion), it is important that we always enable caching by default, so I prefer the current behaviour where  model.compile() ensures that CACHE_DIR is set, with the addition that this should be modifyable. Maybe it can just be e.g. `self.ov_config.setdefault("CACHE_DIR", default_cache_dir)` in model.compile()?

echarlaix (approved):

echarlaix:
To fix the quality test, please run the `make style` [command](https://github.com/huggingface/optimum-intel/blob/v1.9.3/Makefile#L27).
Also to install the required dependencies you can run `pip install .[quality]`

optimum/intel/openvino/modeling_base.py
peterchen-intel: Agree. I don't aware that ov_config is a public member, I don't know why Python doesn't like private member. I changed it back to previous implementation. From my point of view, we should not change self.ov_config in compile.

helena-intel (approved):
Thanks @peterchen-intel this looks good to me!

peterchen-intel:
@echarlaix Can you review and merge?

PR #368, opened 2023-07-03
Added Support for Timm models

sawradip:
# What does this PR do?
This PR modifies the `_from_transformers` method of `OVBaseModel` class using `use_safetensors` and `num_labels`, two parameters that are supported by original `AutoModelForImageClassification.from_pretrained ` method from Huggngface.

This enables loading the Timm models through 'from_pretrained' method on `OVBaseModel` models.

This PR is part of an ongoing GSOC project for seamless integration of TIMM library to intel-optimum, and these parameters are necessary to have enough control to load timm modelsusing existing classes.

Example Usage:
```
import torch
from transformers import AutoConfig
from optimum.intel import OVModelForImageClassification

model_id = "timm/vit_tiny_patch16_224.augreg_in21k"

hf_model = OVModelForImageClassification.from_pretrained(model_id, export=True)
out = hf_model(pixel_values = torch.zeros((5, 3, 224, 224)))
print(out.logits.shape)
```
Test the PR in this [Colab Notebook](https://colab.research.google.com/drive/1CT3d7UKxRfaBS4UmPlHh5lV2GeqaYuqG?usp=sharing)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -290,7 +290,7 @@ class OVBaseModel(PreTrainedModel):
             "subfolder": subfolder,
             "local_files_only": local_files_only,
             "force_download": force_download,
-            "trust_remote_code": trust_remote_code
+            "trust_remote_code": trust_remote_code,
         }
 
         if model_id.startswith("timm/"):
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -261,6 +261,8 @@ class OVBaseModel(PreTrainedModel):
         local_files_only: bool = False,
         task: Optional[str] = None,
         trust_remote_code: bool = False,
+        use_safetensors: bool = None,
+        num_lables: int = None,
         **kwargs,
     ):
         """
@@ -291,8 +293,12 @@ class OVBaseModel(PreTrainedModel):
             "local_files_only": local_files_only,
             "force_download": force_download,
             "trust_remote_code": trust_remote_code,
+            "use_safetensors": use_safetensors
         }
 
+        if num_lables:
+            model_kwargs["num_lables"] = num_lables
+            
         model = TasksManager.get_model_from_task(task, model_id, **model_kwargs)
         model_type = model.config.model_type.replace("_", "-")
 
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -263,12 +263,10 @@ class OVBaseModel(PreTrainedModel):
             "force_download": force_download,
             "trust_remote_code": trust_remote_code,
         }
-
         # Fix the mismatch between timm_config and huggingface_config
         if not os.path.isdir(model_id) and model_info(model_id).library_name == "timm":
             model_kwargs["use_safetensors"] = False
             model_kwargs["num_labels"] = config.num_classes
-            
         model = TasksManager.get_model_from_task(task, model_id, **model_kwargs)
         model_type = model.config.model_type.replace("_", "-")
 
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -19,7 +19,7 @@ from tempfile import TemporaryDirectory
 from typing import Dict, Optional, Union
 
 import openvino
-from huggingface_hub import hf_hub_download
+from huggingface_hub import hf_hub_download, model_info
 from openvino._offline_transformations import apply_moc_transformations, compress_model_transformation
 from openvino.runtime import Core
 from transformers import PretrainedConfig
@@ -264,7 +264,8 @@ class OVBaseModel(PreTrainedModel):
             "trust_remote_code": trust_remote_code,
         }
 
-        if model_id.startswith("timm/"):
+        # Fix the mismatch between timm_config and huggingface_config
+        if not os.path.isdir(model_id) and model_info(model_id).library_name == "timm":
             model_kwargs["use_safetensors"] = False
             model_kwargs["num_labels"] = config.num_classes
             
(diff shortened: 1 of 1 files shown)


AlexKoff88:
thanks @sawradip for the PR.

Is this a pre-work that is required to enable `OVModelForImageClassification`?

Ultimately, we need something like this to be working:
```python
import torch
from transformers import AutoConfig
from optimum.intel import OVModelForImageClassification

model_id = "timm/vit_tiny_patch16_224.augreg_in21k"

ov_model = OVModelForImageClassification.from_transformers(model_id, export=True)

out = ov_model(torch.zeros((1, 3, 224, 224)))
```

sawradip:
Hello! Correct me if I am wrong, but I see that the public method `from_transformers` is not available currently.or do you mean the `from_pretrained` method?

image

sawradip:
@AlexKoff88 I have updated the code as well as the PR. Please take a look and let me know.

optimum/intel/openvino/modeling_base.py
```diff
@@ -264,6 +264,10 @@ def _from_transformers(
             "trust_remote_code": trust_remote_code,
         }
 
+        if model_id.startswith("timm/"):
+            model_kwargs["use_safetensors"] = False
```
echarlaix, line 268: Why only for timm models ?

optimum/intel/openvino/modeling_base.py
```diff
@@ -264,6 +264,10 @@ def _from_transformers(
             "trust_remote_code": trust_remote_code,
         }
 
+        if model_id.startswith("timm/"):
```
echarlaix, line 267: We can check the library_name value of the model card directly instead : 
```suggestion
       if not os.path.isdir(model_id) and huggingface_hub.model_info(model_id).library_name == 'timm'
```

optimum/intel/openvino/modeling_base.py
```diff
@@ -264,6 +264,10 @@ def _from_transformers(
...
+            model_kwargs["use_safetensors"] = False
+            model_kwargs["num_labels"] = config.num_classes
```
echarlaix, line 269: Could you add a comment to explain that this comes from a mismatch between timm and transformers models ?

optimum/intel/openvino/modeling_base.py
```diff
@@ -264,6 +264,10 @@ def _from_transformers(
...
+
```
echarlaix, line 270: Also could you add a test to verify that `timm` models are indeed compatible with `OVModels` ? I wonder if we shouldn't keep the `OVModelForXxx` API for `transformers` models and have our own `timm.create_model` for `timm` models

echarlaix: @AlexKoff88 @helena-intel what do you think ?

helena-intel: I'm not familiar with `timm` yet. I love that optimum-intel allows using transformer models with only minimum changes to code (change AutoModelFor... to OVAutoModelFor...) and it would be great to have that for `timm` too, that existing `timm` users can just change a tiny bit of code and leverage OpenVINO.  Just changing the timm.create() line in the model card from https://huggingface.co/timm/vit_tiny_patch16_224.augreg_in21k for example.

Though if it is possible to load classification models with OVModelForImageClassification, that would be so much easier than "manual" inference. But it seems you would need support for that, with a configured preprocessor for the model. It doesn't seem like this is supported/expected for these models at the moment though, and it seems it can lead to issues to use it this way.

optimum/intel/openvino/modeling_base.py
sawradip: This is perfect @echarlaix ! I was looking for a huggingface-suggested right way to extract the parent repository, and this was what I needed.

optimum/intel/openvino/modeling_base.py
sawradip: > Also could you add a test to verify that `timm` models are indeed compatible with `OVModels` ? I wonder if we shouldn't keep the `OVModelForXxx` API for `transformers` models and have our own `timm.create_model` for `timm` models

This is definitely down the pipeline, as I plan to also add/test some more model optimization on timm models, during my project. Do you suggest test for any specific feature for OVModels? Because, you can see in the accompanying notebook, that after loading through OVModelForImageClassification, the model gives output of correct size.

optimum/intel/openvino/modeling_base.py
echarlaix: Thanks !

optimum/intel/openvino/modeling_base.py
AlexKoff88: @echarlaix, do you mean we should remove the check of "timm"?

echarlaix: Yes not sure to understand why this is done for `timm` models, is this coming from an incompatibility between `transformers` `AutoClasses` and `timm` models ?

optimum/intel/openvino/modeling_base.py
echarlaix: thanks @sawradip ! what do you think about adding tests in [`test_modeling.py`](https://github.com/huggingface/optimum-intel/blob/v1.10.1/tests/openvino/test_modeling.py) to compare the original timm models with one which was exported to OpenVINO ?

optimum/intel/openvino/modeling_base.py
sawradip: Yes. Please see this issue I raised: https://github.com/huggingface/transformers/issues/25282

These are some of the incompatibility. There are some more.

optimum/intel/openvino/modeling_base.py
sawradip: This is a great idea @echarlaix . I will look into creating test cases. For now I am working for an alternate architecture to support more timm models in [this](https://github.com/sawradip/optimum-intel/tree/timm_integration_custom) branch . As soon I can prepare the PR, I will add the tests and let you know.

PR #370, opened 2023-07-05
OVModelForCausalLM pastkv uses bf16 precision

luo-cheng2021:
# What does this PR do?

When openvino is using bf16 in SPR the pastkv inside the inference is already bf16. Because the input/output parameter format is fp32, there must use reorder to convert fp32 to bf16 in input and bf16 to fp32 in output, furthermore from bf16 to fp32 the memory bandwidth will enlarge 1 time. This PR will use preprocess to modify the input/output precision to bf16 when:
1, `INFERENCE_PRECISION_HINT` in ov_config sets to `bf16`, or
2, `INFERENCE_PRECISION_HINT` in environment variable sets to `bf16`, this will help batch testing in SPR.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -110,6 +110,8 @@ class OVBaseDecoderModel(OVModel):
                 "`dynamic_shapes` was set to `False` but static shapes are not supported for causal language model. Please set `dynamic_shapes=True`."
             )
 
+        model = self._try_modify_pastkv_to_lowprecision(model, ov_config, device)
+
         super().__init__(
             model,
             config,
@@ -136,6 +138,50 @@ class OVBaseDecoderModel(OVModel):
                 f"once again with `use_cache={use_cache}` when calling the `from_pretrained` method. "
                 "To export your model, simply set `export=True`."
             )
+
+    def _try_modify_pastkv_to_lowprecision(
+        self, model: openvino.runtime.Model, ov_config: Optional[Dict[str, str]] = None, device: str = "CPU"
+    ):
+        device = device.upper()
+        pastkv_will_use = core.get_property(device, "INFERENCE_PRECISION_HINT")
+        # ov_config["INFERENCE_PRECISION_HINT"] may override the prefer precision
+        if ov_config:
+            user_precision_hint = ov_config.get("INFERENCE_PRECISION_HINT", "")
+            user_mode_hint = ov_config.get("EXECUTION_MODE_HINT", "")
+            if user_precision_hint in STR_TO_OV_TYPE:
+                pastkv_will_use = STR_TO_OV_TYPE[user_precision_hint]
+            elif user_mode_hint.upper() == "ACCURACY":
+                pastkv_will_use = Type.f32
+
+        use_cache = any("past_key_values" in key.get_any_name() for key in model.inputs)
+        self.model_org = model
+        if pastkv_will_use != Type.f32 and use_cache:
+            ppp = PrePostProcessor(model.clone())
+            need_gen = False
+            for key in model.inputs:
+                if "past_key_values" in key.get_any_name() and pastkv_will_use != key.get_element_type():
+                    need_gen = True
+                    ppp.input(key.get_any_name()).tensor().set_element_type(pastkv_will_use)
+            for key in model.outputs:
+                if "present" in key.get_any_name() and pastkv_will_use != key.get_element_type():
+                    need_gen = True
+                    ppp.output(key.get_any_name()).tensor().set_element_type(pastkv_will_use)
+            if need_gen:
+                model = ppp.build()
+
+        return model
+
+    def _save_pretrained(self, save_directory: Union[str, Path]):
+        """
+        Saves the model to the OpenVINO IR format so that it can be re-loaded using the
+        [`~optimum.intel.openvino.modeling.OVModel.from_pretrained`] class method.
+
+        Arguments:
+            save_directory (`str` or `Path`):
+                The directory where to save the model files.
+        """
+        dst_path = os.path.join(save_directory, OV_XML_FILE_NAME)
+        openvino.runtime.serialize(self.model_org, dst_path)
 
     @classmethod
     def _from_transformers(
@@ -250,67 +296,6 @@ class OVBaseDecoderModel(OVModel):
 class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
     export_feature = "text-generation"
     auto_model_class = AutoModelForCausalLM
-
-    def __init__(
-        self,
-        model: openvino.runtime.Model,
-        config: PretrainedConfig = None,
-        device: str = "CPU",
-        dynamic_shapes: bool = True,
-        ov_config: Optional[Dict[str, str]] = None,
-        model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
-        **kwargs,
-    ):
-        model = self._try_modify_pastkv_to_lowprecision(model, ov_config, device, **kwargs)
-        super().__init__(model, config, device, dynamic_shapes, ov_config, model_save_dir, **kwargs)
-
-    def _try_modify_pastkv_to_lowprecision(
-        self, model: openvino.runtime.Model, ov_config: Optional[Dict[str, str]] = None, device: str = "CPU", **kwargs
-    ):
-        device = device.upper()
-        pastkv_will_use = core.get_property(device, "INFERENCE_PRECISION_HINT")
-        # ov_config["INFERENCE_PRECISION_HINT"] may override the prefer precision
-        if ov_config:
-            user_precision_hint = ov_config.get("INFERENCE_PRECISION_HINT", "")
-            user_mode_hint = ov_config.get("EXECUTION_MODE_HINT", "")
-            if user_precision_hint in STR_TO_OV_TYPE:
-                pastkv_will_use = STR_TO_OV_TYPE[user_precision_hint]
-            if user_mode_hint.upper() == "ACCURACY":
-                pastkv_will_use = Type.f32
-
-        use_cache = kwargs.get("use_cache", True)
-        has_pastkv = any("past_key_values" in key.get_any_name() for key in model.inputs)
-        self.model_org = model
-        if pastkv_will_use != Type.f32 and use_cache and has_pastkv:
-            ppp = PrePostProcessor(model.clone())
-            need_gen = False
-            for key in model.inputs:
-                if "past_key_values" in key.get_any_name() and pastkv_will_use != key.get_element_type():
-                    need_gen = True
-                    ppp.input(key.get_any_name()).tensor().set_element_type(pastkv_will_use)
-            for key in model.outputs:
-                if "present" in key.get_any_name() and pastkv_will_use != key.get_element_type():
-                    need_gen = True
-                    ppp.output(key.get_any_name()).tensor().set_element_type(pastkv_will_use)
-            if need_gen:
-                model = ppp.build()
-
-        return model
-
-    def _save_pretrained(self, save_directory: Union[str, Path], file_name: Optional[str] = None, **kwargs):
-        """
-        Saves the model to the OpenVINO IR format so that it can be re-loaded using the
-        [`~optimum.intel.openvino.modeling.OVModel.from_pretrained`] class method.
-
-        Arguments:
-            save_directory (`str` or `Path`):
-                The directory where to save the model files.
-            file_name(`str`, *optional*):
-                The model file name to use when saving the model. Overwrites the default file names.
-        """
-        file_name = file_name if file_name is not None else OV_XML_FILE_NAME
-        dst_path = os.path.join(save_directory, file_name)
-        openvino.runtime.serialize(self.model_org, dst_path, dst_path.replace(".xml", ".bin"))
 
     @add_start_docstrings_to_model_forward(
         INPUTS_DOCSTRING.format("batch_size, sequence_length")
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -251,7 +251,7 @@ class OVBaseDecoderModel(OVModel):
 class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
     export_feature = "text-generation"
     auto_model_class = AutoModelForCausalLM
-    
+
     def __init__(self,
                  model: openvino.runtime.Model,
                  config: PretrainedConfig = None,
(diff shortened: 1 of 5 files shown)


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -248,6 +251,59 @@ def compile(self):
...
+                                     model: openvino.runtime.Model,
+                                     ov_config: Dict[str, str] | None = None,
+                                     device: str = "CPU",
+                                     **kwargs):
+        if device == 'CPU':
+            # if INFERENCE_PRECISION_HINT in ov_config equals to bf16, pastkv will use bf16
+            pastkv_will_use = Type.bf16 if ov_config and ov_config.get("INFERENCE_PRECISION_HINT", "") == "bf16" else Type.f32
+
+            # for batch testsing can set INFERENCE_PRECISION_HINT in enviroment, pastkv will try to use the specific precision
+            if "INFERENCE_PRECISION_HINT" in os.environ:
```
helena-intel, line 276: Can we use the INFERENCE_PRECISION_HINT from CompiledModel.get_property("INFERENCE_PRECISION_HINT") instead? This would use BF16 by default on CPUs that support it and FP32 on CPUs that don't. You should not need to configure inference precision to get the best performance on SPR, and using the compiled model's inference precision would be more robust towards future API changes too.

It would also be best if the solution is generic enough that it's easy to adapt for other devices (it seem possible that this is relevant for GPU inference too, where inference precision is FP16 by default).

luo-cheng2021: Good idea, use core.get_property("INFERENCE_PRECISION_HINT"). If path is still under 'cpu' path, I'm not sure it works for other plugins.

helena-intel:
@luo-cheng2021 to fix the style CI errors you can run `make style` in the root of the repository. `pip install .[quality]` installs the requirements for that.

luo-cheng2021:
> @luo-cheng2021 to fix the style CI errors you can run `make style` in the root of the repository. `pip install .[quality]` installs the requirements for that.

Thanks, fixed.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -249,6 +250,64 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
...
+        **kwargs,
+    ):
+        model = self._try_modify_pastkv_to_lowprecision(model, ov_config, device, **kwargs)
+        super().__init__(model, config, device, dynamic_shapes, ov_config, model_save_dir, **kwargs)
+
+    def _try_modify_pastkv_to_lowprecision(
+        self, model: openvino.runtime.Model, ov_config: Optional[Dict[str, str]] = None, device: str = "CPU", **kwargs
+    ):
+        device = device.upper() if device else device
+        if device == "CPU":
```
dmitry-gorokhov, line 270: Can we enable the same behavior for GPU as well? 
cc'ed @vladimir-paramuzov

helena-intel:
@eaidova or @slyalin could you help review this PR?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -249,6 +250,64 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
...
+            pastkv_will_use = core.get_property(device, "INFERENCE_PRECISION_HINT")
+            # ov_config["INFERENCE_PRECISION_HINT"] may override the prefer precision
+            if ov_config:
+                user_hint = ov_config.get("INFERENCE_PRECISION_HINT", "")
+                openvino_types_str_map = {
```
eaidova, line 275: is it possible to move type mapping in utils or even on openvino side for avoid duplication if it will requires to get dtype from string or maybe it makes sense to have construction from string for Type for avoid requirements to update mapping if new type will be supported (e.g. i4 for example)

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -271,10 +330,22 @@ def forward(
 
         inputs = {}
         if past_key_values is not None:
-            # Flatten the past_key_values
-            past_key_values = tuple(
-                past_key_value for pkv_per_layer in past_key_values for past_key_value in pkv_per_layer
-            )
+            if len(self.key_value_input_names):
```
eaidova, line 333: if I right remember if use_cache = True during model loading and key_value_input_names are empty, model creating will fail with error that model does not support this model, so possibly this check is not necessary here

helena-intel: Yes, self.use_cache is set based on the presence of pkv in model inputs here https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_decoder.py#L122 and there is a check a few lines later that throws a ValueError if does not match with the use_cache value a user provides.

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Limitation to 'CPU' removed, but not sure it works for other plugin.

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Fixed.

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Updated.

eaidova (approved):

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -249,6 +250,52 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
...
+        **kwargs,
+    ):
+        model = self._try_modify_pastkv_to_lowprecision(model, ov_config, device, **kwargs)
+        super().__init__(model, config, device, dynamic_shapes, ov_config, model_save_dir, **kwargs)
+
+    def _try_modify_pastkv_to_lowprecision(
+        self, model: openvino.runtime.Model, ov_config: Optional[Dict[str, str]] = None, device: str = "CPU", **kwargs
+    ):
+        device = device.upper() if device else device
+        if device:
```
helena-intel, line 270: `device` is defined in the line above so it seems this if statement can be removed now that it is not specific to CPU anymore.

luo-cheng2021: Fixed.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -249,6 +250,51 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
...
+            user_precision_hint = ov_config.get("INFERENCE_PRECISION_HINT", "")
+            user_mode_hint = ov_config.get("EXECUTION_MODE_HINT", "")
+            if user_precision_hint in STR_TO_OV_TYPE:
+                pastkv_will_use = STR_TO_OV_TYPE[user_precision_hint]
+            if user_mode_hint.upper() == "ACCURACY":
+                pastkv_will_use = Type.f32
+
+        use_cache = kwargs.get("use_cache", True)
+        has_pastkv = any("past_key_values" in key.get_any_name() for key in model.inputs)
+        if pastkv_will_use != Type.f32 and use_cache and has_pastkv:
```
helena-intel, line 282: If the user has done inference before with default settings (so BF16 on SPR for example), saved the model, and then later wants to use FP32, the model will have been saved with BF16 inputs and this will not be changed now, even when setting INFERENCE_PRECSION_HINT or EXECUTION_MODE_HINT. Is it possible to remove this check?

Also should we consider changing the `.save_pretrained()`  method for these models and always save models with the original input/output precision? Hugging Face makes it really easy to upload models to the hub, and if you happen to convert the model on SPR, the model on the hub would then have bf16 inputs/outputs.

luo-cheng2021: Change `.save_pretrained()` to save original model. Although each instance of OVModelForCausalLM on SPR will have a clone of model, `clone` itself is a very light weight operation: cloned model will share the weights with the original one.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -249,6 +251,67 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
...
+        self,
+        model: openvino.runtime.Model,
+        config: PretrainedConfig = None,
+        device: str = "CPU",
+        dynamic_shapes: bool = True,
+        ov_config: Optional[Dict[str, str]] = None,
+        model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
+        **kwargs,
+    ):
+        model = self._try_modify_pastkv_to_lowprecision(model, ov_config, device, **kwargs)
```
echarlaix, line 264: It could make sense to have this directly in the `__init__()` of `OVBaseDecoderModel` so that it can be reused for every architecture with a decoder (and pkv) what do you think ?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -249,6 +251,67 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
...
+        if ov_config:
+            user_precision_hint = ov_config.get("INFERENCE_PRECISION_HINT", "")
+            user_mode_hint = ov_config.get("EXECUTION_MODE_HINT", "")
+            if user_precision_hint in STR_TO_OV_TYPE:
+                pastkv_will_use = STR_TO_OV_TYPE[user_precision_hint]
+            if user_mode_hint.upper() == "ACCURACY":
+                pastkv_will_use = Type.f32
+
+        use_cache = kwargs.get("use_cache", True)
+        has_pastkv = any("past_key_values" in key.get_any_name() for key in model.inputs)
```
echarlaix, line 282: here you only need `has_pastkv`  (that can be renamed to `use_cache`) as both will always have the same value (see https://github.com/huggingface/optimum-intel/blob/v1.9.3/optimum/intel/openvino/modeling_decoder.py#L130)

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -249,6 +251,67 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
...
+    ):
+        device = device.upper()
+        pastkv_will_use = core.get_property(device, "INFERENCE_PRECISION_HINT")
+        # ov_config["INFERENCE_PRECISION_HINT"] may override the prefer precision
+        if ov_config:
+            user_precision_hint = ov_config.get("INFERENCE_PRECISION_HINT", "")
+            user_mode_hint = ov_config.get("EXECUTION_MODE_HINT", "")
+            if user_precision_hint in STR_TO_OV_TYPE:
+                pastkv_will_use = STR_TO_OV_TYPE[user_precision_hint]
+            if user_mode_hint.upper() == "ACCURACY":
```
echarlaix, line 278: I think if `INFERENCE_PRECISION_HINT` is specified in the `ov_config`, we might want to use this value even if  `ov_config["EXECUTION_MODE_HINT"]=="ACCURACY"` what do you think?

```suggestion
            elif user_mode_hint.upper() == "ACCURACY":
```

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -249,6 +251,67 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
...
+            for key in model.outputs:
+                if "present" in key.get_any_name() and pastkv_will_use != key.get_element_type():
+                    need_gen = True
+                    ppp.output(key.get_any_name()).tensor().set_element_type(pastkv_will_use)
+            if need_gen:
+                model = ppp.build()
+
+        return model
+
+    def _save_pretrained(self, save_directory: Union[str, Path], file_name: Optional[str] = None, **kwargs):
```
echarlaix, line 300: Could you adapt it to https://github.com/huggingface/optimum-intel/blob/735bae1a0179974e9e96df82ab18b6570c99a945/optimum/intel/openvino/modeling_base.py#L139 ?

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Good idea, updated.

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Fixed.

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Fixed.

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Fixed.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -135,6 +139,50 @@ def __init__(
...
+        if ov_config:
+            user_precision_hint = ov_config.get("INFERENCE_PRECISION_HINT", "")
+            user_mode_hint = ov_config.get("EXECUTION_MODE_HINT", "")
+            if user_precision_hint in STR_TO_OV_TYPE:
+                pastkv_will_use = STR_TO_OV_TYPE[user_precision_hint]
+            elif user_mode_hint.upper() == "ACCURACY":
+                pastkv_will_use = Type.f32
+
+        use_cache = any("past_key_values" in key.get_any_name() for key in model.inputs)
+        self.model_org = model
```
echarlaix, line 157:
```suggestion
        self._original_model = model
```

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -135,6 +139,50 @@ def __init__(
...
             )
 
+    def _try_modify_pastkv_to_lowprecision(
+        self, model: openvino.runtime.Model, ov_config: Optional[Dict[str, str]] = None, device: str = "CPU"
+    ):
+        device = device.upper()
+        pastkv_will_use = core.get_property(device, "INFERENCE_PRECISION_HINT")
+        # ov_config["INFERENCE_PRECISION_HINT"] may override the prefer precision
+        if ov_config:
+            user_precision_hint = ov_config.get("INFERENCE_PRECISION_HINT", "")
```
echarlaix, line 149: very minor but would rename for clarity to something like 
```suggestion
            config_precision_hint = ov_config.get("INFERENCE_PRECISION_HINT", "")
```

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -135,6 +139,50 @@ def __init__(
                 "To export your model, simply set `export=True`."
             )
 
+    def _try_modify_pastkv_to_lowprecision(
+        self, model: openvino.runtime.Model, ov_config: Optional[Dict[str, str]] = None, device: str = "CPU"
+    ):
+        device = device.upper()
+        pastkv_will_use = core.get_property(device, "INFERENCE_PRECISION_HINT")
```
echarlaix, line 146: very minor but would rename for clarity to something like 
```suggestion
        pkv_precision = core.get_property(device, "INFERENCE_PRECISION_HINT")
```

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -271,10 +319,19 @@ def forward(
 
         inputs = {}
         if past_key_values is not None:
-            # Flatten the past_key_values
-            past_key_values = tuple(
-                past_key_value for pkv_per_layer in past_key_values for past_key_value in pkv_per_layer
-            )
+            pastkv_will_use = self.model.input(self.key_value_input_names[0]).get_element_type()
```
echarlaix, line 322: very minor but would rename for clarity to something like 

```suggestion
            pkv_precision = self.model.input(self.key_value_input_names[0]).get_element_type()
```
also can be computed only once when instantiating the OVModel

echarlaix (approved):
LGTM, thanks for iterating on it ! Left minor comments

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Fixed.

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Fixed.

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Updated.

optimum/intel/openvino/modeling_decoder.py
luo-cheng2021: Fixed.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -473,3 +483,8 @@ def _convert_to_standard_cache(
     def can_generate(self):
         """Returns True to validate the check that the model using `GenerationMixin.generate()` can indeed generate."""
         return True
+
+
+# bf16 may be in used when quantization, disable pastkv lower precision optimization
+class OVModelForCausalLMDisablePastKVOpt(OVModelForCausalLM):
```
helena-intel: Can you explain why this is needed? 
If it is necessary to disable this for quantized models, can it be done without requiring a separate class? For example with a `disable_pkv_optimizations` parameter to the standard class? It is very unexpected to have to use a different model class for quantizided models. I'm also wondering if we can detect this case automatically.

luo-cheng2021: Quantization will directly access the `self.model.model`, `self.model.request` and assign quantized model to `self.model.model`. If the optimization is not disabled, `self.model.model` will stand the modified model which nncf will not recognize it. Even if we save the modified model to `self.model.model_new`, keep the original model in `self.model.model`, `_quantize_ovcausallm` will collect inputs/outputs which may contain bf16 data and nncf will pass collected inputs/outputs to `self.model.model`, it will fail because datatype is mismatched.
Simply disabling the optimization can avoid the difficult situation when doing quantization, and it should be reasonable, quantized model should be mixed f32 and int8, but bf16 and int8.
If new class is not acceptable, may I use `ov_config['EXECUTION_MODE_HINT']='ACCURACY'` to disable the optimization when doing quantization?

AlexKoff88: Can you remove the new class and change `_quantize_ovcausallm` accordingly? Let me know if you need any help here.

luo-cheng2021: nncf's input parameter `model` should be in type `openvino.runtime.ie_api.Model`, and the modified model has the type `openvino._pyopenvino.Model`. It means the original unmodified model should be used, so I'd suggest to use `ov_config['EXECUTION_MODE_HINT']='ACCURACY'` disable the modified model since it's no use when doing quantization, and the new class will be removed also.

luo-cheng2021:
All needed logic is included in #393 .

PR #374, opened 2023-07-19
Cap transformers to <4.31 for [nncf]

helena-intel:
NNCF does not yet support transformers 4.31, this PR caps the transformers version for now for users who install the NNCF extra.


echarlaix (approved):

PR #376, opened 2023-07-20
Fix OV dataloader

echarlaix:


helena-intel (approved):
Thank you for the very quick fix!

PR #388, opened 2023-07-27
Fix typos, broken documentation links, and clarify instructions for building docs

dmsuehir:
There are a few of changes in this PR:
* Fixed a few typos
* Fixing 2 broken links in the [Optimization INC](https://huggingface.co/docs/optimum/intel/optimization_inc) documentation:
  * The `INCTrainer` link under the "During training optimization" doesn't work. It seems like URL for this link was intended to be autogenerated and point to the reference page, however that autogenerated URL doesn't work
  * Under the "Pruning" header there's a link to neural-compressor documentation that's going to a GitHub URL that no longer exists, so I fixed that
* After making the above changes, I tried to build the docs to locally test the links and ran into a couple of issues.
  * Specifying `VERSION` is required otherwise the make command gives a "VERSION is empty" error 
  * The README does not mention that command will clone a new copy of the repo in a docker container (versus using the local copy of the optimum-intel code). When I followed the instructions, I was confused why I wasn't seeing my local changes in the built html. Only after inspecting the Makefile and Dockerfile did I understand what was going on. I added a couple sentences to explain this, so that others don't run into the same issue.

## Before submitting

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,14 +22,17 @@ To generate the documentation for 🤗 Optimum Intel, simply run the following
 command from the root of the `optimum-intel` repository:
 
 ```bash
-make doc BUILD_DIR=intel-doc-build
+make doc BUILD_DIR=intel-doc-build VERSION=main
 ```
 
 This command will generate the HTML files that will be rendered as the
 documentation on the [Hugging Face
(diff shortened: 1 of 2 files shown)


echarlaix (approved):
LGTM, thanks @dmsuehir  !

PR #390, opened 2023-07-28
fix sdxl model reshape

eaidova:
# What does this PR do?
fix issue for loading SDXL 1.0 refiner model.

model_id = "stabilityai/stable-diffusion-xl-refiner-1.0"
refiner = OVStableDiffusionXLImg2ImgPipeline.from_pretrained(model_id, export=True)

raises: Attribute error: None object does not have attribute model_max_length

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -351,6 +351,7 @@ class OVStableDiffusionPipelineBase(OVBaseModel):
         height: int = -1,
         width: int = -1,
         num_images_per_prompt: int = -1,
+        tokenizer_max_length: int = -1
     ):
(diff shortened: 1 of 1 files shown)


eaidova:
@AlexKoff88  @helena-intel  @echarlaix  could you please take a look?

the root cause of issue that refiner model has only text_encoder 2 and tokenizer 2, so reshaping failed to get tokenizer.model_max_length for this model.

the second observed issue that model has eastetic_score = True in config but requires to have time_ids with 5 len, so I rewrite to get this info from model

AlexKoff88:
LGTM, thanks

AlexKoff88 (approved):

PR #394, opened 2023-08-01
Removed pkv history from quantization statistics of Decoders

AlexKoff88:
It turns out that accuracy results are significantly better if not consider PKV history during quantization, so removing it from Decoders quantization.


AlexKoff88:
It seems like CI is broken by SD update.

echarlaix:
> It seems like CI is broken by SD update.

The NNCF quantization tests are a bit flaky but it doesn't seem related to SD

PR #398, opened 2023-08-02
Tiny test updates

helena-intel:
- Install PyTorch CPU version in modeling test to speed up requirements installation time.
- Decrease pkv speedup minimum to 1.1: inference without pkv has been improved with OpenVINO 2023.1 prerelease, and now the comparison of with_and_without_pkv is sometimes small (this difference is smaller on the GitHub actions runner than locally).

### DIFF
--- a/tests/openvino/test_modeling.py
+++ b/tests/openvino/test_modeling.py
@@ -422,7 +422,7 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "pegasus",
     )
     GENERATION_LENGTH = 100
-    SPEEDUP_CACHE = 1.2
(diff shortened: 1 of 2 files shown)


tests/openvino/test_modeling.py
```diff
@@ -422,7 +422,7 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "pegasus",
     )
     GENERATION_LENGTH = 100
-    SPEEDUP_CACHE = 1.2
+    SPEEDUP_CACHE = 1.1
```
AlexKoff88, line 425: cc'ed @eaidova

echarlaix (approved):
LGTM !

PR #401, opened 2023-08-04
Fix train_text_to_image_qat.py

AlexanderDokuchaev:
### Changes

- Fix incorrect format of `openvino/stable-diffusion/requirements.txt` 
- Change argument name of `Accelerator` from `logging_dir` to `project_dir`, `logging_dir` deprecated since 0.18
- Remove second call `accelerator.prepare` for `unet`

### DIFF
--- a/examples/openvino/stable-diffusion/requirements.txt
+++ b/examples/openvino/stable-diffusion/requirements.txt
@@ -2,4 +2,4 @@ accelerate
 diffusers
 torch~=1.13
 nncf @ git+https://github.com/openvinotoolkit/nncf.git
-tomesd @ git+https://github.com/AlexKoff88/tomesd/tree/openvino
+tomesd @ git+https://github.com/AlexKoff88/tomesd.git@openvino
--- a/examples/openvino/stable-diffusion/requirements.txt
+++ b/examples/openvino/stable-diffusion/requirements.txt
@@ -1,5 +1,6 @@
 accelerate
 diffusers
 torch~=1.13
+torchvision==0.14.1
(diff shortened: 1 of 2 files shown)


AlexKoff88:
As far as I understood model export gets broken after this change so it needs to be fixed before merging this.

examples/openvino/stable-diffusion/requirements.txt
```diff
@@ -2,4 +2,4 @@ accelerate
 diffusers
 torch~=1.13
```
fxmarty, line 3: Is there anything blocking for pytorch 2.0?

AlexKoff88:
> Thank you!
> 
> I can't test those changes as I get an error at this point
> 
> https://github.com/huggingface/optimum-intel/blob/0a80e2053b26d72869710c5f2c14b9ee06d5141e/examples/openvino/stable-diffusion/train_text_to_image_qat.py#L648
> 
> in the script, telling me that the `collate_fn` from the dataloader is not pickable. Have you had the issue @AlexKoff88 ?

I am looking into it. There are multiple issues introduced by updates in `transformers` and `accelerate` libraries. Need to highly refactor the script.

AlexKoff88:
Some issues in this PR should be fixed after merging this PR: https://github.com/huggingface/optimum-intel/pull/397

examples/openvino/stable-diffusion/requirements.txt
AlexanderDokuchaev: Did not check it.

PR #402, opened 2023-08-07
Do not download onnx model in sd pipeline if from_onnx=False

eaidova:
# What does this PR do?

fixes downloading onnx model from hub if both openvino and onnx models are provided. Problem observed in stable diffusion pipeline (especially SDXL base)
model_download

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -217,6 +217,9 @@ class OVStableDiffusionPipelineBase(OVBaseModel):
                     cls.config_name,
                 }
             )
+            ignore_patterns = ["*.msgpack", "*.safetensors", "*pytorch_model.bin"]
+            if not from_onnx:
+                ignore_patterns.extend(["*.onnx", "*.onnx_data"])
             # Downloads all repo's files matching the allowed patterns
             model_id = snapshot_download(
                 model_id,
@@ -225,7 +228,7 @@ class OVStableDiffusionPipelineBase(OVBaseModel):
                 use_auth_token=use_auth_token,
                 revision=revision,
                 allow_patterns=allow_patterns,
-                ignore_patterns=["*.msgpack", "*.safetensors", "*pytorch_model.bin"],
+                ignore_patterns=ignore_patterns,
             )
         new_model_save_dir = Path(model_id)
 


eaidova:
@AlexKoff88  @helena-intel  @echarlaix  could you please take a look?

helena-intel:
Thanks @eaidova ! This was expected to be fixed with https://github.com/huggingface/diffusers/pull/4338 but I also still see ONNX models being downloaded with a source install of diffusers.
@fxmarty could you have a look?

eaidova:
> Thanks @eaidova ! This was expected to be fixed with [huggingface/diffusers#4338](https://github.com/huggingface/diffusers/pull/4338) but I also still see ONNX models being downloaded with a source install of diffusers. @fxmarty could you have a look?

hi @helena-intel, in your referenced pr, fixed downloading onnx if pytorch model used (in our case, when export=True set), in my case model repo contains openvino model (https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0 has openvino model inside) and downloading extra files related to onnx, that is why pr to diffusers does not work in this case

helena-intel:
> hi @helena-intel, in your referenced pr, fixed downloading onnx if pytorch model used (in our case, when export=True set)
Yes, but that may be an oversight there (I was told last week this issue was expected to be fixed with that PR). From my perspective, if it is possible to fix this upstream, that would be nicer. But no strong preference, most important that it is fixed; these files are huge.

AlexKoff88:
I agree with these changes as a stop-gap measure.

AlexKoff88 (approved):

fxmarty (approved):
LGTM thank you!

fxmarty:
Being curious: should we also ignore the OV IR .bin in case we have `from_onnx=True`?

helena-intel:
> Being curious: should we also ignore the OV IR .bin in case we have `from_onnx=True`?

Yes I see no reason why not. We're not really exposing the from_onnx parameter to users though; users will generally use PyTorch models and convert them to OpenVINO on the fly with export=True (which uses from_onnx internally for now) or OpenVINO models.

Thanks for approving the PR @fxmarty . Could you merge it? These models can be so large, it would be really nice to skip downloading unnessary ones.

PR #404, opened 2023-08-09
Timm Models integration to Optimum-intel[Draft]

sawradip:
[Will be added Later]

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -217,9 +217,6 @@ class OVStableDiffusionPipelineBase(OVBaseModel):
                     cls.config_name,
                 }
             )
-            ignore_patterns = ["*.msgpack", "*.safetensors", "*pytorch_model.bin"]
-            if not from_onnx:
-                ignore_patterns.extend(["*.onnx", "*.onnx_data"])
             # Downloads all repo's files matching the allowed patterns
             model_id = snapshot_download(
                 model_id,
@@ -228,7 +225,7 @@ class OVStableDiffusionPipelineBase(OVBaseModel):
                 use_auth_token=use_auth_token,
                 revision=revision,
                 allow_patterns=allow_patterns,
-                ignore_patterns=ignore_patterns,
+                ignore_patterns=["*.msgpack", "*.safetensors", "*pytorch_model.bin"],
             )
         new_model_save_dir = Path(model_id)
 
--- a/optimum/intel/openvino/modeling_timm.py
+++ b/optimum/intel/openvino/modeling_timm.py
@@ -1,0 +1,473 @@
+import os
+import PIL
+import numpy as np
+from pathlib import Path
+from packaging import version
+from collections import OrderedDict
+from typing import Mapping, Any, Dict, List, Optional, Tuple, Union, Callable
+
+import torch
+import torch.nn as nn
+from torch.nn import BCEWithLogitsLoss, CrossEntropyLoss, MSELoss
+
+import timm
+from timm.layers.config import set_fused_attn
+from timm.models._hub import load_model_config_from_hf
+from transformers import PreTrainedModel, PretrainedConfig
+from transformers.utils import TensorType
+from transformers.image_utils import (
+    IMAGENET_STANDARD_MEAN,
+    IMAGENET_STANDARD_STD,
+    ChannelDimension,
+    ImageInput,
+    PILImageResampling,
+    make_list_of_images,
+    to_numpy_array,
+    valid_images, 
+    ImageFeatureExtractionMixin
+    )
+from transformers.image_processing_utils import BaseImageProcessor, BatchFeature, get_size_dict
+from transformers.image_transforms import resize, to_channel_dimension_format
+from transformers.modeling_outputs import BaseModelOutput, ImageClassifierOutput
+
+from optimum.exporters import TasksManager
+from optimum.exporters.onnx.model_configs import ViTOnnxConfig
+
+set_fused_attn(False, False)
+ExportConfigConstructor = Callable[[PretrainedConfig], "ExportConfig"]
+
+
+class TimmConfig(PretrainedConfig):
+    model_type = "timm"
+
+    @classmethod
+    def from_pretrained(
+        cls,
+        pretrained_model_name_or_path: Union[str, os.PathLike],
+        cache_dir: Optional[Union[str, os.PathLike]] = None,
+        force_download: bool = False,
+        local_files_only: bool = False,
+        token: Optional[Union[str, bool]] = None,
+        revision: str = "main",
+        **kwargs,
+    ) -> "PretrainedConfig":
+
+        kwargs["cache_dir"] = cache_dir
+        kwargs["force_download"] = force_download
+        kwargs["local_files_only"] = local_files_only
+        kwargs["revision"] = revision
+
+        config_dict = load_model_config_from_hf(pretrained_model_name_or_path)[0]
+        config_dict["num_labels"] = config_dict.pop("num_classes")
+        config_dict["image_size"] = config_dict.get("input_size")[-1]
+
+        return cls.from_dict(config_dict, **kwargs)
+
+
+class TimmOnnxConfig(ViTOnnxConfig):
+    DEFAULT_TIMM_ONNX_OPSET = 13
+    outputs= OrderedDict([('logits', {0: 'batch_size'})])
+
+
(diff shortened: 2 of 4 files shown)


optimum/intel/openvino/modeling_timm.py
```diff
@@ -0,0 +1,398 @@
...
+    outputs = OrderedDict([("logits", {0: "batch_size"})])
+
+
+class TimmPreTrainedModel(PreTrainedModel):
+    config_class = TimmConfig
+    base_model_prefix = "timm"
+    main_input_name = "pixel_values"
+
+
+class TimmModel(TimmPreTrainedModel):
```
AlexKoff88, line 71: I don't think we need such a model hierarchy, with TimmPreTrainedModel, TimmModel, etc. At least Optimum-Intel is not the place for such classes. What is required is just to put the logic of model creation and prefetch into the .from_timm() method of OVModelForTimm. Moreover, OVModelForTimm itself is not needed since there is no intent to expose it. It can be just a utility function or class that is not inherited from HF hierarchy at all.

optimum/intel/openvino/modeling_timm.py
```diff
@@ -0,0 +1,398 @@
...
+            `config.num_labels > 1` a classification loss is computed (Cross-Entropy).
+        """
+        return_dict = return_dict if return_dict is not None else self.config.use_return_dict
+
+        logits = self.timm(
+            pixel_values,
+            return_dict=return_dict,
+        )
+
+        loss = None
```
AlexKoff88, line 151: Loss is not needed at all in the context of further export to OpenVINO.

optimum/intel/openvino/modeling_timm.py
```diff
@@ -0,0 +1,398 @@
...
+        if not return_dict:
+            return logits
+
+        return ImageClassifierOutput(
+            loss=loss,
+            logits=logits.last_hidden_state,
+        )
+
+
+class TimmImageProcessor(BaseImageProcessor, ImageFeatureExtractionMixin):
```
AlexKoff88, line 185: I didn't expect but this is great that you created it. I just wonder how much from the overall Timm preprocessing options this class supports and how we will validate it?

optimum/intel/openvino/modeling_timm.py
sawradip: I have removed that section.

optimum/intel/openvino/modeling_timm.py
sawradip: I have removed the hierarchy as much as possible, leaving a small placeholder model to contain the `timm` models.

sawradip:
@AlexKoff88  I have removed a of unnecessary code, and abstractions. Please take a look and let me know your feedback.

optimum/intel/openvino/modeling_timm.py
```diff
@@ -0,0 +1,303 @@
...
+        kwargs["revision"] = revision
+
+        config_dict = load_model_config_from_hf(pretrained_model_name_or_path)[0]
+        config_dict["num_labels"] = config_dict.pop("num_classes")
+        config_dict["image_size"] = config_dict.get("input_size")[-1]
+
+        return cls.from_dict(config_dict, **kwargs)
+
+
+class TimmOnnxConfig(ViTOnnxConfig):
```
AlexKoff88, line 59: You used `ViTOnnxConfig` as a base class here. Does it mean that only ViT models from Timm will work here?

AlexKoff88: I would inherit from `VisionOnnxConfig` to make it safer.

tests/openvino/test_modeling.py
```diff
@@ -619,6 +623,27 @@ def test_pipeline(self, model_arch):
         self.assertTrue(isinstance(outputs[0]["label"], str))
         gc.collect()
 
+    @parameterized.expand(TIMM_MODELS)
```
AlexKoff88, line 626: I think we should have a test which:
- load a timm model from hub
- saves it in IR locally (.save_pretrained)
- loads IR and does inference (random data is ok)

sawradip: A test `test_timm_save_and_infer` is added.

optimum/intel/openvino/modeling_timm.py
sawradip: I also wonder. Though I have added and used the class, it would be better if we had any methodical way for validating.

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -217,9 +217,6 @@ def _from_pretrained(
                     cls.config_name,
                 }
             )
-            ignore_patterns = ["*.msgpack", "*.safetensors", "*pytorch_model.bin"]
-            if not from_onnx:
-                ignore_patterns.extend(["*.onnx", "*.onnx_data"])
```
fxmarty, line 222: I think this bit was a fix from a previous PR to avoid downloading ONNX weights when the OV ones are available.

optimum/intel/openvino/modeling_timm.py
```diff
@@ -0,0 +1,338 @@
...
+
+class TimmOnnxConfig(VisionOnnxConfig):
+    DEFAULT_TIMM_ONNX_OPSET = 13
+    outputs = OrderedDict([("logits", {0: "batch_size"})])
+    NORMALIZED_CONFIG_CLASS = NormalizedVisionConfig
+    MIN_TORCH_VERSION = version.parse("1.11")
+
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        return {"pixel_values": {0: "batch_size", 1: "num_channels", 2: "height", 3: "width"}}
```
fxmarty, line 96: Is this to support the ONNX export of timm models? Do you think it would be useful to move it (later) in optimum main repo?

optimum/intel/openvino/modeling_timm.py
```diff
@@ -0,0 +1,338 @@
...
+        logits = self.model(
+            pixel_values,
+        )
+
+        return ImageClassifierOutput(
+            logits=logits,
+        )
+
+
+class TimmImageProcessor(BaseImageProcessor, ImageFeatureExtractionMixin):
```
fxmarty, line 127: Could you put a comment here pointing at what this class is adapted / copied from (if it is)?

fxmarty (approved):
I left a few comments for tiny things to be fixed, but otherwise looks good to me, thank you!

It could be worth mentioning in the documentation as well that `OVModelForImageClassification` can handle timm models.

optimum/intel/openvino/modeling_timm.py
AlexKoff88: @sawradip, can you please answer.

sawradip: It is adapted from ViTImageProcessor - https://github.com/huggingface/transformers/blob/main/src/transformers/models/vit/image_processing_vit.py 

I have added a comment.

optimum/intel/openvino/modeling_diffusion.py
sawradip: Merged the main branch. Hope this is solved.

fxmarty: Have you pushed?

I still see this PR branch reverting the fix: https://github.com/sawradip/optimum-intel/blob/integrate_timm/optimum/intel/openvino/modeling_diffusion.py#L229

sawradip: Now its fixed.   Right?

fxmarty: It is, thank you!

fxmarty:
> It could be worth mentioning in the documentation as well that OVModelForImageClassification can handle timm models.

@sawradip Would you like to precise the doc / add an example, or should I merge as is?

sawradip:
I will add some refernce to the doc as well add some description to the PR.

sawradip:
@fxmarty , I have done some additions in docs, as well as added description to this PR. If everything seems alright, you can merge.

sawradip:
@fxmarty  There was one error with style, fixed it.

Any feedback?

fxmarty (approved):
LGTM!

setup.py
```diff
@@ -17,6 +17,7 @@
     "datasets>=1.4.0",
     "sentencepiece",
     "scipy",
+    "timm",
```
echarlaix, line 20: I don't think `timm` should become a hard dependency

PR #407, opened 2023-08-18
Fix import error is_safetensors_available

eaidova:
# What does this PR do?

fixes failure in tests with import error due to incompatible changes in diffusers 0.20.0


eaidova:
@fxmarty @AlexKoff88  please take a look
it affects testing in https://github.com/huggingface/optimum-intel/pull/404 and https://github.com/huggingface/optimum-intel/pull/397

AlexKoff88:
Nice. Thanks, @eaidova! I've just asked if we need to update dependencies but it seems you made it more generic.

AlexKoff88 (approved):

PR #417, opened 2023-08-27
Integrate weight-only quantizaion of INC

mengniwang95:
This PR integrate weight-only quantization of neural compressor into optimum-intel.

Notice: Need to use the master branch for test

### DIFF
--- a/examples/neural_compressor/language-modeling/run_clm.py
+++ b/examples/neural_compressor/language-modeling/run_clm.py
@@ -202,9 +202,11 @@ class OptimizationArguments:
     )
     weight_only_group: int = field(
         default=-1,
-        metadata={"help": "Group size for weight only quantization. Group_size=[1-N] indicates "
-                    "splitting the input channel elements per group_size. -1 indicates "
-                    "the per-channel quantization per output channel."},
+        metadata={
+            "help": "Group size for weight only quantization. Group_size=[1-N] indicates "
+            "splitting the input channel elements per group_size. -1 indicates "
+            "the per-channel quantization per output channel."
+        },
     )
     weight_only_scheme: str = field(
         default="sym",
@@ -214,6 +216,7 @@ class OptimizationArguments:
         default="RTN",
         metadata={"help": "Scheme for weight only quantization. Choose from 'RTN', 'AWQ' and 'GPTQ'."},
     )
+
 
 @dataclass
 class DataTrainingArguments:
@@ -556,7 +559,9 @@ def main():
             desc=f"Grouping texts in chunks of {block_size}",
         )
 
-    if training_args.do_train or (optim_args.apply_quantization and optim_args.quantization_approach in ["static", "weight_only"]):
+    if training_args.do_train or (
+        optim_args.apply_quantization and optim_args.quantization_approach in ["static", "weight_only"]
+    ):
         if "train" not in tokenized_datasets:
             raise ValueError("--do_train requires a train dataset")
         train_dataset = lm_datasets["train"]
@@ -618,7 +623,7 @@ def main():
             else:
                 recipes = {}
             if optim_args.quantization_approach == "weight_only":
-                op_type_dict={
+                op_type_dict = {
                     ".*": {
                         "weight": {
                             "bits": optim_args.weight_only_bits,
@@ -631,9 +636,7 @@ def main():
             else:
                 op_type_dict = {}
             quantization_config = PostTrainingQuantConfig(
-                approach=optim_args.quantization_approach,
-                op_type_dict=op_type_dict,
-                recipes=recipes
+                approach=optim_args.quantization_approach, op_type_dict=op_type_dict, recipes=recipes
             )
 
     if optim_args.apply_pruning:
@@ -723,9 +726,12 @@ def main():
         quantizer.quantize(
             quantization_config=quantization_config,
             save_directory=training_args.output_dir,
-            calibration_dataset=train_dataset if optim_args.quantization_approach in ["static", "weight_only"] else None,
-            batch_size=1 if optim_args.quantization_approach == "weight_only" and optim_args.weight_only_algorithm == "GPTQ" \
-                    else  training_args.per_device_train_batch_size,
+            calibration_dataset=train_dataset
+            if optim_args.quantization_approach in ["static", "weight_only"]
+            else None,
+            batch_size=1
+            if optim_args.quantization_approach == "weight_only" and optim_args.weight_only_algorithm == "GPTQ"
+            else training_args.per_device_train_batch_size,
         )
         trainer.model = quantizer._quantized_model
     if optim_args.apply_quantization and optim_args.verify_loading:
--- a/examples/neural_compressor/language-modeling/run_clm.py
+++ b/examples/neural_compressor/language-modeling/run_clm.py
@@ -635,6 +635,11 @@ def main():
                         },
                     },
                 }
+                if optim_args.quantization_methodology == "GPTQ":
+                    gptq_args = {
+                        "pad_max_length": block_size,
+                    }
+                    recipes.update({"gptq_args": gptq_args})
             else:
                 op_type_dict = {}
             quantization_config = PostTrainingQuantConfig(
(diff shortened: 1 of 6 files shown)


hshen14:
@echarlaix Could you please help review this PR? INC supports production-level quality of weight-only quantization including INT8 and INT4 for LLMs in latest master (also be released in INC v2.3 in early Sep). Thanks.

examples/neural_compressor/language-modeling/run_clm.py
```diff
@@ -677,19 +714,24 @@ def compute_metrics(eval_preds):
...
 
         quantizer.quantize(
             quantization_config=quantization_config,
             save_directory=training_args.output_dir,
-            calibration_dataset=train_dataset if optim_args.quantization_approach == "static" else None,
-            batch_size=training_args.per_device_train_batch_size,
+            calibration_dataset=train_dataset
+            if optim_args.quantization_approach in ["static", "weight_only"]
+            else None,
+            batch_size=1
```
echarlaix, line 732: is it because GPTQ methology requires a lot of memory resource ? Can you also add a comment to make it clear

examples/neural_compressor/language-modeling/run_clm.py
```diff
@@ -600,7 +622,22 @@ def compute_metrics(eval_preds):
                 recipes = {"smooth_quant": True, "smooth_quant_args": {"alpha": optim_args.smooth_quant_alpha}}
             else:
                 recipes = {}
-            quantization_config = PostTrainingQuantConfig(approach=optim_args.quantization_approach, recipes=recipes)
+            if optim_args.quantization_approach == "weight_only":
+                op_type_dict = {
```
echarlaix, line 626: What happens if `quantization_approach == "weight_only"` and  `op_type_dict`  is not provided ? Should this be also added before quantization in the quantizer when not provided ?

examples/neural_compressor/language-modeling/run_clm.py
```diff
@@ -677,19 +714,24 @@ def compute_metrics(eval_preds):
         trainer.save_metrics("train", metrics)
         trainer.save_state()
 
-    if optim_args.apply_quantization and optim_args.quantization_approach in {"static", "dynamic"}:
+    if optim_args.apply_quantization and optim_args.quantization_approach in {"static", "dynamic", "weight_only"}:
         model = trainer.model if isinstance(trainer.model, PreTrainedModel) else trainer.model._model
+        model.seqlen = block_size
```
echarlaix, line 719: Why is it used for ? Can we include it in the config or quantizer directly instead of modifying the original model ?

optimum/intel/neural_compressor/quantization.py
```diff
@@ -189,6 +190,25 @@ def quantize(
...
+        elif INCQuantizationMode(quantization_config.approach) == INCQuantizationMode.WEIGHT_ONLY:
+            if isinstance(quantization_config.op_type_dict, dict) and len(quantization_config.op_type_dict) > 0:
+                algo = []
+                for _, val in quantization_config.op_type_dict.items():
+                    algo += val.get("weight", {}).get("algorithm", ["RTN"])
+            else:
+                algo = ["RTN"]
+
+            if calibration_dataset is None and ("GPTQ" in algo or "AWQ" in algo):
+                raise ValueError("Weight-only quantizaion needs a calibration dataset.")
```
echarlaix, line 202:
```suggestion
                raise ValueError("Weight-only quantization needs a calibration dataset for both GPTQ and AWQ methodologies.")
```

echarlaix (commented):
Thanks a lot for the addition @mengniwang95

examples/neural_compressor/language-modeling/run_clm.py
echarlaix: Also could make sense to keep the same API as for the NNCF quantizer : https://github.com/huggingface/optimum-intel/blob/v1.10.1/optimum/intel/openvino/quantization.py#L115

mengniwang95: if op_type_dict is not provided, it will use the default configuration: dtype-int4, group_size-32, scheme-sym, algorithm-RTN

examples/neural_compressor/language-modeling/run_clm.py
```diff
@@ -196,6 +196,26 @@ class OptimizationArguments:
...
+            "splitting the input channel elements per group_size. -1 indicates "
+            "the per-channel quantization per output channel."
+        },
+    )
+    weight_only_scheme: str = field(
+        default="sym",
+        metadata={"help": "Scheme for weight only quantization. Choose from 'sym' and 'asym'."},
+    )
+    weight_only_algorithm: str = field(
+        default="RTN",
```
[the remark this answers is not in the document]
mengniwang95: that's because if you don't set the op_type_dict, the default algorithm is 'RTN'

examples/neural_compressor/language-modeling/run_clm.py
mengniwang95: sure, calculation of hessian matrix with bs>1 is WIP

examples/neural_compressor/language-modeling/run_clm.py
mengniwang95: Of course we can add an enhancement in INC to move this feature to config/quantizer. This parameter is used to check calibration data length to ensure they won't exceed model's config acceptable sequence length, since we do not know how users tokenize their own datasets.

echarlaix: Ok I see, then it could make sense to check this in the examples scripts directly, I don't think we need to add this verification in the `INCQuantizer` or in `neural-compressor` (in `GPTQuantizer`) as it belongs to a different scope in my opinion

echarlaix: Once this point is fixed we can merge in my opinion!

tests/neural_compressor/test_optimization.py
```diff
@@ -168,6 +168,86 @@ def test_ipex_static_quantization_with_smoothquant(self, task, model_name, expec
...
+                    "bits": 8,
+                    "group_size": -1,
+                    "scheme": "sym",
+                    "algorithm": "RTN",
+                },
+            },
+        }
+        quantization_config = PostTrainingQuantConfig(approach="weight_only", op_type_dict=op_type_dict)
+        model = AutoModelForCausalLM.from_pretrained(model_name)
+        model.seqlen = 128
```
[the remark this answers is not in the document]
mengniwang95: sure

echarlaix (approved):
Looks great, thanks for the addition @mengniwang95 !

mengniwang95:
Hi Ella, currently UT fails since it doesn't use the latest master code. Do we need to wait neural-compressor 2.3 release for UT test and merge this PR after all test passing? @echarlaix

echarlaix:
> Hi Ella, currently UT fails since it doesn't use the latest master code. Do we need to wait neural-compressor 2.3 release for UT test and merge this PR after all test passing? @echarlaix

For when is  the `neural-compressor` release planned ? This PR is compatible with the current INC latest version so I'm ok to with merging it now (the test can be added in an other PR with INC being installed from source)

mengniwang95:
> > Hi Ella, currently UT fails since it doesn't use the latest master code. Do we need to wait neural-compressor 2.3 release for UT test and merge this PR after all test passing? @echarlaix
> 
> For when is the `neural-compressor` release planned ? This PR is compatible with the current INC latest version so I'm ok to with merging it now (the test can be added in an other PR with INC being installed from source)

Hi Ella, neural-compressor release is planned on 9/15. I add INT4 UT in this branch, but it is not triggered due to neural-compressor < 2.3

hshen14:
@echarlaix it seems some tests failed, while they may not be related with the changes. Could you please help check, or is it okay to get this PR merged?

echarlaix:
> @echarlaix it seems some tests failed, while they may not be related with the changes. Could you please help check, or is it okay to get this PR merged?

Could you update your branch by rebasing from main ? This will fix all unrelated tests. The INC tests are failing, because the release is previewed for tomorrow I think we should install `neural-compressor` from source [here](https://github.com/huggingface/optimum-intel/blob/v1.11.0/.github/workflows/test_inc.yml#L32-L33) to verify all the tests are passing and then we can merge

PR #427, opened 2023-09-08
[Draft]: Added hybrid quantization for seq2seq models

AlexKoff88:
This is a draft of the PR that should be merged after OpenVINO 2023.1 release

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -41,6 +41,7 @@ from ..utils.constant import _TASK_ALIASES
 from .configuration import INT8_WEIGHT_COMPRESSION_CONFIG, OVConfig
 from .modeling_base import OVBaseModel
 from .modeling_decoder import OVBaseDecoderModel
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/quantization.py
```diff
@@ -246,6 +256,42 @@ def _quantize_ovbasemodel(
...
+            quantization_dataset,
+            model_type=nncf.ModelType.TRANSFORMER,
+            fast_bias_correction=kwargs.get("fast_bias_correction", True),
+            **kwargs,
+        )
+
+        # Compress weights of decoders for safity
+        self.model.decoder_model = nncf.compress_weights(self.model.decoder_model)
+        if self.model.use_cache:
+            self.model.decoder_with_past_model = nncf.compress_weights(self.model.decoder_with_past_model)
```
echarlaix, line 291: Would there be a possibility to also quantize the activations for the decoder components?

AlexKoff88: It is possible but it may hurt accuracy.

PR #429, opened 2023-09-11
Update sharing memory flags for OpenVINO

jiwaszki:
Update deprecated `shared_memory` flag with new `share_inputs` which works in the same way. Removes warnings.

Added new flag `share_outputs` to utilize new feature of OpenVINO to reduce copies to minimum.


jiwaszki:
Hi, @echarlaix this PR should go together with OpenVINO 2023.1 (now only pre-release versions are available: https://pypi.org/project/openvino/#history). Will the newest version update trigger automatically when OpenVINO drops? (ofc I will trigger branch update)

AlexKoff88:
I think we should also update the quantization part where the infer request is wrapped. Here, we should only use `share_inputs`.

echarlaix:
Thanks for the addition @jiwaszki 

> Will the newest version update trigger automatically when OpenVINO drops? (ofc I will trigger branch update)

The latest OpenVINO version should be installed, but because these modifications are compatible with `v2023.1` or higher we should also update it [here](https://github.com/huggingface/optimum-intel/blob/v1.11.0/setup.py#L45)

echarlaix (approved):
LGTM, thanks @jiwaszki !

IlyasMoutawwakil:
done in #499

PR #432, opened 2023-09-14
move VAE to fp32 execution precision on GPU

eaidova:
Prevent fp16 overflow for inference OpenVINO on GPU for SDXL and SD2.1


eaidova:
@AlexKoff88  @helena-intel @echarlaix  could you please take a look?
these changes prevent black image output for some stable diffusion models versions on GPU.

cc @vladimir-paramuzov

AlexKoff88 (approved):

PR #434, opened 2023-09-22
OpenVINO device updates

helena-intel:
- Fix error with `OVModelForCausalLM` when using a device like `AUTO:GPU` (which will automatically select a device, but only from available GPUs. For example if you want inference to run on either integrated or discrete GPU, but not on CPU).
- Remove `_SUPPORTED_DEVICES` and `_ensure_supported_device()` since they are not used
- Add device name to "Compiling the model" INFO messages

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -43,17 +43,6 @@ core = Core()
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


echarlaix (approved):
LGTM thanks for the fix @helena-intel !

AlexKoff88 (approved):

PR #437, opened 2023-09-27
Add OpenVINO export CLI

echarlaix:
example : 
```
optimum-cli export openvino --model distilbert-base-uncased-finetuned-sst-2-english distilbert_openvino
```

It could also make sense to add weight only quantization in a following PR, what do you think ? @helena-intel @AlexKoff88

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -58,7 +58,6 @@ def parse_args_openvino(parser: "ArgumentParser"):
         help=(
             "Allows to use custom code for the modeling hosted in the model repository. This option should only be set for repositories you trust and in which "
             "you have read the code, as it will execute on your local machine arbitrary code present in the model repository."
-
         ),
     )
     optional_group.add_argument(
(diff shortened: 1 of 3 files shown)


helena-intel (commented):
Thanks @echarlaix I love this!  I also really like that it exports the tokenizer too. Agreed that weight compression for CausalLM models would be great to add too. It would also be good to have an FP16 option. We could also consider to export to FP16 by default. This is what OpenVINO now does by default too when you convert/save a model, with an option to disable it. 

The example worked without errors, but it exported the distilbert model as model.xml and .bin instead of openvino_model.xml and bin. And for gpt2 it exported decoder_model.xml/bin and decoder_model_with_past xml/bin.

AlexKoff88:
> example :
> 
> ```
> optimum-cli export openvino --model distilbert-base-uncased-finetuned-sst-2-english distilbert_openvino
> ```
> 
> It could also make sense to add weight only quantization in a following PR, what do you think ? @helena-intel @AlexKoff88

It totally makes sense.

AlexKoff88 (approved):

helena-intel (approved):

PR #439, opened 2023-09-27
Fix openvino main export

echarlaix:
Fix `main_export` for decoder models (which were exported twice  : with /without past) as well as the resulting model file name which where wrongfully set for single models (SD and encoder/decoder models were not concerned).  Also now the different `OVModelForXxx` classes all use `main_export` to export the model when needed

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -26,7 +26,7 @@ from optimum.exporters.onnx import __main__ as optimum_main
 from optimum.exporters.onnx.base import OnnxConfig, OnnxConfigWithPast
 from optimum.utils import DEFAULT_DUMMY_SHAPES
 from optimum.utils.save_utils import maybe_save_preprocessors
-
+from ...intel.utils.modeling_utils import patch_decoder_attention_mask
 from .convert import export_models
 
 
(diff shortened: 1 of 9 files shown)


helena-intel:
Thanks Ella!  I noticed one issue with CausalLM export: when loading the model, I got the error "use_cache` was set to `True` but the loaded model only supports `use_cache=False". During export I saw `Overriding 2 configuration item(s) - use_cache -> True` but it appears the model is exported without cache.

echarlaix:
> Thanks Ella! I noticed one issue with CausalLM export: when loading the model, I got the error "use_cache`was set to`True`but the loaded model only supports`use_cache=False". During export I saw `Overriding 2 configuration item(s) - use_cache -> True` but it appears the model is exported without cache.

Thanks a lot for reporting it @helena-intel, could you share a code snippet for me to reproduce this ?

echarlaix:
cc @eaidova

helena-intel:
> Thanks a lot for reporting it @helena-intel, could you share a code snippet for me to reproduce this ?

When I tried to reproduce it it did not occur, so I suppose it was a strange thing in my environment. Sorry for the false alarm!

helena-intel (approved):

echarlaix:
> When I tried to reproduce it it did not occur, so I suppose it was a strange thing in my environment. Sorry for the false alarm!

Thanks a lot for checking @helena-intel

PR #443, opened 2023-09-29
Add fp16 and int8 to OpenVINO models and export CLI

echarlaix:
as discussed in https://github.com/huggingface/optimum-intel/pull/437


AlexKoff88 (approved):
LGTM, thanks Ella!

helena-intel:
Thanks @echarlaix ! I tested it and it worked well. I'm wondering if we should restrict INT8 weights compression to tasks where we are reasonably confident that it works well (LLMs), and show a message  about manually quantizing for other tasks.

optimum/exporters/openvino/__main__.py
```diff
@@ -299,5 +311,18 @@ def main_export(
         output_names=files_subpaths,
         input_shapes=input_shapes,
         device=device,
+        fp16=fp16,
         model_kwargs=model_kwargs,
     )
+    del models_and_onnx_configs
+
+    if int8:
```
AlexKoff88, line 319: One comment from my side. Such an implementation means that we will:
- Convert from PyTorch to OpenVINO with memory reuse (no-copy)
- However, after that, we will serialize to the disk, load the model again, quantize weights, remove files from the disk, and store the compressed version, finally.
- It can be time-consuming for really large models.

Ideally, we should have nncf.compress_weights() right after openvino.convert_model(), the similar to FP16.

AlexKoff88: I opened a follow-up PR #444 for the default 8-bit compression. I tried to follow the approach described above.

AlexKoff88: @eaidova, please take a look as well.

echarlaix: > Ideally, we should have nncf.compress_weights() right after openvino.convert_model(), the similar to FP16.

Yes that's good point, was going to modify this PR but I see that you opened https://github.com/huggingface/optimum-intel/pull/444, thanks @AlexKoff88

optimum/exporters/openvino/__main__.py
```diff
@@ -123,6 +125,19 @@ def main_export(
     >>> main_export("gpt2", output="gpt2_onnx/")
     ```
     """
+    if int8:
+        if fp16:
+                "Both `fp16` and `int8` were both set to `True`, please select only one of these options."
```
eaidova, line 131: is it really impossible to have both fp16 and int8 in the same time?
I think it should be possible

echarlaix: wasn't sure this would work, should we set it by default to True for large models (along with `int8=True`) ? (enabling int8 weights and fp16 computations) ?

echarlaix: enabling it back in  https://github.com/huggingface/optimum-intel/pull/443/commits/eefec6c064057111e17c77e60083d78594da5755

eaidova: from my knowledge, there are some potential accuracy issues with fp16 compression due to floating overflow for some models observable on GPU (like sometimes we see for stable diffusion models), so it may be dangerous to compress the model to fp16 by default. We are still investigating how to improve that and getting more info about which models are affected.

AlexKoff88 (commented):
LGTM, thanks!

eaidova (approved):

PR #445, opened 2023-10-03
Changed the logic of default 8-bit weights compression

AlexKoff88:
@eaidova highlighted the case when `load_in_8bits = False` is passed explicitly. Based on this I revised the whole logic. Now,
- warning is shown if nncf is absent in the text-generation branch
- exception is raised everytime when `load_in_8bits = True` is passed without nncf
- added tests


eaidova (approved):

tests/openvino/test_training.py
```diff
@@ -715,7 +715,7 @@ def check_ovmodel_reshaping(self, ovmodel: OVModel):
         model_id="hf-internal-testing/tiny-random-Wav2Vec2Model",
         nncf_compression_config=[QUANTIZATION_CONFIG_FOR_WAV2VEC2],
         expected_fake_quantize=48,
-        expected_int8=31,
+        expected_int8=30,
```
echarlaix, line 718: why this change ?

echarlaix (approved):
LGTM, thanks @AlexKoff88 !

PR #448, opened 2023-10-10
Use share_inputs to replace shared_memory

wgzintel:
remove the FutureWarning:
optimum/intel/openvino/modeling_decoder.py:364: FutureWarning: `shared_memory` is deprecated and will be removed in 2024.0. Value of `shared_memory` is going to override `share_inputs` value. Please use only `share_inputs` explicitly.


wgzintel:
replaced by  PR429

PR #451, opened 2023-10-13
GPTQ export w/a

AlexKoff88:
- These changes are required to export GPTQ models without GPU

### DIFF
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -226,6 +226,31 @@ class OVBaseDecoderModel(OVModel):
 
             if use_cache:
                 task = task + "-with-past"
+                
+        # Patch the modules to export of GPTQ models w/o GPU
+        do_gptq_patching = False
+        config_dict = config.to_dict()
+        quantization_config = config_dict.get("quantization_config", None)
+        do_gptq_patching = quantization_config and quantization_config["quant_method"] == "gptq"
+        if do_gptq_patching:
+            torch.set_default_dtype(torch.float32) 
+            orig_cuda_check = torch.cuda.is_available
+            torch.cuda.is_available = lambda: True
+            
+            from optimum.gptq import GPTQQuantizer
+            orig_post_init_model = GPTQQuantizer.post_init_model
+            def post_init_model(self, model):
+                from auto_gptq import exllama_set_max_input_length
+                class StoreAttr(object):
+                    pass
+
+                model.quantize_config = StoreAttr()
+                model.quantize_config.desc_act = self.desc_act
+                if self.desc_act and not self.disable_exllama and self.max_input_length is not None:
+                    model = exllama_set_max_input_length(model, self.max_input_length)
+                return model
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -227,6 +227,34 @@ def _from_transformers(
...
 
+        # Patch the modules to export of GPTQ models w/o GPU
+        do_gptq_patching = False
+        config_dict = config.to_dict()
+        quantization_config = config_dict.get("quantization_config", None)
+        do_gptq_patching = quantization_config and quantization_config["quant_method"] == "gptq"
+        if do_gptq_patching:
+            torch.set_default_dtype(torch.float32)
+            orig_cuda_check = torch.cuda.is_available
+            torch.cuda.is_available = lambda: True
```
echarlaix, line 238: why is this needed ?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -238,10 +266,14 @@ def _from_transformers(
...
             trust_remote_code=trust_remote_code,
-            model_kwargs=kwargs,
             int8=load_in_8bit,
         )
 
+        # Unpatch modules after GPTQ export
+        if do_gptq_patching:
+            torch.cuda.is_available = orig_cuda_check
+            GPTQQuantizer.post_init_model = orig_post_init_model
+
```
echarlaix, line 276: I think it could make sense to add a test as well to verify we don't break anything in the future

echarlaix (commented):
Thanks for this addition @AlexKoff88 !

optimum/intel/openvino/modeling_decoder.py
AlexKoff88: GPTQ-optimized models from Transformers require NVidia GPU to be available for inference. We are introducing two changes to overcome this limitation:
- Patching some functions that prevent instantiating AutoModelForCausalLM without GPUs
- Patching QLinear forward in AutoGPTQ to enable model tracing via TorchScript

The first part comes to Optimum while the second is to OpenVINO PyTorch Frontend.

optimum/intel/openvino/modeling_decoder.py
AlexKoff88: I can add a test that checks that AutoModelForCausalLM can be instantiated without GPU. What do you think?

optimum/intel/openvino/modeling_decoder.py
echarlaix: thanks for the explanation !

optimum/intel/openvino/modeling_decoder.py
echarlaix: We can add a test for GPTQ quantized model export + inference once the PR is merged in AutoGPTQ then

echarlaix: My bad, it was clearly explained in your other comment. Let's merge, we can add a test once enabled in the next openvino release

PR #456, opened 2023-10-17
add position_ids in forward

jiqing-feng:
Hi @echarlaix 

Do you think that we should add `position_ids` in the forward of the generation model? The `optimum` has supported to generate `position_ids` in this [PR](https://github.com/huggingface/optimum/pull/1381).

cc @changwangss


optimum/intel/generation/modeling.py
```diff
@@ -88,7 +89,7 @@ def jit_trace(model: PreTrainedModel, task: str, use_cache: bool = False):
     traced_model(**model_inputs)
     traced_model(**model_inputs)
 
-    return traced_model
+    return traced_model, has_position_ids
```
echarlaix, line 92: I think we should keep `jit_trace` as it is 
```suggestion
    return traced_model

```

optimum/intel/generation/modeling.py
```diff
@@ -116,6 +118,7 @@ def __init__(
         self._device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
         self.normalized_config = NormalizedConfigManager.get_normalized_config_class(config.model_type)(config)
         self.model_dtype = kwargs.get("model_dtype", None)
+        self.has_position_ids = has_position_ids
```
echarlaix, line 121: no need to have an attribute, we can use  [`MODEL_TYPES_REQUIRING_POSITION_IDS`](https://github.com/huggingface/optimum/blob/e7bd60dd2c1e295263ba57a4e468a62ab5b179e8/optimum/exporters/onnx/utils.py#L69)  directly

echarlaix (commented):
thanks for the addition @jiqing-feng, you're right we need to add the support of `position_ids` now that it has been integrated in optimum

optimum/intel/generation/modeling.py
jiqing-feng: Thanks, I will use it.

optimum/intel/generation/modeling.py
```diff
@@ -305,6 +316,23 @@ def forward(
...
+            if not self.use_cache:
+                past_key_values_length = 0
+            else:
+                past_key_values_length = past_key_values[0][1].shape[-2]
+            position_ids = torch.arange(
+                past_key_values_length, seq_length + past_key_values_length, dtype=torch.long, device=self._device
+            ).unsqueeze(0)
+            inputs.update({"position_ids": position_ids})
+        elif not self.has_position_ids and position_ids is not None:
+            logger.warning("You miss the position_ids in the inputs")
```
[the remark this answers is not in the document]
jiqing-feng: Yes, it is more reasonable. However, for generation tasks, different decoding way will cause different inputs. For example, llama in [greedy_search](https://github.com/huggingface/transformers/blob/main/src/transformers/generation/utils.py#L2501-L2505) contains `position_ids` in inputs but [assisted_decoding](https://github.com/huggingface/transformers/blob/main/src/transformers/generation/utils.py#L4466) only have `input_ids`. Besides, we already generate `attention_mask` in the [forward](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/generation/modeling.py#L264). WDYT?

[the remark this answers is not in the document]
echarlaix: I see your point, I'm ok with the modification but think we need to add a test for every architecture to verify we create it correctly. For example is `past_key_values_length = past_key_values[0][1].shape[-2]` for every architecture ?

jiqing-feng:
Hi @echarlaix . I have fixed all tests, would you please help me review these changes? Thx!

jiqing-feng:
Hi @echarlaix , could you have a look at this change? Thx 😄 !

jiqing-feng:
Hi @echarlaix . Would you please help me review these changes? This change could avoid forward failure because of the `position_ids`. For now, `position_ids` is just like `past_key_values` in some models, it should be contained in the model inputs since jit trace dummy inputs have `position_ids`.

optimum/intel/generation/modeling.py
```diff
@@ -296,8 +295,24 @@ def forward(
...
+                    past_key_values[0].shape[-2]
+                    if model_type.replace("-", "_") in MULTI_QUERY_ATTN_MODELS
+                    else past_key_values[0][1].shape[-2]
+                )
+            position_ids = torch.arange(
+                past_key_values_length, seq_length + past_key_values_length, dtype=torch.long, device=self._device
+            ).unsqueeze(0)
+            inputs.update({"position_ids": position_ids})
+        elif not has_position_ids and position_ids is not None:
+            logger.warning("You miss the position_ids in the inputs")
```
echarlaix, line 315: I think it would be better to check directly in the model graph to check whether `position_ids` is one of the model's expected input if that can be done. If not this new addition will create issues for all the previously exported models (the ones that were exported without any `position_ids)for all architectures from `MODEL_TYPES_REQUIRING_POSITION_IDS`
```suggestion
        if "position_ids" in self.input_names:
            if position_ids is None:
                position_ids = ...
            inputs["position_ids"] = position_ids
```

optimum/intel/generation/modeling.py
```diff
@@ -264,6 +262,7 @@ def forward(
         }
 
         model_type = self.config.model_type.replace("_", "-")
+        has_position_ids = True if model_type in MODEL_TYPES_REQUIRING_POSITION_IDS else False
```
echarlaix, line 265:
```suggestion
        has_position_ids = model_type in MODEL_TYPES_REQUIRING_POSITION_IDS

```

optimum/intel/generation/modeling.py
echarlaix: also concerning the `position_ids` (and the `attention_mask`) computation I think we should do as follow : https://github.com/huggingface/optimum-intel/blob/819e3e824bdcef574a1487f700acd9a2fe39ddac/optimum/intel/openvino/modeling_decoder.py#L365-L383

jiqing-feng: I am afraid there is no `input_names` attr in this class.

echarlaix: yes could this be added you think ?

jiqing-feng: Yes, I can try it

jiqing-feng:
Hi @echarlaix . I am afraid `input_names` is not a great way. For example, the `model_input_names` in `GPT2` is `["input_ids", "attention_mask"]`, see [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/tokenization_gpt2.py#L159). However, `GPT2` will prepare `position_ids` in generation which is not in the `model_input_names`, see [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/modeling_gpt2.py#L1016). 

I think it would be better to keep this way and don't use `input_names`, WDYT?

echarlaix:
> Hi @echarlaix . I am afraid `input_names` is not a great way. For example, the `model_input_names` in `GPT2` is `["input_ids", "attention_mask"]`, see [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/tokenization_gpt2.py#L159). However, `GPT2` will prepare `position_ids` in generation task which is not in the `model_input_names`, see [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/modeling_gpt2.py#L1016).
> 
> I think it would be better to keep this way and don't use `input_names` since our goal is to enable jit trace model support generation task, WDYT?

I was suggesting to check the model graph directly as done in https://github.com/huggingface/optimum-intel/blob/v1.12.1/optimum/intel/openvino/modeling_base.py#L82 (to check whether position_ids is one of the model's expected input) If that can't be done, this PR might results in issues for all the previously exported models (the ones that were exported without any position_ids) for all architectures from `MODEL_TYPES_REQUIRING_POSITION_IDS`

jiqing-feng:
Hi @echarlaix . I think I got what you mean. I checked the forward inputs by the graph model inputs, could you please help me to review these changes? Thx!

jiqing-feng:
Hi @Ella . Sorry for the misunderstanding. I just found that there is no way to get the input names from a Torch Script model, so I can only get the input names when tracing the model. Would like to hear your opinion. Thx!

echarlaix:
> Hi @echarlaix . Sorry for the misunderstanding. I just found that there is no way to get the input names from a Torch Script model, so I can only get the input names when tracing the model. Would like to hear your opinion. Thx!

I was able to have something with :   
```python
 input_names = [inputs.debugName() for inputs in model.graph.inputs()]
```
can you check it out ?

jiqing-feng:
Hi @echarlaix . Thanks for your advice, it perfectly fixed my problem. Would you please review these changes? Thx!

echarlaix:
> Hi @echarlaix . Thanks for your advice, it perfectly fixed my problem. Would you please review these changes? Thx!
> 
> And the failed CIs are not related to my changes

Added updates in https://github.com/jiqing-feng/optimum-intel/pull/2, can you take a look ?

echarlaix:
Also could you add a test before we can merge ?

jiqing-feng:
Hi @echarlaix . I have merged your changes and also added the tests. Would you please help to review the test function? Thx!

optimum/intel/generation/modeling.py
```diff
@@ -275,6 +272,15 @@ def forward(
...
         }
 
+        if "position_ids" in self.input_names:
+            if position_ids is None:
+                position_ids = attention_mask.long().cumsum(-1) - 1
+                position_ids.masked_fill_(attention_mask == 0, 1)
+                if past_key_values:
+                    position_ids = position_ids[:, -1].unsqueeze(-1)
+
+            inputs["position_ids"] = position_ids
```
echarlaix, line 282: To fix the CI you can do :

```suggestion
        if "position_ids" in self.input_names and position_ids is None:
            position_ids = attention_mask.long().cumsum(-1) - 1
            position_ids.masked_fill_(attention_mask == 0, 1)
            if past_key_values:
                position_ids = position_ids[:, -1].unsqueeze(-1)
        if position_ids is not None:
            inputs["position_ids"] = position_ids

```

optimum/intel/generation/modeling.py
```diff
@@ -275,6 +272,15 @@ def forward(
...
         }
 
+        if "position_ids" in self.input_names and position_ids is None:
+            position_ids = attention_mask.long().cumsum(-1) - 1
+            position_ids.masked_fill_(attention_mask == 0, 1)
+            if past_key_values:
+                position_ids = position_ids[:, -1].unsqueeze(-1)
+        if position_ids is not None:
+            inputs["position_ids"] = position_ids
+
```
echarlaix, line 282:
```suggestion

```

optimum/intel/generation/modeling.py
```diff
@@ -275,6 +272,14 @@ def forward(
             "attention_mask": attention_mask,
         }
 
+        if "position_ids" in self.input_names and position_ids is None:
+            position_ids = attention_mask.long().cumsum(-1) - 1
+            position_ids.masked_fill_(attention_mask == 0, 1)
+            if past_key_values:
+                position_ids = position_ids[:, -1].unsqueeze(-1)
+        if position_ids is not None:
+            inputs["position_ids"] = position_ids
```
echarlaix, line 281:
```suggestion
        if "position_ids" in self.input_names and position_ids is None:
            position_ids = attention_mask.long().cumsum(-1) - 1
            position_ids.masked_fill_(attention_mask == 0, 1)
            if past_key_values:
                position_ids = position_ids[:, -1].unsqueeze(-1)

        if "position_ids" in self.input_names or not self.input_names:
            inputs["position_ids"] = position_ids
```

PR #457, opened 2023-10-19
allow optionally fill attention mask in forward

eaidova:
# What does this PR do?
allow passing inputs without attention mask for OVModelForCausalLM

For inference openvino expect that all inputs that exists in model filled before starting inference.
Input filling for text generation controlled when you run generate function and guarantee that all inputs provided.
But when user trying to use forward directly (building own custom generation pipeline for example), there is no knowledge which inputs available in model.
PyTorch models allow run model without passing some optional inputs, like attention_mask (filling it inside with default values). In our case, if user do not provide this input, it leads to error that do not have any context what went wrong.

These changes adding attention_mask filling inside forward for improving user experience 
## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -352,7 +352,10 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
             input_ids = input_ids[:, -1:]
 
         inputs = {}
+        past_len = 0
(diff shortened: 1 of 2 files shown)


eaidova:
@echarlaix @AlexKoff88 @helena-intel could you please take a look?

AlexKoff88 (approved):

echarlaix (approved):
Thanks for the addition !

PR #459, opened 2023-10-23
Enable openvino inference for GPTBigCode models

echarlaix:


tests/openvino/test_modeling.py
```diff
@@ -459,6 +460,7 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "gpt_neox",
         "llama",
         "marian",
+        # "mistral",
```
echarlaix, line 463: will be supported in optimum next release

PR #465, opened 2023-10-30
move gptq patching to export stage

eaidova:
move GPTQ patching to export inside export_main to allow convert models using optimum-cli
changed default factory class in OVModelForCausalLM to be able override it (in case if used model class inherited from OVModelForCausalLM, returns instance of this class in backward compatible way, current changes leads to getting always OVModelForCausalLM that may be unexpected from user side)


eaidova:
@AlexKoff88 @helena-intel @echarlaix could you please take a look?

AlexKoff88 (approved):

echarlaix (approved):
LGTM, thanks @eaidova

PR #470, opened 2023-11-03
add support whisper for openvino

eaidova:
# What does this PR do?

Adding support for inference whisper models family with openvino.
Verified on [distil-whisper-large-v2](https://huggingface.co/distil-whisper/distil-large-v2)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -23,23 +23,21 @@ from transformers import AutoConfig, AutoTokenizer
 from optimum.exporters import TasksManager
 from optimum.exporters.onnx import __main__ as optimum_main
 from optimum.exporters.onnx.base import OnnxConfig, OnnxConfigWithPast
+from optimum.utils import DEFAULT_DUMMY_SHAPES
+from optimum.utils.save_utils import maybe_load_preprocessors, maybe_save_preprocessors
+
+from ...intel.utils.import_utils import is_nncf_available, is_optimum_version, is_transformers_version
+from .convert import export_models
 
 
-try:
+if is_optimum_version(">=", "1.16.0"):
     from optimum.exporters.onnx.constants import SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED
-except ImportError:
-    # Duplicated from https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/constants.py
-    # until it is not part of package
+else:
+    # Copied from https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/constants.py
     SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED = [
         "bart",
         "whisper",
     ]
-from optimum.utils import DEFAULT_DUMMY_SHAPES
-from optimum.utils.save_utils import maybe_load_preprocessors, maybe_save_preprocessors
-
-from ...intel.utils.import_utils import is_nncf_available, is_transformers_version
-from .convert import export_models
-
 
 OV_XML_FILE_NAME = "openvino_model.xml"
 
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -23,10 +23,21 @@ from transformers import AutoConfig, AutoTokenizer
 from optimum.exporters import TasksManager
 from optimum.exporters.onnx import __main__ as optimum_main
 from optimum.exporters.onnx.base import OnnxConfig, OnnxConfigWithPast
+
+
+try:
+    from optimum.exporters.onnx.constants import SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED
+except ImportError:
+    # Duplicated from https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/constants.py
+    # until it is not part of package
+    SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED = [
+        "bart",
+        "whisper",
+    ]
 from optimum.utils import DEFAULT_DUMMY_SHAPES
 from optimum.utils.save_utils import maybe_load_preprocessors, maybe_save_preprocessors
 
-from ...intel.utils.import_utils import is_nncf_available
+from ...intel.utils.import_utils import is_nncf_available, is_transformers_version
 from .convert import export_models
 
 
@@ -140,10 +151,12 @@ def main_export(
     do_gptq_patching = False
     try:
         config = AutoConfig.from_pretrained(model_name_or_path, trust_remote_code=trust_remote_code)
+        model_type = config.model_type.replace("_", "-")
         config_dict = config.to_dict()
         quantization_config = config_dict.get("quantization_config", None)
         do_gptq_patching = quantization_config and quantization_config["quant_method"] == "gptq"
     except Exception:
+        model_type = None
         pass
 
     if do_gptq_patching:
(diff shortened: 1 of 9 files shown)


eaidova:
@AlexKoff88  @echarlaix  @helena-intel could you please take a look?

eaidova:
@echarlaix  I pin transformers<4.35 version as workaround for incompatibility issues to be able pass ci checks (removing SSDP, refactoring models that lead to previously used patching approach does not work anymore for llama, opt, and other models).
Is there any plans to fix these issues and alignment with optimum package?

AlexKoff88 (approved):

adrianboguszewski:
@echarlaix @helena-intel could we prioritize this PR? We need it very much. It will simplify using OpenVINO for Whisper a lot :)

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -610,3 +670,138 @@ def _reshape(self, model: openvino.runtime.Model, batch_size: int, sequence_leng
...
+        top_inheritance_index = OVModelForSpeechSeq2Seq.__mro__.index(GenerationMixin)
+        return (
+            (cls,)
+            + OVModelForSpeechSeq2Seq.__mro__[:top_inheritance_index]
+            + (WhisperForConditionalGeneration,)
+            + OVModelForSpeechSeq2Seq.__mro__[top_inheritance_index:]
+        )
+
+
+class OVModelForWhisper(
```
echarlaix, line 793: might make sense to use instead :
```suggestion
class _OVModelForWhisper(
```
in case we want to deprecate in the future

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -610,3 +670,138 @@ def _reshape(self, model: openvino.runtime.Model, batch_size: int, sequence_leng
...
+)
+class OVModelForSpeechSeq2Seq(OVModelForSeq2SeqLM):
+    auto_model_class = AutoModelForSpeechSeq2Seq
+    main_input_name = "input_features"
+    export_feature = "automatic-speech-recognition"
+
+    def prepare_inputs_for_generation(
+        self,
+        input_ids,
+        input_features: Optional[torch.FloatTensor] = None,
```
echarlaix, line 689: Is this needed ?
```suggestion

```
I don't see it  in
https://github.com/huggingface/transformers/blob/main/src/transformers/models/whisper/modeling_whisper.py#L2494
or 
https://github.com/huggingface/transformers/blob/a49f4acab3c1eea82907e12f82eafbd4673deb39/src/transformers/models/speech_to_text/modeling_speech_to_text.py#L1339

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -610,3 +670,138 @@ def _reshape(self, model: openvino.runtime.Model, batch_size: int, sequence_leng
...
+        cross_attn_head_mask=None,
+        use_cache=None,
+        encoder_outputs=None,
+        **kwargs,
+    ) -> Dict:
+        if decoder_attention_mask is None:
+            decoder_attention_mask = torch.ones_like(input_ids).to(input_ids.device)
+
+        return {
+            "input_features": input_features,
```
echarlaix, line 704:
```suggestion

```

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -610,3 +670,138 @@ def _reshape(self, model: openvino.runtime.Model, batch_size: int, sequence_leng
...
+                decoder_attention_mask=decoder_attention_mask,
+                past_key_values=past_key_values,
+                encoder_hidden_states=encoder_outputs.last_hidden_state,
+                encoder_attention_mask=attention_mask,
+            )
+
+        return Seq2SeqLMOutput(
+            logits=decoder_outputs.logits,
+            past_key_values=decoder_outputs.past_key_values,
+        )
```
echarlaix, line 763: very close to`OVModelForSeq2SeqLM.forward` and `OVModelForPix2Struct.forward`, it would be nice to uniformize everything and re-use `OVModelForSeq2SeqLM.forward` when possible (here with something like `super().forward(input_ids=input_features, ...)`

echarlaix (commented):
very nice PR @eaidova

optimum/intel/openvino/modeling_seq2seq.py
eaidova: applyed

optimum/intel/openvino/modeling_seq2seq.py
eaidova: removed

optimum/intel/openvino/modeling_seq2seq.py
eaidova: removed

optimum/intel/openvino/modeling_seq2seq.py
eaidova: renamed

eaidova:
@echarlaix looks like transformers 4.36 break some model export, including whisper support. I added workaround similar to https://github.com/huggingface/optimum/pull/1590, please take a look

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -610,3 +652,526 @@ def _reshape(self, model: openvino.runtime.Model, batch_size: int, sequence_leng
...
+        and put GenerationMixin at the end of the MRO
+        """
+        top_inheritance_index = OVModelForSpeechSeq2Seq.__mro__.index(GenerationMixin)
+        return (
+            (cls,)
+            + OVModelForSpeechSeq2Seq.__mro__[:top_inheritance_index]
+            + (WhisperForConditionalGeneration,)
+            + OVModelForSpeechSeq2Seq.__mro__[top_inheritance_index:]
+        )
+
```
echarlaix, line 750: Thanks a lot for fixing the whisper generation (now that the model attribute is no longer accessible with latest transformers release) we can also remove `MetaClassRemoveParentsAndReorder`
```suggestion
```

optimum/exporters/openvino/__main__.py
```diff
@@ -23,10 +23,21 @@
 from optimum.exporters import TasksManager
 from optimum.exporters.onnx import __main__ as optimum_main
 from optimum.exporters.onnx.base import OnnxConfig, OnnxConfigWithPast
+
+
+try:
+    from optimum.exporters.onnx.constants import SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED
+except ImportError:
```
echarlaix, line 30: we can also check the optimum version with [`is_optimum_version`](https://github.com/huggingface/optimum-intel/blob/a30e337542c5d81c1ba0c6545fa0a5651f046dec/optimum/intel/utils/import_utils.py#L179) (might be useful in the future in case something breaks)

optimum/intel/openvino/modeling_seq2seq.py
echarlaix: thanks!

echarlaix (approved):
Looks great, thanks a lot for the addition @eaidova  :fire:

optimum/intel/openvino/modeling_seq2seq.py
eaidova: removed

optimum/exporters/openvino/__main__.py
eaidova: refactored

PR #477, opened 2023-11-07
fix causallm models export

eaidova:
# What does this PR do?

fix export LLMs after update on optimum 1.14.0

I observe that all CausalLM models currently have fallback on onnx export instead of using direct pytorch export. It is connected with latest optimum changes that uses model patcher for patching attention mask in decoder. OpenVINO pytorch export functionality based on torchscript now and can not use models with return_dict = True due to TS tracing limitations (requires that all values in output dict have the same type, which is not achievable with models outputs because it may contain None for loss or tuple with past_key_values in the same time with pytorch tensors), so previously if custom patching used, then we apply patch on top of original patcher forward. If I right understand, now patcher can override not only forward, but any other method and also there is DecoderModelPatcher that is used for CauslaLM models does not override forward at all.
This PR should resolve proper method patching for decoder models

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -28,6 +28,7 @@ from optimum.exporters.onnx.base import OnnxConfig
 from optimum.exporters.onnx.convert import check_dummy_inputs_are_allowed
 from optimum.exporters.onnx.convert import export_pytorch as export_pytorch_to_onnx
 from optimum.exporters.onnx.convert import export_tensorflow as export_tensorflow_onnx
+from optimum.exporters.onnx.model_patcher import DecoderModelPatcher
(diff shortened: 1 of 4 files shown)


eaidova:
@helena-intel  @AlexKoff88  @echarlaix could you please take a look?

helena-intel:
Thanks for the quick fix @eaidova ! It worked well in my test. Would it be good to have a test that verifies that a known working model is indeed exported with PyTorch, to prevent this in the future? We could just check the openvino_model.xml file for 			`<framework value="pytorch" />` for example. Not asking to add it to this PR, could be done later.

AlexKoff88 (approved):

echarlaix (approved):
Thanks for the fix @eaidova

PR #481, opened 2023-11-29
pkv dtype

jiqing-feng:
Hi, @echarlaix . I changed the past key values data type to the model's data type, otherwise the trace will fail if we use bf16 model.


jiqing-feng:
Hi @echarlaix . The failed CIs do not relate to my changes, would you please have a look? Thx!

echarlaix (approved):

PR #486, opened 2023-12-07
Stateful models by default

slyalin:
# What does this PR do?

Support to convert and run stateful models.
Similar to https://github.com/huggingface/optimum-intel/pull/493, but makes stateful models by default in optimum-cli and from_pretrained.

## TODO
- [x] Command line interface update
- [ ] Testing on a wider range of topologies

## Before submitting


slyalin:
> Besides the first two items described in TODO we should revise Seq2seq models as well and make decoding part stateful as well: https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_seq2seq.py#L412

@AlexKoff88, is it going to just cover more model architectures to be stateful without providing extra value to those architectures that have already been enabled in this PR? Can we do that in a separate PR? So we will introduce stateful capabilities gradually providing necessary API changes in this first PR?

AlexKoff88:
> > Besides the first two items described in TODO we should revise Seq2seq models as well and make decoding part stateful as well: https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_seq2seq.py#L412
> 
> @AlexKoff88, is it going to just cover more model architectures to be stateful without providing extra value to those architectures that have already been enabled in this PR? Can we do that in a separate PR? So we will introduce stateful capabilities gradually providing necessary API changes in this first PR?

I think we can do it in a separate PR when reworking Seq2seq models and getting rid of two decoder sub-models.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -316,49 +342,66 @@ def forward(
...
+        else:
+            # past_key_values are not used explicitly, instead they are handled inside the model
+            if past_key_values is None:
+                # Need a marker to differentiate the first generate iteration from the others in
+                # the first condition at the function beginning above.
+                # It should be something that is not None and it should be True when converted to Boolean.
+                past_key_values = ((),)
+                # This is the first iteration in a sequence, reset all states
+                for state in self.request.query_state():
+                    state.reset()
```
usstq, line 401: Maybe we can call [set_state](https://docs.openvino.ai/2023.2/openvino_docs_OV_UG_model_state_intro.html#openvino-state-api) instead of reset, because set_state can provide exact static shape with correct batch_size and 0 as length, but reset() on dynamic shape state-variable put the state-variable into a "undefined" shape state (what will happen if we call get_state() after reset()? ).

slyalin: Not required, we cover this case with properly constructed state initialization expression right inside the model. It will obtain correct shapes even in case if batch dimension is dynamic -- it will be propagated to initial state tensor shape. It is well known part for both CPU and GPU and all necessary optimizations should be applied inside the plugins.

optimum/exporters/openvino/convert.py
```diff
@@ -364,6 +382,13 @@ def ts_patched_forward(*args, **kwargs):
             inp_tensor.get_node().set_partial_shape(static_shape)
             inp_tensor.get_node().set_element_type(get_element_type(inp_data.cpu().numpy().dtype))
         ov_model.validate_nodes_and_infer_types()
+
+        if stateful:
```
eaidova, line 386: should we also check task before applying patch? for making this transformation, the model should have with_past in task type and if I right understand current changes are targeted only for text-generation-with-past, there is also seq2seq models with decoder with past as @AlexKoff88 already mentioned in comment, where this transformation should be applied on only one from 3 models

optimum/exporters/openvino/stateful.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+import numpy as np
+from packaging import version
+import openvino as ov
+from openvino.runtime import opset13
+from optimum.intel.utils.import_utils import is_openvino_version
+
+
+def model_has_name(ov_model: ov.Model, name: str):
```
eaidova, line 23:
```suggestion
def model_has_input_output_name(ov_model: ov.Model, name: str):
```

optimum/exporters/openvino/stateful.py
```diff
@@ -0,0 +1,146 @@
...
+    return name in sum([list(t.get_names()) for t in ov_model.inputs + ov_model.outputs], list())
+
+
+def model_has_input(ov_model: ov.Model, name: str):
+    return name in sum([list(t.get_names()) for t in ov_model.inputs], list())
+
+
+def model_has_cache_reorder(ov_model):
+    return model_has_input(ov_model, 'beam_idx')
```
Wovchena, line 32: The name should be `beam_id` or `beam_ids` if you count batch to stay aligned with `input_ids`

Wovchena: Never mind. It's reasonable to state that `idx` is singular form of the word `ids`

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -316,49 +342,66 @@ def forward(
...
+                # Set initial value for the next beam_idx input that will be used at the current iteration
+                # and will be optionally updated by _reorder_cache at the next iterations if beam_search is used
+                self.next_beam_idx = np.array(range(batch_size), dtype=int)
```
echarlaix, line 404: would prefer to had it defined in `__init__` directly (can be set to `None`)

echarlaix (commented):
Thanks for the addition @slyalin !!

optimum/exporters/openvino/stateful.py
```diff
@@ -0,0 +1,146 @@
...
+        # Re-validation model if shapes are altered above
+        ov_model.validate_nodes_and_infer_types()
+
+    apply_make_stateful_transformation(ov_model, input_output_map)
+    if num_beams_and_batch is None:
+        build_state_initializer(ov_model, batch_dim)
+
+
+def raise_if_openvino_is_too_old():
+    if is_openvino_version("<=", "2023.2"):
```
Wovchena, line 123: That prohibits https://storage.openvinotoolkit.org/repositories/openvino/packages/nightly/2023.3.0-13432-a6ea22ad0e6/l_openvino_toolkit_ubuntu20_2023.3.0.dev20231129_x86_64.tgz usage. I get `ValueError: Could not create or use stateful model when using old version of openvino==2023.3.0-13432-a6ea22ad0e6. Install openvino>=2023.3.0.`

optimum/exporters/openvino/convert.py
slyalin: As it is implemented now, it is a user responsibility to set stateful=True only for right models. To set it by default there should be more adjustments in the code, which I cannot provide in this PR. Part of the models will not be supported. I am just not so familiar with the code base to provide necessary changes to enable by default. We need someone who can do that.

optimum/exporters/openvino/stateful.py
slyalin: I wasn't super creative when choosing this name, and just borrowed it from `_reorder_cache` method argument name. So this is aligned with that part of the code that this thing comes from. And this `idx` is not the same as those `id` or `ids` they are indexes for a different thing.

optimum/exporters/openvino/stateful.py
slyalin: As I remember the previous version of this condition was different and didn't work with nightly. I have changed it and tested with nightly -- everything worked. Not sure why it doesn't work now -- probably I messed up with the versions when checked on my side.

optimum/exporters/openvino/better_transformer_patch.py
```diff
@@ -0,0 +1,41 @@
...
+    def pathed_generate_dummy_inputs(self, *args, **kwargs):
+        dummy_inputs = self._original_generate_dummy_inputs(*args, **kwargs)
+        if 'input_ids' in dummy_inputs and dummy_inputs['input_ids'].shape[1] == 1:
+            dummy_inputs['input_ids'] = torch.cat([dummy_inputs['input_ids'], dummy_inputs['input_ids']], dim=-1)
+            attention_mask = dummy_inputs['attention_mask']
+            dummy_inputs['attention_mask'] = torch.cat([attention_mask, attention_mask.new_ones((attention_mask.shape[0], 1))], dim=-1)
+        return dummy_inputs
+
+    model_config._original_generate_dummy_inputs = model_config.generate_dummy_inputs
+    model_config.generate_dummy_inputs = types.MethodType(pathed_generate_dummy_inputs, model_config)
```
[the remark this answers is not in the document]
slyalin: Probably it can be used now. When this code was first written it was applied externally as a patch for the model. Now it is most likely can be implemented directly in export function.

optimum/exporters/openvino/stateful.py
slyalin: @eaidova reproduced this issue when both openvino and openvino-nightly installed. @Wovchena please uninstall the packages and install openvino-nightly only.

optimum/intel/openvino/modeling_decoder.py
slyalin, line 401: We can use one call of recently added `reset_state` method: https://github.com/openvinotoolkit/openvino/pull/21660/files#diff-a52b634025ad126ebd274f6f1dd7d516c25deda996a180a0b10535c845a25a53R439.

optimum/commands/export/openvino.py
```diff
@@ -139,5 +144,6 @@ def run(self):
             pad_token_id=self.args.pad_token_id,
             compression_option=self.args.weight_format,
             compression_ratio=self.args.ratio
```
Wovchena, line 146:
```suggestion
            compression_ratio=self.args.ratio,
```

optimum/exporters/openvino/stateful.py
ilya-lavrenov: I suppose we need to fix it to work with different OpenVINO installations.

optimum/exporters/openvino/stateful.py
```diff
@@ -0,0 +1,146 @@
...
+
+def model_has_input(ov_model: ov.Model, name: str):
+    return name in sum([list(t.get_names()) for t in ov_model.inputs], list())
+
+
+def model_has_cache_reorder(ov_model):
+    return model_has_input(ov_model, 'beam_idx')
+
+
+def model_has_state(ov_model):
```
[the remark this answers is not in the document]
eaidova: We want to use this method externally, so that is why naming without underscore is preferable for us.

TO DO inside means that possibly the realization changed in future, but as part of API, this function is very useful

optimum/intel/openvino/modeling_decoder.py
eaidova: could you please explain? I do not understand your suggestion

optimum/exporters/openvino/stateful.py
eaidova: renamed in https://github.com/huggingface/optimum-intel/pull/493

optimum/exporters/openvino/convert.py
```diff
@@ -364,6 +382,13 @@ def ts_patched_forward(*args, **kwargs):
...
+            # Patching model according to stateful parameters
+            model.key_value_input_names = [name for name in input_names if name.startswith('past_key_values.')]
+            model.key_value_output_names = [name for name in output_names if name.startswith('present.')]
```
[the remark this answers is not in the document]
eaidova: rewritten in https://github.com/huggingface/optimum-intel/pull/493

optimum/commands/export/openvino.py
eaidova: fixed in https://github.com/huggingface/optimum-intel/pull/493

optimum/exporters/openvino/stateful.py
eaidova: I have no issues with detection version using python libs from archive. The problem looks like happens in case if you have installed openvino from multiple sources in the same time (e.g. pypi openvino + pypi openvino-nightly or pypi openvino + PYTHONPATH to openvino libs) )and their versions are different

eaidova: in my PR with fixes for this branch,
https://github.com/huggingface/optimum-intel/pull/493
 I changed logic for version verification. previously used approach checked only packages that exists in site-packages of python environment (only installed as wheels), but does not take into account that user (in development case for example) install openvino as archive or build from source without installing wheels.

Now, version openvino should be aligned with import order

optimum/intel/openvino/modeling_decoder.py
eaidova: yes, but we already started to use this branch in benchmarking, so we need to be able validate it on previous package, that may not contains these changes yet.

My suggestion to add check that this method exists and as workaround perform reset in cycle if not, until ov release will not happen, after release we change lower bound for ov and remove this workaround

eaidova: done in https://github.com/huggingface/optimum-intel/pull/493

optimum/exporters/openvino/better_transformer_patch.py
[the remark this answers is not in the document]
eaidova: that will not work, because there is additional logic that modify sequence_len in OnnxConfigWithPast after original shapes specification
https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/base.py#L663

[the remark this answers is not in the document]
eaidova: starting with optimum 1.14.0 this behaviour triggered only by legacy path, but we do not use it, so it can be removed at all

optimum/intel/openvino/modeling_decoder.py
echarlaix: I would expect to have all attributes defined in the `__init__`, which is not the case here

slyalin:
@echarlaix, I cannot push changes to https://github.com/huggingface/optimum-intel/pull/493, so I've continued development in my original PR with all changes from that PR. Could you approve workflows to allow checks running here?

optimum/intel/openvino/quantization.py
```diff
@@ -312,10 +312,12 @@ def start_async(
...
                 inputs: Any = None,
                 userdata: Any = None,
-                shared_memory: bool = False,
+                share_inputs: bool = False,
+                *,
+                shared_memory: Any = None,
             ):
                 data_cache.append(inputs)
-                self.request.infer(inputs, shared_memory)
+                self.request.infer(inputs, share_inputs, share_outputs=True, shared_memory=shared_memory)
```
slyalin, line 320: @AlexKoff88, this is how I have fixed it. Please provide feedback if something looks weird from your perspective.

optimum/intel/openvino/modeling_decoder.py
eaidova: added in https://github.com/huggingface/optimum-intel/pull/493

slyalin:
This PR is completely superseded by https://github.com/huggingface/optimum-intel/pull/493.

PR #495, opened 2023-12-25
fix load_in_int8 behaviour for large models

eaidova:
# What does this PR do?

update load_in_int8 logic in from_pretrained for getting the following behavior:

if user specified explicitly load_in_int8=True - model weights will be compressed to int8 after conversion
if user specified explicitly load_in_int8=Flase - model weights will not be compressed
if user does not specify this argument, allow to decide based on model size 

This behavior is also aligned with optimum-cli export for openvino

Also changed providing load_in_8bit argument in _from_pretrained method after _from_transformers for avoiding double weights compression during model loading

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -302,7 +302,7 @@ class OVBaseModel(OptimizedModel):
         )
 
         config.save_pretrained(save_dir_path)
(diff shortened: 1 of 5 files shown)


AlexKoff88 (approved):

AlexKoff88:
thanks for fixing that

AlexKoff88 (approved):

AlexKoff88 (approved):

PR #500, opened 2024-01-05
Convert tokenizers with openvino_tokenizers

slyalin:
# What does this PR do?

Export tokenizer and detokenizer as OpenVINO models using openvino_tokenizers from https://github.com/openvinotoolkit/openvino_contrib/tree/master/modules/custom_operations/user_ie_extensions/tokenizer/python. Activated by default as a part of optimum-cli export openvino command line tool. Compatible with https://github.com/openvinotoolkit/openvino.genai/tree/master/text_generation/causal_lm/cpp.

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -348,7 +348,7 @@ def main_export(
             generation_config.save_pretrained(output)
         maybe_save_preprocessors(model_name_or_path, output)
         try:
-            # Avoid loding it for the second time if loaded before
+            # TODO: Avoid loading the tokenizer again if loaded before
             tokenizer = AutoTokenizer.from_pretrained(model_name_or_path)
             tokenizer_export(tokenizer, output)
         except:
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -19,6 +19,7 @@ from typing import Any, Callable, Dict, Optional, Union
 
 from requests.exceptions import ConnectionError as RequestsConnectionError
 from transformers import AutoConfig, AutoTokenizer
+from openvino import save_model
 
 from optimum.exporters import TasksManager
 from optimum.exporters.onnx import __main__ as optimum_main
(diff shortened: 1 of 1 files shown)


slyalin:
@apaniukov will cover this functionality. This PR is just to initiate discussion.

optimum/exporters/openvino/__main__.py
```diff
@@ -328,6 +347,12 @@ class StoreAttr(object):
         if generation_config is not None:
             generation_config.save_pretrained(output)
         maybe_save_preprocessors(model_name_or_path, output)
+        try:
+            # Avoid loding it for the second time if loaded before
```
slyalin, line 351:
```suggestion
            # TODO: Avoid loading the tokenizer again if loaded before
```

optimum/exporters/openvino/__main__.py
```diff
@@ -46,6 +47,24 @@
...
+        from openvino_tokenizers import convert_tokenizer
+        ov_tokenizer, ov_detokenizer = convert_tokenizer(tokenizer, with_detokenizer=True)
+        if isinstance(output, str):
+            output = Path(output)
+        tokenizer_path = output.joinpath("openvino_tokenizer" + suffix + ".xml")
+        detokenizer_path = output.joinpath("openvino_detokenizer" + suffix + ".xml")
+        save_model(ov_tokenizer, tokenizer_path)
+        save_model(ov_detokenizer, detokenizer_path)
+    except Exception as exception:
+        print("[ WARNING ] OpenVINO tokenizer/detokenizer models couldn't be exported because of exception:", exception)
```
eaidova, line 65:
```suggestion
        logger.warning(OpenVINO tokenizer/detokenizer models couldn't be exported because of exception:", exception)
```

optimum/exporters/openvino/__main__.py
```diff
@@ -328,6 +347,12 @@ class StoreAttr(object):
         if generation_config is not None:
             generation_config.save_pretrained(output)
         maybe_save_preprocessors(model_name_or_path, output)
+        try:
+            # TODO: Avoid loading the tokenizer again if loaded before
+            tokenizer = AutoTokenizer.from_pretrained(model_name_or_path)
+            tokenizer_export(tokenizer, output)
+        except:
+            print("[ WARNING ] Could not load tokenizer using specified model ID or path. OpenVINO tokenizer/detokenizer models won't be generated.")
```
eaidova, line 355: tokenizer already loaded in maybe_load_preprocessors function, I recommend to check this function result instead. Also possibly you should take into account trust_remote_code parameter if you want load tokenizer explicitly

optimum/exporters/openvino/__main__.py
```diff
@@ -46,6 +47,24 @@
 logger = logging.getLogger(__name__)
 
 
+def tokenizer_export(
```
eaidova, line 50: I think it should be in convert.py file together with other conversion functions.

optimum/exporters/openvino/__main__.py
apaniukov: Reuse tokenizer from `maybe_load_preprocessors` result. But a new problem is that there are two identical tokenizers there - one from `AutoTokenizer` and one from `AutoProcessor`.

apaniukov: Here is the PR: https://github.com/slyalin/optimum-intel/pull/2

optimum/exporters/openvino/__main__.py
apaniukov: Moved.

optimum/exporters/openvino/__main__.py
apaniukov: Fixed.

slyalin:
Closing in favor of #513

PR #506, opened 2024-01-09
Add openvino-nightly to automated tests

helena-intel:
Add openvino-nightly to automated tests (in addition to the openvino that is installed by setup.py).
We may not want to keep this for all the tests forever since it doubles the amount of tests on every PR - but for now it would be useful to check PRs with openvino-nightly before the OpenVINO 2023.0 release, and to catch any issues with openvino-nightly in existing code.
Also changed Python versions to test to 3.8 and 3.11.

### DIFF
--- a/.github/workflows/test_openvino.yml
+++ b/.github/workflows/test_openvino.yml
@@ -17,8 +17,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.8, 3.11]
         os: [ubuntu-latest]
+        openvino: ["openvino", "openvino-nightly"]
 
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/test_openvino.yml
```diff
@@ -33,12 +34,9 @@ jobs:
         # install PyTorch CPU version to avoid installing CUDA packages on GitHub runner without GPU
         pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
         pip install .[openvino,nncf,tests,diffusers]
+    - name: Install openvino-nightly (optional)
```
echarlaix, line 37: It make sense to have tests with `openvino-nightly`, could make sense to add it in the nightly tests instead, wdyt ?

helena-intel: @echarlaix I think that for now it would be useful to have it here, so we can test the PRs that target 2023.3. And then after 2023.3 we can move this to nightly tests?

echarlaix: Yes the only issue is that the CI will be red until the next release (as the quantization tests will fail for one at least of the openvino version), will merge it anyway if that helps on your side!

PR #508, opened 2024-01-09
f32 precision for compare-with-transformers tests

helena-intel:
Allow these tests to pass locally on devices where inference is run by default in FP16 (GPU) or BF16 (latest Xeon).

Also added an OpenVINO version check to make a test pass in 2023.3 (get_property now returns a string instead of an object).

### DIFF
--- a/tests/openvino/test_modeling.py
+++ b/tests/openvino/test_modeling.py
@@ -90,6 +90,8 @@ TENSOR_ALIAS_TO_TYPE = {
 
 SEED = 42
 
+F32_CONFIG = {"CACHE_DIR": "", "INFERENCE_PRECISION_HINT": "f32"}
+
 
 class Timer(object):
     def __enter__(self):
@@ -251,9 +253,7 @@ class OVModelForSequenceClassificationIntegrationTest(unittest.TestCase):
     def test_compare_to_transformers(self, model_arch):
         model_id = MODEL_NAMES[model_arch]
         set_seed(SEED)
-        ov_model = OVModelForSequenceClassification.from_pretrained(
-            model_id, export=True, ov_config={"CACHE_DIR": "", "INFERENCE_PRECISION_HINT": "f32"}
-        )
+        ov_model = OVModelForSequenceClassification.from_pretrained(model_id, export=True, ov_config=F32_CONFIG)
         self.assertIsInstance(ov_model.config, PretrainedConfig)
         transformers_model = AutoModelForSequenceClassification.from_pretrained(model_id)
         tokenizer = AutoTokenizer.from_pretrained(model_id)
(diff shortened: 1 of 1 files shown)


tests/openvino/test_modeling.py
```diff
@@ -430,7 +440,9 @@ class OVModelForFeatureExtractionIntegrationTest(unittest.TestCase):
     def test_compare_to_transformers(self, model_arch):
         model_id = MODEL_NAMES[model_arch]
         set_seed(SEED)
-        ov_model = OVModelForFeatureExtraction.from_pretrained(model_id, export=True)
+        ov_model = OVModelForFeatureExtraction.from_pretrained(
+            model_id, export=True, ov_config={"CACHE_DIR": "", "INFERENCE_PRECISION_HINT": "f32"}
```
dtrawins, line 444: I suggest using OV_CONFIG const in utils and loading it via:
ov_model = OVModelForFeatureExtraction.from_pretrained(
            model_id, export=True, ov_config=OV_CONFIG)

helena-intel: > I suggest using OV_CONFIG const in utils and loading it via:
ov_model = OVModelForFeatureExtraction.from_pretrained(
model_id, export=True, ov_config=OV_CONFIG)

I like being explicit in tests, especially when we deviate from defaults. But I can add an F32_CONFIG at the top of this file.

> also setting NUN_STREAMS = 1 could be useful

Agreed that this can be useful, but we should run some tests and assuming no issues, set this by default in optimum-intel, not just for the tests.

tests/openvino/test_modeling.py
```diff
@@ -125,7 +128,10 @@ def test_load_from_hub_and_save_model(self):
         loaded_model = OVModelForSequenceClassification.from_pretrained(self.OV_MODEL_ID, ov_config=ov_config)
         self.assertTrue(manual_openvino_cache_dir.is_dir())
         self.assertGreaterEqual(len(list(manual_openvino_cache_dir.glob("*.blob"))), 1)
-        self.assertEqual(loaded_model.request.get_property("PERFORMANCE_HINT").name, "THROUGHPUT")
+        if get_version() < "2023.3":
```
echarlaix, line 131: minor comment : it could be replaced with [`is_openvino_version`](https://github.com/huggingface/optimum-intel/blob/c64025dae99596c82fd79cba42b674a1576455f4/optimum/intel/utils/import_utils.py#L192) for clarity
```suggestion
        if is_openvino_version("<", "2023.3"):
```

echarlaix (approved):
Thanks for the addition @helena-intel

tests/openvino/test_modeling.py
helena-intel: > I like being explicit in tests, especially when we deviate from defaults. But I can add an F32_CONFIG at the top of this file.

Done

echarlaix:
Thanks a lot @helena-intel

tests/openvino/test_modeling.py
dtrawins: the same will be needed also in stable diffusion tests

helena-intel: > the same will be needed also in stable diffusion tests

Thanks @dtrawins I did that in this PR https://github.com/huggingface/optimum-intel/pull/518

PR #512, opened 2024-01-12
add IPEX model and README

jiqing-feng:
Hi @echarlaix . I added IPEX model by following the INC model and OpenVino model. Basically, I removed the ipex optimization from `inference.py` to `modeling_base.py`. It is better to let users know how to use IPEX optimized model just like INC and OpenVino. Would you please help to review it? Thx!


echarlaix:
Included in https://github.com/huggingface/optimum-intel/pull/516 so closing

PR #516, opened 2024-01-16
Add IPEX models

echarlaix:
continuing work from https://github.com/huggingface/optimum-intel/pull/512

cc @ofirzaf


tests/ipex/test_modeling.py
```diff
@@ -0,0 +1,66 @@
...
+class Timer(object):
+    def __enter__(self):
+        self.elapsed = time.perf_counter()
+        return self
+
+    def __exit__(self, type, value, traceback):
+        self.elapsed = (time.perf_counter() - self.elapsed) * 1e3
+
+
+class INCModelingTest(unittest.TestCase):
```
jiqing-feng, line 34: May be we can rename it to `IPEXModelingTest` because we use `IPEXModelForCausalLM` to load the model.

jiqing-feng:
Hi @echarlaix I have tested the example in README, and it works well. I am OK with this PR, we could merge it and move to the next step.

PR #517, opened 2024-01-16
Fix loading Timm models with ov_config

helena-intel:
`ov_config` was not passed to `_to_load` in the case where `local_timm_model or (not os.path.isdir(model_id) and model_info(model_id).library_name == "timm")`. This PR fixes that.


echarlaix (approved):

PR #518, opened 2024-01-16
Use f32 inference for some OpenVINO stable diffusion/training tests

helena-intel:
Follow-up on https://github.com/huggingface/optimum-intel/pull/508 , this PR adds an f32 config for compare-with-transformers training and stable diffusion tests, to allow them to pass on systems where BF16 inference is the default.


echarlaix (approved):
LGTM!

PR #520, opened 2024-01-17
Enable automatic CACHE_DIR for GPU inference only

helena-intel:
Setting CACHE_DIR used to speed up model loading on both CPU and GPU, but recent model loading improvements in OpenVINO mean that on CPU it is now faster to load the model without setting CACHE_DIR. I tested this with several kinds of models, on Core and Xeon. Loading without cache can be about three times faster. On GPU loading from cache is still about three times faster (tested with ARC770). This PR enables automatic CACHE_DIR only when device is explicitly set to GPU.

For seq2seq models the ov_config was set on init, but because models can also be moved to GPU with `.to()` the setting of CACHE_DIR should now be done on compiling the model. The OVDecoder and OVEncoder had no concept of a model_save_dir. I added a parent_model to the constructor to get access to this.

### DIFF
--- a/optimum/intel/openvino/modeling_seq2seq.py
+++ b/optimum/intel/openvino/modeling_seq2seq.py
@@ -266,34 +266,11 @@ class OVModelForSeq2SeqLM(OVBaseModelForSeq2SeqLM, GenerationMixin):
         self.device = torch.device("cpu")
         self.decoder_with_past = None
         enable_compilation = kwargs.get("compile", True)
-        encoder_cache_dir = Path(self.model_save_dir).joinpath("encoder_cache")
-        ov_encoder_config = {**self.ov_config}
-
-        if "CACHE_DIR" not in ov_encoder_config.keys() and not str(self.model_save_dir).startswith(gettempdir()):
-            ov_encoder_config["CACHE_DIR"] = str(encoder_cache_dir)
-
-        self.encoder = OVEncoder(
-            self.encoder_model, self._device, ov_encoder_config, main_input_name=self.main_input_name
-        )
-
-        decoder_cache_dir = Path(self.model_save_dir).joinpath("decoder_cache")
-        ov_decoder_config = {**self.ov_config}
-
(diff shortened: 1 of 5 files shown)


tests/openvino/test_modeling.py
```diff
@@ -598,11 +593,15 @@ def test_compare_with_and_without_past_key_values(self):
         gc.collect()
 
     def test_auto_device_loading(self):
-        model_id = MODEL_NAMES["gpt2"]
         for device in ("AUTO", "AUTO:CPU"):
-            model = OVModelForCausalLM.from_pretrained(model_id, export=True, use_cache=True, device=device)
+            OV_MODEL_ID = "echarlaix/distilbert-base-uncased-finetuned-sst-2-english-openvino"
```
echarlaix, line 597: can be moved outside loop

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -435,13 +412,13 @@ class OVEncoder:
             The OpenVINO inference request associated to the encoder.
     """
 
-    def __init__(self, model: openvino.runtime.Model, device: str, ov_config: Dict, main_input_name="input_ids"):
+    def __init__(self, model: openvino.runtime.Model, device: str, parent_model: OVModelForSeq2SeqLM):
         self.model = model
         self._device = device
+        self.parent_model = parent_model
```
echarlaix, line 418: can be used to get the device also wdyt ?

echarlaix (approved):
LGTM, thanks @helena-intel !

optimum/intel/openvino/modeling_seq2seq.py
helena-intel: Yes, I was a bit hesitant to make more changes than needed, but should have done this :-) Thanks, I'll change it.

echarlaix: perfect, thanks @helena-intel !

PR #522, opened 2024-01-18
Return the original forward in PyTorch model after exporting to OpenVINO.

alexsu52:
# What does this PR do?
`export_pytorch` can patch forward function of the input Pytorch model for correct exporting to OpenVINO, but does not return the original forward. It leads to undefined behavior if user uses PyTorch model after export. For example to the following error:
```
'tuple' object has no attribute 'values'.
```

This PR corrects it.

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -428,6 +428,9 @@ def export_pytorch(
                 compression_option=compression_option,
                 compression_ratio=compression_ratio,
             )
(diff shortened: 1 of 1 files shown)


alexsu52:
cc' @AlexKoff88 @eaidova

eaidova (approved):

AlexKoff88 (approved):

echarlaix (approved):
LGTM, thanks @alexsu52

PR #524, opened 2024-01-18
Add warning message when using transformers < 4.35

helena-intel:
There is a large performance degradation when using transformers < 4.36 and optimum 1.16 to export stateful models. This PR adds a warning message when exporting stateful models with transformers < 4.36 or PyTorch<2.1.1. I made the warning bright red because without that it would definitely be lost between the other output, and this is really important to not miss. The ANSI color codes should work well on most systems (Windows and Linux terminal, Jupyter). In the case that it doesn't work, it's just a few numbers printed before and after the message. 

image


echarlaix (approved):
LGTM!

PR #528, opened 2024-01-19
Add test for _print_compiled_model_properties

helena-intel:
Add test that checks that setting OPENVINO_LOG_LEVEL=3 or calling `_print_compiled_model_properties()` directly does not raise an Exception.


echarlaix (approved):

echarlaix:
tests failing coming from https://github.com/huggingface/optimum-intel/blob/ff1e38233270c4c579ae94593c07a20ff88030d6/optimum/intel/openvino/modeling_base.py#L356

PR #529, opened 2024-01-22
[OpenVINO]: Updated documentation about weight compression

AlexKoff88:


AlexKoff88:
@ljaljushkin, please take a look as well.

docs/source/optimization_ov.mdx
```diff
@@ -62,6 +62,36 @@ tokenizer.save_pretrained(save_dir)
...
+
+For the 4-bit weight quantization we recommend using the NNCF API like below:
+```python
+from optimum.intel import OVModelForCausalLM
+import nncf
+
+model = OVModelForCausalLM.from_pretrained(model_id, load_in_8bit=False)
+model.model = nncf.compress_weights(
+        model.model,
+        mode=nncf.CompressWeightsMode.INT4_SYM,
```
echarlaix, line 85: why not add a `load_in_4bit` argument ?

docs/source/optimization_ov.mdx
```diff
@@ -62,6 +62,36 @@ tokenizer.save_pretrained(save_dir)
 
 The `quantize()` method applies post-training static quantization and export the resulting quantized model to the OpenVINO Intermediate Representation (IR). The resulting graph is represented with two files: an XML file describing the network topology and a binary file describing the weights. The resulting model can be run on any target Intel device.
 
+## LLM weight-only quantization
```
echarlaix, line 65: Not specific to LLM so would replace with :

```suggestion
##  Weight-only quantization
```

docs/source/inference.mdx
```diff
@@ -112,6 +112,23 @@ from optimum.intel import OVModelForCausalLM
...
 
+> **NOTE:** `load_in_8bit` is enabled by default for the models larger than 1 billion parameters.
+
+There are also alternative compression options for a different performance-accuracy trade-off:
+
+| Option                                 | Description                        |
+|--------------------------------------|--------------------------------------|
+| `fp16`                               | Float16 weights.   |
+| `int8`                               | INT8 weights.      |
+| `int4_sym_g128`                      | INT4 *symmetrically* quantized weights of most of the Linear layers with group size 128. The rest of the Linear layers and all Embeddings are quantized to INT8.   |
```
echarlaix, line 123: would keep this explanation outside the table :
```suggestion
| `int4_sym_g128`                      | *symmetric* INT4  with group size 128 |
```
and add the second sentence after the table (+ could be nice to extend a bit on "most of the Linear layers")

docs/source/inference.mdx
```diff
@@ -96,13 +96,13 @@ tokenizer.save_pretrained(save_directory)
...
-You can also apply INT8 quantization on your models weights when exporting your model with the CLI:
+You can also apply 8-bit or 4-bit weight quantization when exporting your model with the CLI:
 
 ```bash
-optimum-cli export openvino --model gpt2 --int8 ov_model
+optimum-cli export openvino --model gpt2 --weight-format int8 ov_model
 ```
 
-This will results in the exported model linear and embedding layers to be quantized to INT8, the activations will be kept in floating point precision.
+This will results in the exported model linear and embedding layers to be quantized to INT8, the activations will be kept in floating point precision.
```
echarlaix, line 105:
```suggestion
This will results in the exported model linear and embedding layers to be quantized to INT8 or INT4, the activations will be kept in floating point precision.
```

docs/source/optimization_ov.mdx
AlexKoff88: we will add a bit later when finalize INT4 workflow in NNCF

docs/source/inference.mdx
AlexKoff88: addressed

echarlaix: thanks!

docs/source/optimization_ov.mdx
echarlaix: might make sense to wait for it no what do you think ?

helena-intel: @echarlaix could we merge this now (after review) and then update for INT4 later with a new PR? The docs are incorrect now and that's causing some confusion, so it would be really good to have correct docs again.

docs/source/optimization_ov.mdx
```diff
@@ -62,6 +62,36 @@ tokenizer.save_pretrained(save_dir)
...
+
+```python
+from optimum.intel import OVModelForCausalLM
+
+model = OVModelForCausalLM.from_pretrained(model_id, load_in_8bit=True)
+```
+
+> **NOTE:** `load_in_8bit` is enabled by default for models larger than 1 billion parameters.
+
+For the 4-bit weight quantization we recommend using the NNCF API like below:
```
helena-intel, line 77: I would much recommend optimum-cli over NNCF API for this. It's such a quick and easy method. And (unless that has been fixed very recently) NNCF fails on SPR/EMR with a BF16 error and it's not easy to know how to work around that.

docs/source/inference.mdx
```diff
@@ -96,13 +96,13 @@ tokenizer.save_pretrained(save_directory)
 
 ### Weight only quantization
 
-You can also apply INT8 quantization on your models weights when exporting your model with the CLI:
+You can also apply 8-bit or 4-bit weight quantization when exporting your model with the CLI:
```
helena-intel, line 99: Could we mention that this is primarily useful for specific model types (CausalLM, ...) with a large number of weights?

docs/source/optimization_ov.mdx
AlexKoff88: I guess we fixed this issue in the recent version of NNCF. @alexsu52, please confirm.

docs/source/inference.mdx
AlexKoff88: Addressed

docs/source/optimization_ov.mdx
echarlaix: works for me

docs/source/optimization_ov.mdx
alexsu52: Yes, you are right. Models with float16 and float32 weigths work on SPR/EMR.

AlexKoff88:
I think we can merge this. @echarlaix

PR #530, opened 2024-01-24
Fix ov device

echarlaix:
cc @eaidova

### DIFF
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -129,8 +129,12 @@ class OVModel(OVBaseModel):
         Use the specified `device` for inference. For example: "cpu" or "gpu". `device` can
         be in upper or lower case. To speed up first inference, call `.compile()` after `.to()`.
         """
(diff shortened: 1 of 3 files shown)


optimum/intel/openvino/modeling.py
```diff
@@ -129,8 +129,12 @@ def to(self, device: str):
         Use the specified `device` for inference. For example: "cpu" or "gpu". `device` can
         be in upper or lower case. To speed up first inference, call `.compile()` after `.to()`.
         """
-        self._device = str(device).upper()
-        self.request = None
+        if isinstance(device, str):
+            self._device = device.upper()
+            self.request = None
+        else:
+            logger.warning(f"device must be of type {str} but got {type(device)} instead")
```
helena-intel, line 136: The issue is fixed (thank you!), I tested that with this PR these pipelines execute on GPU again, but this warning shows every  time:

image

AlexKoff88 (approved):

PR #532, opened 2024-01-26
fix OVCasualLM model inference without generate

eaidova:
# What does this PR do?

fix position_ids initialization for first inference stateful models


echarlaix (approved):
thanks for the fix @eaidova

PR #533, opened 2024-01-26
Expose InferRequestWrapper class so it can be imported from elsewhere

nikita-savelyevv:
# What does this PR do?

`InferRequestWrapper` class can be useful in other places. Now it's moved from inside a method and can be imported if needed.

## Before submitting


nikita-savelyevv:
@AlexKoff88, please review

AlexKoff88:
This class can be helpful for statistic collection in the scripts that use Optimum for inference. I am ok with this.

AlexKoff88 (approved):

PR #538, opened 2024-01-30
[WIP]: load_in_4bit option for OVModelForCausalLM

AlexKoff88:


optimum/intel/openvino/quantization.py
```diff
@@ -119,7 +121,8 @@ def quantize(
         self,
         calibration_dataset: Dataset = None,
         save_directory: Union[str, Path] = None,
-        quantization_config: OVConfig = None,
+        quantization_config: QuantizationConfigMixin = None,
+        ov_config: OVConfig = None,
```
echarlaix, line 125: Not sure we need both an `ov_config` and `quantization_config`, would prefer to keep only one if that's feasible

echarlaix (commented):
looks really great @AlexKoff88 !

optimum/intel/openvino/modeling_base.py
```diff
@@ -186,11 +186,13 @@ def _from_pretrained(
...
                 Whether or not to force the (re-)download of the model weights and configuration files, overriding the
                 cached versions if they exist.
-            file_name(`str`, *optional*):
+            file_name (`str`, *optional*):
                 The file name of the model to load. Overwrites the default file name and allows one to load the model
                 with a different name.
-            local_files_only(`bool`, *optional*, defaults to `False`):
+            local_files_only (`bool`, *optional*, defaults to `False`):
                 Whether or not to only look at local files (i.e., do not try to download the model).
+            load_in_8bit (`bool`, *optional*, defaults to `False`):
```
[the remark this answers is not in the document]
AlexKoff88: we have no idea how it works and impacts accuracy. So, we would like to enable it for decoders only for now.

optimum/intel/__init__.py
```diff
@@ -171,9 +174,15 @@
         if not (is_openvino_available() and is_nncf_available()):
             raise OptionalDependencyNotAvailable()
     except OptionalDependencyNotAvailable:
-        from .utils.dummy_openvino_and_nncf_objects import OVConfig, OVQuantizer, OVTrainer, OVTrainingArguments
+        from .utils.dummy_openvino_and_nncf_objects import (
```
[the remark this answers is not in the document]
AlexKoff88: Done

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -350,15 +360,14 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
             checkpoint="gpt2",
         )
     )
-    def forward(
+    def prepare_forward_inputs(
```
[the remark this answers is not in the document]
AlexKoff88: sure

optimum/intel/openvino/quantization.py
AlexKoff88: @echarlaix, how about deprecating `ov_config` here and migrating to different types of `QuantizationConfigMixin` so that there will be only `quantization_config` parameter?

optimum/intel/openvino/modeling_base.py
```diff
@@ -285,7 +287,7 @@ def _from_transformers(
 
         compression_option = None
         if load_in_8bit is not None:
-            compression_option = "int8" if load_in_8bit else "fp32"
+            compression_option = "fp32"
```
echarlaix, line 290: could you add a comment to specify that if `load_in_8bit` is not specified, then it'll be set as default (depending on the model size)

optimum/intel/openvino/modeling_base.py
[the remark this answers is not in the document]
echarlaix: I see, could we add a warning to specify that it'll be ignored when specified for other models then ?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -543,7 +574,7 @@ def _from_pretrained(
             local_files_only=local_files_only,
         )
 
-        model = cls.load_model(model_cache_path, load_in_8bit=load_in_8bit)
+        model = cls.load_model(model_cache_path, load_in_8bit=False if load_in_4bit else load_in_8bit)
```
echarlaix, line 583: should we raise an error to ensure `load_in_4bit` and `load_in_8bit` are not both set to True

optimum/intel/openvino/quantization.py
echarlaix: was actually thinking about the opposite as `ov_config` have a quantization section, need to think a bit about this

optimum/intel/openvino/modeling_base.py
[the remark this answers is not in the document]
AlexKoff88: sure

optimum/intel/openvino/modeling_decoder.py
AlexKoff88: sure

optimum/intel/openvino/modeling_base.py
AlexKoff88: There is a message about that [here](https://github.com/huggingface/optimum-intel/blob/374b1fc71301f8f60440416648114c571f7063f0/optimum/exporters/openvino/__main__.py#L308) which is the only one entry point for auto identification of 8-bit compression.

AlexKoff88:
CI is ready. The two failed quantization tests will be fixed by #535.

optimum/intel/openvino/configuration.py
```diff
@@ -83,6 +84,7 @@ def __init__(
         compression: Union[List[Dict], Dict, None] = None,
         input_info: Optional[List] = None,
         save_onnx_model: bool = False,
+        weight_quantization_config: Optional[QuantizationConfigMixin] = None,
```
AlexKoff88, line 87: @echarlaix, fyi, if I name it `quantization_config` I get an error during the serialization process due to the logic in transformers library that checks the availability of `quantization_config` field and fails if it is None.

optimum/intel/openvino/quantization.py
```diff
@@ -157,7 +158,7 @@ def quantize(
         self,
         calibration_dataset: Dataset = None,
         save_directory: Union[str, Path] = None,
-        quantization_config: OVConfig = None,
+        ov_config: OVConfig = None,
```
echarlaix, line 161: should we add a warning in `quantize` that `quantization_config` is deprecated like

```
quantization_config = kwargs.pop("quantization_config", None)
if quantization_config is not None:
    logger.warning("The argument `quantization_config` is deprecated, and will be removed in optimum-intel v1.6.0, please use `ov_config` instead"")
ov_config = ov_config or quantization_config
```

setup.py
```diff
@@ -45,7 +45,7 @@
         "transformers>=4.34.0",
     ],
     "openvino": ["openvino>=2023.2", "onnx", "onnxruntime", "transformers>=4.36.0", "optimum>=1.16.1"],
-    "nncf": ["nncf>=2.7.0"],
+    "nncf": ["nncf @ git+https://github.com/openvinotoolkit/nncf.git"],
```
echarlaix, line 48: for when is the next nncf release planned ?

optimum/intel/openvino/configuration.py
echarlaix: Could having only one argument make sense instead of having both `compression` and `weight_quantization_config` (we could create an instance of `WeightQuantizationConfig` if needed in the init directly). Also could you add a test to verify serialization is working ?

echarlaix (approved):
thanks for iterating, it looks great

optimum/intel/openvino/quantization.py
AlexKoff88: Done

setup.py
AlexKoff88: The release should be in 2-3 weeks. I would like to migrate to develop version for now to have time for integration.

optimum/intel/openvino/configuration.py
AlexKoff88: Fixed

optimum/intel/openvino/modeling_base.py
AlexKoff88: Done

echarlaix: thanks!

setup.py
echarlaix: OK would prefer to wait for nncf next release but 2/3 weeks sounds reasonable

BreidyAriasD (approved):

PR #541, opened 2024-01-30
Show device name in _print_compiled_model_properties

helena-intel:
- Show device name in _print_compiled_model_properties
- Enable CACHE_DIR also for devices like "GPU:0"

It is useful to be able to quickly see the exact device name of the inference device. For example, when doing inference on a machine with several GPUs, to doublecheck whether iGPU or dGPU is used. This PR adds a line that shows the full name of the execution device in _print_compiled_model_properties. Works with CPU, GPU and AUTO. I also changed the logic for checking if CACHE_DIR should be enabled to also enable CACHE_DIR by default when device is specified as e.g. "GPU:0". 

```
EXECUTION_DEVICES:                                                                                                                                                                                                                    
  CPU: 11th Gen Intel(R) Core(TM) i9-11900KF @ 3.50GHz   
....
EXECUTION_DEVICES:
  OCL_GPU.0: Intel(R) Arc(TM) A770 Graphics (dGPU)
```

### DIFF
--- a/optimum/intel/openvino/modeling_seq2seq.py
+++ b/optimum/intel/openvino/modeling_seq2seq.py
@@ -568,7 +568,7 @@ class OVDecoder:
         if (
             "CACHE_DIR" not in ov_config.keys()
             and not str(self.parent_model.model_save_dir).startswith(gettempdir())
-            and self._device.lower().startswith("gpu")
+            and "gpu" in self._device.lower()
         ):
             cache_dir = Path(self.parent_model.model_save_dir).joinpath("model_cache")
             ov_config["CACHE_DIR"] = str(cache_dir)
--- a/optimum/intel/openvino/modeling_seq2seq.py
+++ b/optimum/intel/openvino/modeling_seq2seq.py
@@ -456,7 +456,7 @@ class OVEncoder:
         if (
             "CACHE_DIR" not in ov_config.keys()
             and not str(self.parent_model.model_save_dir).startswith(gettempdir())
-            and self._device.lower() == "gpu"
+            and self._device.lower().split(":")[0] == "gpu"
         ):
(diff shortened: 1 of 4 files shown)


optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -456,7 +456,7 @@ def _compile(self):
         if (
             "CACHE_DIR" not in ov_config.keys()
             and not str(self.parent_model.model_save_dir).startswith(gettempdir())
-            and self._device.lower() == "gpu"
+            and self._device.lower().split(":")[0] == "gpu"
```
echarlaix, line 459: what about replacing it with for cases like device is set to "GPU.0" or "AUTO:GPU,CPU"
```suggestion
            and "gpu" in self._device.lower()
```

optimum/intel/openvino/utils.py
```diff
@@ -145,3 +145,9 @@ def _print_compiled_model_properties(compiled_model):
                 logger.info(f"  {k}: {value}")
         except Exception:
             logger.error(f"[error] Get property of '{k}' failed")
+    try:
+        logger.info("EXECUTION_DEVICES:")
+        for device in compiled_model.get_property("EXECUTION_DEVICES"):
+            logger.info(f"  {device}: {Core().get_property(device, 'FULL_DEVICE_NAME')}")
+    except Exception:
+        logger.error("[error] Get FULL_DEVICE_NAME failed")
```
echarlaix, line 153: in which circonstance would this fail ?

helena-intel: > in which circonstance would this fail ?

If there is a bug (or API change) in OpenVINO. I did the warning like this to be consistent with the other warning when getting a property fails. I'm also fine with setting this to pass, I just don't want any exceptions from this function to propagate.

PR #542, opened 2024-01-30
Automatic `torch.autocast` for IPEXModel

ofirzaf:
# What does this PR do?
Handle `torch.autocast` for `IPEXModel`s automatically. If a model was traced with autocasting enabled you must also enable autocasting when forwarding with the model, otherwise an error will be thrown.
@echarlaix 

## Before submitting


echarlaix (approved):
thnaks @ofirzaf

PR #543, opened 2024-01-31
Add an initial warmup step to `IPEXModel`s

ofirzaf:
# What does this PR do?
The first 2 forwards of an IPEXModel after trace/load includes background optimizations steps that make the output of these forwards unpredictable and non consistent with the model after the optimizations. To fix that, an initial warmup step was added to the `__init__` of IPEXModels

Depends on PR #542 

@echarlaix 

## Before submitting

### DIFF
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -67,7 +67,7 @@ class IPEXModel(OptimizedModel):
         OptimizedModel.__init__(self, model=model, config=config)
         # To do: add XPU support
         self._device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
-        self._dtype = self.config.torch_dtype
+        self._dtype = self.config.torch_dtype if self.config.torch_dtype is not None else torch.float32
         self.model.to(self._device)
         self.model_save_dir = model_save_dir
 
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -227,7 +227,7 @@ class IPEXModel(OptimizedModel):
     def _init_warmup(self):
         # warmup, the first 2 forwards of an IPEX model include some preprocessing steps and
         # the results of the compute are unpredictable
-        use_cache = getattr(self, "use_cache", getattr(self.config, "use_cache", False))
+        use_cache = "past_key_values" in self.input_names
         dummy_inputs = prepare_jit_inputs(self, self.export_feature, use_cache)
         for _ in range(2):
             self(**dummy_inputs)
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -67,6 +67,7 @@ class IPEXModel(OptimizedModel):
         OptimizedModel.__init__(self, model=model, config=config)
         # To do: add XPU support
         self._device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+        self._dtype = self.config.torch_dtype
         self.model.to(self._device)
         self.model_save_dir = model_save_dir
 
@@ -190,7 +191,7 @@ class IPEXModel(OptimizedModel):
         if "token_type_ids" in self.input_names:
             inputs["token_type_ids"] = token_type_ids
 
-        outputs = self.model(**inputs)
+        outputs = self._call_model(**inputs)
         return ModelOutput(**outputs) if isinstance(outputs, dict) else ModelOutput(logits=outputs[0])
 
     def eval(self):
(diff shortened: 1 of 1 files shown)


optimum/intel/ipex/modeling_base.py
```diff
@@ -62,11 +63,13 @@ def __init__(
         model,
         config: PretrainedConfig = None,
         model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
+        initial_warmup: bool = True,
```
echarlaix, line 66: would rename to 
```suggestion
        warmup: bool = True,
```

echarlaix (approved):
LGTM

optimum/intel/ipex/modeling_base.py
ofirzaf: Done

optimum/intel/ipex/modeling_base.py
```diff
@@ -209,6 +218,22 @@ def to(self, device: Union[torch.device, str]):
...
+            with torch.autocast(self.device.type, self.dtype):
+                out = self.model(*args, **kwargs)
+        except RuntimeError:
+            out = self.model(*args, **kwargs)
+        return out
+
+    def _init_warmup(self):
+        # warmup, the first 2 forwards of an IPEX model include some preprocessing steps and
+        # the results of the compute are unpredictable
+        use_cache = getattr(self, "use_cache", getattr(self.config, "use_cache", False))
```
echarlaix, line 232: I think we should use `self.use_cache` here directly or "past_key_values" in self.input_names

optimum/intel/ipex/modeling_base.py
```diff
@@ -259,16 +284,17 @@ def forward(
...
 
 
 class IPEXModelForQuestionAnswering(IPEXModel):
     auto_model_class = AutoModelForQuestionAnswering
     export_feature = "question-answering"
 
+    @wraps(IPEXModel.forward)
     def forward(self, *args, **kwargs):
-        outputs = self.model(*args, **kwargs)
+        outputs = super().forward(*args, **kwargs)
```
echarlaix, line 297: why is this needed ?

optimum/intel/ipex/modeling_base.py
ofirzaf: I think we can't use `self.use_cache` cause it doesn't exist in IPEXModels that are not causal. The other option can work. Will try that

optimum/intel/ipex/modeling_base.py
ofirzaf: The `prepare_jit_inputs` looks at the signature of the function and the wraps and super help avoid code copy

echarlaix: would prefer we avoid as it will fail in case `outputs` is not a dict https://github.com/huggingface/optimum-intel/blob/8ee487dc2ade5bd0023d1bbe0a0103d6af8821e0/optimum/intel/ipex/modeling_base.py#L193

echarlaix: also not sure to see the link with `prepare_jit_inputs`

optimum/intel/ipex/modeling_base.py
ofirzaf: On second thought, why not leave it as is? We are checking if the model has `use_cache` attribute and if not we take the value from the `self.config`. What do you think?

optimum/intel/ipex/modeling_base.py
ofirzaf: 1. In `_init_warmup` we call `prepare_jit_inputs` which examines the passed model's `forward` signature to see which dummy inputs exists in the signature. If we don't use wraps we get the signature of 
    ```
    self, *args, **kwargs
    ```
     instead of 
    ```
    self, input_ids: torch.Tensor, attention_mask: torch.Tensor, token_type_ids: torch.Tensor = None, **kwargs,
    ```
2. `outputs` will always be a dict because this is the output of `IPEXModel.forward`, no?

echarlaix: 1. OK I understand, was thinking that `prepare_jit_inputs` was only used for the torchscript export but I see that it's also used in `_init_warmup`, thanks for the clarification
2. here I'm talking about outputs https://github.com/huggingface/optimum-intel/blob/8ee487dc2ade5bd0023d1bbe0a0103d6af8821e0/optimum/intel/ipex/modeling_base.py#L192C9-L192C16

optimum/intel/ipex/modeling_base.py
echarlaix: here the model is already traced so we should check the model directly (no need to get the info from the config which could be giving incorrect information)

PR #547, opened 2024-02-01
Relax dependency on accelerate and datasets in OVQuantizer

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


optimum/intel/openvino/quantization.py
```diff
@@ -51,6 +51,12 @@
 )
 
 
+if is_datasets_available():
+    try:
+        from datasets import Dataset
+    except ImportError:
+        pass
```
echarlaix, line 58: in favor of :
```suggestion
    from datasets import Dataset
```

optimum/intel/openvino/quantization.py
```diff
@@ -526,7 +539,7 @@ def get_calibration_dataset(
 
     def _get_calibration_dataloader(
         self,
-        calibration_dataset: Dataset,
+        calibration_dataset: "Dataset",
```
echarlaix, line 542: I think we can also soon deprecate the `input_names` attribute as it looks unused (to remove the constraint that `calibration_dataset` should have `column_names` attribute, making it optional)

echarlaix (approved):
looks great, thanks @eaidova

optimum/intel/openvino/quantization.py
eaidova: I added try except intentionally here for avoid issue that @helena-intel described in internal conversation (when import datasets failed in sdxl pipeline due to python compiled without lzma package)

echarlaix: OK not sure to have all context here, it seems like we don't need to import `Dataset` at runtime as it's only used for type hints, then can we replace it with :

```suggestion
    if TYPE_CHECKING:
        from datasets import Dataset
```

also is this issue specific to `Dataset` or is it for every imports from `datasets` ?

eaidova: based on what I can see on screen shot issue should be reproducible for any import from datasets due to this import
https://github.com/huggingface/datasets/blob/main/src/datasets/__init__.py#L18 that finally by chain import this 
https://github.com/huggingface/datasets/blob/main/src/datasets/utils/extract.py#L3

and this package does not work if python installed without _lzma lib (it is rare case and I can not reproduce it locally, but looks like that what we have)

PR #548, opened 2024-02-01
Fix openvino/test_training.py

daniil-lyakhov:
# What does this PR do?

* Ignored scope is updated for `CUSTOMIZED_QUANTIZATION_CONFIG`
* Number of expected quantizers is decreased for the impacted cases.

### Reasons:

After nncf update (2.7.0 -> 2.8.0), NNCFGraph was fixed for the Bert model: now head masks path is visible for the quantization algorithm:
Screenshot 2024-02-01 121631
This led to head masks path quantization: 
Screenshot 2024-02-01 115618
I've updated Ignored scope with IS from the default config, and now amount of quantizers was decreased.

CC: @AlexKoff88

### DIFF
--- a/tests/openvino/test_training.py
+++ b/tests/openvino/test_training.py
@@ -276,7 +276,7 @@ class OVTrainerBaseTrainingTest(unittest.TestCase, ABC):
         shutil.rmtree(self.output_dir)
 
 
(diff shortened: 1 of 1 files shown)


AlexKoff88:
It looks good and fixes a subset of tests. Let's merge. The rest of them are fixed in #538 and #535.

PR #560, opened 2024-02-14
Add bits and sym parameters to the OV quantization config

echarlaix:


optimum/intel/openvino/configuration.py
```diff
@@ -127,3 +149,91 @@ def _enable_standard_onnx_export_option(self):
...
+        dataset: Optional[str] = None,
+        ratio: Optional[float] = None,
+        group_size: Optional[int] = None,
+        all_layers: Optional[bool] = None,
+        sensitivity_metric: Optional[str] = None,
+        ignored_scope: Optional[dict] = None,
+        **kwargs,
+    ):
+        self.bits = bits
+        self.sym = sym
```
echarlaix, line 208: `CompressWeightsMode` is not JSON serializable so replace mode with `bits` and `sym` argument

optimum/exporters/openvino/convert.py
```diff
@@ -509,6 +509,8 @@ def export_models(
     Returns:
         list of input_names and output_names from ONNX configuration
     """
+
+    # TODO : modify compression_option to quantization_config
```
AlexKoff88, line 513: This is what I planned as well after merging changes in the export API.

AlexKoff88:
@echarlaix, all the failures in CI are related to the fact that the GitHub version of NNCF enables weight quantization of Conv layers by default while the officially released version does not have it. Realistically, we will have a new NNCF version released by the end of Feb but we will create a release branch this week, so we can just install NNCF from the release branch and no changes in the test references are required.

optimum/intel/openvino/configuration.py
AlexKoff88: @alexsu52, @ljaljushkin, fyi. We need to adjust `CompressWeightsMode` accordingly.

AlexKoff88: @echarlaix, I think we can quickly fix it in NNCF. The idea i behind the mode is that it is more flexible and allows the introducing of various types of weight compression, e.g. non-uniform data types, lookups, etc.

AlexKoff88: It seems I got where the problem is coming from and made a [PR](https://github.com/openvinotoolkit/nncf/pull/2474) to NNCF.

optimum/intel/openvino/quantization.py
```diff
@@ -334,7 +335,7 @@ def _quantize_ovcausallm(
                 quantization_config = OVWeightQuantizationConfig(mode=nncf.CompressWeightsMode.INT8_SYM)
                 self.model.model = nncf.compress_weights(self.model.model)
             else:
-                compress_decoder_weights(self.model, quantization_config)
+                _int4_weight_only_quantization(self.model, quantization_config)
```
AlexKoff88, line 338: We also need to support INT8_SYM and INT8_ASYM the same way as you did it for INT4 precision. I mean something like `mode=CompressWeightsMode.INT8_SYM if config.sym else CompressWeightsMode.INT8_ASYM`

optimum/intel/openvino/configuration.py
AlexKoff88: Per offline discussion, we can go with the current implementation.

optimum/intel/openvino/quantization.py
echarlaix: yes totally makes sense !

optimum/intel/openvino/configuration.py
echarlaix: works for me, and thanks again for the quick fix

optimum/intel/openvino/quantization.py
echarlaix: added `_weight_only_quantization` to support both

optimum/exporters/openvino/convert.py
echarlaix: yes I agree, will add it in a following PR (to deprecate both `compression_option` and `load_in_8bit`)

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -594,7 +595,18 @@ def _from_pretrained(
...
+            if not is_nncf_available():
+                raise ImportError(
+                    "Quantization of the weights requires nncf, please install it with `pip install nncf`"
+                )
+            from .quantization import _weight_only_quantization
+
+            default_config = _check_default_4bit_configs(config)
+
+            if default_config:
+                logger.info(f"For the given model, we recommend the following `quantization_config` : {default_config}")
```
echarlaix, line 607: I don't think we should overwrite the `quantization_config` wdyt @AlexKoff88

AlexKoff88 (approved):

optimum/intel/openvino/modeling_decoder.py
AlexKoff88: Actually, this was an idea behind the combination of `load_in_4bit` + `quantizaiton_config`. If the later one is None we use the pre-defined default config, otherwise we override it an let the user to use a custom config.

PR #563, opened 2024-02-15
Deprecate _get_compression_options

echarlaix:
In this PR :

* we enable the `ov_config` argument for all OVModels (previously reserved to `OVModelForCausalLM`) in order to give the possibility to pass any `quantization_config`
* we deprecate the  `_get_compression_options`  method

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -54,15 +54,6 @@ from .utils import (
 )
 
 
-# TODO : remove as unused
-_COMPRESSION_OPTIONS = {
-    "int8": {"mode": nncf.CompressWeightsMode.INT8},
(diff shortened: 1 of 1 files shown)


AlexKoff88:
@echarlaix, what is the plan for CLI? Will we have all the granular options  exposed to the user, e.g. `--bits, --sym, --dataset, etc.`

AlexKoff88 (approved):

echarlaix:
> @echarlaix, what is the plan for CLI? Will we have all the granular options exposed to the user, e.g. `--bits, --sym, --dataset, etc.`

Yes I'm working on a PR to deprecate `compression_options` and `compression_ratio`, will then do a second PR to deprecate add `--bits, --sym` parameters

PR #564, opened 2024-02-16
Concurrency support using model clone 

dtrawins:
# Support for execution with multi concurrency
- Adding clone operation to model objects which creates new execution context without duplicating compiled_model and memory usage. It enables multi-concurrency in multithreaded applications.
- request attribute is now deprecated - new attributes compiled_model and infer_request are added instead - they match OpenVINO objects
- improved performance for decoder models by eliminating creating new requests for each inference

## Before submitting


dtrawins:
This PR is a replacement for https://github.com/huggingface/optimum-intel/pull/519

PR #565, opened 2024-02-16
Deprecate compression options

echarlaix:
In this PR we deprecate `compression_option` and `compression_ratio` during export, in favor of `ov_config` which will handle all quantization parameters (`group_size`, `ratio`, sym, ...) through `quantization_config` and `dtype` (for fp16 export)

Also add an `quantization_config` argument for all `OVModels` to enable 8bit and 4bit quantization
```python
model = OVModelForXxx.from_pretrained(model_id, quantization_config={"bits" : 4})
```

### DIFF
--- a/docs/source/inference.mdx
+++ b/docs/source/inference.mdx
@@ -116,16 +116,20 @@ model = OVModelForCausalLM.from_pretrained(model_id, load_in_8bit=True)
 
 There are also alternative compression options for a different performance-accuracy trade-off:
 
-| Option                                                              | Description       |
-|---------------------------------------------------------------------|-------------------|
-| `fp16`                                                              | Float16 weights   |
-| `int8`                                                              | INT8 weights      |
-| `int4_sym_g128`, `int4_asym_g128`, `int4_sym_g64`, `int4_asym_g64`* | INT4 weights      |
+| Option  | Description       |
+|---------|-------------------|
+| `fp16`  | Float16 weights   |
+| `int8`  | INT8 weights      |
+| `int4`  | INT4 weights      |
 
-*`sym` and `asym` stand for symmetric and asymmetric quantization, `g128` and `g64` means the group size `128` and `64` respectively. 
+The `--sym` parameter stands for symmetric quantization, asymmetric quantization will be applied by default if not provided.
 
-`--ratio` CLI parameter controls the ratio between 4-bit and 8-bit quantized layers and can also change performance-accuracy trade-off for the optimized model. It is valid only for INT4 quantization options.
+For INT4 quantization :
+* The `--group-size` parameter will define the group size to use for quantization, the rcommended value is 128 and -1 will results in per-column quantization.
+* The `--ratio` CLI parameter controls the ratio between 4-bit and 8-bit quantized layers and can also change performance-accuracy trade-off for the optimized model. It is valid only for INT4 quantization options.
 
+
+controls the ratio between 4-bit and 8-bit quantized layers and can also change performance-accuracy trade-off for the optimized model. It is valid only for INT4 quantization options.
 
 To apply quantization on both weights and activations, you can use the `OVQuantizer`, more information in the [documentation](https://huggingface.co/docs/optimum/main/en/intel/optimization_ov#optimization).
 
--- a/docs/source/inference.mdx
+++ b/docs/source/inference.mdx
@@ -47,6 +47,8 @@ Here we set the `task` to `text-generation-with-past`, with the `-with-past` suf
 optimum-cli export openvino --model local_path --task text-generation-with-past ov_model
 ```
 
+To export your model in fp16, you can add `--weight-format fp16` when exporting your model.
+
 Once the model is exported, you can load the OpenVINO model using :
 
 ```python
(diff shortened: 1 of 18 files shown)


docs/source/inference.mdx
```diff
@@ -96,15 +98,23 @@ tokenizer.save_pretrained(save_directory)
...
-This can also be done when loading your model by setting the `load_in_8bit` argument when calling the `from_pretrained()` method.
+By default the quantization scheme will be [assymmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#asymmetric-quantization), to make it [symmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#symmetric-quantization) you can add `--sym`.
+
+For INT4 quantization you can also specify the following arguments :
+* The `--group-size` parameter will define the group size to use for quantization, `-1` it will results in per-column quantization.
+* The `--ratio` CLI parameter controls the ratio between 4-bit and 8-bit quantization. If set to 0.9, it means that 90% of the layers will be quantized to `int4` while 10% will be quantized to `int8`.
+
+Smaller `group_size` and `ratio` of usually improve accuracy at the sacrifice of the model size and inference latency.
+
+You can also apply apply 8-bit quantization on your model's weight when loading your model by setting the `load_in_8bit=True` argument when calling the `from_pretrained()` method.
```
AlexKoff88, line 117:
```suggestion
You can also apply 8-bit quantization on your model's weight when loading your model by setting the `load_in_8bit=True` argument when calling the `from_pretrained()` method.
```

AlexKoff88:
Thanks, @echarlaix. It looks great!

Here are a couple of comments from my side:
- How are the default 4-bit configs supposed to work on the API level? Like `model=OVModelForCausalLM.from_pretrained(model_id, ov_config=dict(dtype="int4"))`?
- It is worth adding a test that checks this flow on the API level. We can do it with dummy models and patching `_DEFAULT_4BIT_CONFIGS` with `unittest.mock.patch()`.

echarlaix:
>    How are the default 4-bit configs supposed to work on the API level? Like `model=OVModelForCausalLM.from_pretrained(model_id, ov_config=dict(dtype="int4"))`?

Currently the user needs to specify the `quantization_config` :
```python
model = OVModelForSequenceClassification.from_pretrained("/tmp/test", quantization_config={"bits" : 4})
```
do you think `ov_config` should be added as an argument ?
 
>    It is worth adding a test that checks this flow on the API level. We can do it with dummy models and patching `_DEFAULT_4BIT_CONFIGS` with `unittest.mock.patch()`.

`_DEFAULT_4BIT_CONFIGS` is used in the CLI so are you thinking about adding a test in [test_exporters_cli.py](https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_exporters_cli.py) to see if the correct quantization config is given for a dummy model (after patching `_DEFAULT_4BIT_CONFIGS`) ? We would also need to add the saving of the `OVConfig` after export (was planning to add it in a following PR but can add it in this one if needed)

AlexKoff88:
> > How are the default 4-bit configs supposed to work on the API level? Like `model=OVModelForCausalLM.from_pretrained(model_id, ov_config=dict(dtype="int4"))`?
> 
> Currently the user needs to specify the `quantization_config` :
> 
> ```python
> model = OVModelForSequenceClassification.from_pretrained("/tmp/test", quantization_config={"bits" : 4})
> ```
> 
> do you think `ov_config` should be added as an argument ? (I think we should have either `quantization_config` or `ov_config`)
> 
> > It is worth adding a test that checks this flow on the API level. We can do it with dummy models and patching `_DEFAULT_4BIT_CONFIGS` with `unittest.mock.patch()`.
> 
> `_DEFAULT_4BIT_CONFIGS` is used in the CLI so are you thinking about adding a test in [test_exporters_cli.py](https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_exporters_cli.py) to see if the correct quantization config is given for a dummy model (after patching `_DEFAULT_4BIT_CONFIGS`) ? We would also need to add the saving of the `OVConfig` after export (was planning to add it in a following PR but can add it in this one if needed)

Now, I see that `_DEFAULT_4BIT_CONFIGS` are used only in CLI. On the API level, we still have a warning message about the recommended config. Can we align the CLI and API behaviors by any chance?

I am ok with adding the test in the follow-up PR.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -575,11 +576,18 @@ def _from_pretrained(
             local_files_only=local_files_only,
         )
 
+        # Give default quantization config if not provided and load_in_8bit=True
+        if load_in_8bit:
+            quantization_config = quantization_config or {"bits": 8}
+
         if isinstance(quantization_config, dict):
+            if quantization_config == {"bits": 4} and config.name_or_path in _DEFAULT_4BIT_CONFIGS:
```
AlexKoff88, line 584: thanks, @echarlaix. I'm fine with merging this if it works.

AlexKoff88 (approved):

optimum/intel/openvino/modeling_decoder.py
echarlaix: perfect, will add the config saving + test mentioned above in a following PR

PR #566, opened 2024-02-18
fix jit model

jiqing-feng:
Hi @echarlaix . This PR fix the jit model:
1. Enable model_dtype to generate the correct pkv data type in forward
2. Enable _dtype to generate the correct pkv data type in prepare jit inputs
3. Use our own prepare_inputs_for_generation to make sure the jit trace work.
4. Remove `autocast` in the class as users can use it outside, and we shouldn't use it if the model type is `bf16` or `fp16`

Would you please help me to review it? Thx!


optimum/intel/ipex/modeling_base.py
```diff
@@ -325,9 +318,11 @@ def __init__(
     ):
         # Perform the initial warmup at the end of __init__
         super().__init__(model, config, model_save_dir=model_save_dir, warmup=False)
+        GenerationMixin.__init__(self)
 
         self.normalized_config = NormalizedConfigManager.get_normalized_config_class(config.model_type)(config)
         self.model_dtype = kwargs.get("model_dtype", self.dtype)
+        self._dtype = self.model_dtype
```
echarlaix, line 325: I think we should deprecate the `model_dtype` attribute (redundant with `_dtype`), adding a warning that it will be removed after v1.18.0 

```
    @property
    def model_dtype(self):
        # add a warning 
        return self._dtype 
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -346,15 +341,44 @@ def __init__(
             )
         except AttributeError:
             self.model_cls = get_model_class(self.config, AutoModelForCausalLM._model_mapping)
-        self._reorder_cache = self.model_cls._reorder_cache.__get__(self)
-        self.prepare_inputs_for_generation = self.model_cls.prepare_inputs_for_generation.__get__(self)
```
echarlaix, line 350: I think it makes sense to keep it as `_reorder_cache` and `prepare_inputs_for_generation` can be different depending on the modeling

optimum/intel/ipex/modeling_base.py
```diff
@@ -346,15 +341,44 @@ def __init__(
...
         if hasattr(self.model_cls, "_convert_to_standard_cache"):
             self._convert_to_standard_cache = self.model_cls._convert_to_standard_cache
         if hasattr(self.model_cls, "_convert_to_bloom_cache"):
             self._convert_to_bloom_cache = self.model_cls._convert_to_bloom_cache
         if warmup:
             self._init_warmup()
 
+    def prepare_inputs_for_generation(self, input_ids, past_key_values=None, **kwargs):
```
echarlaix, line 351: I think we can remove

optimum/intel/ipex/modeling_base.py
```diff
@@ -215,14 +216,6 @@ def to(self, device: Union[torch.device, str]):
     def can_generate(self):
         return isinstance(self, GenerationMixin)
 
-    def _call_model(self, *args, **kwargs):
```
echarlaix, line 218: also might make sense to add a test to verify this doesn't break support for model traced with with autocasting enabled cc @ofirzaf do you know if there any tiny model on thus hub we can use for this (https://huggingface.co/Intel/q8_tiny_starcoder_py/ ?)

echarlaix (commented):
thaks for the fix!

optimum/intel/ipex/modeling_base.py
ofirzaf: You can use this one, yes

ofirzaf:
What this PR is trying to fix?

The autocasting was added here because autocasting with a traced model is not optional, either the model runs with autocasting or the model runs without, it is an attribute of the model and not something the user can decide upon.

Regarding `prepare_inputs_for_generation`, the current implementation derives the correct implementation from the original's model architecture. This enables a wider support for more model architectures and scalability to support new model architectures out of the box. The proposed implementation might not support a wide variety of models and disable support for assisted generation methods.

jiqing-feng:
Hi @echarlaix @ofirzaf . What I did is based on the latest transformers and try to enable llama jit model.
1. I rewrite `prepare_inputs_for_generation` because if this function in [llama](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L1236) will failed if we use jit because jit model do not have attr layer
2. Remove autocast is because pytorch do not recommend use autocast on a bf16 model, and it also avoid CI failed at [here](https://github.com/huggingface/optimum-intel/blob/main/tests/ipex/test_modeling.py#L338-L354). The `transformers_outputs.logits` is fp32 and the `outputs.logits` is bf16 because of the autocast.

Here is my script and traceback:
```python
import torch
from transformers import AutoTokenizer, pipeline
from optimum.intel import IPEXModelForCausalLM

model_id = "meta-llama/Llama-2-7b-chat-hf"
model = IPEXModelForCausalLM.from_pretrained(model_id, export=True, torch_dtype=torch.bfloat16)
generation_kwargs = {"do_sample": False, "max_new_tokens": 32, "num_beams": 4}
text_generator = pipeline("text-generation", model=model, tokenizer=tokenizer)

print(text_generator("I am happy today because ", **generation_kwargs))
```

traceback
```
  File "/home/jiqingfe/test.py", line 13, in <module>
    print(text_generator("I am happy today because ", **generation_kwargs))
  File "/home/jiqingfe/transformers/src/transformers/pipelines/text_generation.py", line 241, in __call__
    return super().__call__(text_inputs, **kwargs)
  File "/home/jiqingfe/transformers/src/transformers/pipelines/base.py", line 1196, in __call__
    return self.run_single(inputs, preprocess_params, forward_params, postprocess_params)
  File "/home/jiqingfe/transformers/src/transformers/pipelines/base.py", line 1203, in run_single
    model_outputs = self.forward(model_inputs, **forward_params)
  File "/home/jiqingfe/transformers/src/transformers/pipelines/base.py", line 1102, in forward
    model_outputs = self._forward(model_inputs, **forward_params)
  File "/home/jiqingfe/transformers/src/transformers/pipelines/text_generation.py", line 328, in _forward
    generated_sequence = self.model.generate(input_ids=input_ids, attention_mask=attention_mask, **generate_kwargs)
  File "/home/jiqingfe/miniconda3/envs/ipex/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/jiqingfe/transformers/src/transformers/generation/utils.py", line 1603, in generate
    return self.beam_search(
  File "/home/jiqingfe/transformers/src/transformers/generation/utils.py", line 2967, in beam_search
    model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
  File "/home/jiqingfe/transformers/src/transformers/models/llama/modeling_llama.py", line 1237, in prepare_inputs_for_generation
    if past_key_value := getattr(self.model.layers[0].self_attn, "past_key_value", None):
  File "/home/jiqingfe/miniconda3/envs/ipex/lib/python3.10/site-packages/torch/jit/_script.py", line 823, in __getattr__
    return super().__getattr__(attr)
  File "/home/jiqingfe/miniconda3/envs/ipex/lib/python3.10/site-packages/torch/jit/_script.py", line 530, in __getattr__
    return super().__getattr__(attr)
  File "/home/jiqingfe/miniconda3/envs/ipex/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1688, in __getattr__
    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
AttributeError: 'RecursiveScriptModule' object has no attribute 'layers'
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -346,15 +341,44 @@ def __init__(
...
         if hasattr(self.model_cls, "_convert_to_bloom_cache"):
             self._convert_to_bloom_cache = self.model_cls._convert_to_bloom_cache
         if warmup:
             self._init_warmup()
 
+    def prepare_inputs_for_generation(self, input_ids, past_key_values=None, **kwargs):
+        past_key_values = past_key_values or kwargs.get("past", None)
+
+        if self.use_cache and past_key_values is not None:
+            input_ids = input_ids[:, -1:]
```
ofirzaf, line 355: This line disables the possibility to use the model with assisted generation, since in AG you need to be able to process more than 1 token + `past_key_values` per step.

ofirzaf:
@jiqing-feng I think the problem might be in the tracing process. the `layers` attribute is part of the new KVCache abstraction in HF/transformers. I think you need to make sure that you don't hit that if in the tracing and then it will be ignored and then you shouldn't even execute it when you infer with the traced model (need to check that). Anyway, if you use the proposed implementation for `prepare_inputs_for_generation`, you limit the usability of `IPEXModel`s to a narrow familiy of models

jiqing-feng:
Hi @echarlaix @ofirzaf Thanks for your review. 
1. For assisted decoding, I have fixed it with minor changes, it now supports assisted decoding.
2. For `prepare_inputs_for_generation` and `reorder_cahche` and so on... . We didn't expect the `IPEXModel` to support all models, only some model types in our support list should be enough. There is no way to avoid using `prepare_inputs_for_generation` and `reorder_cahche` when the model is traced to graph mode, so I can only copy these functions from [BaseModelForCausalLM](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/generation/modeling.py#L140-L253) to make sure the generation works.

Hope you understand, thx!

jiqing-feng:
Hi @echarlaix @ofirzaf . Thanks for your fix by adding `_prepare_inputs_for_generation_for_llama`, it is great!

I have some minor changes to support low-precision(bf16) model, would you please help me to review it? Thx!

optimum/intel/ipex/modeling_base.py
```diff
@@ -414,7 +413,7 @@ def forward(
             position_ids = attention_mask.long().cumsum(-1) - 1
             position_ids.masked_fill_(attention_mask == 0, 1)
             if past_key_values:
-                position_ids = position_ids[:, -1].unsqueeze(-1)
+                position_ids = position_ids[:, -input_ids.shape[-1] :]
```
echarlaix, line 416: why is this modification needed ?

optimum/intel/ipex/modeling_base.py
echarlaix: we actually should have a test for this to verify no modifications will break support, would you mind adding it @jiqing-feng in https://github.com/huggingface/optimum-intel/blob/v1.15.2/tests/ipex/test_modeling.py#L201 ? Also given @ofirzaf above explanations removing `_call_model` will likely broke support for [`q8_tiny_starcoder_py`](https://huggingface.co/Intel/q8_tiny_starcoder_py/)

echarlaix (commented):
thanks for iterating on it @jiqing-feng !

optimum/intel/ipex/modeling_base.py
jiqing-feng: `position_ids` should always have the same size as `input_ids`, we cannot assume the length is `1` while pkv exists (for example, assisted decoding).

optimum/intel/ipex/modeling_base.py
jiqing-feng: have fixed it

jiqing-feng:
Hi @echarlaix refer to this [comment](https://github.com/huggingface/optimum-intel/pull/566#discussion_r1504095476), do you mind making it clear what kind of tests should I add? Thx! 
Besides, I am not saying that we cannot use `autocast`, I propose that the `autocast` should be outside the model like:
```python
with torch.autocast(device.type, type):
    model = IPEXModelForCausalLM.from_pretrained(model_id, export=True, torch_dtype=torch.bfloat16)
    model(**inputs)
```

jiqing-feng:
Hi @echarlaix , I have reverted "_call_model", would you please help to review and merge it? Thx!

jiqing-feng:
Hi @echarlaix . Do you mind having a look at this PR? I think it is ready to merge.

jiqing-feng:
Hi @ofirzaf , Do you mind confirming whether it could be merged? Thx!

ofirzaf:
LGTM

optimum/intel/ipex/modeling_base.py
ofirzaf: Shouldn't this be taken care by `prepare_inputs_for_generation`?

jiqing-feng:
Hi @echarlaix . I have added the testing for assisted decoding which will input both `past_key_values` and `input_ids` with `length>1`.

optimum/intel/ipex/modeling_base.py
jiqing-feng: It's also an option, WDYT @echarlaix ?

echarlaix: it's already integrated in `prepare_inputs_for_generation` for each modeling 
https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L1237, could make sense to remove it all actually

jiqing-feng:
Hi @echarlaix  @ofirzaf . You were right, we don't need to prepare postion_ids in the forward because `prepare_inputs_for_generation` already did it. Would you please review it again? It should be ready to merge : ) Thanks!

echarlaix (approved):

PR #568, opened 2024-02-19
add openvino export configs

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


helena-intel (approved):
Thanks @eaidova this is awesome! I tested optimum-cli with Mixtral and it works well.

optimum/exporters/openvino/model_configs.py
```diff
@@ -0,0 +1,250 @@
...
+        if cond1 and cond2 and cond3 and cond4:
+            # Obtain the past sequence length from the value instead of the key (Bloom).
+            past_length = dummy_inputs["past_key_values"][0][1].shape[0]
+            for k, v in dummy_inputs.items():
+                if k not in ["attention_mask", "past_key_values"]:
+                    dummy_inputs[k] = v[:, -1:]
+
+            dummy_inputs["attention_mask"] = DummyInputGenerator.pad_input_on_dim(
+                dummy_inputs["attention_mask"],
+                desired_length=past_length + 1,
```
echarlaix, line 189: shouldn't it be instead (decoder input_ids sequence length can be different from 1)
```suggestion
                desired_length=past_length + dummy_inputs["input_ids"].shape[1],
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -0,0 +1,250 @@
...
+class ChatGLM2DummyTextInputGenerator(DummyTextInputGenerator):
+    SUPPORTED_INPUT_NAMES = {
+        "input_ids",
+        "attention_mask",
+        "token_type_ids",
+        "position_ids",
+    }
+
+    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+        import torch
```
echarlaix, line 88: we could raise an error if `framework!="pt"`

optimum/exporters/openvino/model_configs.py
```diff
@@ -0,0 +1,250 @@
...
+
+        input = super().generate(input_name, framework, int_dtype, float_dtype)
+        if input_name == "attention_mask":
+            input = torch.ones(input.shape, dtype=input.dtype)
+        if input_name == "position_ids":
+            bs = input.shape[0]
+            input = torch.range(0, input.shape[1], dtype=input.dtype).repeat(bs, 1)
```
echarlaix, line 95: why is  this needed ? couldn't we keep the inputs generated with `self.random_mask_tensor` and `self.random_int_tensor` ?

tests/openvino/test_modeling.py
```diff
@@ -486,6 +486,7 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "llama_gptq",
         "marian",
         "mistral",
+        "mixtral",
```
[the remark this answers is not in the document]
echarlaix: yes agreed, would be nice to have a test for all the new architectures supported in this PR

echarlaix (commented):
loooks great, thanks a lot @eaidova :fire:

optimum/exporters/openvino/model_configs.py
```diff
@@ -0,0 +1,250 @@
...
+            for _ in range(self.num_layers)
+        ]
+
+
+@register_in_tasks_manager("chatglm", *["text-generation", "text-generation-with-past"])
+class ChatGLM2OpenVINOConfig(TextDecoderOnnxConfig):
+    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig.with_args(vocab_size="padded_vocab_size", num_layers="num_layers")
+    DUMMY_INPUT_GENERATOR_CLASSES = (ChatGLM2DummyTextInputGenerator, ChatGLM2DummyPastKeyValuesGenerator)
+    DUMMY_PKV_GENERATOR_CLASS = ChatGLM2DummyPastKeyValuesGenerator
+    no_position_ids = False
```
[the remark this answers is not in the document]
eaidova: chatglm tokenizer always returns position_ids input, so normally, it is always provided on pytorch level for chatglm.

I will remove this initialization, I prepared this config some time ago, where we have no official support position_ids yet, I belive it is not needed anymore as it is provided by optimum

optimum/exporters/openvino/model_configs.py
eaidova: unfortunatly, model has check that attention mask contains zeros and in this case failed and makes additional actions expecting that attention mask will be passed directly in 4d format

optimum/exporters/openvino/convert.py
```diff
@@ -48,11 +51,7 @@
 )
 
 
-if is_optimum_version(">=", "1.16.99"):
-    from optimum.exporters.onnx.utils import _get_submodels_and_onnx_configs
-
-else:
-    from optimum.exporters.onnx.__main__ import _get_submodels_and_onnx_configs
+init_model_configs()
```
echarlaix, line 54: shouldn't it be called only once in `model_configs` so that we can overwritte ONNX config in case we want to ?

optimum/exporters/openvino/convert.py
```diff
@@ -32,10 +32,13 @@
...
 from optimum.exporters.onnx.convert import export_pytorch as export_pytorch_to_onnx
 from optimum.exporters.onnx.convert import export_tensorflow as export_tensorflow_onnx
+from optimum.exporters.utils import _get_submodels_and_export_configs
 from optimum.utils import DEFAULT_DUMMY_SHAPES, is_diffusers_available
 from optimum.utils.save_utils import maybe_save_preprocessors
 
-from ...intel.utils.import_utils import is_nncf_available, is_optimum_version
+from ...intel.utils.import_utils import is_nncf_available
+from .base import init_model_configs
+from .model_configs import *  # noqa: F403
```
echarlaix, line 41:
```suggestion

```

optimum/exporters/openvino/model_configs.py
```diff
@@ -0,0 +1,250 @@
...
+            inputs_or_outputs[f"{name}.{i}.key"] = {1: "batch_size", 0: decoder_sequence_name}
+            inputs_or_outputs[f"{name}.{i}.value"] = {1: "batch_size", 0: decoder_sequence_name}
+
+    def patch_model_for_export(
+        self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ) -> "ModelPatcher":
+        return ChatGLMModelPatcher(self, model, model_kwargs=model_kwargs)
+
+
+@register_in_tasks_manager("mixtral", *["text-generation", "text-generation-with-past"])
```
echarlaix, line 234: could also make sense to provide `library_name`
```suggestion
@register_in_tasks_manager("mixtral", *["text-generation", "text-generation-with-past"], library_name="transformers")
```

optimum/exporters/openvino/convert.py
eaidova: we need this for registering configs, the same made in neuron 
https://github.com/huggingface/optimum-neuron/blob/881d399372f750652fd53887b763106c1561b176/optimum/neuron/modeling_base.py#L31

optimum/exporters/openvino/convert.py
eaidova: if you can point me place, where it can be called to be triggered once, I can move

echarlaix: wouldn't there work https://github.com/eaidova/optimum-intel/blob/ea/ov_export_configs/optimum/exporters/openvino/model_configs.py#L43

optimum/exporters/openvino/model_configs.py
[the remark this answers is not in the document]
echarlaix: We add it by default when needed to have correct output for `batch_size` > 1

hat all sequences are of the same length if position_ids is not passed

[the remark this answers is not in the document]
eaidova: it is not related to issue with these models,chatglm tokenizer returns position_ids explicitly and it is always in use on pytorch level, there is no logic with generation position_ids inside model as I know, this model should be always exported with position_ids to match with expected behaviour with torch

optimum/exporters/openvino/model_configs.py
eaidova: updated

optimum/exporters/openvino/model_configs.py
eaidova: non-applicable anymore

optimum/exporters/openvino/model_configs.py
eaidova: updated according updated optimum code

optimum/exporters/openvino/convert.py
eaidova: I'll try to experiment with that

eaidova: reworked

optimum/exporters/openvino/convert.py
eaidova: @echarlaix without import model_configs module decorators can not to be executed. I reworked do not use asterix for import, but is it fine if i stay 

`import optimu.exportes.openvino.model_confis` inside `optimum/exportes/openvino/__init__.py`, that should be enough for trigger decorator

optimum/exporters/openvino/model_patcher.py
```diff
@@ -52,3 +69,166 @@ def patch_model_with_bettertransformer(model):
         return model
 
     return model
+
+
+def mixtral_sparse_moe_block_forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
```
echarlaix, line 74:
```suggestion
def _mixtral_sparse_moe_block_forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -0,0 +1,266 @@
...
+        "input_ids",
+        "attention_mask",
+        "token_type_ids",
+        "position_ids",
+    }
+
+    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+        input = super().generate(input_name, framework, int_dtype, float_dtype)
+        if input_name == "attention_mask":
+            input = self.random_int_tensor(input.shape, max_value=1, min_value=1)
```
echarlaix, line 122: could you add a comment or link for the model attention mask zeros value check ? (so that we know why this was added in the future)

optimum/exporters/openvino/model_configs.py
```diff
@@ -0,0 +1,266 @@
...
+    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
+
+
+class ChatGLM2DummyTextInputGenerator(DummyTextInputGenerator):
+    SUPPORTED_INPUT_NAMES = {
+        "input_ids",
+        "attention_mask",
+        "token_type_ids",
+        "position_ids",
+    }
```
echarlaix, line 117: is it needed ?

```suggestion
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -0,0 +1,266 @@
...
+@register_in_tasks_manager("jais", *["text-generation", "text-generation-with-past"])
+class JaisOpenVINOConfig(TextDecoderOnnxConfig):
+    DEFAULT_ONNX_OPSET = 13
+    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig.with_args(
+        num_layers="n_layer", num_attention_heads="n_head", hidden_size="n_embd"
+    )
+
+
+@register_in_tasks_manager("qwen2", *["text-generation", "text-generation-with-past"])
+class Qwen2OpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
```
echarlaix, line 85: for all configs inheriting from `TextDecoderWithPositionIdsOnnxConfig`, is it optional or is the exported model's outputs  incorrect for `batch_size>1` ? If optional I would prefer to remove them for now (to keep consistency as as that's we do for all models where it's optional)

echarlaix (commented):
Looks great @eaidova

tests/openvino/test_modeling.py
```diff
@@ -556,8 +571,10 @@ def test_pipeline(self, model_arch):
     def test_multiple_inputs(self, model_arch):
         model_id = MODEL_NAMES[model_arch]
         set_seed(SEED)
-        model = OVModelForCausalLM.from_pretrained(model_id, export=True, compile=False)
-        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        if model_arch in self.REMOTE_CODE_MODELS:
```
eaidova, line 574:
```suggestion
        model_kwargs = {}
        if model_arch in self.REMOTE_CODE_MODELS:
```

optimum/exporters/openvino/model_configs.py
eaidova: yes, all of them inherit this issue from llama

tests/openvino/test_modeling.py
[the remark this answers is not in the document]
eaidova: all supported models have tests now

PR #569, opened 2024-02-19
[OV]: Fixed inference after 4 bit weight compression

AlexKoff88:
Fixed inferencing the baseline model instead of the quantized version (delayed compilation).

### DIFF
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -612,7 +612,6 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
                 )
 
             _weight_only_quantization(causal_model, quantization_config)
-            causal_model.request = None
         return causal_model
 
 
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -592,8 +592,9 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
         else:
             init_cls = cls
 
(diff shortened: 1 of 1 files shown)


eaidova (approved):

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -592,7 +592,9 @@ def _from_pretrained(
         else:
             init_cls = cls
 
-        causal_model = init_cls(model=model, config=config, model_save_dir=model_cache_path.parent, **kwargs)
+        causal_model = init_cls(
+            model=model, config=config, model_save_dir=model_cache_path.parent, compile=not load_in_4bit, **kwargs
+        )
```
echarlaix, line 597:
```suggestion
        enable_compilation = kwargs.pop("compile", True) and not load_in_4bit
        causal_model = init_cls(
            model=model, config=config, model_save_dir=model_cache_path.parent, compile=enable_compilation, **kwargs
        )
```

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -609,6 +611,7 @@ def _from_pretrained(
                 )
 
             _weight_only_quantization(causal_model, quantization_config)
+            causal_model.request = None
```
echarlaix, line 614: why is this needed ?

echarlaix (approved):
thanks a lot for the fix!

echarlaix:
thanks for iterating @AlexKoff88, a last modification is needed here to support cases where self.request is None https://github.com/huggingface/optimum-intel/blob/8f7d016fbeb0cd185e80b9852207c2b530d3d394/optimum/intel/openvino/modeling_decoder.py#L422

PR #567, opened 2024-02-19
Model patcher

jiqing-feng:
This PR enables ipex llama model by patching functions and classes, and it has 30% speed-up than the original optimization. 
The ipex optimization ops will be released soon, and I will add the CI tests once it released. We can focus on the integration for now.

BTW, this PR includes #566, and I will rebase it after #566 is merged.

PR #573, opened 2024-02-27
concurrency without model cloning

dtrawins:
# Support for multi threading in execution?

### DIFF
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -17,6 +17,7 @@ import os
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Dict, Optional, Tuple, Union
+from datetime import datetime
 
 import numpy as np
 import openvino
@@ -210,6 +211,7 @@ class OVBaseDecoderModel(OVModel):
                 if self.is_dynamic:
                     self.model = self._reshape(self.model, -1, -1)
                 self.request = None
+                self.compiled_model = None
 
     def _save_pretrained(self, save_directory: Union[str, Path]):
         """
@@ -335,6 +337,7 @@ class OVBaseDecoderModel(OVModel):
     def compile(self):
         if self.request is None:
             super().compile()
+            self.compiled_model =self.request
             self.request = self.request.create_infer_request()
 
     def _make_stateful(self):
@@ -352,6 +355,18 @@ class OVBaseDecoderModel(OVModel):
 class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
     export_feature = "text-generation"
     auto_model_class = AutoModelForCausalLM
+
+    def generate(self, *args, **kwargs):
+        self.compile()
+        infer_context = [self.compiled_model.create_infer_request()]
+        kwargs["infer_context"] = infer_context
+        return super().generate(*args, **kwargs)
+
+    def __call__(self, *args, **kwargs):
+        self.compile()
+        infer_context = [self.compiled_model.create_infer_request()]
+        kwargs["infer_context"] = infer_context
+        return super().__call__(*args, **kwargs)
 
     @add_start_docstrings_to_model_forward(
         INPUTS_DOCSTRING.format("batch_size, sequence_length")
@@ -375,10 +390,13 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
         batch_size = input_ids.shape[0]
         if self.config.model_type == "bloom":
             batch_size *= self.config.num_attention_heads
-
+        #print("prepare inputs - input_ids:",input_ids)
         inputs = {}
         past_len = 0
+        #print("model stateful", self.stateful)
+        #print("use cache", self.use_cache)
         if not self.stateful:
+            #print("prepare inputs - past_key_values:",past_key_values)
             if past_key_values is not None:
                 if self.config.model_type not in MULTI_QUERY_ATTN_MODELS:
                     past_len = past_key_values[0][1].shape[-2]
@@ -417,13 +435,16 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
                     inputs[input_name] = Tensor(model_inputs.get_element_type(), shape.get_shape())
         else:
             # past_key_values are not used explicitly, instead they are handled inside the model
-            if past_key_values is None:
+            #print("past_values", past_key_values)
+            #if past_key_values is None:
                 # This is the first iteration in a sequence, reset all states
-                if self.request is not None:
-                    self.request.reset_state()
+                #if infer_request is not None:
+                #    infer_request.reset_state()
+                #    print("reseting state")
                 # Set initial value for the next beam_idx input that will be used at the current iteration
                 # and will be optionally updated by _reorder_cache at the next iterations if beam_search is used
-                self.next_beam_idx = np.arange(batch_size, dtype=int)
+                #past_key_values = [np.arange(batch_size, dtype=int)]
+            ...
 
         inputs["input_ids"] = np.array(input_ids)
         # Add the attention_mask inputs when needed
@@ -451,8 +472,10 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
 
         if "beam_idx" in self.input_names:
             inputs["beam_idx"] = (
-                self.next_beam_idx if self.next_beam_idx is not None else np.arange(batch_size, dtype=int)
+                past_key_values[0] if past_key_values is not None else np.arange(batch_size, dtype=int)
             )
+            #if past_key_values is not None:
+            #    print("type",type(past_key_values[0]))
 
         return inputs
 
@@ -462,10 +485,10 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
         attention_mask: Optional[torch.LongTensor] = None,
         past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None,
         position_ids: Optional[torch.LongTensor] = None,
+        infer_context: Optional[list[openvino.runtime.InferRequest]] = None,
         **kwargs,
     ) -> CausalLMOutputWithPast:
         self.compile()
-
         inputs = self.prepare_inputs(
             input_ids=input_ids,
             attention_mask=attention_mask,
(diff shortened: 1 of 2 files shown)


dtrawins:
@slyalin @eaidova  can you add your comments?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -354,6 +356,18 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
     export_feature = "text-generation"
     auto_model_class = AutoModelForCausalLM
 
+    def generate(self, *args, **kwargs):
+        self.compile()
+        infer_context = [self.compiled_model.create_infer_request()]
+        kwargs["infer_context"] = infer_context
+        return super().generate(*args, **kwargs)
+
+    def __call__(self, *args, **kwargs):
```
slyalin, line 365: @dtrawins, please explain why `__call__` should be different from `forward` behavior. `__call__` will eventually call `forward` without any added semantics on top of that. So why cannot we move this code to forward?

dtrawins: @slyalin that is indeed not required. potentially it could be used to pass the infer_request context. Assuming we create new request in the forward method if generate method didn't pass the context, that is not required.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -463,32 +461,41 @@ def forward(
...
+                infer_request = self.compiled_model.create_infer_request()
+        infer_request.start_async(inputs, share_inputs=True)
+        infer_request.wait()
+        logits = torch.from_numpy(infer_request.get_tensor("logits").data).to(self.device)
         if self.stateful:
             # Need a marker to differentiate the first generate iteration from the others in
             # the first condition at the function beginning above.
             # It should be something that is not None and it should be True when converted to Boolean.
-            past_key_values = ((),)
+            past_key_values = ((inputs["beam_idx"]), infer_request)
```
echarlaix, line 493: not related to `past_key_values` so I don't think we should update `past_key_values` here, the resulting output will not be what it's expected for example :
```
output = model(**tokens)
pkv = output.past_key_values
```

echarlaix (commented):
THanks a lot for your work @dtrawins

optimum/intel/openvino/modeling_base.py
```diff
@@ -86,6 +86,7 @@ def __init__(
 
         self.model = model
         self.request = None
+        self.compiled_model = None
```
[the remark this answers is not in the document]
dtrawins: It is needed to create new infer_request in the context of generate method for each concurrent thread. So far we had in the model class request attribute which was pointing to a static infer_request and can not be used to allocate new request. Generally there is a bit confusing setup when the request attribute is set to the compiled_model object in the based class but latest it is overwritten to become the infer_request. Eventually the recommendation would be to switch to using compiled_model attribute instead and create infer_requests dynamically. It was proposed to make this switch in a separate PR.

optimum/intel/openvino/modeling_decoder.py
echarlaix: My point is that it's not related to `past_key_values` so we shouldn't update it with beam_idx / inference execution context

dtrawins: @slyalin can you add your comments here? The idea was to reused this variable for stateful models because they don't use it at all. That was the only method we found that could be used to pass the beam_idx and execution context (which includes the state data) without changing the model API. The other alternative was with using model.clone() method for each thread which would also using a separate execution context without duplicating memory consumption https://github.com/huggingface/optimum-intel/pull/564. Would cloning be better method to support concurrency in the execution? Is there some other option we are not aware of? I guess it is a bit unique situation with the stateful models in openvino so probably it is not handled in transformers lib.

slyalin: > My point is that it's not related to past_key_values

Definitely it _is_ related to `past_key_values` even more than the old `((),)` value. `beam_idx` together with `infer_request` are used to track `past_key_values` for a particular sequence. Literally, `infer_request` has a model state that consists of `past_key_values` tensors, and `beam_idx` allows indirect rows reordering in that state in case of beam search. This PR just makes it more explicit than it was before and moving this attributes from the model class instance to each sequence, which allows having multiple sequences for a single model class instance.

@echarlaix, do you have a better alternative to pass these values?

slyalin: > we shouldn't update this variable with beam_idx / inference execution context

I saw another example where `beam_idx` is stored directly in `past_key_values` here: https://github.com/huggingface/optimum-intel/pull/567/files#diff-96778abd26ec589f358d62af391a34fdda4f7e4734cca48b421c567f5f3042e6R591-R600. So indeed it was treated as a detail of implementation that depends on backend used for inference.

echarlaix: > @echarlaix, do you have a better alternative to pass these values?

why not introducing a class inheriting from `ModelOutput` like `CausalLMOutputWithPast` https://github.com/huggingface/transformers/blob/536ea2aca234fb48c5c69769431d643b0d93b233/src/transformers/modeling_outputs.py#L678 with dedicated `beam_idx` / `inference_request` arguments ?

slyalin: @echarlaix, thanks for the suggestion. But not sure that I understand the flow in this case. Do you have examples where such an inheritance is used for similar purposes?

echarlaix: here for example : [BertForPreTrainingOutput](https://github.com/huggingface/transformers/blob/536ea2aca234fb48c5c69769431d643b0d93b233/src/transformers/models/bert/modeling_bert.py#L738)

dtrawins: @echarlaix if we create new modelOutput data class and it is returned by the Forward method, how it could be passed back to the Forward method in the next cycle?

echarlaix: can you try something like :
```python
from dataclasses import dataclass
from transformers.modeling_outputs import ModelOutput

@dataclass
class CausalLMOutputWithPast(ModelOutput):
    logits: torch.FloatTensor = None
    past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None
    beam_idx: Optional[int] = None
    inference_session = None
```



and then overwritte `_update_model_kwargs_for_generation` https://github.com/huggingface/transformers/blob/45c065109074d60c587d3e562f16531d02a422f6/src/transformers/generation/utils.py#L630 by adding somethign like :   

```python
    def _update_model_kwargs_for_generation(
        self,
        outputs: ModelOutput,
        model_kwargs: dict[str],
        is_encoder_decoder: bool = False,
        standardize_cache_format: bool = False,
    ) -> dict[str]:
        model_kwargs = super()._update_model_kwargs_for_generation(
            outputs=outputs,
            model_kwargs=model_kwargs,
            is_encoder_decoder=is_encoder_decoder,
            standardize_cache_format=standardize_cache_format,
        )

        if "beam_idx" in outputs:
            model_kwargs["beam_idx"] = outputs["beam_idx"]

        return model_kwargs
```

echarlaix: Let me know if you need help on this @dtrawins

dtrawins: @echarlaix @eaidova @slyalin Could you have a look if the latest version is passing the context fine now?
I'm not reusing past_key_values for stateful models with th generation context. There are additional fields in the `forward` output beam_idx and infer_request. Now only 9 tests is left to fix but seams unrelated to concurrency. Probably rebase from main is needed.
Anyway can one comment if beam_idx would be populated correctly. It is not defined now in reorder_caches for stateful models.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -322,6 +340,7 @@ def normalized_config(self):
     def compile(self):
         if self.request is None:
             super().compile()
+            self.compiled_model = self.request
             self.request = self.request.create_infer_request()
```
echarlaix, line 344: why not remove this
```suggestion
```
and use [self.request](https://github.com/huggingface/optimum-intel/blob/ff1d94b5522a2bdda042f02564b76686bf3add6c/optimum/intel/openvino/modeling_base.py#L408) instead of `self.compiled_model` ? (`self.request` doesn't seem to be used anywhere)

tests/openvino/test_modeling.py
```diff
@@ -578,6 +578,72 @@ def test_compare_to_transformers(self, model_arch):
...
+                transformers_outputs = transformers_model(**tokens)
+            # Compare tensor outputs
+            self.assertTrue(torch.allclose(ov_outputs.logits, transformers_outputs.logits, atol=1e-4))
+            # self.assertTrue(False)
+
+        run_on_multiple_threads(run_ov_model, tokens_list, (transformers_model, ov_model))
+
+        del transformers_model
+        del ov_model
+        gc.collect()
```
echarlaix, line 645: The time taken to run all tests is already non negligible so I think we should merge it with `test_compare_to_transformers` (to not duplicate steps like export)

tests/openvino/test_modeling.py
```diff
@@ -608,6 +674,42 @@ def test_pipeline(self, model_arch):
         del model
         gc.collect()
 
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    def test_pipeline_multithreading(self, model_arch):
```
echarlaix, line 678: same comment, can be merged with `test_pipeline`

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -44,6 +45,23 @@
...
+        infer_request(`openvino.runtime.InferRequest` to be reused in the generation cycles.
+        beam_idx (`torch.Tensor` beam search algorimth context for the generation using stateful models
+    """
+
+    loss: Optional[torch.FloatTensor] = None
+    logits: torch.FloatTensor = None
+    past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None
+    hidden_states: Optional[Tuple[torch.FloatTensor, ...]] = None
+    attentions: Optional[Tuple[torch.FloatTensor, ...]] = None
+    infer_request: Optional[openvino.runtime.InferRequest] = None
```
echarlaix, line 63: could we rename it to something like `request` or `inference_request` ?

echarlaix (approved):
Looks great, thanks for iterating on this @dtrawins !

optimum/intel/openvino/modeling_decoder.py
dtrawins: What i confirmed is beam_idx was not passed correctly. The same, initial beam_idx was circulating for the whole pipeline resulting in incorrect accuracy with beam search. Somehow it was not detected by functional tests. 
Anyway my proposal is to pass the beam_idx content from reorder_caches method inside past_key_value. I tested it gives correct results and the code is in my opinion clean. The forward method returns empty past_key_values as expected for stateful models. In case someone would like to manage the pipeline for stateless models outside of transformers using just forward method, it would be still possible - beam_idx should be passed inside past_key_value and inference_request context via model_args. Anyway that is probably unlikely use case scenario. Would it be acceptable?

optimum/intel/openvino/modeling_decoder.py
eaidova: current name is aligned with openvino api name, so for me infer_request sounds better

echarlaix: I think that'd be clearer for users who are not familiar with the openvino ecosystem, also we don't use `infer_request` anywhere in optimum-intel so was thinking about something a bit more explicit

echarlaix:
I think we're close to merge, just waiting for couple of points above to be addressed, let me know if you need any help from my side @dtrawins (fixing conflicts / applying suggested changes)

PR #575, opened 2024-02-28
Telemetry sending

popovaan:
# What does this PR do?

Adding of Telemetry sending.

## Before submitting

### DIFF
--- a/optimum/intel/utils/import_utils.py
+++ b/optimum/intel/utils/import_utils.py
@@ -73,23 +73,27 @@ _openvino_available = importlib.util.find_spec("openvino") is not None
 _openvino_version = "N/A"
 if _openvino_available:
     try:
-        import openvino_telemetry as tm
         from openvino.runtime import get_version
 
-        tm = tm.Telemetry(
-            tid="G-W5E9RNLD4H",
-            app_name="Optimum Intel",
-            app_version=get_version(),
-            backend="ga4",
-            enable_opt_in_dialog=False,
-            disable_in_ci=True,
-        )
+        try:
+            import openvino_telemetry as tm
 
-        tm.send_event(
-            "optimum_intel",
-            "openvino_imported",
-            "OpenVino {}, Optimum {}.".format(get_version(), importlib_metadata.version("optimum")),
-        )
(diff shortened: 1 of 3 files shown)


optimum/intel/utils/import_utils.py
```diff
@@ -74,6 +74,21 @@
...
+                          backend='ga4',
+                          enable_opt_in_dialog=False,
+                          disable_in_ci=True)
+
+        # temporal code to check that "CI" variable is set in CI pipelines,
+        # needs to be removed before merge
+        import os
+        assert "CI" in os.environ and os.environ["CI"].lower() == "true"
+
+        tm.send_event("optimum_intel", "openvino_imported", "OpenVino {}, Optimum {}.".format(get_version(), importlib_metadata.version("optimum")))
```
echarlaix, line 91: Not sure this is something we want to add in optimum

popovaan: There was a request to add Telemetry from Gorbachev Yury so we could track the number of users in optimum-intel.

AlexKoff88 (approved):

eaidova (approved):

optimum/intel/utils/import_utils.py
```diff
@@ -73,8 +73,24 @@
 _openvino_version = "N/A"
 if _openvino_available:
     try:
+        import openvino_telemetry as tm
```
eaidova, line 76: should be here additional try-except for import telemetry only?

if user has openvino but not install openvino-telemetry it now can fail reporting that openvino is not available

optimum/intel/utils/import_utils.py
AlexKoff88: @echarlaix, any concerns about this request? We track OpenVINO usage on a regular basis and we would like to know how much inflow comes from HF ecosystem.

AlexKoff88: @echarlaix, can you please take a look again? As @popovaan mentioned, openvino-telemetry is installed always when openvino is installed in the environment. We are just enabling it here to track optimum usage. CI jobs are excluded.

PR #577, opened 2024-02-28
Fix collecting duplicate tensors in quantization calibration dataset

nikita-savelyevv:
# What does this PR do?

There appear to be duplicates data tensors in collected calibration data for quantization.

For example, when collecting calibration data for whisper decoder on a dataset of audio samples, all tensors for input `encoder_hidden_states` point to the same memory view (the last one encountered). 

I suspect sharing of inputs/outputs to be the reason. However, setting `share_inputs=False, share_outputs=False` for all request calls inside `InferRequestWrapper` class did not help. So I added deepcopying of inputs as a fix.

Added a test covering the fixed issue.

## Before submitting

### DIFF
--- a/tests/openvino/test_quantization.py
+++ b/tests/openvino/test_quantization.py
@@ -18,7 +18,6 @@ import tempfile
 import unittest
 from collections import defaultdict
 from functools import partial
-from itertools import islice
 
 import evaluate
 import numpy as np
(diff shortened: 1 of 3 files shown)


nikita-savelyevv:
@AlexKoff88 @eaidova please, take a look

AlexKoff88 (approved):

eaidova (approved):

tests/openvino/test_quantization.py
```diff
@@ -589,3 +594,38 @@ def compute_metrics(p):
...
+            ov_model.generate(input_features)
+
+        data_hashes_per_key = defaultdict(list)
+        for inputs_dict in calibration_data:
+            for k, v in inputs_dict.items():
+                x = (v.numpy() if isinstance(v, torch.Tensor) else v).copy()
+                data_hashes_per_key[k].append(hash(x.tobytes()))
+        for k, data_hashes in data_hashes_per_key.items():
+            # All hashes can not be equal because calibration dataset contains at least 2 different samples
+            assert any(data_hashes[0] != it for it in data_hashes), f"Collected tensors are all equal for input {k}"
```
echarlaix, line 631: can you please replace it with 
```suggestion
            self.assertTrue(any(data_hashes[0] != it for it in data_hashes))
```
also shouldn't it be :
```suggestion
            self.assertTrue(len(set(data_hashes)) == num_samples)
```
with here num_samples = 2 ?

nikita-savelyevv: > can you please replace it with

Replaced

> also shouldn't it be :

This condition wouldn't hold because during a single whisper model inference, the decoder model is inferenced multiple times autoregressively. So in total the length of `calibration_data` list is larger than the number of used audio samples, in this case 2. Moreover, for some input dict keys there are more unique tensors collected than for others so there's no single `num_samples` value for all of them. That's why I just check that they are not all equal for some key.

PR #578, opened 2024-02-28
Save an openvino config summarizing all information related to quantization when saving model

echarlaix:
When quantization is applied with `load_in_8bit=True` or by  specifying a `quantization_argument`, an `openvino_config.json` file will be saved when the model  / model's config is saved through the `save_pretrained` method


docs/source/optimization_ov.mdx
helena-intel, line 1: Do I understand correctly that with these changes you need to add a quantization_config to .from_pretrained when loading a model, but that doesn't actually do anything until you call quantizer.quantize? 
What is the reason for that change? If the motivation is to save the config, it seems we could just update that after quantizing the model?

setup.py
```diff
@@ -39,18 +39,11 @@
...
-        "openvino>=2023.3",
-        "onnx",
-        "onnxruntime",
-    ],
+    "neural-compressor": ["neural-compressor>=2.2.0", "onnx", "onnxruntime<1.15.0"],
+    "openvino": ["openvino>=2023.3", "onnx", "onnxruntime"],
     "openvino-tokenizers": ["openvino-tokenizers[transformers]"],
-    "nncf": ["nncf>=2.8.1"],
+    # "nncf": ["nncf>=2.8.1"],
+    "nncf": ["nncf @ git+https://github.com/openvinotoolkit/nncf.git"],
```
echarlaix, line 46: (will be removed before merging)

docs/source/optimization_ov.mdx
echarlaix: There is no modification of the quantization process in this PR (you can still apply quantization by providing the `quantization_config` or `load_in_8bit` arguments when creating an  instance of `OVModel`), the main addition is that we save the configuration containing all the information related to quantization when saving the model with `model.save_pretrained(output_dir)`. I also removed the default compression value that was default when creating an `OVConfig` instance as it's not used (we should start moving all quantization parameters in `quantization_config` in the future).

docs/source/optimization_ov.mdx
```diff
@@ -38,8 +38,6 @@ save_dir = "ptq_model"
 def preprocess_function(examples, tokenizer):
     return tokenizer(examples["sentence"], padding="max_length", max_length=128, truncation=True)
 
-# Load the default quantization configuration detailing the quantization we wish to apply
-quantization_config = OVConfig()
```
echarlaix, line 42: removed as currently unused 

https://github.com/huggingface/optimum-intel/blob/v1.15.2/optimum/intel/openvino/quantization.py#L268

AlexKoff88:
@nikita-savelyevv, @l-bat as they plan to introduce relevant changes as well.

optimum/intel/openvino/modeling_base.py
```diff
@@ -203,12 +217,28 @@ def _from_pretrained(
...
+        return cls(
+            model,
+            config=config,
+            model_save_dir=model_cache_path.parent,
+            quantization_config=quantization_config,
+            **kwargs,
+        )
+
+    @staticmethod
+    def _prepare_quantization_config(
```
AlexKoff88, line 232: it s probably worth calling it as `_prepare_weight_quantization_config`

AlexKoff88 (approved):

optimum/intel/openvino/modeling_base.py
echarlaix: updated !

PR #579, opened 2024-02-28
Update  stable diffusion example requirements

helena-intel:
torchvision is needed for running the example. 
nncf from source was needed when this example was created, but  has had a new release since.


AlexKoff88 (approved):

echarlaix (approved):
LGTM, thanks @helena-intel

PR #580, opened 2024-02-29
Convert Tokenizers By Default

apaniukov:
# What does this PR do?

Make tokenizer conversion [the default behaviour](https://github.com/huggingface/optimum-intel/pull/513#discussion_r1507086074). 
- Add `--disable-convert-tokenizer` option.
- Add a warning for the old `--convert-tokenizer` option.
- Add detailed advanced check for tokenizers version compatibility and detailed instructions for troubleshooting.

Covered scenarios for different OpenVINO Tokenizers \ OpenVINO combinations:
| Tokenizers \ OpenVINO   | Release PyPI | Pre-release Simple PyPI | -nightly PyPI | Archive |
| ----------------------- | :----------: | :---------------------: | :-----------: | :-----: |
| Release PyPI            |      +       |            +            |       +       |    +    |
| Pre-release Simple PyPI |      +       |            +            |       +       |    +    |
| Build                   |              |                         |               |    +    |

The tokenizer conversion adds a couple of seconds and the resulting files are a couple of megabytes. For opt-125m (no pytorch weights download), the conversion without tokenizer took 13.340s compared to 13.969s with tokenizer conversion, which is less than 5%. If we take weights downloading time into account, the percentage is even lower.
The resulting folder size increases from 482 MB to 484 MB.

## Before submitting


eaidova (approved):

eaidova:
it seems useless until we need to install openvino tokenizers as extra explicitly, please consider including it in openvino extra

slyalin:
> it seems useless until we need to install openvino tokenizers as extra explicitly, please consider including it in openvino extra

@apaniukov, please add openvino_tokenizer in openvino extra in this PR.

optimum/intel/utils/import_utils.py
```diff
@@ -94,21 +94,43 @@
...
-        (
-            requirement.split("==")[-1]
-            for requirement in importlib_metadata.requires("openvino-tokenizers")
-            if requirement.startswith("openvino==")
-        ),
-        "",
-    )
-    _openvino_tokenizers_available = _compatible_openvino_version == ov_major_version
+    _is_ovt_dev_version = "dev" in _openvino_tokenizers_version
+    _ov_version = importlib_metadata.version("openvino")
```
helena-intel, line 98: This will break when users install OpenVINO with other methods than `pip install openvino`.

apaniukov: Update checks. Add a table for covered scenarios.

optimum/intel/utils/import_utils.py
```diff
@@ -196,8 +170,108 @@ def is_openvino_available():
...
+            "OpenVINO and OpenVINO Tokenizers versions are not binary compatible.\n"
+            f"OpenVINO version:            {_openvino_version}\n"
+            f"OpenVINO Tokenizers version: {tokenizers_version}\n"
+            "First 3 numbers should be the same. Update OpenVINO Tokenizers to compatible version. "
+        )
+        if not pip_metadata_version:
+            message += (
+                "For archive installation of OpenVINO try to build OpenVINO Tokenizers from source: "
+                "https://github.com/openvinotoolkit/openvino_tokenizers/tree/master?tab=readme-ov-file"
+                "#build-and-install-from-source"
```
helena-intel, line 247: It would be good to explain which branch to pick, because building the source version only fixes a version conflict if the version the user has installed happens to be the same as the then current source version.

optimum/intel/utils/import_utils.py
```diff
@@ -196,8 +170,108 @@ def is_openvino_available():
...
+        if tokenizers_version == "0.0.0.0":
+            try:
+                tokenizers_version = importlib_metadata.version("openvino_tokenizers") or tokenizers_version
+            except importlib_metadata.PackageNotFoundError:
+                pass
+        message = (
+            "OpenVINO and OpenVINO Tokenizers versions are not binary compatible.\n"
+            f"OpenVINO version:            {_openvino_version}\n"
+            f"OpenVINO Tokenizers version: {tokenizers_version}\n"
+            "First 3 numbers should be the same. Update OpenVINO Tokenizers to compatible version. "
```
helena-intel, line 241: image
Since the error message said "first 3 numbers" I expected these slightly different nightly versions would work, but it doesn't. So then it makes more sense to say that the versions have to be exactly the same.

I get this error even when I use an archive of March 27 https://storage.openvinotoolkit.org/repositories/openvino/packages/nightly/2024.1.0-14876-fc5da9d9359 and openvino tokenizers  2024.1.0.0.dev20240327 
image

helena-intel (commented):
I created comments about the error messages as I was testing this, but there are quite a few things that can go wrong or cause confusion. IMO it is a better user experience to just show one line like "OpenVINO tokenizer model could not be exported. This does not impact model export. See [documentation url] for more info" and then explain there how to make sure all versions are in sync.

optimum/intel/utils/import_utils.py
```diff
@@ -196,8 +170,108 @@ def is_openvino_available():
...
+            "pip install --pre -U openvino openvino-tokenizers"
+            "--extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly\n"
+            "If you using archive OpenVINO distribution install release, nightly version or build from source:\n"
+            f"Release: pip install --no-deps openvino-tokenizers~={ov_major_version}\n"
+            f"Nightly: pip install --pre openvino-tokenizers~={ov_major_version}.0.dev --extra-index-url "
+            f"https://storage.openvinotoolkit.org/simple/wheels/nightly\n"
+            "Build from source instructions: https://github.com/openvinotoolkit/openvino_tokenizers"
+            "?tab=readme-ov-file#build-and-install-from-source\n"
+        )
+        return False
```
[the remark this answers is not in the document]
apaniukov: That is why it is just `info`, not a warning. But I agree, replace the message with a link to installation instructions.

optimum/intel/utils/import_utils.py
```diff
@@ -196,8 +170,108 @@ def is_openvino_available():
...
+                "#build-and-install-from-source"
+            )
+        else:
+            message += "It is recommended to use the same day builds for pre-release version. "
+            if is_nightly:
+                message += (
+                    "openvino-nightly package detected - use nightly drops from PyPI Simple "
+                    "Repo (PEP 503) implementation on AWS S3 instead. "
+                )
+            message += "To update both OpenVINO and OpenVINO Tokenizers to release version perform:\n"
```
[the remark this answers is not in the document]
apaniukov: Yes, but there is the next line with the command to install the pre-release version, which is identical to nightly but won't cause problems with two openvino versions in one environment.

optimum/intel/utils/import_utils.py
apaniukov: This is because of the different Linux distros. Add this message if Linux platform is detected:
> The PyPI version of OpenVINO Tokenizers is built on CentOS and may not be compatible with other Linux distributions; rebuild OpenVINO Tokenizers from source.

optimum/intel/utils/import_utils.py
apaniukov: This is not the case, the compatibility is defined by the OpenVINO version, not OpenVINO Tokenizers version. The only issue is if the archive OpenVINO version is old and OpenVINO Tokenizers use a newer API.

I can add a message about release branches in the OpenVINO Tokenizers repo instead.

tests/openvino/test_exporters_cli.py
```diff
@@ -122,7 +122,7 @@ def test_exporters_cli(self, task: str, model_type: str):
     def test_exporters_cli_tokenizers(self, task: str, model_type: str):
```
echarlaix, line 122: we shoudl remove :
```diff
- @unittest.skipIf(not is_openvino_tokenizers_available(), reason="OpenVINO Tokenizers not available")
```

optimum/intel/utils/import_utils.py
```diff
@@ -196,8 +170,100 @@ def is_openvino_available():
     return _openvino_available
 
 
+def _create_pip_update_command(package_name: str, compatible_version: str, is_dev_version: bool = False) -> str:
```
echarlaix, line 173: where is this used ?

optimum/intel/utils/import_utils.py
```diff
@@ -196,8 +170,100 @@ def is_openvino_available():
...
+
+@functools.lru_cache(1)
 def is_openvino_tokenizers_available():
-    return _openvino_tokenizers_available
+    if not is_openvino_available():
+        return False
+
+    try:
+        import openvino_tokenizers
+    except ImportError:
```
echarlaix, line 197: would prefer
```suggestion
    if importlib.util.find_spec("openvino_tokenizers") is None:
```

optimum/intel/utils/import_utils.py
```diff
@@ -196,8 +170,100 @@ def is_openvino_available():
...
+    try:
+        pip_metadata_version = importlib_metadata.version("openvino")
+        is_nightly = False
+    except importlib_metadata.PackageNotFoundError:
+        try:
+            pip_metadata_version = importlib_metadata.version("openvino-nightly")
+            is_nightly = True
+        except importlib_metadata.PackageNotFoundError:
+            pip_metadata_version = False
+            is_nightly = False
```
echarlaix, line 214: this should be done outside in my opinion (check if nightly)

optimum/intel/utils/import_utils.py
```diff
@@ -196,8 +170,100 @@ def is_openvino_available():
...
+                    "openvino-nightly package will be deprecated in the future - use pre-release drops instead. "
+                )
+            message += "To update both OpenVINO and OpenVINO Tokenizers to the latest pre-release version perform:\n"
+            if is_nightly:
+                message += "pip uninstall -y openvino-nightly && "
+            message += (
+                "pip install --pre -U openvino openvino-tokenizers "
+                "--extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly\n"
+                "Tokenizer won't be converted."
+            )
```
echarlaix, line 262: this warnings should only be displayed during the export step so should be moved somewhere else

optimum/intel/utils/import_utils.py
apaniukov: Update readme in [this PR](https://github.com/openvinotoolkit/openvino_tokenizers/pull/104)

tests/openvino/test_exporters_cli.py
apaniukov: Removed and add logic checks for ["Test openvino-nightly" stage](https://github.com/huggingface/optimum-intel/blob/a48e0ca3177075235c2a67ee770e07bd751e49dd/.github/workflows/test_openvino.yml#L39), where the libraries are not compatible.

optimum/intel/utils/import_utils.py
apaniukov: Fixed

optimum/intel/utils/import_utils.py
apaniukov: Do you mean in the separate function? It doesn't matter for other components, so I check it here.

optimum/intel/utils/import_utils.py
apaniukov: Moved [here](https://github.com/huggingface/optimum-intel/blob/eb055941103db6a2101df93741afd0c6be34dabd/optimum/exporters/openvino/__main__.py#L347).

optimum/exporters/openvino/__main__.py
```diff
@@ -327,7 +320,7 @@ class StoreAttr(object):
         **kwargs_shapes,
     )
 
-    if convert_tokenizer:
+    if convert_tokenizer and is_openvino_tokenizers_available():
```
echarlaix, line 323: The tokenizer export results in additional files being included which could be confusing for users, can this be moved to to an `openvino_tokenizers` directory ? 

```suggestion
    if convert_tokenizer and is_openvino_tokenizers_available():
         output = Path(output) / "openvino_tokenizers"
```

tests/openvino/test_exporters_cli.py
```diff
@@ -118,21 +118,23 @@ def test_exporters_cli(self, task: str, model_type: str):
...
-                number_of_tokenizers,
-                f"OVT: {is_openvino_tokenizers_available() }",
-            )
+            if not is_openvino_tokenizers_available():
+                self.assertTrue(
+                    "OpenVINO Tokenizers is not available." in output
+                    or "OpenVINO and OpenVINO Tokenizers versions are not binary compatible." in output,
+                    msg=output,
+                )
+                return
```
echarlaix, line 134: would prefer to have it removed to make sure this is always tested (we should always have a compatible version of `openvino` / `openvino-tokenizer` when testing)
```suggestion
```

tests/openvino/test_exporters_cli.py
```diff
@@ -118,21 +118,23 @@ def test_exporters_cli(self, task: str, model_type: str):
         for arch in SUPPORTED_ARCHITECTURES
         if not arch[0].endswith("-with-past") and not arch[1].endswith("-refiner")
     )
-    @unittest.skipIf(not is_openvino_tokenizers_available(), reason="OpenVINO Tokenizers not available")
     def test_exporters_cli_tokenizers(self, task: str, model_type: str):
         with TemporaryDirectory() as tmpdir:
             output = subprocess.check_output(
```
echarlaix, line 123: we should also check whether no error was raised here

echarlaix: With
```bash
optimum-cli export openvino --model hf-internal-testing/tiny-random-t5 --task text2text-generation ov_model
```

I'm getting the following error : 
```
OpenVINO Tokenizer export for T5TokenizerFast is not supported. Exception: [Errno 2] No such file or directory: '/tmp/tmprj8zsg44/spiece.model'
```

echarlaix: Would like to have this fixed before making the tokenizer export default / merging this PR

optimum/exporters/openvino/__main__.py
apaniukov: This is similar to the original tokenizer, so we replicate an existing pattern here.

tests/openvino/test_exporters_cli.py
apaniukov: It is not a bug, we don't support the Unigram model from `tokenizer.json` yet, only from sentencepiece model file. If you use a [repository](https://huggingface.co/google-t5/t5-base/tree/main) that has such a file, the tokenizer will be converted.

I added one more check for the vocab file here: https://github.com/openvinotoolkit/openvino_tokenizers/pull/116
This will transform
```shell
OpenVINO Tokenizer export for T5TokenizerFast is not supported. Exception: [Errno 2] No such file or directory: '/tmp/tmprj8zsg44/spiece.model'
```
into:
```shell
OpenVINO Tokenizer export for T5TokenizerFast is not supported. Exception: Cannot convert tokenizer of this type without `.model` file.
```

The issue is that the original tokenizer has info about `.model` file in the tokenizer object, but it does not have it in practice:
image

I would argue that this is an edge case of a test repo and most tokenizers with info about the `.model` file have it, so it should not block the merge.

optimum/exporters/openvino/__main__.py
echarlaix: I'm thinking about something similar than what's done for [SD models](https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0/tree/main/tokenizer) is this something that sounds reasonable to you? I'm fine with adding it in a following PR if you prefer not including it in this one

apaniukov: Yes, I can do that in a separate PR.

tests/openvino/test_exporters_cli.py
apaniukov: It doesn't throw an error in this case, just prints an error message from a tokenizer conversion process. The model is converted fine:
```bash
>> optimum-cli export openvino --model hf-internal-testing/tiny-random-t5 --task text2text-generation openvino_model
INFO:nncf:NNCF initialized successfully. Supported frameworks detected: torch, onnx, openvino
The task `text2text-generation` was manually specified, and past key values will not be reused in the decoding. if needed, please pass `--task text2text-generation-with-past` to export using the past key values.
Using the export variant default. Available variants are:
    - default: The default ONNX variant.
Using framework PyTorch: 2.1.2+cu121
Overriding 1 configuration item(s)
        - use_cache -> False
Using framework PyTorch: 2.1.2+cu121
Overriding 1 configuration item(s)
        - use_cache -> False
/home/apaniuko/python/optimum-intel/venv/lib/python3.10/site-packages/transformers/modeling_utils.py:908: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if causal_mask.shape[1] < attention_mask.shape[1]:
OpenVINO Tokenizer export for T5TokenizerFast is not supported. Exception: [Errno 2] No such file or directory: '/tmp/tmpyww_b5i_/spiece.model'

>> ls openvino_model/
config.json             openvino_decoder_model.bin  openvino_encoder_model.bin  special_tokens_map.json  tokenizer.json
generation_config.json  openvino_decoder_model.xml  openvino_encoder_model.xml  tokenizer_config.json
```

The confusion comes from the fact that this is the last message from the conversion process. Maybe adding the "The model converted successfully" message will be beneficial. This message can be added in a separate PR as well.

echarlaix: Can we either disable this warning or disable export for unsupported cases ?

apaniukov: This warning is for any exception that appeared during conversion. After seeing this message, the user can create an issue for a particular model/tokenizer support.
I also prefer to tell the user that the tokenizer is not converted, rather than silently not including it without warning, so the lack of a (de)tokenizer model won't be a surprise.

echarlaix: I'd agree if the export of the tokenizer was an explicit choice of the user (current `--convert-tokenizer`) but as this PR makes it default I think having such warning can be an issue as it makes it look like the export failed

apaniukov: I changed the log level to debug, the message won't be visible by default.

echarlaix: thanks @apaniukov

echarlaix:
Couple of tests are failing, will merged once fixed, let me know if you need any help @apaniukov

echarlaix:
Now that https://github.com/huggingface/optimum-intel/pull/618 is merged, the tokenizer won't be exported by default when hybrid quantization is applied for [SD models](https://github.com/huggingface/optimum-intel/blob/4651ac2c4a05e047c86c66a3347b39cb6384cd0b/optimum/commands/export/openvino.py#L214-L247) I'm fine with merging it now and for us to add this in a following PR (with https://github.com/huggingface/optimum-intel/pull/580#discussion_r1562231029) what do you think @apaniukov ?

apaniukov:
> Now that #618 is merged, the tokenizer won't be exported by default when hybrid quantization is applied for [SD models](https://github.com/huggingface/optimum-intel/blob/4651ac2c4a05e047c86c66a3347b39cb6384cd0b/optimum/commands/export/openvino.py#L214-L247) I'm fine with merging it now and for us to add this in a following PR (with [#580 (comment)](https://github.com/huggingface/optimum-intel/pull/580#discussion_r1562231029)) what do you think @apaniukov ?

No problem, let's do that and I will create a new PR with the fixes we discussed here.

PR #585, opened 2024-03-05
change llama test model to cover MQA

jiqing-feng:
Hi @echarlaix  . This PR changed the llama test model to align with llama2 (llama2 and llama share the same model arch but have little difference in config, we would like to change to llama2 as most people use llama2 now.). The new llama test model is quite small and also covers the MQA which met failed issues before. Would you please review it? Thx.

New llama test model link: [Jiqing/tiny_random_llama2](https://huggingface.co/Jiqing/tiny_random_llama2). It is more similar to llama2-(7b, 13b, 70b)

(Might need to re-run CI)

### DIFF
--- a/tests/ipex/test_modeling.py
+++ b/tests/ipex/test_modeling.py
@@ -66,7 +66,7 @@ MODEL_NAMES = {
     "gpt_neox": "hf-internal-testing/tiny-random-GPTNeoXForCausalLM",
     "gptj": "hf-internal-testing/tiny-random-GPTJModel",
     "levit": "hf-internal-testing/tiny-random-LevitModel",
-    "llama": "fxmarty/tiny-llama-fast-tokenizer",
+    "llama": "Jiqing/tiny_random_llama2",
     "marian": "sshleifer/tiny-marian-en-de",
     "mbart": "hf-internal-testing/tiny-random-mbart",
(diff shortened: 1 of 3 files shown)


tests/ipex/test_modeling.py
```diff
@@ -66,7 +66,7 @@
     "gpt_neox": "hf-internal-testing/tiny-random-GPTNeoXForCausalLM",
     "gptj": "hf-internal-testing/tiny-random-GPTJModel",
     "levit": "hf-internal-testing/tiny-random-LevitModel",
-    "llama": "fxmarty/tiny-llama-fast-tokenizer",
+    "llama": "Jiqing/tiny_random_llama2",
```
echarlaix, line 69: would be in favor of keeping the current one and adding an other test instead
```suggestion
    "llama": "fxmarty/tiny-llama-fast-tokenizer",
    "llama2": "Jiqing/tiny_random_llama2",
```

echarlaix: also to fix the test we would need to update the tokenizer from https://huggingface.co/Jiqing/tiny_random_llama2/

echarlaix (approved):

jiqing-feng:
Hi @echarlaix . Thanks for your review. I think we could merge it now :)

PR #586, opened 2024-03-05
include nncf in openvino extras

eaidova:
# What does this PR do?
install nncf with openvino extras. role nncf in model optimization during openvino model export and loading is increased in last releases (e.g. export model with int8 or int4 weights compression for large models), it gives more benefits to users with having it from scratch 

## Before submitting

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ To install the latest release of 🤗 Optimum Intel with the corresponding requi
 | Accelerator                                                                                                      | Installation                                                         |
 |:-----------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------|
 | [Intel Neural Compressor](https://www.intel.com/content/www/us/en/developer/tools/oneapi/neural-compressor.html) | `pip install --upgrade-strategy eager "optimum[neural-compressor]"`  |
-| [OpenVINO](https://docs.openvino.ai/latest/index.html)                                                           | `pip install --upgrade-strategy eager "optimum[openvino,nncf]"`      |
+| [OpenVINO](https://docs.openvino.ai/latest/index.html)                                                           | `pip install --upgrade-strategy eager "optimum[openvino]"`      |
 | [Intel Extension for PyTorch](https://intel.github.io/intel-extension-for-pytorch/#introduction)                 | `pip install --upgrade-strategy eager "optimum[ipex]"`               |
 
 The `--upgrade-strategy eager` option is needed to ensure `optimum-intel` is upgraded to the latest version.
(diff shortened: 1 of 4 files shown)


AlexKoff88 (approved):

helena-intel:
> it gives more benefits to users with having it from scratch

Users already get NNCF by default, unless they choose to not install the nncf extra, even though that is what we recommend in documentation (or if they don't read documentation, but then they would also not know to install the openvino extra)

Two downsides of including NNCF in [openvino]:
- NNCF adds 33 new dependencies (of dependencies). All these dependencies can have bugs, or incompatibilities. NNCF is imported on importing OVModelFor.... so these issues then can also affect users who do not even need NNCF (this happens from time to time).  
- Having nncf installed increases import time (again because when it is available, it is imported). On my laptop (10th gen Core) `from optimum.intel import OVModelForSequenceClassification` taskes  4.5-5.5 seconds without nncf installed, and 7.5-10 seconds with nncf installed. Not a problem when you're exporting a model once and need NNCF, but when you're writing code and have to run an inference script or notebook cell many times, this is noticeable.

AlexKoff88:
> > it gives more benefits to users with having it from scratch
> 
> Users already get NNCF by default, unless they choose to not install the nncf extra, even though that is what we recommend in documentation (or if they don't read documentation, but then they would also not know to install the openvino extra)
> 
> Two downsides of including NNCF in [openvino]:
> 
> * NNCF adds 33 new dependencies (of dependencies). All these dependencies can have bugs, or incompatibilities. NNCF is imported on importing OVModelFor.... so these issues then can also affect users who do not even need NNCF (this happens from time to time).
> * Having nncf installed increases import time (again because when it is available, it is imported). On my laptop (10th gen Core) `from optimum.intel import OVModelForSequenceClassification` taskes  4.5-5.5 seconds without nncf installed, and 7.5-10 seconds with nncf installed. Not a problem when you're exporting a model once and need NNCF, but when you're writing code and have to run an inference script or notebook cell many times, this is noticeable.

Thanks, @helena-intel for providing your opinion here. 
Regarding the additional 33 dependencies you mentioned, I would like to comment that originally it is much less (see https://github.com/openvinotoolkit/nncf/blob/481ce9d9624c4b699878aa337d2e9b78dc25a517/setup.py#L104). We are also validating compatibility of NNCF with OpenVINO on regular basis since this is a part of the product. Moreover, we rely only on official releases of NNCF in Optimum. If you know the cases where import of NNCF causes problems please share with reproducible step,
As for the import time comment, I don't think it is a stopper but we will investigate it. cc'ed @alexsu52, @l-bat please take a look.

Again, our motivation is to provide a smooth UX when working with LLMs and optimizing them. We are also enabling SD optimization in the separate PR. It involves NNCF as well.

PR #589, opened 2024-03-07
change model output parameter to last_hidden_states

jiqing-feng:
`AutoModel` without any `Head` will output `hidden_state` instead of `logits`。

### DIFF
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -193,7 +193,7 @@ class IPEXModel(OptimizedModel):
             inputs["token_type_ids"] = token_type_ids
 
         outputs = self._call_model(**inputs)
-        return ModelOutput(**outputs) if isinstance(outputs, dict) else ModelOutput(logits=outputs[0])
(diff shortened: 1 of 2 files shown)


optimum/intel/ipex/modeling_base.py
```diff
@@ -193,7 +193,7 @@ def forward(
             inputs["token_type_ids"] = token_type_ids
 
         outputs = self._call_model(**inputs)
-        return ModelOutput(**outputs) if isinstance(outputs, dict) else ModelOutput(logits=outputs[0])
+        return ModelOutput(**outputs) if isinstance(outputs, dict) else ModelOutput(last_hidden_state=outputs[0])
```
echarlaix, line 196: it should be `logits` for all other classes (like `IPEXModelForSequenceClassification`, `IPEXModelForTokenClassification`, ...) can you update it accordingly ?

jiqing-feng: Add a class parameter `output_name` would be more convenient and clean. Please take a review and let me know your opinion:)

echarlaix (approved):
Thanks for the fix @jiqing-feng !

PR #590, opened 2024-03-07
Remove accelerate and onnxruntime from required dependencies

echarlaix:


helena-intel (approved):
Thanks @echarlaix ! Somewhat related: what do you think about moving onnx to INSTALL_REQUIRE? It is required for INC, IPEX and OpenVINO, so it might as well be in the main dependencies? It would help user experience a bit, because pip will show ONNX as a dependency of optimum-intel, and will check compatibility if there are ever version limits, and it makes it easier to install optimum-intel in an existing OpenVINO environment for advanced users (we should still recommend optimum[openvino] to users, but it would make it possible to skip the extra).

echarlaix:
> Thanks @echarlaix ! Somewhat related: what do you think about moving onnx to INSTALL_REQUIRE? It is required for INC, IPEX and OpenVINO, so it might as well be in the main dependencies? It would help user experience a bit, because pip will show ONNX as a dependency of OpenVINO, and will check compatibility if there are ever version limits, and it makes it easier to install optimum-intel in an existing OpenVINO environment for advanced users (we should still recommend optimum[openvino] to users, but it would make it possible to skip the extra).

Makes sense! Added it in https://github.com/huggingface/optimum-intel/pull/590/commits/0b50ed5cdfc87e6cfc68d1370946904f32b74d53

AlexKoff88:
We use `datasets` in the data-aware weight quantization of LLMs. Shall we throw a meaningful exception in this case?

I also looked at the CI errors, we should update the references caused by the changes in OpenVINO and NNCF.

echarlaix:
> We use `datasets` in the data-aware weight quantization of LLMs. Shall we throw a meaningful exception in this case?

Currently if `weights_only=True` then `_get_calibration_dataloader` will not be called so modifications from this PR shouldn't have any impact (would definitely make sense to also enable the possibility to provide the `calibration_dataset` to the `OVQuantizer` for data-aware weight only quantization)

> I also looked at the CI errors, we should update the references caused by the changes in OpenVINO and NNCF.

Added the fix in https://github.com/huggingface/optimum-intel/pull/590/commits/94a990f2d08e1b68a2f78ccd1fe437236309caf2

PR #592, opened 2024-03-07
Fix default int8 quantization for CLI

echarlaix:


AlexKoff88 (approved):

helena-intel (approved):
Thanks for the quick fix! I tested it and it works as expected again.

PR #594, opened 2024-03-11
add ipex model example for text-generation

jiqing-feng:
Hi @echarlaix . I suppose we'd better have an ipex model example for different tasks since the integration of `IPEXModel` has finished. For now, I only added the text-generation task and will support more tasks in the examples, WDYT?


examples/ipex/text-generation/ipex_text-generation.ipynb
```diff
@@ -0,0 +1,96 @@
+{
```
echarlaix, line 1: looks great, could you mode it to https://github.com/huggingface/optimum-intel/tree/main/notebooks ?

echarlaix (approved):

examples/ipex/text-generation/ipex_text-generation.ipynb
jiqing-feng: Done

notebooks/ipex/ipex_text-generation.ipynb
```diff
@@ -0,0 +1,96 @@
...
+   "metadata": {},
+   "source": [
+    "# IPEX model for text-generation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "IPEX model will replace the linears and some ops. Please note that IPEXModel uses a graph mode model to inference to accelerate the generation."
```
echarlaix, line 14: what do you think about something like : 
```suggestion
    "To load your IPEX model, you can just replace your `AutoModelForXxx` class with the corresponding `IPEXModelForXxx` class. You can set `export=True` to load a PyTorch checkpoint, export your model via TorchScript and apply IPEX optimizations : both operators optimization (replaced with customized IPEX operators) and graph-level optimization (like operators fusion) will be applied on your model."
```

PR #595, opened 2024-03-11
add ipex readme

jiqing-feng:
Hi @echarlaix . I'd like to introduce ipex model to users in the readme,hope to hear your opinion. Thx!


README.md
```diff
@@ -44,6 +44,34 @@ where `extras` can be one or more of `ipex`, `neural-compressor`, `openvino`, `n
...
+Below are examples of how to use IPEX model to generate texts.
+### generate
+```diff
+import torch
+from transformers import AutoTokenizer, AutoConfig
+- from transformers import AutoModelForCausalLM
++ from optimum.intel.ipex import IPEXModelForCausalLM
+
+config = AutoConfig.from_pretrained("gpt2")
+model = IPEXModelForCausalLM.from_pretrained(
```
echarlaix, line 57:
```suggestion
- model = AutoModelForCausalLM.from_pretrained(
+ model = IPEXModelForCausalLM.from_pretrained(
```

README.md
```diff
@@ -44,6 +44,34 @@ where `extras` can be one or more of `ipex`, `neural-compressor`, `openvino`, `n
...
+  "gpt2",
```
echarlaix, line 58: do you have a quantized model under the intel organization https://huggingface.co/Intel that we can directly load instead ?

jiqing-feng: I am afraid not, but I could upload a quantized one. 
Inputting a common model_id will be more acceptable for users because they might want to try different models with the API. Would like to hear your opinion : )

echarlaix: I'm not sure I agree here, it would be nice to load an IPEX quantized model here to promote this feature support. If we don't do it we should then explain the gain in using IPEX vs torch : what optimization are applied depending on whether we epxort the model to torchscript or not.

Also could you move this section after the OpenVINO section so that we keep the same structure ?

README.md
```diff
@@ -44,6 +44,33 @@ where `extras` can be one or more of `ipex`, `neural-compressor`, `openvino`, `n
...
+from transformers import AutoTokenizer
+- from transformers import AutoModelForCausalLM
++ from optimum.intel.ipex import IPEXModelForCausalLM
+
+- model = AutoModelForCausalLM.from_pretrained(
++ model = IPEXModelForCausalLM.from_pretrained(
+  "gpt2",
+  torch_dtype=torch.bfloat16,
+  export=True,
+)
```
echarlaix, line 61:
```suggestion

  model_id = "gpt2"
- model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.bfloat16)
+ model = IPEXModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.bfloat16, export=True)
```

README.md
```diff
@@ -202,6 +203,33 @@ Quantization aware training (QAT) is applied in order to simulate the effects of
 You can find more examples in the [documentation](https://huggingface.co/docs/optimum/intel/index).
 
 
+## IPEX
+With `export=True`, IPEX model will replace torch linear to ipex linear which prepacks the weights. It will also apply linear fusioin and [IAKV](https://intel.github.io/intel-extension-for-pytorch/cpu/latest/tutorials/llm.html#indirect-access-kv-cache) for generation. Finally, jit.trace will be applied to change the model to graph mode.
```
echarlaix, line 207:
```suggestion
With `export=True`, IPEX model will replace torch linear to ipex linear which prepacks the weights. It will also apply linear fusion and [IAKV](https://intel.github.io/intel-extension-for-pytorch/cpu/latest/tutorials/llm.html#indirect-access-kv-cache) for generation. Finally, jit.trace will be applied to change the model to graph mode.
```

echarlaix: > With `export=True`, IPEX model will replace torch linear to ipex linear which prepacks the weights. 

Isn't it also the case for `export=False` ?

jiqing-feng: The Ipex optimization should be combined with jit.trace. Otherwise, there are no big differences compared to stock PyTorch.

README.md
```diff
@@ -202,6 +203,29 @@ Quantization aware training (QAT) is applied in order to simulate the effects of
 You can find more examples in the [documentation](https://huggingface.co/docs/optimum/intel/index).
 
 
+## IPEX
+With `export=True`, IPEX model will replace torch linear to ipex linear which prepacks the weights. It will also apply linear fusion and [IAKV](https://intel.github.io/intel-extension-for-pytorch/cpu/latest/tutorials/llm.html#indirect-access-kv-cache) for generation. Finally, jit.trace will be applied to change the model to graph mode.
+Here is the example of how to use IPEX optimized model to generate texts.
+### generate
+```diff
+  import torch
```
echarlaix, line 211:
```suggestion
```

README.md
```diff
@@ -202,6 +203,28 @@ Quantization aware training (QAT) is applied in order to simulate the effects of
 You can find more examples in the [documentation](https://huggingface.co/docs/optimum/intel/index).
 
 
+## IPEX
+With `export=True`, IPEX model will replace torch linear to ipex linear which prepacks the weights. It will also apply linear fusion and [IAKV](https://intel.github.io/intel-extension-for-pytorch/cpu/latest/tutorials/llm.html#indirect-access-kv-cache) for generation. Finally, jit.trace will be applied to change the model to graph mode.
+Here is the example of how to use IPEX optimized model to generate texts.
+### generate
```
echarlaix, line 209:
```suggestion

```

README.md
```diff
@@ -202,6 +203,27 @@ Quantization aware training (QAT) is applied in order to simulate the effects of
 You can find more examples in the [documentation](https://huggingface.co/docs/optimum/intel/index).
 
 
+## IPEX
+With `export=True`, IPEX model will replace torch linear to ipex linear which prepacks the weights. It will also apply linear fusion and [IAKV](https://intel.github.io/intel-extension-for-pytorch/cpu/latest/tutorials/llm.html#indirect-access-kv-cache) for generation. Finally, jit.trace will be applied to change the model to graph mode.
+Here is the example of how to use IPEX optimized model to generate texts.
```
echarlaix, line 208:
```suggestion
To load your IPEX model, you can just replace your `AutoModelForXxx` class with the corresponding `IPEXModelForXxx` class. You can set `export=True` to load a PyTorch checkpoint, export your model via TorchScript and apply IPEX optimizations : both operators optimization (replaced with customized IPEX operators) and graph-level optimization (like operators fusion) will be applied on your model.
```

README.md
```diff
@@ -202,6 +203,26 @@ Quantization aware training (QAT) is applied in order to simulate the effects of
 You can find more examples in the [documentation](https://huggingface.co/docs/optimum/intel/index).
 
 
+## IPEX
+To load your IPEX model, you can just replace your `AutoModelForXxx` class with the corresponding `IPEXModelForXxx` class. You can set `export=True` to load a PyTorch checkpoint, export your model via TorchScript and apply IPEX optimizations : both operators optimization (replaced with customized IPEX operators) and graph-level optimization (like operators fusion) will be applied on your model.
+```diff
+  from transformers import AutoTokenizer, pipeline
+- from transformers import AutoModelForCausalLM
++ from optimum.intel.ipex import IPEXModelForCausalLM
```
echarlaix, line 211:
```suggestion
+ from optimum.intel import IPEXModelForCausalLM
```

README.md
```diff
@@ -44,6 +44,7 @@ where `extras` can be one or more of `ipex`, `neural-compressor`, `openvino`, `n
 
 # Quick tour
 
+
```
echarlaix, line 47:
```suggestion
```

PR #596, opened 2024-03-11
hot fix for weights compression

eaidova:
# What does this PR do?
fix export large models with quantization config and prevent overriding ov_config provided into model class if used export=True (it may have some additional parameters for ov runtime e.g. inference_precision_hint)

Fixes # (issue)

## Before submitting

### DIFF
--- a/tests/openvino/test_quantization.py
+++ b/tests/openvino/test_quantization.py
@@ -490,6 +490,27 @@ class OVWeightCompressionTest(unittest.TestCase):
                     }
                     save_model_patch.aasert_called_with(saving_params)
 
+    def test_ovmodel_load_large_model_with_additional_quantization_config(self):
+        with unittest.mock.patch("transformers.modeling_utils.ModuleUtilsMixin") as model_mixin_patch:
+            model_mixin_patch.num_parameters.return_value = 2e9
+            with unittest.mock.patch("openvino.runtime.ie_api.Core.read_model") as core_patch:
+                with unittest.mock.patch("optimum.exporters.openvino.convert._save_model") as save_model_patch:
+                    _ = OVModelForCausalLM.from_pretrained(
+                        MODEL_NAMES["llama"],
+                        export=True,
+                        compile=False,
+                        use_cache=False,
+                        quantization_config=OVWeightQuantizationConfig(bits=4, sym=True, group_size=-1, ratio=0.8),
+                    )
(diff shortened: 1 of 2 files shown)


tests/openvino/test_quantization.py
```diff
@@ -490,6 +490,27 @@ def test_ovmodel_load_large_model_with_uncompressed_weights(self):
...
+                        quantization_config=OVWeightQuantizationConfig(bits=4, sym=True, group_size=-1, ratio=0.8),
+                    )
+                    # quantization will be performed later, using load_model
+                    saving_params = {
+                        "model": unittest.mock.ANY,
+                        "path": unittest.mock.ANY,
+                        "compression_option": "fp32",
+                        "compression_ratio": None,
+                    }
+                    save_model_patch.aasert_called_with(saving_params)
```
echarlaix, line 512:
```suggestion
                    save_model_patch.assert_called_with(saving_params)
```

eaidova: what changed?

tests/openvino/test_quantization.py
```diff
@@ -490,6 +490,27 @@ def test_ovmodel_load_large_model_with_uncompressed_weights(self):
...
+                        compile=False,
+                        use_cache=False,
+                        quantization_config=OVWeightQuantizationConfig(bits=4, sym=True, group_size=-1, ratio=0.8),
+                    )
+                    # quantization will be performed later, using load_model
+                    saving_params = {
+                        "model": unittest.mock.ANY,
+                        "path": unittest.mock.ANY,
+                        "compression_option": "fp32",
+                        "compression_ratio": None,
```
echarlaix, line 510: should be replaced with `ov_config`

eaidova: I think it should be take care in separated pr with other tests, I keep it in the same way how it was used in other tests

eaidova: I want to secure a case in  which I met an issue (I tried to compress the model to int4 and noted that the model was compressed into int8 instead).

AlexKoff88:
@echarlaix, it seems like .from_pretrained() method does not work correctly with quantization options exposed in it. We have multiple concerns about it and should fix ASAP. Please take a look again.

echarlaix (approved):
Thanks for the fix @eaidova

PR #597, opened 2024-03-11
add commit id into dev version setup

eaidova:
# What does this PR do?
Add commit_id into the dev version during installation to prevent pip caching installation from the source 

Fixes # (issue)

## Before submitting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,6 @@
+import os
 import re
+import subprocess
 
 from setuptools import find_namespace_packages, setup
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -8,6 +10,18 @@
     filepath = "optimum/intel/version.py"
     with open(filepath) as version_file:
         (__version__,) = re.findall('__version__ = "(.*)"', version_file.read())
+    if __version__.endswith(".dev0"):
+        dev_version_id = "unknown_version"
```
echarlaix, line 14: would prefer to not update `__version__` if we're not able to get the latest commit

eaidova: updated

echarlaix (approved):

PR #599, opened 2024-03-11
Add warning for transformers>=4.38 and OpenVINO 2024.0

helena-intel:
Add a warning to downgrade transformers to 4.37 or use OpenVINO nightly for stateful models. Not ideal, but better than nothing.

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -41,7 +41,7 @@ def patch_model_with_bettertransformer(model):
             + COLOR_RESET
         )
 
-    if is_transformers_version(">=", "4.38") and is_openvino_version("<", "2024.1.0-14612"):
+    if getattr(model.config, "model_type") in {"gpt_bigcode", "llama"} and is_transformers_version(">=", "4.38") and is_openvino_version("<", "2024.1.0-14612"):
         log.warn(
             COLOR_RED + f"[WARNING] Stateful models are not supported with Transformers {_transformers_version} and "
             "this OpenVINO version. For good performance, consider using a nightly OpenVINO build: "
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -15,6 +15,7 @@
 import logging as log
 
 from optimum.intel.utils.import_utils import (
+    _openvino_version,
     _torch_version,
     _transformers_version,
     is_torch_version,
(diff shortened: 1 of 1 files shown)


optimum/exporters/openvino/model_patcher.py
```diff
@@ -39,6 +41,14 @@ def patch_model_with_bettertransformer(model):
             + COLOR_RESET
         )
 
+    if is_transformers_version(">=", "4.38") and _openvino_version < "2024.1.0-14612":
+        log.warn(
+            COLOR_RED + f"[WARNING] Stateful models are not supported with Transformers {_transformers_version} and "
+            "this OpenVINO version. For good performance, consider using a nightly OpenVINO build: "
```
echarlaix, line 47:
```suggestion
            f"this OpenVINO version {_openvino_version}. For good performance, consider using a nightly OpenVINO build: "
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -39,6 +41,14 @@ def patch_model_with_bettertransformer(model):
             + COLOR_RESET
         )
 
+    if is_transformers_version(">=", "4.38") and _openvino_version < "2024.1.0-14612":
```
echarlaix, line 44: why not :
```suggestion
    if is_transformers_version(">=", "4.38") and is_optimum_version("<", "2024.1.0-14612"):
```

echarlaix: also this concerns only a subset of architectures, so would make sense to only have a warning for those wdyt ?

optimum/exporters/openvino/model_patcher.py
helena-intel: I didn't do that because if you install 2024.0.0 release, _openvino_version is '2024.0.0-14509' and I didn't think that version with commit-id would be recognizable, and it's not useful for users, because they cannot do `pip install openvino==2024.0.0-15000` for example. But if you want I can change it.

optimum/exporters/openvino/model_patcher.py
helena-intel: Changed both. Do models based on llama keep model_type llama?

optimum/exporters/openvino/model_patcher.py
echarlaix: I see, we could use `_openvino_version.split("-")[0]` but no strong opinion on this,current integration also works for me

echarlaix (approved):

optimum/exporters/openvino/model_patcher.py
helena-intel: Now showing full version if people have installed a prerelease, but only main version for 2024.0.0 or lower installs.

optimum/exporters/openvino/model_patcher.py
echarlaix: [Here](https://github.com/huggingface/transformers/blob/b382a09e28c7e59129246ccdf4b00f2cac4547a4/src/transformers/models/auto/configuration_auto.py#L274) you can find the model's type list

PR #601, opened 2024-03-12
Improve documentation

echarlaix:


docs/source/inference.mdx
```diff
@@ -99,21 +99,22 @@ tokenizer.save_pretrained(save_directory)
 
 ### Weight-only quantization
 
-You can also apply 8-bit or 4-bit weight quantization when exporting your model with the CLI by setting the `weight-format` argument to respectively `int8` or `int4`:
+You can also apply fp16, 8-bit or 4-bit weight quantization on the linear and embedding layers when exporting your model with the CLI by setting `--weight-format` to respectively `fp16`, `int8` or `int4`:
```
AlexKoff88, line 102: convolutional layers are supported as well

docs/source/optimization_ov.mdx
```diff
@@ -19,27 +19,95 @@ limitations under the License.
...
+model = OVModelForCausalLM.from_pretrained(model_id, quantization_config=quantization_config)
+```
+
+You can tune quantization parameters to achieve a better performance accuracy trade-off as follows:
+
+```python
+quantization_config = OVWeightQuantizationConfig(bits=4, sym=False, ratio=0.8, dataset="ptb")
+```
+
+By default the quantization scheme will be [assymmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#asymmetric-quantization), to make it [symmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#symmetric-quantization) you can add `sym=True`.
```
AlexKoff88, line 70:
```suggestion
By default the quantization scheme will be [asymmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#asymmetric-quantization), to make it [symmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#symmetric-quantization) you can add `sym=True`.
```

AlexKoff88:
Thanks, @echarlaix. Overall it looks good, I just made some minor comments.

AlexKoff88 (approved):

docs/source/inference.mdx
helena-intel, line 102:
```suggestion
You can also apply fp16, 8-bit or 4-bit weight compression on the linear and embedding layers when exporting your model with the CLI by setting `--weight-format` to respectively `fp16`, `int8` or `int4`:
```
Not a strong opinion if others disagree, but I prefer compression because FP16 is still continuous. We're also calling it compression in other docs.

helena-intel (approved):
Thanks @echarlaix this is a great improvement!

docs/source/inference.mdx
echarlaix: added it in https://github.com/huggingface/optimum-intel/pull/601/commits/be866d49e75cd83771d0bc990760d05ef48e7b92

PR #603, opened 2024-03-13
Fix OpenVINO notebooks

helena-intel:
- use streaming dataset for audio example to fix permission issue with URL
- change pytorch_model.bin to model.safetensors for size comparison
- speed up quantization notebook test by using only 10 samples
- update half() messaging (no longer needed to do half() on GPU)

The diff for the inference notebook is large because I ran it again and outputs/metadata changed.


echarlaix (approved):
Thanks a lot for the fixes @helena-intel

PR #606, opened 2024-03-14
Phi-2 on Intel's MTL iGPU demo notebook

ofirzaf:
# What does this PR do?
We added a notebook that demonstrates how to quantize Microsoft's Phi-2 model to 4bit Weight Only Quantization using OpenVINO. Then we show how to use the quantized model for generating text and code on Intel's iGPU found on the latest Intel's client CPUs. Lastly, we show a simple Gradio demo for chatbot with the quantized Phi-2 locally on your own laptop.

@echarlaix 

## Before submitting


notebooks/openvino/phi-2_on_mtl.ipynb
```diff
@@ -0,0 +1,571 @@
...
+    "device = 'gpu'  # choose from ['cpu', 'gpu']\n",
+    "stateful = True  # Optimize model by setting the KV cache as part of the models state instead of as an input"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1f398868-93d7-4c2d-9591-9bac8e9b701c",
+   "metadata": {},
+   "source": [
+    "With this configuration we expect the model size to reduce to around to 1.62GB:  $0.8 \\times 2.7{\\times}10^3 \\times \\frac{1}{2}\\text{B} + 0.2 * 2.7{\\times}10^3 \\times 1\\text{B} = 1.62{\\times}10^3\\text{B} = 1.62\\text{GB}$"
```
echarlaix, line 82: nice addition ! could make sense to specify that the model has  2.7 billion parameters and to give a bit more context so that it's clearer for users, wdyt ?

notebooks/openvino/phi-2_on_mtl.ipynb
```diff
@@ -0,0 +1,571 @@
+{
```
echarlaix, line 1: also might make sense to rename the file to something like code_generation_weight_quantization.ipynb

echarlaix (approved):
Looks good, thanks for your work @ofirzaf !

notebooks/openvino/phi-2_on_mtl.ipynb
ofirzaf: Added some information about the model in the first paragraph.

notebooks/openvino/phi-2_on_mtl.ipynb
ofirzaf: We prefer the notebook will stay with the current name, will that be a problem?

ofirzaf:
Hi @echarlaix, thanks for the review. I uploaded a revised version. Please let me know if you think any other changes are needed. Thanks!

notebooks/openvino/phi-2_on_mtl.ipynb
```diff
@@ -0,0 +1,585 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Load kwargs\n",
+    "load_kwargs = {\n",
+    "    'device': device,\n",
+    "    'ov_config': {\n",
+    "        \"PERFORMANCE_HINT\": \"LATENCY\",\n",
+    "        \"INFERENCE_PRECISION_HINT\": precision,\n",
+    "        \"CACHE_DIR\": os.path.join(save_name, \"model_cache\"),  # OpenVINO will use this directory as cache\n",
```
helena-intel, line 122: These settings should be good by default. What is the reason for setting them here? When settings are set to a specific value users will wonder "when should I use other settings", "what do these settings mean", "what other values can I choose" but that's not explained in the notebook (and would be a distraction to explain too from my perspective)

notebooks/openvino/phi-2_on_mtl.ipynb
```diff
@@ -0,0 +1,585 @@
...
+    "    },\n",
+    "    'stateful': stateful,\n",
```
helena-intel, line 124: Statefulness is enabled by default and there is no reason to disable statefulness, so I would remove this option.

notebooks/openvino/phi-2_on_mtl.ipynb
ofirzaf: We don't use the precision default, last time I checked the code, if I override the ov_config to change 1 of the defaults, the other defaults do not get updated. Please check if the default `PERFORMANCE_HINT` and `CACHE_DIR` still apply while I change the default `INFERENCE_PRECISION_HINT`. Thanks

notebooks/openvino/phi-2_on_mtl.ipynb
ofirzaf: We have plans to update this notebook to showcase other uses that require stateful to be off and this kind of left over. We can remove this for now if it is an issue.

notebooks/openvino/phi-2_on_mtl.ipynb
helena-intel: Aah that is true.  We should fix that in optimum-intel and not completely override the config but  add the default CACHE_DIR and PERFORMANCE_HINT if they are not given. I will fix that. So yes, keep for now, maybe the notebook can be updated when we improve the default experience.

notebooks/openvino/phi-2_on_mtl.ipynb
helena-intel: I would really prefer to remove it in this notebook for now. Curious about the usecase that requires stateful to be off!

ofirzaf: Removed

notebooks/openvino/phi-2_on_mtl.ipynb
ofirzaf: `optimized_chatbot.ipynb` sounds good?

guybd: @echarlaix how about optimized_llm_inference.ipynb?
we are ok with whatever you decide 
anyway - can you make the change and accept the PR before the weekend?
@ofirzaf

echarlaix: works for me ! will merge it now to not block anything on your side

PR #609, opened 2024-03-14
Add Gemma to list of models to show openvino/transformers version warning for

helena-intel:
Add Gemma to the list of models to show a bright red warning for, because statefulness is not supported with 2024.0


echarlaix (approved):
LGTM !

PR #605, opened 2024-03-14
apply bettertransformer by default for causallm

eaidova:
# What does this PR do?

align stateful and stateless IRs for causallm

PR #612, opened 2024-03-18
Revert "apply bettertransformer by default for causallm"

eaidova:
Reverts huggingface/optimum-intel#605


slyalin:
@echarlaix, please revert because it breaks the expected form of stateful models. It requires more thorough validation on our side.

echarlaix (approved):

PR #620, opened 2024-03-21
rework inputs preparation for OVModelForCausalLM

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -136,6 +136,7 @@ def __init__(
         self._original_model = self.model.clone()  # keep original model for serialization
         self._pkv_precision = Type.f32
         self.next_beam_idx = None
+        self.past_len = 0
```
echarlaix, line 139:
```suggestion
        self._past_length = 0
```

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -491,6 +495,7 @@ def forward(
             # the first condition at the function beginning above.
             # It should be something that is not None and it should be True when converted to Boolean.
             past_key_values = ((),)
+            self.past_len += input_ids.shape[1]
```
echarlaix, line 498: I would prefer to have a method `_get_past_length` with `input_ids` and `past_key_values` arguments returning the expected past_length  instead of having a `past_len` attribute which needs to be updated at different steps (which will get complicated for maintenance in my opinion). WHat do you think @eaidova ?

eaidova: @echarlaix  I'm not sure that it is possible, past_len was introduced primarily for stateful model, because we can not touch past_key_values that are stored inside stateful model (it can become complicated and non-effective because the conception of stateful model provides managing memory and format for them on runtime device level, API that returns kv cache back to users will break it and requires to repack them in user-friendly format it can cost additional resources and time)

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -501,24 +506,38 @@ def forward(
...
 
-    # Adapted from transformers.models.gpt2.modeling_gpt2.GPT2LMHeadModel.prepare_inputs_for_generation
+    # Adapted from transformers.models.llama.modeling_llama.LlamaForCausalLM.prepare_inputs_for_generation
     def prepare_inputs_for_generation(self, input_ids, past_key_values=None, **kwargs):
         # if model is used as a decoder in encoder-decoder model, the decoder attention mask is created on the fly
         attention_mask = kwargs.get("attention_mask", None)
         use_cache = kwargs.get("use_cache", None)
 
+        if past_key_values is not None:
+            # Keep only the unprocessed tokens:
```
ofirzaf, line 523: Hi, I used this PR to try speculative decoding with OV models. This PR fixes most of the problems that existed before when running with SD. I would like to suggest to add the following condition to better support SD:
```python
if past_key_values is not None:
  if not self.stateful:
    self.past_len = past_key_values[0][0].shape[2]
  ...
```
In the current implementation we remember the last sequence length the model saw but in SD the next input length is different according to the number of tokens that were accepted.

optimum/intel/openvino/modeling_decoder.py
ofirzaf: @eaidova the following might help, you can find the current length in stateful models by using the following: 
```python
self.request.query_state()[0].state.shape[2]
```

eaidova: @ofirzaf i know how to query state, what I'm trying to say that we need to check size on each inference, it is not recommended to do it so often and will lead to perf and memory menegment issus due to packing/unpacking states, that is why using calculation of shapes without quiry state is preferred from openvino runtime perspective

optimum/intel/openvino/modeling_decoder.py
eaidova: renamed

optimum/intel/openvino/modeling_decoder.py
eaidova: moved to usage function

optimum/intel/openvino/modeling_decoder.py
eaidova: updated

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -402,14 +398,13 @@ def prepare_inputs(
...
                         )
-                else:
-                    past_len = past_key_values[0].shape[-2]
 
                 # Add the past_key_values to the decoder inputs
                 inputs = dict(zip(self.key_value_input_names, past_key_values))
 
             # Create empty past_key_values for decoder_with_past first generation step
             elif self.use_cache:
+                past_len = 0
```
eaidova, line 407:
```suggestion
```

eaidova:
@echarlaix looks like phi model used in tests was removed  from hub
`FAILED tests/openvino/test_modeling.py::OVModelForCausalLMIntegrationTest::test_multiple_inputs_24_phi - huggingface_hub.utils._errors.RepositoryNotFoundError: 401 Client Error. (Request ID: Root=1-66066c80-77cc281202da82206bf94efb;6b43617e-63e2-4432-95ce-1b591c2093f4)`

Do you know which model can be used as replacement?

echarlaix:
> @echarlaix looks like phi model used in tests was removed from hub `FAILED tests/openvino/test_modeling.py::OVModelForCausalLMIntegrationTest::test_multiple_inputs_24_phi - huggingface_hub.utils._errors.RepositoryNotFoundError: 401 Client Error. (Request ID: Root=1-66066c80-77cc281202da82206bf94efb;6b43617e-63e2-4432-95ce-1b591c2093f4)`
> 
> Do you know which model can be used as replacement?

Let me check whether [`bumblebee-testing/tiny-random-PhiForCausalLM`](https://huggingface.co/bumblebee-testing/tiny-random-PhiForCausalLM) could be used instead

echarlaix:
should be fixed in https://github.com/huggingface/optimum-intel/pull/643

echarlaix (approved):

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -630,22 +654,12 @@ def _from_pretrained(
...
     def prepare_inputs_for_generation(self, input_ids, past_key_values=None, **kwargs):
-        attention_mask = kwargs.get("attention_mask", None)
-        use_cache = kwargs.get("use_cache", None)
-
         # only last token for input_ids if past is not None
         if past_key_values and not self.stateful:
             # the cache may be in the stardard format (e.g. in contrastive search), convert to bloom's format if needed
             if past_key_values[0][0].shape[0] == input_ids.shape[0]:
                 past_key_values = self._convert_to_bloom_cache(past_key_values)
-
```
echarlaix, line 641: @eaidova could you also add a test for verify that calling `.generate()` twice in a row gives expected results ? (`self._past_length` correctly re-initialized)

echarlaix: also current tests are failing due to what seems to be a mismatch in the shape of generated `position_ids`

PR #626, opened 2024-03-22
disable export=False in IPEXModel

jiqing-feng:
Hi @echarlaix . I think we should disable `export=False` in `IPEXModel` because we propose that the ipex optimization contains jit.trace. After this PR merged, I could update the ipex example and readme (#594 #595 ) to clarify it. Thx

### DIFF
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -204,7 +204,9 @@ class IPEXModel(OptimizedModel):
         **kwargs,
     ):
         if not getattr(config, "torchscript", False):
-            raise ValueError("`config.torchscript` should be set to `True`, if your model is not a TorchScript model and needs to be traced please set `export=True` when loading it with `.from_pretrained()`")
+            raise ValueError(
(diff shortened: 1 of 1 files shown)


optimum/intel/ipex/modeling_base.py
```diff
@@ -203,6 +203,9 @@ def _from_pretrained(
         subfolder: str = "",
         **kwargs,
     ):
+        if not getattr(config, "torchscript", False):
+            raise ValueError("`torchscript` should be set to True to load TorchScript model")
+
```
echarlaix, line 208: Why do we need this ?

echarlaix, line 208:
```suggestion
        if not getattr(config, "torchscript", False):
            raise ValueError("`config.torchscript` should be set to `True`, if your model is not a TorchScript model and needs to be traced please set `export=True` when loading it with `.from_pretrained()`")

```

echarlaix (approved):

echarlaix (approved):
LGTM !

PR #627, opened 2024-03-22
fix jit memory issue

### LINKED ISSUE #604: Memory not release after jit.trace
Hi @echarlaix . When I use `IPEXModelForCausalLM.from_pretrained(model_id)` or `TSModelForCausalLM.from_pretrained(model_id)`, I found the memory usage is double as the original model loaded by transformers. This is because of the [jit issue](https://github.com/pytorch/pytorch/issues/96726). Should we refactor our from_pretrained API to work around this issue? For example, we return the traced model directly instead of saving and loading, because the previous traced model cannot be released and the memory will be double if we load the jit model which we saved before.

jiqing-feng:
Fix #604 

Hi @echarlaix . As we discussed before, the jit trace memory issue can only workaround for now. I think the best way is to skip the save and load step and just return the cls after jit tracing. WDYT?

Jit trace memory issue: [96726](https://github.com/pytorch/pytorch/issues/96726)

### DIFF
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -176,7 +176,7 @@ class IPEXModel(OptimizedModel):
         config.torchscript = True
         config.torch_dtype = torch_dtype
 
-        return cls(traced_model, config=config, model_save_dir=model_id, use_cache=use_cache)
+        return cls(traced_model, config=config, model_save_dir=model_id, use_cache=use_cache, warmup=False)
 
     @classmethod
     def _from_pretrained(
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -90,6 +90,8 @@ def ipex_jit_trace(model, task, use_cache):
     model.config.return_dict = False
 
     model = ipex.optimize(model.eval(), dtype=model.dtype, inplace=True)
+    # Disable repack while jit tracing to reduce the memory
+    ipex._C.disable_jit_linear_repack()
     with torch.no_grad():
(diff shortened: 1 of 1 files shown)


yao-matrix:
@jiqing-feng , do you observe memory saving with your PR? any before vs after proof data?

jiqing-feng:
> @jiqing-feng , do you observe memory saving with your PR? any before vs after proof data?

Script:
```python
import torch
from optimum.intel.ipex import IPEXModelForCausalLM
model = IPEXModelForCausalLM.from_pretrained("meta-llama/Llama-2-7b-chat-hf", torch_dtype=torch.bfloat16, export=True)
```

Memory size before change: 51.7G
image

Memory size after my change: 13.6G
image

optimum/intel/ipex/modeling_base.py
```diff
@@ -171,23 +173,10 @@ def _from_transformers(
...
-            config=config,
-            use_auth_token=use_auth_token,
-            revision=revision,
-            force_download=force_download,
-            cache_dir=cache_dir,
-            local_files_only=local_files_only,
-            use_cache=use_cache,
-            model_dtype=torch_dtype,
-        )
+        return cls(traced_model, config=config, model_save_dir=model_id, use_cache=use_cache)
```
echarlaix, line 179: shouldn't we set `warmup=False` here as some warmup is already done in `ipex_jit_trace` ?

```suggestion
        return cls(traced_model, config=config, model_save_dir=model_id, use_cache=use_cache, warmup=False)
```

echarlaix (approved):

echarlaix (approved):
Looks great, thanks @jiqing-feng

PR #628, opened 2024-03-22
support more models in openvino export

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


optimum/exporters/openvino/__main__.py
```diff
@@ -50,7 +50,7 @@ def main_export(
     device: str = "cpu",
     framework: Optional[str] = None,
     cache_dir: Optional[str] = None,
-    trust_remote_code: bool = False,
+    trust_remote_code: bool = None,
```
eaidova, line 53:
```suggestion
    trust_remote_code: bool = False,
```

eaidova:
@echarlaix  @AlexKoff88 could you please take a look?

tests/openvino/test_modeling.py
```diff
@@ -521,10 +521,11 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "stablelm",
         "starcoder2",
         "phi",
+        "internlm2",
```
echarlaix, line 524: would be nice to also add :

```suggestion
        "internlm2",
        "decilm",
        "olmo",
        "orion",
```

optimum/exporters/openvino/__main__.py
```diff
@@ -203,6 +203,8 @@ def main_export(
         do_gptq_patching = quantization_config and quantization_config["quant_method"] == "gptq"
         model_type = config.model_type.replace("_", "-")
 
+        if model_type in {"falcon", "mpt"} and trust_remote_code:
+            trust_remote_code = False
```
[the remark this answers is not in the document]
eaidova: mpt and falcon export failed if trust_remote_code used due to different in formats of past key values

tests/openvino/test_modeling.py
eaidova: I plan to add it after spliting tests on nightly to not increase testing time too much

optimum/exporters/openvino/model_configs.py
```diff
@@ -74,7 +75,7 @@ def init_model_configs():
 
 
 @register_in_tasks_manager("baichuan", *["text-generation", "text-generation-with-past"], library_name="transformers")
-class BaichaunOpenVINOConfig(TextDecoderOnnxConfig):
+class BaichaunOpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
```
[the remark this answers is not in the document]
eaidova: due to accuracy issues in some baichuan model versions (previosly I analyzed bauchuan2, but baichuan has the same model_type and difference in position_ids calculation

optimum/exporters/openvino/model_patcher.py
```diff
@@ -513,5 +514,264 @@ def __init__(
...
+    attention_bias: Optional[torch.Tensor] = None,
+    past_key_values: Optional[Tuple[Tuple[torch.Tensor, torch.Tensor]]] = None,
+    use_cache: bool = False,
+    last_logits_only: bool = False,
+    output_hidden_states: Optional[bool] = None,
+):
+    output_hidden_states = output_hidden_states if output_hidden_states is not None else False
+
+    if past_key_values:
+        assert len(past_key_values) == self.config.n_layers
```
echarlaix, line 564: could you raise an error instead (with an explicit message) something like :
```suggestion
        raise ValueError(f"The number of layers and `past_key_values` first dimension expected to be {self.config.n_layers}, but got: {len(past_key_values)} instead.")
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -513,5 +514,264 @@ def __init__(
...
+        for block_idx, block in enumerate(self.transformer.blocks):
+            if output_hidden_states:
+                # add hidden states
+                all_hidden_states.append(x)
+
+            layer_past = None if past_key_values is None else past_key_values[block_idx]
+            # shape: (batch_size, seq_len, d_model)
+            x, cache = block(x, attention_bias=attention_bias, layer_past=layer_past, use_cache=use_cache)
+            if attn_key_values is not None:
+                assert cache is not None
```
echarlaix, line 638: same comment

optimum/exporters/openvino/model_patcher.py
```diff
@@ -513,5 +514,264 @@ def __init__(
...
+            layers_past = (
+                None
+                if past_key_values is None
+                else past_key_values[
+                    group_idx * self.config.block_group_size : (group_idx + 1) * self.config.block_group_size
+                ]
+            )
+            x, cache = block_group(x, attention_bias=attention_bias, layers_past=layers_past, use_cache=use_cache)
+            if attn_key_values is not None:
+                assert cache is not None
```
echarlaix, line 655: same comment

optimum/exporters/openvino/__main__.py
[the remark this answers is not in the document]
echarlaix: > mpt and falcon export failed if trust_remote_code used due to difference in formats of past key values

I imagine it could be the case for all architectures no ? Also is this only for stateful models ? (should work for MPT at least for the case with and without pkv for non stateful models as I see here for ONNX https://github.com/huggingface/optimum/blob/253c6c2ff0f834b6fb03557add6ba4ae1272714d/tests/exporters/onnx/test_onnx_export.py#L570)

optimum/exporters/openvino/model_patcher.py
eaidova: I'm tryed to preserve original code as close as possible change only relevant parts that should be changed. this assert comes from original model code

optimum/exporters/openvino/__main__.py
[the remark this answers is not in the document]
eaidova: I updated condition and added warning for using this models, users still can convert models with trust_remote_code for other tasks or if they provide custom configs

optimum/exporters/openvino/model_patcher.py
eaidova: removed

optimum/exporters/openvino/model_patcher.py
eaidova: removed

tests/openvino/test_modeling.py
eaidova: for some models I have difficulties with creating valid dummy models, do not block merging by this issue, I'll remove these models support

optimum/exporters/openvino/model_patcher.py
eaidova: removed

tests/openvino/test_modeling.py
echarlaix: thanks @eaidova

optimum/exporters/openvino/__main__.py
```diff
@@ -220,6 +219,20 @@ def main_export(
...
+        # there are some difference between remote and in library representation of past key values for some models,
+        # for avoiding confusion we disable remote code for them
+        if (
+            trust_remote_code
+            and model_type in {"falcon", "mpt", "phi"}
+            and ("with-past" in task or original_task == "auto")
+            and not custom_export_configs
+        ):
+            logger.warning(
+                "Model type `{model_type}` export for task `{task}` is not supported for loading with `trust_remote_code=True`"
```
echarlaix, line 231:
```suggestion
                f"Model type `{model_type}` export for task `{task}` is not supported for loading with `trust_remote_code=True`"
```

optimum/exporters/openvino/__main__.py
```diff
@@ -220,6 +219,20 @@ def main_export(
...
+                "using default export configuration, `trust_remote_code` will be disabled. "
+                "Please provide custon export config if you want load model with remote code."
```
echarlaix, line 233:
```suggestion
                "Please provide custom export config if you want load model with remote code."
```

echarlaix (approved):

PR #630, opened 2024-03-22
Transition to a newer NNCF API for PyTorch model quantization

nikita-savelyevv:
Post-Training Quantization for PyTorch backend with `nncf.create_compressed_model()` API is obsolete and should be replaced with `nncf.quantize()` call which is used for OV backend.

# What does this PR do?

Replace `nncf.create_compressed_model()` call with `nncf.quantize()` call for quantization of PyTorch models.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -24,10 +24,9 @@ import nncf
 import openvino
 import torch
 import transformers
-from nncf import CompressWeightsMode, IgnoredScope, NNCFConfig, SensitivityMetric
+from nncf import CompressWeightsMode, IgnoredScope, SensitivityMetric
 from nncf.quantization.advanced_parameters import AdvancedSmoothQuantParameters
-from nncf.torch import create_compressed_model, register_default_init_args, register_module
-from nncf.torch.dynamic_graph.io_handling import wrap_nncf_model_inputs_with_objwalk
+from nncf.torch import register_module
 from nncf.torch.initialization import PTInitializingDataLoader
 from openvino._offline_transformations import compress_quantize_weights_transformation
 from openvino.runtime import Core, Tensor
@@ -47,7 +46,7 @@ from ...exporters.openvino.stateful import ensure_export_task_support_stateful, 
 from ..utils.constant import _TASK_ALIASES
 from ..utils.import_utils import DATASETS_IMPORT_ERROR, is_datasets_available
 from ..utils.modeling_utils import get_model_device
-from .configuration import DEFAULT_QUANTIZATION_CONFIG, OVConfig, OVWeightQuantizationConfig
+from .configuration import OVConfig, OVWeightQuantizationConfig
 from .modeling_base import OVBaseModel
 from .utils import (
     MAX_ONNX_OPSET,
@@ -240,8 +239,6 @@ class OVQuantizer(OptimumQuantizer):
         if ov_config is not None:
             if not isinstance(ov_config, OVConfig):
                 raise TypeError(f"`ov_config` should be an `OVConfig`, but got: {type(ov_config)} instead.")
-            elif ov_config.compression is None:
-                ov_config.compression = DEFAULT_QUANTIZATION_CONFIG
 
         if isinstance(self.model, OVBaseModel):
             self._quantize_ovbasemodel(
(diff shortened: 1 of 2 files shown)


nikita-savelyevv:
@AlexKoff88 could you please review this PR?

AlexKoff88:
@echarlaix, as we agreed, this is the first PR in the series of changes to deprecate legacy NNCF functions and move all the actual optimization functionality under OVQuantizer API. Please take a look.

AlexKoff88 (approved):

optimum/intel/openvino/quantization.py
```diff
@@ -409,25 +407,25 @@ def _quantize_torchmodel(
...
-            nncf_config = register_default_init_args(nncf_config, calibration_dataloader)
-            controller, model = create_compressed_model(
-                model, nncf_config, wrap_inputs_fn=wrap_nncf_model_inputs_with_objwalk
+            quantization_dataset = nncf.Dataset(calibration_dataloader)
+            model = nncf.quantize(
+                model,
+                quantization_dataset,
+                model_type=nncf.ModelType.TRANSFORMER if not kwargs.get("model_type") else kwargs.get("model_type"),
+                fast_bias_correction=kwargs.get("fast_bias_correction", True),
+                **kwargs,
```
echarlaix, line 416: would prefer to have the specifics arguments supported here, also some could be provided through the `ov_config`

optimum/intel/openvino/quantization.py
```diff
@@ -360,7 +358,7 @@ def _quantize_torchmodel(
             logger.info(
                 "No configuration describing the quantization process was provided, a default OVConfig will be generated."
             )
-            ov_config = OVConfig(compression=DEFAULT_QUANTIZATION_CONFIG)
+            ov_config = OVConfig()
```
echarlaix, line 361: not used for quantization (but now only for the `save_onnx_model` parameter) so not sure we need to create an instance when not provided, we can only give `save_onnx_model` a default value instead. Also no need to save the configuration after the quantization + export steps for the same reason  : would remove saving the config here 
https://github.com/huggingface/optimum-intel/blob/c2d267aef2fc2185be8cece275fc27e13f9b3602/optimum/intel/openvino/quantization.py#L444

optimum/intel/openvino/quantization.py
```diff
@@ -409,25 +407,25 @@ def _quantize_torchmodel(
                 data_collator=data_collator,
             )
 
-            model_inputs = next(iter(calibration_dataloader))
-            ov_config.add_input_info(model_inputs)
-            nncf_config = NNCFConfig.from_dict(ov_config.__dict__)
-            nncf_config = register_default_init_args(nncf_config, calibration_dataloader)
-            controller, model = create_compressed_model(
```
echarlaix, line 416: if create_compressed_model is deprecated, we should also update the trainer : 

https://github.com/huggingface/optimum-intel/blob/c2d267aef2fc2185be8cece275fc27e13f9b3602/optimum/intel/openvino/trainer.py#L259

echarlaix (approved):
Looks good thanks @nikita-savelyevv

optimum/intel/openvino/quantization.py
nikita-savelyevv: The current call is aligned with the one at `_quantize_ovbasemodel()`. Call arguments for both backends are planned to be updated in another future PR.

optimum/intel/openvino/quantization.py
nikita-savelyevv: Made the suggested changes, also removed ov_config from list of arguments altogether. In the future will bring it back to pass quantization parameters through it.

optimum/intel/openvino/quantization.py
nikita-savelyevv: This unfortunately can't be replaced, because in the trainer `create_compressed_model()` is used for other compression methods besides quantization. 

As discussed with @AlexKoff88, we plan to keep `create_compressed_model()` in the OVTrainer for now. But to refactor the way config is provided to it.

echarlaix: got it !

echarlaix (approved):

PR #632, opened 2024-03-25
Add tensor caching option to InferRequestWrapper

nikita-savelyevv:
# What does this PR do?

**Problem**
`InferRequestWrapper` object is imported and used in whisper notebook at [openvino_notebooks](https://github.com/openvinotoolkit/openvino_notebooks/tree/master/notebooks/distil-whisper-asr).

After PR #577 was merged, the memory requirements for quantizing whisper large-v2 model increased dramatically. This leads to OOM when running quantization on a decent sized calibration dataset. For example, I get OOM on a dataset of 30 calibration audio files on machine with 125 GB.

The reason is that the whisper `decoder_with_past` model has KV input for each decoder block. There are 32 blocks, so 64 inputs in total. All these inputs accept tensors of at least 12 x 512 dimension. Also, since single audio input generates tens of decoder inputs the factor in increased even more. 

**Solution**
For that particular use case input tensors overlap between neighboring samples, so I added a tensor cache to `InferRequestWrapper` object. This significantly reduces memory overhead, but results in slight performance overhead due to tensor hash computation. 

I disabled this option by default and now it's only used in tests. This is because of the reason above and the fact that it's not actually clear whether this has any benefits for the models supported for quantization by optimum.  Please note, that quantization of Seq2Seq models for which this fix is added is not yet supported, but rather `InferRequestWrapper` is imported elsewhere for manual quantization. Despite this, I still think that this functionality may become useful in the future, for example when we support quantization of Seq2Seq models directly through optimum.

I also added docs for `InferRequestWrapper` class.

**Related ticket**
136705

## Before submitting

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -18,7 +18,7 @@ import logging
 import os
 from collections import deque
 from pathlib import Path
-from typing import Any, Callable, Dict, Optional, Tuple, Union, List
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import nncf
 import openvino
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -86,10 +86,13 @@ class InferRequestWrapper:
     Wrapper class for OV InferRequest or CompiledModel objects that collects inputs which they were called with to
     a list.
     """
-    def __init__(self,
-                 request: Union[openvino.InferRequest, openvino.CompiledModel],
-                 collected_inputs: List = None,
-                 apply_caching: bool = False):
+
+    def __init__(
+        self,
+        request: Union[openvino.InferRequest, openvino.CompiledModel],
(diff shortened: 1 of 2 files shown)


nikita-savelyevv:

optimum/intel/openvino/quantization.py
```diff
@@ -82,21 +82,62 @@ def batch_size(self):
 
 
 class InferRequestWrapper:
-    def __init__(self, request, data_cache=None):
+    """
+    Wrapper class for OV InferRequest or CompiledModel objects that collects inputs which they were called with to
```
AlexKoff88, line 86: We introduced .prepare_inputs() method for decoder models some time ago that can be used to collect inputs (and it is used actually). @nikita-savelyevv, can you please check if we still need this w/a with patching InferRequest?

AlexKoff88 (approved):

echarlaix (approved):

PR #634, opened 2024-03-27
Add support of OVModelForVision2Seq class

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -235,6 +253,50 @@
     ```
 """
 
+IMAGE_TO_TEXT_EXAMPLE = r"""
+    Example of text generation:
+
+    ```python
+    >>> from transformers import {processor_class}, {tokenizer_class}
+    >>> from optimum.onnxruntime import {model_class}
```
AlexKoff88, line 261:
```suggestion
    >>> from optimum.intel import {model_class}
```

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -235,6 +253,50 @@
...
+    >>> gen_tokens = model.generate(**inputs)
+    >>> outputs = tokenizer.batch_decode(gen_tokens, skip_special_tokens=True)
+
+    ```
+
+    Example using `transformers.pipeline`:
+
+    ```python
+    >>> from transformers import {processor_class}, {tokenizer_class}, pipeline
+    >>> from optimum.onnxruntime import {model_class}
```
AlexKoff88, line 283:
```suggestion
    >>> from optimum.intel import {model_class}
```

AlexKoff88 (approved):

optimum/intel/__init__.py
```diff
@@ -121,6 +121,7 @@
             "OVModelForQuestionAnswering",
             "OVModelForSeq2SeqLM",
             "OVModelForSpeechSeq2Seq",
+            "OVModelForVision2Seq",
```
echarlaix, line 124: can you add `OVModelForVision2Seq` here as well https://github.com/huggingface/optimum-intel/blob/a3bf1727b78056283a4ad0cf77b6a0ff7519aeea/optimum/intel/__init__.py#L242

tests/openvino/test_modeling.py
```diff
@@ -1402,3 +1405,102 @@ def test_pipeline(self, model_arch):
...
+            self.assertTrue(torch.allclose(torch.Tensor(ov_outputs.logits), transformers_outputs.logits, atol=1e-3))
+
+        gc.collect()
+
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    def test_pipeline_image_to_text(self, model_arch: str):
+        model_id = MODEL_NAMES[model_arch]
+        ov_model = OVModelForVision2Seq.from_pretrained(model_id, export=True)
+        feature_extractor, tokenizer = self._get_preprocessors(model_id)
+
```
echarlaix, line 1493: could you add a `model.reshape(..)` in one test to make sure it works

echarlaix (approved):
Looks great, thanks for the addition @eaidova !

optimum/intel/__init__.py
eaidova: added

tests/openvino/test_modeling.py
eaidova: added

tests/openvino/test_modeling.py
```diff
@@ -1402,3 +1405,113 @@ def test_pipeline(self, model_arch):
...
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    def test_model_is_reshapable(self, model_arch: str):
+        model_id = MODEL_NAMES[model_arch]
+        ov_model = OVModelForVision2Seq.from_pretrained(model_id, export=True, compile=False)
+        ov_model.reshape(5, -1)
+        ov_model.compile()
+        models = [ov_model.encoder.model, ov_model.decoder.model, ov_model.decoder_with_past.model]
+        for model in models:
+            for inp_tensor in model.inputs:
+                inp_tensor.partial_shape[0] == 5
```
echarlaix, line 1457: I was thinking about including it in a test like `test_pipeline_image_to_text` to check that it's compatible with inference but we can have the additional test that you added. To not increase too much the time of the github action workflows on push with additional test like `test_model_is_reshapable` I would be in favor of running them nightly instead like https://github.com/huggingface/optimum/blob/dedb852eaea3d3d899920f4017b9fd899eb2756b/tests/exporters/onnx/test_onnx_export.py#L334 and https://github.com/huggingface/optimum/blob/dedb852eaea3d3d899920f4017b9fd899eb2756b/.github/workflows/test_exporters_slow.yml#L33C48-L33C63

optimum/intel/__init__.py
echarlaix: not updated on my side!

tests/openvino/test_modeling.py
eaidova: I do not think that this PR is right place for making such infra changes (because it requires at least make some estimation about which tests should be run on a nightly basis instead of precommit), I also does not think that this specific test makes big impact on testing performance, but if you want I can remove it and add reshape into pipeline tests

optimum/intel/__init__.py
eaidova: please chack one more time

tests/openvino/test_modeling.py
echarlaix: works for me!

tests/openvino/utils_tests.py
```diff
@@ -70,7 +71,7 @@
     "nystromformer": "hf-internal-testing/tiny-random-NystromformerModel",
     "pegasus": "hf-internal-testing/tiny-random-pegasus",
     "pix2struct": "fxmarty/pix2struct-tiny-random",
-    "phi": "hf-internal-testing/tiny-random-PhiForCausalLM",
+    "phi": "Xenova/tiny-random-PhiForCausalLM",
```
eaidova, line 74:
```suggestion
    "phi": "hf-internal-testing/tiny-random-PhiForCausalLM",
```

echarlaix:
failing tests unrelated so merging

PR #638, opened 2024-03-27
Introduce OVQuantizationConfig for nncf.quantize() parameters

nikita-savelyevv:
# What does this PR do?

- Existing `OVWeightQuantizationConfig` is now intended only for weight compression and the new `OVQuantizationConfig` is used for full quantization.
- Parameters for inner `nncf.quantize()` call should from now on be given by providing an instance of `OVQuantizationConfig` to `OVQuantizer.quantize()` via `OVConfig`.
- The config objects are serializable, but the saved quantization config is intended only for information purposes. It can't be loaded and used for quantization again. This is because it may contain some custom objects, like instances of `nncf.Dataset` or `transformers.PreTrainedTokenizer`, which can't be serialized so they are omitted during serialization.
- The previously used `DEFAULT_QUANTIZATION_CONFIG ` is moved to `trainer.py` where its only usage currently is.
- Added a test for quantization config serialization. Updated all other quantization tests to be aligned with the new format.

## Before submitting


optimum/intel/openvino/trainer.py
```diff
@@ -136,6 +136,25 @@
 NNCF_LOG_FILE_NAME = "nncf_output.log"
 
 
+DEFAULT_QUANTIZATION_CONFIG = {
+    "algorithm": "quantization",
```
AlexKoff88, line 140: @echarlaix, the idea is to keep the config inside the example for a while. Once we are ready with a "smooth" transition from PTQ to QAT in NNCF we will revise it as well.

echarlaix: ok you mean keeping the examples https://github.com/huggingface/optimum-intel/blob/v1.16.0/examples/openvino/text-classification/run_glue.py#L539 compatible with the `compression` argument for some time?

echarlaix: also let me know when the PR is ready for review @nikita-savelyevv

optimum/intel/openvino/quantization.py
```diff
@@ -691,23 +675,26 @@ def _hybrid_quantization(
...
         model=compressed_model,
         calibration_dataset=nncf.Dataset(dataset),
         model_type=nncf.ModelType.TRANSFORMER,
         ignored_scope=ptq_ignored_scope,
-        # The SQ algo should be disabled for MatMul nodes because their weights are already compressed
-        advanced_parameters=nncf.AdvancedQuantizationParameters(AdvancedSmoothQuantParameters(matmul=-1)),
+        # SQ algo should be disabled for MatMul nodes because their weights are already compressed
+        advanced_parameters=nncf.AdvancedQuantizationParameters(
+            smooth_quant_alphas=AdvancedSmoothQuantParameters(matmul=-1)
+        ),
```
nikita-savelyevv, line 697: There was a small bug here

tests/openvino/test_quantization.py
```diff
@@ -110,6 +118,10 @@ def preprocess_function(examples, tokenizer):
             outputs = model(**tokens)
             self.assertTrue("logits" in outputs)
 
+            # Verify that the configuration is correctly saved and loaded
+            loaded_config = OVConfig.from_pretrained(tmp_dir)
+            self.assertEqual(ov_config.quantization_config.to_dict(), loaded_config.quantization_config)
+
```
nikita-savelyevv, line 124: Brought back after #630

nikita-savelyevv:
@echarlaix @AlexKoff88 PR is ready for review

AlexKoff88 (approved):
I mostly agree with the changes. @echarlaix, please take a look and merge if it is ok.

optimum/intel/openvino/configuration.py
```diff
@@ -137,38 +180,38 @@ def add_input_info(self, model_inputs: Dict, force_batch_one: bool = False):
...
-            isinstance(self.compression, dict)
-            and "algorithm" in self.compression
-            and self.compression["algorithm"] == "quantization"
-        ):
-            self.compression["export_to_onnx_standard_ops"] = self.save_onnx_model
-        elif isinstance(self.compression, list):
-            for i, algo_config in enumerate(self.compression):
-                if algo_config["algorithm"] == "quantization":
-                    self.compression[i]["export_to_onnx_standard_ops"] = self.save_onnx_model
+    def to_dict_safe(self, to_diff_dict: bool = False) -> Dict[str, Any]:
```
echarlaix, line 183: (same comment everywhere)
```suggestion
    def _to_dict_safe(self, to_diff_dict: bool = False) -> Dict[str, Any]:
```

optimum/intel/openvino/configuration.py
```diff
@@ -100,31 +52,122 @@
...
-        if isinstance(quantization_config, QuantizationConfigMixin):
-            bits = self.quantization_config.bits
-        else:
-            bits = self.quantization_config.get("bits", None)
+        if isinstance(self.quantization_config, dict):
+            # Config is loaded as dict during deserialization
+            logger.info(
+                "`quantization_config` was provided as a dict, in this form it can't be used for quantization. "
+                "Please provide config as an instance of OVWeightQuantizationConfig or OVQuantizationConfig"
+            )
```
echarlaix, line 166: why is it disabled for dict ?

optimum/intel/openvino/quantization.py
```diff
@@ -203,37 +203,31 @@ def from_pretrained(cls, model: PreTrainedModel, **kwargs):
...
                 The directory where the quantized model should be saved.
-            quantization_config (`OVConfig`, *optional*):
-                The configuration containing the parameters related to quantization.
+            ov_config (`OVConfig`, *optional*):
+                The configuration containing the parameters related to quantization. If not provided, 8-bit symmetric
+                weight-only quantization will be applied.
             file_name (`str`, *optional*):
                 The model file name to use when saving the model. Overwrites the default file name `"model.onnx"`.
-            batch_size (`int`, defaults to 8):
+            batch_size (`int`, defaults to 1):
```
echarlaix, line 225: nice catch !

optimum/intel/openvino/quantization.py
```diff
@@ -252,45 +247,39 @@ def quantize(
...
         ```
         """
+        if "calibration_dataset" in kwargs:
+                "`calibration_dataset` argument is deprecated. Please provide calibration dataset "
+                "with `ov_config.quantization_config.dataset`."
+            )
+        if "weights_only" in kwargs:
+                "`weights_only` argument is deprecated. Please provide `ov_config.quantization_config` "
```
echarlaix, line 262: might make sense to leave at least one or two minor release to warn people about this before deprecating it, what do you think ?

echarlaix (commented):
looks great, thanks @nikita-savelyevv

optimum/intel/openvino/configuration.py
```diff
@@ -100,31 +52,122 @@
...
+        """
+        if len(property_names) == 0:
+            return super().to_dict()
+        with replace_properties_values(self, property_names, [None] * len(property_names)):
+            result = super().to_dict()
+        return result
+
+    def to_dict(self) -> Dict[str, Any]:
+        properties_to_omit = [] if is_serializable(self.dataset) else ["dataset"]
+        if isinstance(self.ignored_scope, nncf.IgnoredScope):
```
echarlaix, line 136: do you think we need it in the configuration (we could create an instance of `IgnoredScope` directly in the quantizer)

optimum/intel/openvino/configuration.py
nikita-savelyevv: Mainly, because of the `dataset` property of the quantization config. It can't be serialized in general scenario. But if we move it to a separate parameter of `.quantize()` method as you have suggested, then I believe it will be possible.

optimum/intel/openvino/configuration.py
```diff
@@ -179,62 +222,64 @@ class OVWeightQuantizationConfig(QuantizationConfigMixin):
...
         self.tokenizer = tokenizer
-        self.dataset = dataset
         self.group_size = group_size or (-1 if bits == 8 else 128)
         self.ratio = ratio
         self.all_layers = all_layers
         self.sensitivity_metric = sensitivity_metric
-        self.ignored_scope = ignored_scope
-        self.num_samples = num_samples
-        self.quant_method = "default"  # TODO : enable AWQ after nncf v2.9.0 release
+        self.subset_size = subset_size
```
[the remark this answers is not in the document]
nikita-savelyevv: Done

optimum/intel/openvino/configuration.py
```diff
@@ -100,31 +52,122 @@
...
+                "Dataset must be a instance of either string, list of strings, nncf.Dataset or "
+                f"dataset.Dataset, but found {type(self.dataset)}"
+            )
+        if not (self.ignored_scope is None or isinstance(self.ignored_scope, nncf.IgnoredScope)):
+            raise ValueError(
+                "Ignored scope must be a instance of either dict, or nncf.IgnoredScope but found "
+                f"{type(self.dataset)}"
+            )
+
+    def to_dict_without_properties(self, property_names: Union[List[str], Tuple[str]]) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
nikita-savelyevv: Thanks, renamed relevant names

optimum/intel/openvino/quantization.py
nikita-savelyevv: Replaced raising `ValueError` by giving warning. This should be enough since if `ov_config.quantization_config` is omitted, weight compression is applied.

optimum/intel/openvino/quantization.py
```diff
@@ -252,45 +247,39 @@ def quantize(
...
         >>> quantizer = OVQuantizer.from_pretrained(model, task="text-generation")
-        >>> quantizer.quantize(save_directory="./quantized_model", weights_only=True)
+        >>> ov_config = OVConfig(quantization_config=OVWeightQuantizationConfig(bits=8, sym=True))
+        >>> quantizer.quantize(ov_config=ov_config, save_directory="./quantized_model")
         >>> optimized_model = OVModelForCausalLM.from_pretrained("./quantized_model")
         ```
         """
+        if "calibration_dataset" in kwargs:
+                "`calibration_dataset` argument is deprecated. Please provide calibration dataset "
```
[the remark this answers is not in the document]
nikita-savelyevv: Done, as you suggested

optimum/intel/openvino/configuration.py
nikita-savelyevv: Agree, done

nikita-savelyevv:
@echarlaix please review after addressed changes

echarlaix (approved):
Looks great, thanks @nikita-savelyevv

optimum/intel/openvino/configuration.py
```diff
@@ -259,6 +302,86 @@ def post_init(self):
...
+            overflow_fix (`nncf.OverflowFix`, default to OverflowFix.DISABLE):
+                Parameter for controlling overflow fix setting.
+            weight_only (`bool`, *optional*):
+                Used to explicitly specify type of quantization (weight-only of full) to apply.
+        """
+        if weight_only is True:
+            logger.warning(
+                "Trying to create an instance of `OVQuantizationConfig` with `weight_only` being True. "
+                "Please check your configuration."
+            )
```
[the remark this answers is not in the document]
nikita-savelyevv: It's just to make sure that nobody creates the config with the wrong `weight_only` property, e.g. `OVWeightQuantizationConfig(weight_only=False)`. Wrong property in turn will impact deserialization of quantization config: https://github.com/nikita-savelyevv/optimum-intel/blob/introduce-ov-quantization-config/optimum/intel/openvino/configuration.py#L152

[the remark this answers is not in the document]
nikita-savelyevv: I remembered the reason this is needed. In the scenario when someone provides quantization config as dictionary instead of a Config object, there is some logic that tries to infer the type of the config based on the provided keys. However, this is not 100% reliable, so there's an additional flag to distinguish between weight-only and full quantization.

For example, if config is given as `dict(ignored_scope={"names": ["op_name"]})` than it's ambiguous which kind of quantization is intended because both kinds accept an ignored scope parameter. By default in such case `OVWeightQuantizationConfig` will be created and a warning [will be given](https://github.com/nikita-savelyevv/optimum-intel/blob/introduce-ov-quantization-config/optimum/intel/openvino/configuration.py#L146). 

To run full quantization with only ignored scope given, it's required to provide config as `dict(weight_only=False, ignored_scope={"names": ["op_name"]})`. In such case an instance of `OVQuantizationConfig` will be built and full quantization is run.

The warning you mention here happens when for example the config is given as `dict(bits=8, sym=True, weight_only=False)` which is confusing since parameters for weight only quantization are given, but `weight_only` hints at that the full quantization is intended. In such case, weight only quantization will be executed nevertheless, but the warning is given to avoid such confusion.

optimum/intel/openvino/modeling_base.py
```diff
@@ -18,6 +18,7 @@
 from tempfile import TemporaryDirectory, gettempdir
 from typing import Dict, Optional, Union
 
+import nncf
```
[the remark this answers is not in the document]
nikita-savelyevv: Removed

optimum/intel/openvino/configuration.py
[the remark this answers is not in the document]
echarlaix: I see, I think it would make sense to always create an instance of `OVWeightQuantizationConfig` by default and remove this parameter

PR #646, opened 2024-04-02
add protein preediction task for ipex

jiqing-feng:
Hi @echarlaix . I want to add an example of protein prediction to show users how to apply ipex optimization on a loaded remote model. I choose [protst model](https://huggingface.co/mila-intel/ProtST-esm1b), which is a SOTA model for protein prediction. Would you please help to review and merge it? Thx!

### DIFF
--- a/notebooks/ipex/protein_prediction.ipynb
+++ b/notebooks/ipex/protein_prediction.ipynb
@@ -1,0 +1,257 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Use ipex optimization on a remote model"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


IlyasMoutawwakil:
Hi @jiqing-feng !
The example seems to introduce a lot of task and model implementation specific code for which the demand is low (45 downloads in a month). 
To demonstrate that ipex works on arbitrary remote modeling code, it would be better to target an architecture which is not natively integrated in `transformers` but has a lot of demand like ChatGLM for example. 
I think It would also make more sense to add this example in ipex examples https://github.com/intel/intel-extension-for-pytorch/tree/main/examples/cpu/inference/python since the only use of optimum-intel here is [`jit_trace`](https://github.com/huggingface/optimum-intel/blob/972491991710f8a92cdef35e0914de92a88995a4/optimum/intel/generation/modeling.py#L67) which is just a utility function.

echarlaix:
Agreed with @IlyasMoutawwakil, I don't think `optimum-intel` is the most adapted place for this example

jiqing-feng:
Thanks, I will try another way : )

PR #647, opened 2024-04-04
Add notebook for local RAG

ofirzaf:
# What does this PR do?

In this notebook we give an example how to run RAG locally on your laptop using OpenVINO/LangChain/Phi-2

@echarlaix 

## Before submitting


ofirzaf:
Hi @echarlaix , the PR is ready for review, will appreciate your comments

IlyasMoutawwakil:
Hi @ofirzaf very cool app !
I don't think this should be in the notebooks examples since it's mostly ChromaDB+LangCha code, which is still unstable and will be difficult to maintain. It makes more sense to demonstrate complicated applications like these on a space in `intel` or `optimum-intel` huggingface organizations and/or with community blog post.

PR #648, opened 2024-04-04
Added passing of CI var to docker.

popovaan:
# What does this PR do?

Pass CI variable to docker container.

Fixes # 

OpenVino may send telemetry data. To disable telemetry sending in CI "CI" variable is used, that is set in most CI pipelines.
Docker does not pass outer env variables to container, so "CI" var is needed to be passed explicitly to prevent telemetry sending from docker container. 

## Before submitting


eaidova (approved):

eaidova:

PR #650, opened 2024-04-07
indicate the correct ipex model class in testing

jiqing-feng:
Hi @echarlaix . It's just a minor change to indicate the correct ipex model class in the testing. Fixed the wrong match of the token classification task and sequence classification task.


echarlaix (approved):
Thanks for the fix @jiqing-feng !

PR #652, opened 2024-04-08
Replace deprecated transformers is_torch_tpu_available

echarlaix:
`is_torch_tpu_available` will be officially deprecated in `transformers` v4.41.0

https://github.com/huggingface/transformers/blob/f5658732d5bd38657b6cf3138f69373a94866c61/src/transformers/utils/import_utils.py#L521

### DIFF
--- a/optimum/intel/neural_compressor/trainer.py
+++ b/optimum/intel/neural_compressor/trainer.py
@@ -62,7 +62,7 @@ from transformers.utils import (
     is_accelerate_available,
     is_apex_available,
     is_sagemaker_mp_enabled,
-    is_torch_tpu_available,
+    is_torch_xla_available,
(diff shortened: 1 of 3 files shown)


eaidova:
@echarlaix it looks like, these changes breaks backward compatibility with transformers 4.37
here no this imported function
https://github.com/huggingface/transformers/blob/v4.37.2/src/transformers/utils/import_utils.py

echarlaix:
> @echarlaix it looks like, these changes breaks backward compatibility with transformers 4.37 here no this imported function https://github.com/huggingface/transformers/blob/v4.37.2/src/transformers/utils/import_utils.py

Thanks for reporting @eaidova, will open a PR to fix this!

PR #654, opened 2024-04-08
fix causal mask update bf16 accuracy issue in gemma

eaidova:
# What does this PR do?
gemma and llama may fail with RuntimeError: probability tensor contains either `inf`, `nan` or element < 0 during generation on some platforms due to using constant that can not be fit into bf16 and fp16 dtype and represented in these precisions as -inf. Provided patch for avoiding this issue

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -288,19 +288,24 @@ class ChatGLMModelPatcher(DecoderModelPatcher):
             block.self_attention.core_attention.forward = block.self_attention.core_attention._orig_forward
 
 
-def _gemma_update_causal_mask(self, attention_mask, input_tensor, cache_position):
+# adopted from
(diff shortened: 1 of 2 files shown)


eaidova:
@echarlaix could you please take a look? It will be very appreciated if we can merge it ASAP

optimum/exporters/openvino/model_patcher.py
```diff
@@ -288,10 +288,74 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        causal_mask = AttentionMaskConverter._unmask_unattended(causal_mask, min_dtype)
+
+    return causal_mask
+
+
 class GemmaModelPatcher(DecoderModelPatcher):
     def __enter__(self):
         super().__enter__()
 
+        # gemma has some accuracy issues with bf16 with transformers >= 4.39
```
echarlaix, line 351: could make sense for `GemmaModelPatcher` to inherit from `LlamaModelPatcher` to avoid code redundancy

echarlaix (approved):
LGTM, thanks for the fix @eaidova

PR #655, opened 2024-04-09
[Gemma][Llama] fix bf16 min_dtype

usstq:
# What does this PR do?

`min_dtype` of torch.float32 type in Llama/Gemma's _update_causal_mask() will converted into bf16's `-infinity` by CPU plugin (due to design of https://www.felixcloutier.com/x86/vcvtneps2bf16 instruction).

this PR fixes it by using `torch.finfo(torch.bfloat16).min` to replace `torch.finfo(torch.float32).min`, they are very close in magnitude but won't cause `-infinity` issue.

```python
>>> torch.finfo(torch.float).min
-3.4028234663852886e+38
>>> torch.finfo(torch.bfloat16).min
-3.3895313892515355e+38
```

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -287,6 +287,72 @@ class ChatGLMModelPatcher(DecoderModelPatcher):
         for block in self._model.transformer.encoder.layers:
             block.self_attention.core_attention.forward = block.self_attention.core_attention._orig_forward
 
+# transformers/models/llama/modeling_llama.py
+# transformers/models/gemma/modeling_gemma.py
+def _update_causal_mask(self, attention_mask, input_tensor, cache_position):
+    if self.config._attn_implementation == "flash_attention_2":
+        if attention_mask is not None and 0.0 in attention_mask:
+            return attention_mask
+        return None
+
+    dtype, device = input_tensor.dtype, input_tensor.device
+    min_dtype = torch.finfo(dtype).min
+
+    # workaround CVS-137270
+    if dtype == torch.float32 or dtype == torch.float:
+        print("use torch.finfo(torch.bfloat16).min as WA for issue CVS-137270")
+        min_dtype = torch.finfo(torch.bfloat16).min
+
+    sequence_length = input_tensor.shape[1]
+    if hasattr(getattr(self.layers[0], "self_attn", {}), "past_key_value"):  # static cache
+        target_length = self.config.max_position_embeddings
+    else:  # dynamic cache
+        target_length = (
+            attention_mask.shape[-1] if isinstance(attention_mask, torch.Tensor) else cache_position[-1] + 1
+        )
+
+    causal_mask = torch.full((sequence_length, target_length), fill_value=min_dtype, dtype=dtype, device=device)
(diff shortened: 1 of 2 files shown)


eaidova:
@usstq , thanks, I prepared similar changes with only difference that it uses fp16 as input dtype as minimal for all supported devices #654

optimum/exporters/openvino/model_patcher.py
```diff
@@ -288,6 +289,73 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+def _update_causal_mask(self, attention_mask, input_tensor, cache_position):
+    if self.config._attn_implementation == "flash_attention_2":
+        if attention_mask is not None and 0.0 in attention_mask:
+            return attention_mask
+        return None
+
+    dtype, device = input_tensor.dtype, input_tensor.device
+    min_dtype = torch.finfo(dtype).min
+
+    # workaround CVS-137270
```
eaidova, line 303: I'm not sure that it is good idea to mention openvino jira ticket in external repo

optimum/exporters/openvino/model_patcher.py
```diff
@@ -301,6 +369,27 @@ def __enter__(self):
                     rotary_emb.base ** (torch.arange(0, rotary_emb.dim, 2, dtype=torch.int64).float() / rotary_emb.dim)
                 )
 
+        self._model.model.original_update_causal_mask = self._model.model._update_causal_mask
+        self._model.model._update_causal_mask = types.MethodType(_update_causal_mask, self._model.model)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+        self._model.model._update_causal_mask = self._model.model.original_update_causal_mask
+        del self._model.model.original_update_causal_mask
```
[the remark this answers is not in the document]
usstq: changed!

optimum/exporters/openvino/model_patcher.py
usstq: removed!

optimum/exporters/openvino/model_patcher.py
```diff
@@ -288,7 +289,82 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    if self.config._attn_implementation == "flash_attention_2":
+        if attention_mask is not None and 0.0 in attention_mask:
+            return attention_mask
+        return None
+
+    dtype, device = input_tensor.dtype, input_tensor.device
+    min_dtype = torch.finfo(dtype).min
+
+    if dtype == torch.float32 or dtype == torch.float:
+        min_dtype = torch.finfo(torch.bfloat16).min
```
eaidova, line 304: what do you think about using torch.finfo(torch.float16) instead? We tested yesterday and sow that it also works without accuracy issue. gemma has 2b variant that can be fit to arm cpu with fp16, I think it also can get the same overflow on this platform (just for avoiding fixing that twice)

eaidova (approved):

optimum/exporters/openvino/model_patcher.py
usstq: Good point, fp16 should be better!

PR #656, opened 2024-04-09
Update pip install command in docs

helena-intel:
Just adding `--upgrade-strategy eager` by itself doesn't upgrade, it's necessary to also add `--upgrade`.


echarlaix (commented):
Thanks @helena-intel !

PR #664, opened 2024-04-16
Remove check as nncf now a required dependency

echarlaix:
nncf added in openvino extra in  https://github.com/huggingface/optimum-intel/pull/586 

cc @helena-intel @eaidova


helena-intel:
NNCF is a standard dependency in the [openvino] extra now, but from my perspective that doesn't make it a "required dependency". You can still use optimum-intel without NNCF. It's a pretty common use-case for me to convert models offline and then in the inference environment just do `pip install optimum-intel openvino`.

echarlaix:
> NNCF is a standard dependency in the [openvino] extra now, but from my perspective that doesn't make it a "required dependency". You can still use optimum-intel without NNCF. It's a pretty common use-case for me to convert models offline and then in the inference environment just do `pip install optimum-intel openvino`.
> 
> [EDIT] I see that with optimum-intel from source NNCF is already required just for importing . That is unexpected to me. I mentioned in the linked PR too that there was still a method to install without NNCF.

Yes it's the case since https://github.com/huggingface/optimum-intel/pull/638, added a fix in https://github.com/huggingface/optimum-intel/pull/668 in case we don't want to merge this PR (and make `nncf` officially a required dependency)

PR #658, opened 2024-04-11
Fix compatibility for latest itrex version

echarlaix:
cc @PenghuiCheng @changwangss

PR #666, opened 2024-04-17
Fixed load issue and update docs for weight-only quantization with intel-extension-for-transformers

PenghuiCheng:
# What does this PR do?
This PR fixed load issue for weight-only quantized model and update documents for weight-only quantization with intel-extension-for-transformers.
This PR dependence on https://github.com/huggingface/optimum-intel/pull/658.

## Before submitting

### DIFF
--- a/docs/source/optimization_inc.mdx
+++ b/docs/source/optimization_inc.mdx
@@ -126,6 +126,33 @@ mpirun -np <number_of_processes> <RUN_CMD>
 
 Please refer to INC [documentation](https://github.com/intel/neural-compressor/blob/master/docs/source/tuning_strategies.md#distributed-tuning) and [text-classification](https://github.com/huggingface/optimum-intel/tree/main/examples/neural_compressor/text-classification) example for more details. 
 
+## Weight-only quantization
+As large language models (LLMs) become more prevalent, there is a growing need for new and improved quantization methods that can meet the computational demands of these modern architectures while maintaining the accuracy. Compared to normal quantization like W8A8, weight only quantization is probably a better trade-off to balance the performance and the accuracy. Up to now, we support "GPTQ" and "RTN" method. 
+
+```python
+from intel_extension_for_transformers.transformers.utils.config import GPTQConfig, RtnConfig
+# for GPTQ method
+quantization_config = GPTQConfig(
+                          damp_percent=0.01,
+                          weight_dtype="int4_clip",
+                      )
+
+# for RTN method
+quantization_config = RtnConfig(
+                          weight_dtype="int4_clip",
+                      )
+quantizer = INCQuantizer.from_pretrained(model)
+quantizer.quantize(
+    quantization_config=quantization_config,
+    save_directory="output_dir",
+    calibration_dataset=(
+        train_dataset if quantization_config.quant_metod == "gptq" else None
+    ),
+)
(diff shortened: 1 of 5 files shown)


examples/neural_compressor/text-generation/run_generation.py
```diff
@@ -281,6 +289,69 @@ def main():
     )
     parser.add_argument("--dataset_name", nargs="?", default="NeelNanda/pile-10k", const="NeelNanda/pile-10k")
     parser.add_argument("--calib_iters", default=100, type=int, help="calibration iters.")
+    parser.add_argument(
```
echarlaix, line 292: I would prefer to keep this example for post-training as ITREX is currently not a required dependency, what do you think about adding this example directly to https://github.com/intel/intel-extension-for-transformers/tree/main/examples/huggingface/pytorch and add a link to these examples in the README? For example I see https://github.com/intel/intel-extension-for-transformers/blob/main/examples/huggingface/pytorch/text-generation/quantization/run_generation.py

docs/source/optimization_inc.mdx
```diff
@@ -126,6 +126,33 @@ mpirun -np <number_of_processes> <RUN_CMD>
 
 Please refer to INC [documentation](https://github.com/intel/neural-compressor/blob/master/docs/source/tuning_strategies.md#distributed-tuning) and [text-classification](https://github.com/huggingface/optimum-intel/tree/main/examples/neural_compressor/text-classification) example for more details. 
 
+## Weight-only quantization
```
echarlaix, line 129: Let's wait for this feature to be more stable before adding it to the documentation (currently not compatible with optimum-intel latest release)

optimum/intel/neural_compressor/quantization.py
```diff
@@ -75,10 +75,10 @@
 INTEL_EXTENSION_FOR_TRANSFORMERS_MINIMUM_VERSION = "1.4.0"
 
 if is_intel_extension_for_transformers_available():
-    if is_intel_extension_for_transformers_version("!=", INTEL_EXTENSION_FOR_TRANSFORMERS_MINIMUM_VERSION):
+    if is_intel_extension_for_transformers_version("<", INTEL_EXTENSION_FOR_TRANSFORMERS_MINIMUM_VERSION):
         raise ImportError(
             f"Found an incompatible version of `intel-extension-for-transformers`. Found version {_intel_extension_for_transformers_version}, "
-            f"but only version {INTEL_EXTENSION_FOR_TRANSFORMERS_MINIMUM_VERSION} is supported."
+            f"but only version {INTEL_EXTENSION_FOR_TRANSFORMERS_MINIMUM_VERSION} or higher is supported."
```
echarlaix, line 81: I think it makes sense to fix the ITREX version at the moment as it will avoid any undesired impact resulting from potential breaking changes from ITREX (as it was the case for ITREX `v1.3.0` -> `v1.4.0`)

optimum/intel/neural_compressor/quantization.py
```diff
@@ -297,6 +297,7 @@ def quantize(
             )
 
             self._quantized_model.quantization_config = quantization_config
+            self._quantized_model.config.quantization_config = quantization_config
```
echarlaix, line 300: I think we should keep the model configs separated from the quantization config

PR #668, opened 2024-04-17
Remove nncf dependency from openvino configs

echarlaix:
Fixes https://github.com/huggingface/optimum-intel/actions/workflows/test_openvino_basic.yml

cc @helena-intel @nikita-savelyevv

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -16,16 +16,18 @@ import inspect
 import logging
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
-import nncf
 import torch
-from nncf.quantization.advanced_parameters import OverflowFix
 from transformers import PretrainedConfig
 from transformers.utils.quantization_config import QuantizationConfigMixin, QuantizationMethod
 
 from optimum.configuration_utils import BaseConfig
 
+from ..utils.import_utils import is_nncf_available
+
+if is_nncf_available():
+    import nncf
 
 logger = logging.getLogger(__name__)
 
@@ -91,7 +93,7 @@ class OVQuantizationConfigBase(QuantizationConfigMixin):
         if not (self.num_samples is None or isinstance(self.num_samples, int) and self.num_samples > 0):
             raise ValueError(f"`num_samples` is expected to be a positive integer, but found: {self.num_samples}")
 
-    def get_ignored_scope_instance(self) -> nncf.IgnoredScope:
+    def get_ignored_scope_instance(self) -> "nncf.IgnoredScope":
         if self.ignored_scope is None:
             return nncf.IgnoredScope()
         return nncf.IgnoredScope(**copy.deepcopy(self.ignored_scope))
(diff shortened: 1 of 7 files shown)


optimum/intel/openvino/configuration.py
```diff
@@ -63,6 +70,7 @@ def __init__(
         ignored_scope: Optional[dict] = None,
         num_samples: Optional[int] = None,
         weight_only: Optional[bool] = None,
+        quant_method: Union[QuantizationMethod, OVQuantizationMethod] = OVQuantizationMethod.DEFAULT,
```
nikita-savelyevv, line 73: Why is this moved here?

optimum/intel/openvino/configuration.py
```diff
@@ -240,7 +245,7 @@ def __init__(
         sensitivity_metric: Optional[str] = None,
         ignored_scope: Optional[dict] = None,
         num_samples: Optional[int] = None,
-        quant_method: Optional[Union[QuantizationMethod, OVQuantizationMethod]] = OVQuantizationMethod.DEFAULT,
+        quant_method: Union[QuantizationMethod, OVQuantizationMethod] = OVQuantizationMethod.DEFAULT,
```
nikita-savelyevv, line 248: Why not `Optional`?

nikita-savelyevv (commented):
Nice changes! Didn't consider this in my PR. Just couple questions from my side

AlexKoff88 (approved):

optimum/intel/openvino/configuration.py
echarlaix: [`OVQuantizationConfigBase`](https://github.com/huggingface/optimum-intel/blob/228a3e080922fe327c1d516eccbc5856178d2f1c/optimum/intel/openvino/configuration.py#L56)  inherits from `QuantizationConfigMixin` which has [`quant_method`](https://github.com/huggingface/transformers/blob/9459efb807fef4960eaf58d35e52fac4459933d2/src/transformers/utils/quantization_config.py#L74) as class attribute but no default value so results in `AttributeError: 'OVQuantizationConfig' object has no attribute 'quant_method'` when creating an instance of `OVQuantizationConfig`. I think it can make sense to have `quant_method` that can be set for every quantization config so that we can easily enable more quantization methodologies when integrated to nncf but if that's not planned for static PTQ for example I can revert it and just set the class attribute to a default value like `quant_method = OVQuantizationMethod.DEFAULT` what do you think ?

optimum/intel/openvino/configuration.py
echarlaix: I removed it as currently it's not expected for `quant_method` to be set to None  (`Optional[OVQuantizationMethod]` <-> `Union[OVQuantizationMethod, None]`), let me know if you think this should be a possibility

nikita-savelyevv: Got it, thanks.

optimum/intel/openvino/configuration.py
nikita-savelyevv: Thanks for explaining! I believe we don't plan to extend PTQ in the near future for use cases where `quant_method` will be a relevant parameter. But nevertheless since we inherit from `QuantizationConfigMixin` and it has a such property, we should support it I guess. I'm ok with this change.

nikita-savelyevv (approved):

optimum/intel/openvino/configuration.py
echarlaix: OK thanks for the information, because there is no such plan for PTQ I added : https://github.com/huggingface/optimum-intel/pull/668/commits/2916165df0fbe2722d05254e049c5ef213903d99

PR #676, opened 2024-04-22
apply sdpa for mpt and internlm

eaidova:
# What does this PR do?

optimize mpt and internlm models with scaled dot product attention
fixed export baichuan-13b model

## Before submitting


eaidova:

optimum/exporters/openvino/convert.py
```diff
@@ -388,15 +388,16 @@ def ts_patched_forward(*args, **kwargs):
...
-            input_name = ordered_input_names[idx]
-            inp_tensor.get_tensor().set_names({input_name})
-            inp_data = flatten_inputs[idx]
-            static_shape = PartialShape(inp_data.shape)
-            dims = inputs[input_name]
-            for dim in dims:
-                static_shape[dim] = -1
-            inp_tensor.get_node().set_partial_shape(static_shape)
-            inp_tensor.get_node().set_element_type(get_element_type(inp_data.cpu().numpy().dtype))
+            if idx < len(ordered_input_names):
```
AlexKoff88, line 391: Shall we handle the situation when there is no index in the `ordered_input_names`, i.e. when idx >= len(ordered_input_names)

AlexKoff88:
Can we have a test for each model architecture that is updated in this PR?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -611,3 +653,217 @@ def __init__(
...
+
+    attn_output = torch.nn.functional.scaled_dot_product_attention(
+        query_states, key_states, value_states, attention_mask, scale=(1 / math.sqrt(self.head_dim))
+    )
+    attn_output = attn_output.transpose(1, 2).contiguous()
+    attn_output = attn_output.reshape(bsz, q_len, self.hidden_size)
+
+    attn_output = self.wo(attn_output)
+
+    attn_weights = None
```
echarlaix, line 851: what if `output_attentions` is True ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -600,6 +602,46 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+    past_key_value = (key_states, value_states) if use_cache else None
+    attn_output = F.scaled_dot_product_attention(query_states, key_states, value_states, attn_mask=attention_mask)
+    attn_output = attn_output.transpose(1, 2)
+    attn_weights = None
+    attn_output = attn_output.reshape(bsz, q_len, self.hidden_size)
+    attn_output = self.o_proj(attn_output)
+
+    if not output_attentions:
+        attn_weights = None
```
echarlaix, line 640: already set to None

optimum/exporters/openvino/model_patcher.py
eaidova: as I know by default in transformers model classes gives warning for this and moves on eager attention instead sdpa, but  during export this is always False (export configs outputs is not designed for handlering additional outputs introduced by this flag), so I do not provided this branch of code

optimum/exporters/openvino/convert.py
eaidova: removed

optimum/exporters/openvino/model_patcher.py
eaidova: removed

AlexKoff88 (approved):

optimum/exporters/openvino/model_patcher.py
eaidova: I provided output_attentions support

tests/openvino/utils_tests.py
```diff
@@ -83,7 +84,7 @@
     "poolformer": "hf-internal-testing/tiny-random-PoolFormerModel",
     "qwen": "katuni4ka/tiny-random-qwen",
     "qwen2": "Qwen/Qwen1.5-0.5B",
-    "resnet": "hf-internal-testing/tiny-random-resnet",
+    "resnet": "hf-internal-testing/tiny-rantiny-random-baichuan2-13bdom-resnet",
```
echarlaix, line 87:
```suggestion
    "resnet": "hf-internal-testing/tiny-random-resnet",
```

echarlaix (approved):

PR #684, opened 2024-04-23
Deprecate use_auth_token

IlyasMoutawwakil:
# What does this PR do?
This PR deprecates `use_auth_token` in favor of `token`

Fixes # (issue)

## Before submitting


IlyasMoutawwakil:
I tried to add warning at the beginning of every function/method that used to accept `use_auth_token`, then only pass `token` wich occasionally gets assigned `use_auth_token`'s value. This works with `transformers` and `hugingface_hub` functions/methods but not diffusers (some still take `use_auth_token` only)

echarlaix (commented):
Looks great @IlyasMoutawwakil, could you add a test to [OVModelIntegrationTest](https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_modeling.py#L112) as well as done [here](https://github.com/huggingface/optimum/blob/c55f8824f58db1a2f1cfc7879451b4743b8f206b/tests/onnxruntime/test_modeling.py#L940)

IlyasMoutawwakil:
i added a test (that passes) but some cases still fail, like loading+exporting from a private repo, but it's an issue from optimum side.

echarlaix (approved):

PR #685, opened 2024-04-23
fix input generator for falcon40b

eaidova:
# What does this PR do?
this is temporal workaround until we can not use https://github.com/huggingface/optimum/pull/1825

falcon dummy input generator is not able to handle falcon-40b model (case with new_decoder_architecture and multi_query)

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -19,13 +19,14 @@ from packaging import version
 from transformers.utils import is_tf_available
 
 from optimum.exporters.onnx.config import TextDecoderOnnxConfig, TextDecoderWithPositionIdsOnnxConfig
-from optimum.exporters.onnx.model_configs import GemmaOnnxConfig, LlamaOnnxConfig
+from optimum.exporters.onnx.model_configs import FalconOnnxConfig, GemmaOnnxConfig, LlamaOnnxConfig
 from optimum.exporters.tasks import TasksManager
 from optimum.utils import DEFAULT_DUMMY_SHAPES
 from optimum.utils.input_generators import (
     DummyInputGenerator,
     DummyPastKeyValuesGenerator,
     DummyTextInputGenerator,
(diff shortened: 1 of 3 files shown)


eaidova:

AlexKoff88 (approved):

optimum/exporters/openvino/model_configs.py
```diff
@@ -437,3 +438,52 @@ class OrionOpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
...
+            batch_size=batch_size,
+            sequence_length=sequence_length,
+            random_batch_size_range=random_batch_size_range,
+            random_sequence_length_range=random_sequence_length_range,
+            **kwargs,
+        )
+        if normalized_config.new_decoder_architecture and normalized_config.multi_query:
+            self.num_kv_heads = normalized_config.num_attention_heads
+        elif normalized_config.new_decoder_architecture and not normalized_config.multi_query:
+            self.num_kv_heads = normalized_config.num_kv_heads
```
echarlaix, line 466: can we simplify like 
```suggestion
        if normalized_config.new_decoder_architecture:
            self.num_kv_heads = normalized_config.num_attention_heads if normalized_config.multi_query else normalized_config.num_kv_heads
```

echarlaix (approved):

optimum/exporters/openvino/model_configs.py
```diff
@@ -437,3 +438,52 @@ class OrionOpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
...
+        else:
+            self.num_kv_heads = 1
```
echarlaix, line 468: also shouldn't it be 
```suggestion
        if normalized_config.new_decoder_architecture:
            self.num_kv_heads = normalized_config.num_attention_heads
        else:
            self.num_kv_heads = normalized_config.num_kv_heads if not normalized_config.multi_query else 1
```
?

eaidova: this variant is correct

PR #687, opened 2024-04-24
Update imports in examples/notebooks

helena-intel:
Update examples/notebooks to use `from optimum.intel import ...` instead of `from optimum.intel.openvino import ...`. This is best practice, and it prevents difficult to debug errors. For example, if accelerate is not installed and you import `OVTrainer`, you get a friendly error with the new import but not with the old version.

Also 
- added accelerate to example requirements
- updated Ubuntu/Python versions in the workflow files.
- fixed running example tests on changing example files

@AlexKoff88 unrelated to this PR, the audio classification example fails with this error:

```
FAILED examples/openvino/test_examples.py::TestExamples::test_audio_classification - nncf.errors.ValidationError: No match has been found among the model operations for the following ignored/target scope definitions:
 - ignored_scope: ['{re}.*__truediv__.*', '{re}.*matmul_1']
Refer to the original_graph.dot to discover the operations in the model currently visible to NNCF and specify the ignored/target scopes in terms of the names there.
```


.github/workflows/test_openvino_examples.yml
```diff
@@ -22,9 +22,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.10"]
+        python-version: ["3.8", "3.11"]
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
AlexKoff88, line 27: why are we moving to Ubuntu 22? is it aligned with OpenVINO OS support matrix?

helena-intel: 24.04 is around the corner and 20.04 is pretty old now.  22.04 is officially supported and we release archives for it too. NPU is not supported with 20.04, so 22.04 is arguably the best version at the moment.

But if there is any reason to keep 20.04 happy to move it back.

AlexKoff88: Thanks for the context. It makes sense.

AlexKoff88 (approved):

echarlaix (approved):
Looks good, thanks @helena-intel !!

echarlaix:
Can you take a look at the failing `test_audio_classification` examples before we can merge @helena-intel ?

helena-intel:
> Can you take a look at the failing `test_audio_classification` examples before we can merge @helena-intel ?

It looks like the issue is caused by the nncf_compression_config not being used. In https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py#L313 compression is set to None. I changed it to  `kwargs.get("compression", None)`.

echarlaix:
thanks a lot for the fix @helena-intel

PR #688, opened 2024-04-25
Pin Pytorch 2.2 in IPEX Tests

IlyasMoutawwakil:
# What does this PR do?

Until a compatible ipex is released.

## Before submitting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ except Exception as error:
     assert False, "Error: Could not open '%s' due %s\n" % (filepath, error)
 
 INSTALL_REQUIRE = [
-    "torch>=1.11,<2.3",
+    "torch>=1.11",
     "transformers>=4.36.0,<4.41.0",
     "optimum~=1.19",
     "datasets>=1.4.0",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -27,7 +27,7 @@
     assert False, "Error: Could not open '%s' due %s\n" % (filepath, error)
 
 INSTALL_REQUIRE = [
-    "torch>=1.11",
+    "torch>=1.11,<2.3",
```
echarlaix, line 30: I think it make sense to pin the torch version for IPEX tests only, the IPEX version should always match the torch version (it's not related to optimum-intel)

helena-intel:
OpenVINO tests pass with PyTorch 2.3 ( https://github.com/huggingface/optimum-intel/actions/runs/8821417491/job/24217231927 ). @eaidova do you know of any issues with OpenVINO and PyTorch 2.3? If there are no issues with OpenVINO the PyTorch version should be limited only in the INC/IPEX extra.

setup.py
IlyasMoutawwakil: thanks makes sense

echarlaix (approved):
Thanks @IlyasMoutawwakil !

PR #690, opened 2024-04-25
Add version info to OpenVINO models

helena-intel:
Versions of optimum-intel and dependencies can have an impact on performance/accuracy. This PR adds transformers, torch, optimum and optimum-intel dependencies to the openvino_model.xml file (the OpenVINO version is already there). This is really useful when debugging issues, or looking into differences.

Example (`<optimum>` is added):
image

@eaidova @AlexKoff88

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -20,9 +20,10 @@ import os
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
+import onnx
 from transformers.utils import is_tf_available, is_torch_available
 
-from openvino.runtime import PartialShape, save_model
+from openvino.runtime import Model, PartialShape, save_model
 from openvino.runtime.exceptions import OVTypeError
 from openvino.runtime.utils.types import get_element_type
 from openvino.tools.ovc import convert_model
@@ -33,8 +34,10 @@ from optimum.exporters.onnx.convert import export_pytorch as export_pytorch_to_o
 from optimum.exporters.onnx.convert import export_tensorflow as export_tensorflow_onnx
 from optimum.exporters.utils import _get_submodels_and_export_configs
 from optimum.intel.utils.import_utils import (
+    _nncf_version,
     _optimum_intel_version,
     _optimum_version,
+    _timm_version,
     _torch_version,
     _transformers_version,
 )
@@ -87,10 +90,8 @@ def _save_model(model, path: str, ov_config: Optional["OVConfig"] = None):
 
         compress_to_fp16 = ov_config.dtype == "fp16"
 
-    model.set_rt_info(_transformers_version, ["optimum", "transformers_version"])
-    model.set_rt_info(_torch_version, ["optimum", "pytorch_version"])
-    model.set_rt_info(_optimum_intel_version, ["optimum", "optimum_intel_version"])
-    model.set_rt_info(_optimum_version, ["optimum", "optimum_version"])
+    library_name = TasksManager.infer_library_from_model(Path(path).parent)
+    model = _add_version_info_to_model(model, library_name)
     save_model(model, path, compress_to_fp16)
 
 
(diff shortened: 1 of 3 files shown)


optimum/exporters/openvino/convert.py
```diff
@@ -81,6 +87,10 @@ def _save_model(model, path: str, ov_config: Optional["OVConfig"] = None):
 
         compress_to_fp16 = ov_config.dtype == "fp16"
 
+    model.set_rt_info(_transformers_version, ["optimum", "transformers_version"])
+    model.set_rt_info(_torch_version, ["optimum", "pytorch_version"])
+    model.set_rt_info(_optimum_intel_version, ["optimum", "optimum_intel_version"])
+    model.set_rt_info(_optimum_version, ["optimum", "optimum_version"])
```
eaidova, line 93: do we need additionally add info about nncf here?
and as some other libs may be used as model source for conversion (diffusers for sd, timm for image classification and sentence_transformers), I think it can be useful to provide this libs info too (if they are available or based on source lib for conversion)

optimum/intel/utils/import_utils.py
```diff
@@ -33,6 +33,7 @@
 STR_OPERATION_TO_FUNC = {">": op.gt, ">=": op.ge, "==": op.eq, "!=": op.ne, "<=": op.le, "<": op.lt}
 
 _optimum_version = importlib_metadata.version("optimum")
+_optimum_intel_version = importlib_metadata.version("optimum-intel")
```
eaidova, line 36: why not from optimum.intel.version import __version__?

optimum/exporters/openvino/convert.py
AlexKoff88: +1 for including NNCF version

helena-intel: I added NNCF, diffusers, timm, sentence-transformers and onnx (for models converted from ONNX). I prefer not to add them just when the dependency is available, it's a bit strange to me to have timm_version in a BERT model for example. For timm and diffusers I now get the library name from the model path. Seems to work well, but let me know if there is a better way.

optimum/intel/utils/import_utils.py
helena-intel: I liked the consistency with _optimum_version :) But also the importlib_metadata version gives the commit ID:

```
>>> from optimum.intel.version import __version__
>>> from optimum.intel.utils.import_utils import _optimum_intel_version
>>> __version__
'1.17.0.dev0'
>>> _optimum_intel_version
'1.17.0.dev0+9d58b66'
```

eaidova (approved):

optimum/exporters/openvino/convert.py
```diff
@@ -689,3 +700,34 @@ def export_tokenizer(
...
+        model.set_rt_info(_transformers_version, ["optimum", "transformers_version"])
+        model.set_rt_info(_torch_version, ["optimum", "pytorch_version"])
+        model.set_rt_info(_optimum_intel_version, ["optimum", "optimum_intel_version"])
+        model.set_rt_info(_optimum_version, ["optimum", "optimum_version"])
+
+        if any("token_embeddings" in output.get_names() for output in model.outputs):
+            import sentence_transformers
+
+            model.set_rt_info(sentence_transformers.__version__, ["optimum", "sentence_transformers_version"])
+        if library_name == "diffusers":
```
echarlaix, line 719: why not :
```suggestion
       if library_name == "sentence_transformers":
            import sentence_transformers

            model.set_rt_info(sentence_transformers.__version__, ["optimum", "sentence_transformers_version"])
        elif library_name == "diffusers":
```

echarlaix (approved):
Looks great, thanks @helena-intel

optimum/exporters/openvino/convert.py
helena-intel: That was my first attempt, but that didn't work, after `optimum-cli export openvino -m BAAI/bge-base-en-v1.5 --library sentence_transformers bge-base-en-ov-st` library_name is detected as transformers by `TasksManager.infer_library_from_model`. If I copy the [config_sentence_transformers.json](https://huggingface.co/BAAI/bge-base-en-v1.5/blob/main/config_sentence_transformers.json) to the model directory, the model is detected as sentence-transformers. Should we save that file for models exported with sentence_transformers library?

echarlaix: I see, an option could be to re-use the `library_name` inferred from the original model, this solution also works for me

PR #692, opened 2024-04-29
make unet and vae number of channels static

eaidova:
# What does this PR do?

uses static number of channels for unet and vae as this value is important for model architecture (number of channels for convolutions) and can not be changed in runtime and always expected to be the same like provided in config

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ INSTALL_REQUIRE = [
 
 TESTS_REQUIRE = [
     "accelerate",
-    "pytest",
+    "pytest<8.2",
     "parameterized",
     "Pillow",
     "evaluate",


slyalin (approved):
Based on the description, it will simplify application code when exported OV model is being deployed without the need to hardcode internal parameters in the application or reading them from config files.

echarlaix (approved):
LGTM, why not adding it to optimum instead ?

eaidova:
> LGTM, why not adding it to optimum instead ?

@echarlaix, I'll open the same PR to optimum too if you agree, I just want unblocking some work on openvino side where these changes can simplify demo workflow as soon as possible

echarlaix:
> > LGTM, why not adding it to optimum instead ?
> 
> @echarlaix, I'll open the same PR to optimum too if you agree, I just want unblock some work on openvino side where these changes can simplify demo workflow as soon as possible

works for me thanks @eaidova

PR #695, opened 2024-04-30
optimize first latency beam search for OVModelForCausalLM

eaidova:
# What does this PR do?
this PR reduces first token latency for OVModelForCausalLM class if beam search decoding selected. Beam search represented during generation as batch of sequences (generation batch size = [num_input_promts * num_beams]). Generation API duplicates initial input sequence for promoting them for each beam before starting work, while on the first step all sequences are equal (in the same time, the first inference for models with cache is the most time-consuming part). The idea is postpone sequence duplication for beams after first iteration done (including duplication of past key values and logits in outputs) 

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -17,7 +17,7 @@ import os
 import warnings
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import Dict, Optional, Tuple, Union, Any, List, Callable, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import openvino
@@ -28,12 +28,22 @@ from openvino.runtime import Core, Tensor, Type
 from transformers import AutoModelForCausalLM, PretrainedConfig
 from transformers.file_utils import add_start_docstrings, add_start_docstrings_to_model_forward
 from transformers.generation import GenerationMixin
-from transformers.modeling_outputs import CausalLMOutputWithPast
+from transformers.generation.beam_search import BeamScorer, ConstrainedBeamSearchScorer
 from transformers.generation.configuration_utils import GenerationConfig, GenerationMode
-from transformers.generation.utils import GenerateOutput, GenerateBeamDecoderOnlyOutput, GenerateBeamOutput, _split_model_inputs, stack_model_outputs
-from transformers.generation.stopping_criteria import StoppingCriteriaList, EosTokenCriteria
 from transformers.generation.logits_process import LogitsProcessorList
-from transformers.generation.beam_search import BeamScorer
+from transformers.generation.stopping_criteria import (
+    EosTokenCriteria,
+    StoppingCriteriaList,
+    validate_stopping_criteria,
+)
+from transformers.generation.utils import (
+    GenerateBeamDecoderOnlyOutput,
+    GenerateBeamOutput,
+    GenerateOutput,
+    _split_model_inputs,
+    stack_model_outputs,
+)
+from transformers.modeling_outputs import CausalLMOutputWithPast
 
 from optimum.utils.normalized_config import NormalizedConfigManager
 
@@ -44,6 +54,7 @@ from ..utils.modeling_utils import MULTI_QUERY_ATTN_MODELS
 from .configuration import _DEFAULT_4BIT_CONFIGS, OVConfig, OVWeightQuantizationConfig, _check_default_4bit_configs
 from .modeling import _TOKENIZER_FOR_DOC, INPUTS_DOCSTRING, MODEL_START_DOCSTRING, OVModel
 from .utils import ONNX_WEIGHTS_NAME, OV_XML_FILE_NAME, STR_TO_OV_TYPE
+
 
 if TYPE_CHECKING:
     from transformers.modeling_utils import PreTrainedModel
@@ -528,7 +539,7 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
             position_ids.masked_fill_(attention_mask == 0, 1)
             if past_key_values:
                 position_ids = position_ids[:, -input_ids.shape[1] :]
-        
+
         model_inputs = {
             "input_ids": input_ids,
             "past_key_values": past_key_values,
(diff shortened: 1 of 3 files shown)


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -651,6 +764,954 @@ def _from_pretrained(
 
         return causal_model
 
+    def _beam_search(
```
echarlaix, line 767: Could you specify using comments which part has been modified from the original code, this would help a lot for review and future maintenance

tests/openvino/test_modeling.py
```diff
@@ -757,6 +757,87 @@ def test_default_filling_attention_mask_and_position_ids(self):
...
+
+        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        tokenizer.pad_token = tokenizer.eos_token
+        tokens = tokenizer(["Today is a nice day and I am longer", "This is me"], return_tensors="pt", padding=True)
+        ov_model_stateful.generation_config.eos_token_id = None
+        ov_model_stateless.generation_config.eos_token_id = None
+        transformers_model.generation_config.eos_token_id = None
+        ov_model_stateful.config.eos_token_id = None
+        ov_model_stateless.config.eos_token_id = None
+        transformers_model.config.eos_token_id = None
```
echarlaix, line 774: why is this needed ?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -651,6 +764,954 @@ def _from_pretrained(
...
+
+        while self._has_unfinished_sequences(this_peer_finished, synced_gpus, device=input_ids.device):
+            model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
+
+            # if sequential is True, split the input to batches of batch_size and run sequentially
+            if sequential:
+                inputs_per_sub_batches = _split_model_inputs(
+                    model_inputs,
+                    split_size=batch_size,
+                    full_batch_size=batch_beam_size if not self._first_iteration else batch_size,
```
echarlaix, line 868: shouldn't it be 
```suggestion
                    full_batch_size=batch_beam_size if not first_iteration else batch_size,
```

tests/openvino/test_modeling.py
eaidova: I copied it from test above where comparison with transformers. For my test, it is not need, for that place, i think it is used for generate exact number of tokens without early stopping

echarlaix (commented):
What is the observed performance gain for the first token generation ? There is a lot of method overwritten in this PR which could lead to potential issues in terms of future transformers compatibility for generation, so would like to make sure the performance gain is significant before considering merging

optimum/intel/openvino/modeling_decoder.py
IlyasMoutawwakil: +1

IlyasMoutawwakil:
IMO this will be very heavy to maintain with the constant changes in transformers lib, especially since the text generation api will be undergoing heavy refactorization soon. 

Would it not make sense to instead of optimizing the generation strategy, rather optimize the first forward pass, with something along the lines of:

```python
def generate():
    if beam_search: # or any generation strategy where this issue is observed
        self.first_beam_search_iteration = True
     else:
        self.first_beam_search_iteration = False
    
    return super().generate()

def forward():
    if self.first_beam_search_iteration :
        unique_inputs, inverse_order = torch.unique(inputs, dim=0, return_inverse=True) 
        # we can also use what we know about how the inputs are duplicated to deduplicate them
        unique_outputs = super().forward(unique_inputs)
        outputs = unique_inputs[inverse_order]
        self.first_beam_search_iteration = False
    else:
        outputs = super().forward(inputs)
    return outputs
```

I admit that this is more stateful and hacky than what's suggested in the PR, but it requires maintaining less code, until this duplication issue with beam search gets fixed in transformers.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -651,6 +764,954 @@ def _from_pretrained(
...
+                " `stopping_criteria=StoppingCriteriaList([MaxLengthCriteria(max_length=max_length)])` instead.",
+                UserWarning,
+            )
+            stopping_criteria = validate_stopping_criteria(stopping_criteria, max_length)
+        if len(stopping_criteria) == 0:
+            warnings.warn("You don't have defined any stopping_criteria, this will likely loop forever", UserWarning)
+        pad_token_id = pad_token_id if pad_token_id is not None else self.generation_config.pad_token_id
+        if eos_token_id is not None:
+            logger.warning_once(
+                "`eos_token_id` is deprecated in this function and will be removed in v4.41, use"
```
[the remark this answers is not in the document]
eaidova: this is copied from original transformers code

eaidova:
@IlyasMoutawwakil, thank you for your suggestion, that is from what I begin, but problem that we need to know how inputs was duplicated for nonstateful case to duplicate past key values and this required additional context for that (from generation config) that is not provided inside forward

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -651,6 +764,954 @@ def _from_pretrained(
...
+            )
+            stopping_criteria.append(EosTokenCriteria(eos_token_id=eos_token_id))
+        else:
+            eos_token_id = [
+                criteria.eos_token_id.tolist() for criteria in stopping_criteria if hasattr(criteria, "eos_token_id")
+            ]
+            eos_token_id = eos_token_id[0] if eos_token_id else None
+            if eos_token_id is None and self.generation_config.eos_token_id is not None:
+                eos_token_id = self.generation_config.eos_token_id
+                stopping_criteria.append(EosTokenCriteria(eos_token_id=eos_token_id))
```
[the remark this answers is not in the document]
eaidova: removed

eaidova:
@IlyasMoutawwakil @echarlaix  please take a look one more time, I significantly updated code for reducing overriding beam search methods

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -460,14 +477,13 @@ def forward(
     ) -> CausalLMOutputWithPast:
         self.compile()
 
-        inputs = self.prepare_inputs(
+        inputs, duplication_idicies = self.prepare_inputs(
```
echarlaix, line 480:
```suggestion
        inputs, duplication_indices = self.prepare_inputs(
```

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -448,7 +463,9 @@ def prepare_inputs(
                 self.next_beam_idx if self.next_beam_idx is not None else np.arange(batch_size, dtype=int)
             )
 
-        return inputs
+        if self._first_iter_beam_search:
+            inputs, duplication_indices = self._deduplicate_inputs(inputs)
+        return inputs, duplication_indices
```
echarlaix, line 468: to avoid a breaking change could this be moved to the `forward` instead ?

tests/openvino/test_modeling.py
```diff
@@ -778,6 +778,104 @@ def test_default_filling_attention_mask_and_position_ids(self):
...
+            num_beams=4,
+            do_sample=False,
+            eos_token_id=None,
+        )
+
+        transformers_outputs = transformers_model.generate(**tokens, generation_config=gen_config)
+        ov_stateful_outputs = ov_model_stateful.generate(**tokens, generation_config=gen_config)
+        self.assertTrue(torch.allclose(ov_stateful_outputs, transformers_outputs))
+        ov_stateless_outputs = ov_model_stateless.generate(**tokens, generation_config=gen_config)
+        self.assertTrue(torch.allclose(ov_stateless_outputs, transformers_outputs))
```
echarlaix, line 827: looks the same everywhere so would iterate on the different gen_config instead

echarlaix (approved):
Looks great thanks for iterating on this @eaidova

PR #698, opened 2024-05-02
Adding more INC tests

IlyasMoutawwakil:
# What does this PR do?
This PR adds and enables more tests for INC

## Before submitting

### DIFF
--- a/tests/neural_compressor/test_optimization.py
+++ b/tests/neural_compressor/test_optimization.py
@@ -94,7 +94,7 @@ class QuantizationTest(INCTestMixin):
 
         quantized_model = None
         save_onnx_model = False
-        model_kwargs = {"use_cache": False} if task == "text-generation" else {}
+        model_kwargs = {"use_cache": False, "use_io_binding": False} if task == "text-generation" else {}
         quantization_config = PostTrainingQuantConfig(approach="dynamic")
 
         with tempfile.TemporaryDirectory() as tmp_dir:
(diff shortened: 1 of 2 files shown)


tests/neural_compressor/test_optimization.py
```diff
@@ -130,28 +133,26 @@ def test_static_quantization(self, task, model_arch, expected_quantized_matmuls)
...
-        op_type_dict = (
-            {"Embedding": {"weight": {"dtype": ["fp32"]}, "activation": {"dtype": ["fp32"]}}}
-            if save_onnx_model
-            else None
-        )
-        quantization_config = PostTrainingQuantConfig(approach="static", op_type_dict=op_type_dict)
         quantized_model = None
+        save_onnx_model = False
+        quantization_config = PostTrainingQuantConfig(approach="static")
+        model_kwargs = {"use_cache": False, "use_io_binding": False} if task == "text-generation" else {}
```
echarlaix, line 139:
```suggestion
        model_kwargs = {"use_cache": False} if task == "text-generation" else {}
```

echarlaix (approved):
Looks greeat, thans a lot @IlyasMoutawwakil

tests/neural_compressor/test_optimization.py
IlyasMoutawwakil: it was in the dynamic quant tests, I'll remove it

IlyasMoutawwakil: it seems we only do that with text generation because `use_cache=False` imposes it:
```
ValueError: The parameters combination use_cache=False, use_io_binding=True is not supported. Please either pass use_cache=True, use_io_binding=True (default), or use_cache=False, use_io_binding=False.
```

PR #700, opened 2024-05-06
Refactor SD calibration data collection

nikita-savelyevv:
# What does this PR do?

- Calibration dataset collection logic moved from `OVStableDiffusionPipelineBase` to `OVQuantizer`
- Added `OVQuantizationMethod.HYBRID`. Hybrid quantization is now called through `OVQuantizer`.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -50,7 +50,7 @@ from ...exporters.openvino.stateful import ensure_export_task_support_stateful, 
 from ..utils.constant import _TASK_ALIASES
 from ..utils.import_utils import DATASETS_IMPORT_ERROR, is_datasets_available
 from ..utils.modeling_utils import get_model_device
-from .configuration import OVConfig, OVQuantizationConfig, OVWeightQuantizationConfig, OVQuantizationMethod
+from .configuration import OVConfig, OVQuantizationConfig, OVQuantizationMethod, OVWeightQuantizationConfig
 from .modeling_base import OVBaseModel
 from .utils import (
     MAX_ONNX_OPSET,
@@ -339,8 +339,8 @@ class OVQuantizer(OptimumQuantizer):
 
             if isinstance(self.model, OVStableDiffusionPipelineBase):
                 calibration_dataset = self._prepare_unet_dataset(
-                    quantization_config.num_samples,
-                    dataset=calibration_dataset)
+                    quantization_config.num_samples, dataset=calibration_dataset
+                )
             elif Dataset is not None and isinstance(calibration_dataset, Dataset):
                 calibration_dataloader = self._get_calibration_dataloader(
                     calibration_dataset=calibration_dataset,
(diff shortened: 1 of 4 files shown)


nikita-savelyevv:
@l-bat @AlexKoff88 @echarlaix please take a look

optimum/intel/openvino/quantization.py
```diff
@@ -318,80 +321,95 @@ def _quantize_ovbasemodel(
...
         quantization_config = ov_config.quantization_config
+
+        if calibration_dataset is not None:
+            # Process custom calibration dataset
+
+            if isinstance(self.model, OVStableDiffusionPipelineBase):
+                calibration_dataset = self._prepare_unet_dataset(
+                    quantization_config.num_samples, dataset=calibration_dataset
+                )
+            elif Dataset is not None and isinstance(calibration_dataset, Dataset):
```
AlexKoff88, line 344: Please correct the part `elif Dataset is not None`

optimum/intel/openvino/quantization.py
```diff
@@ -318,80 +321,95 @@ def _quantize_ovbasemodel(
...
+                    data_collator=data_collator,
+                )
+
+                if self.model.export_feature == "text-generation" and self.model.use_cache:
+                    calibration_dataset = self._prepare_text_generation_dataset(
+                        quantization_config, calibration_dataloader
+                    )
+                else:
+                    calibration_dataset = nncf.Dataset(calibration_dataloader)
+            elif isinstance(calibration_dataset, collections.abc.Iterable):
```
AlexKoff88, line 358: I am ok with this elif clause but I assume that it is responsibility of NNCF to do such kind of checks for Itarable and through the meaningful exceptions if it is not correct.

optimum/intel/openvino/quantization.py
```diff
@@ -318,80 +321,95 @@ def _quantize_ovbasemodel(
...
 
-            _weight_only_quantization(self.model.model, quantization_config, calibration_dataset)
+            if quantization_config.quant_method == OVQuantizationMethod.HYBRID:
+                if calibration_dataset is None:
+                    raise ValueError("Calibration dataset is required to run hybrid quantization.")
+                if isinstance(self.model, OVStableDiffusionPipelineBase):
+                    self.model.unet.model = _hybrid_quantization(
+                        self.model.unet.model, quantization_config, calibration_dataset
+                    )
+                else:
```
AlexKoff88, line 394: can you please place a comment about what kind of cases (models) matches this else?

optimum/intel/openvino/quantization.py
```diff
@@ -654,6 +672,105 @@ def _remove_unused_columns(self, dataset: "Dataset"):
         ignored_columns = list(set(dataset.column_names) - set(self._signature_columns))
         return dataset.remove_columns(ignored_columns)
 
+    def _prepare_gptq_dataset(self, quantization_config: OVWeightQuantizationConfig):
```
AlexKoff88, line 675:
```suggestion
    def _prepare_builtin_dataset(self, quantization_config: OVWeightQuantizationConfig):
```

optimum/intel/openvino/quantization.py
nikita-savelyevv: This part is actually correct. This is to adequately process the case when `datasets` is not available. There is such code above:
```
if is_datasets_available():
    from datasets import Dataset
else:
    Dataset = None
```

optimum/intel/openvino/quantization.py
nikita-savelyevv: I agree, this is not strictly required, but it's just a small convenience to avoid the need for users to create an instance of `nncf.Dataset`.

AlexKoff88:
@nikita-savelyevv, please revise this example as well: https://github.com/huggingface/optimum-intel/blob/main/notebooks/openvino/stable_diffusion_hybrid_quantization.ipynb

optimum/intel/openvino/quantization.py
nikita-savelyevv: Done

PR #704, opened 2024-05-11
add XPU support for `IPEXModel.from_pretrained`

faaany:
## What does this PR do?
This PR adds XPU support for loading a Torchscript model using `IPEXModel.from_pretrained`. Below is a test example: 

```python
import torch 
from optimum.intel import IPEXModel

model_id = "faaany/bert-base-uncased-float32-traced"
model_id_tokenizer = "google-bert/bert-base-uncased"
model = IPEXModel.from_pretrained(model_id)
tokenizer = AutoTokenizer.from_pretrained(model_id_tokenizer)
inputs = tokenizer("Paris is the capital of France.", return_tensors="pt").to("xpu")
print("________________")
print(model.device)
with torch.no_grad():
    outputs = model(**inputs)
    embeddings = outputs[0][:, 0]
    print(embeddings)
```

Please note that `faaany/bert-base-uncased-float32-traced` is just a test model, which is traced from "google-bert/bert-base-uncased". Since I didn't load any tokenizer config, so I pass `model_id_tokenizer` to `AutoTokenizer` in the example above.

### DIFF
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -324,7 +324,7 @@ class IPEXModel(OptimizedModel):
         if not self._is_ipex_exported:
             use_cache = "past_key_values" in self.input_names
             dummy_inputs = prepare_jit_inputs(self, self.export_feature, use_cache)
-            if "cpu" not in str(self._device):
+            if self._device.type != "cpu":
                 dummy_inputs = recursive_to_device(value=dummy_inputs, device=self._device)
             for _ in range(2):
                 self(**dummy_inputs)
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -39,12 +39,12 @@ from transformers import (
     GenerationConfig,
     GenerationMixin,
     PretrainedConfig,
+    is_torch_xpu_available,
 )
 from transformers.dynamic_module_utils import get_class_from_dynamic_module
 from transformers.modeling_outputs import CausalLMOutputWithPast, ModelOutput
 from transformers.models.auto.auto_factory import _get_model_class as get_model_class
 from transformers.utils import WEIGHTS_NAME
-from transformers import is_torch_xpu_available
 
 from optimum.exporters import TasksManager
 from optimum.modeling_base import OptimizedModel
@@ -53,7 +53,7 @@ from optimum.utils import NormalizedConfigManager
 from ...exporters.ipex.model_patcher import _IPEX_EXPORTED_TASK, _patch_model
 from ..generation.modeling import prepare_jit_inputs
 from ..utils.import_utils import is_ipex_version, is_torch_version, is_transformers_version
-from ..utils.modeling_utils import MULTI_QUERY_ATTN_MODELS, patch_decoder_attention_mask
+from ..utils.modeling_utils import MULTI_QUERY_ATTN_MODELS, patch_decoder_attention_mask, recursive_to_device
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


faaany:
@echarlaix pls have a review. Thx!

faaany:
@yao-matrix

faaany:
Hi @IlyasMoutawwakil could you help review this PR? Thanks a lot!

optimum/intel/ipex/modeling_base.py
```diff
@@ -319,6 +347,8 @@ def _init_warmup(self):
         if not self._is_ipex_exported:
             use_cache = "past_key_values" in self.input_names
             dummy_inputs = prepare_jit_inputs(self, self.export_feature, use_cache)
+            if "cpu" not in str(self._device):
+                dummy_inputs = {name: tensor.to(self._device) for name, tensor in dummy_inputs.items()}
```
echarlaix, line 351: To also make it work for inputs which are list or tuple (like `past_key_values`) you need to add a [`recursive_to_device`]((https://github.com/huggingface/optimum/blob/b3ecb6c405b7fd5425d79483fd7dc88c0609be8e/optimum/exporters/onnx/utils.py#L120)) function (can be added [here](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/utils/)) and then : 

```python
for key, value in dummy_inputs.items():
    dummy_inputs[key] = recursive_to_device(value=value, device=self._device)
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -128,10 +129,37 @@ def __init__(
         **kwargs,
     ):
         OptimizedModel.__init__(self, model=model, config=config)
-        # To do: add XPU support
-        self._device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
-        self._dtype = self.config.torch_dtype if self.config.torch_dtype is not None else torch.float32
+        device_map = kwargs.pop("device_map", None)
```
echarlaix, line 132: `device_map` is used by `accelerate` to dispatch the weights for each layer across different devices which is not the case here so I think we should remove 

```suggestion
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -128,10 +129,37 @@ def __init__(
...
+                if is_torch_xpu_available(check_device=True):
+                    self._device = torch.device(f"xpu:{device_map}")
+                elif torch.cuda.is_available():
+                    self._device = torch.device(f"cuda:{device_map}")
+                else:
+                    self._device = torch.device("cpu")
+            else:
+                raise ValueError(
+                    f"device_map should be either be a string, an integer or a torch.device object, but found {type(device_map)}"
+                )
```
echarlaix, line 160:
```suggestion
```

echarlaix:
To fix the code style test you can do the following : 
```
pip install .[quality]
make style
```

optimum/intel/ipex/modeling_base.py
faaany: Thx for the suggestion! I added a `recursive_to_device` function, which also works for a dictionary. So we can save `for key, value in dummy_inputs.items():`. Pls have a review.

faaany:
Hi @echarlaix, there are some CI tests failing. Do you know what I can do to fix them? Thx!

optimum/intel/utils/modeling_utils.py
```diff
@@ -169,3 +169,16 @@ def get_model_device(model: torch.nn.Module) -> torch.device:
...
+
+
+def recursive_to_device(value, device):
+    """
+    Recursivley move the tensor element in `value` to `device`
+    """
+    if isinstance(value, (tuple, list)):
+        return type(value)(recursive_to_device(v, device) for v in value)
+    elif isinstance(value, dict):
+        return type(value)({k: recursive_to_device(v, device) for k, v in value.items()})
```
echarlaix, line 181: wouldn't this be equivalent to
```suggestion
        return {k: recursive_to_device(v, device) for k, v in value.items()}
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -319,6 +324,8 @@ def _init_warmup(self):
         if not self._is_ipex_exported:
             use_cache = "past_key_values" in self.input_names
             dummy_inputs = prepare_jit_inputs(self, self.export_feature, use_cache)
+            if "cpu" not in str(self._device):
```
echarlaix, line 327:
```suggestion
            if self._device.type != "cpu":
```

echarlaix (approved):
LGTM, thanks for the addition @faaany !

PR #705, opened 2024-05-13
add sdpa for phi3 openvino model

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -971,7 +971,11 @@ def _phi3_self_attn_sdpa_forward(
             use_cache=use_cache,
         )
 
-    from transformers.models.llama.modeling_llama import apply_rotary_pos_emb, repeat_kv
+    # TO DO: remove llama imports when transformers with phi3 support will be released
+    try:
+        from transformers.models.phi3.modelling_phi3 import apply_rotary_pos_emb, repeat_kv
+    except ImportError:
+        from transformers.models.llama.modeling_llama import apply_rotary_pos_emb, repeat_kv
 
     bsz, q_len, _ = hidden_states.size()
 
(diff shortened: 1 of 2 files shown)


eaidova:
@echarlaix @AlexKoff88  could you please take a look?

AlexKoff88 (approved):

optimum/exporters/openvino/model_patcher.py
```diff
@@ -951,15 +951,103 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        return self._orig_forward(
+            hidden_states=hidden_states,
+            attention_mask=attention_mask,
+            position_ids=position_ids,
+            past_key_value=past_key_value,
+            output_attentions=output_attentions,
+            use_cache=use_cache,
+        )
+
+    from transformers.models.llama.modeling_llama import apply_rotary_pos_emb, repeat_kv
```
echarlaix, line 974: even if currently equivalent shouldn't be replace it with 
```suggestion
    from transformers.models.phi3.modeling_phi3 import apply_rotary_pos_emb, repeat_kv
```
to avoid any issue resulting from potential refactorization

optimum/exporters/openvino/model_patcher.py
```diff
@@ -951,15 +951,103 @@ def __exit__(self, exc_type, exc_value, traceback):
                 block.attention.forward = block.attention._orig_forward
 
 
+# Adapted from Phi3Attention.forward
```
echarlaix, line 954: would be very useful for  maintenance to also add the commit + exact file the original method is from, like for example :
```suggestion
# Adapted from https://github.com/huggingface/transformers/blob/ccdabc5642bf84849af93f591e207dc625c8e1e1/src/transformers/models/phi3/modeling_phi3.py#L426
```

echarlaix (approved):

optimum/exporters/openvino/model_patcher.py
eaidova: I believe it will breaks with current release, because phi3 code in transformers is not released yet... Can we add some TODO to change this in future?

echarlaix: yes works for me, thanks @eaidova !

PR #709, opened 2024-05-15
cover more models with openvino export

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


echarlaix (approved):
Looks great, thanks @eaidova!

tests/openvino/test_modeling.py
```diff
@@ -617,8 +622,11 @@ def test_compare_to_transformers(self, model_arch):
         if model_arch == "qwen":
             return
 
-        if model_arch != "chatglm":
+        if not model_arch in ["chatglm", "persimmon"]:
```
echarlaix, line 625: to fix the style check :
```suggestion
        if model_arch not in ["chatglm", "persimmon"]:
```

PR #710, opened 2024-05-15
prevent loading model for export if it is not supported

eaidova:
# What does this PR do?

Prevent downloading and loading model into RAM if it is not supported for export and custom export config is not provided


eaidova:
@echarlaix  @IlyasMoutawwakil  @AlexKoff88  could you please take a look?

P.S. Possibly the same changes may be useful for onnx too

echarlaix (approved):
LGTM, thanks @eaidova

PR #712, opened 2024-05-15
Fix diffusers requirement for quantizing models

helena-intel:
quantization.py now imports `OVStableDiffusionPipelineBase`  which requires diffusers to be installed, but that should not be required for quantizing non-diffusion models. This PR changes this so `OVStableDiffusionPipelineBase` only gets imported if diffusers is available. This is safe because the class is only used in `isinstance` calls, and if a model is a stable diffusion model, diffusers will definitely be available.


echarlaix (approved):
Thanks for the fix @helena-intel !

PR #713, opened 2024-05-16
[OV] Add --all-layers argument to CLI

nikita-savelyevv:
# What does this PR do?

Add `--all-layers` quantization argument to openvino export CLI interface. 

## Before submitting

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -120,6 +120,14 @@ def parse_args_openvino(parser: "ArgumentParser"):
         ),
     )
     optional_group.add_argument(
+        "--all-layers",
+        action="store_true",
+        default=None,
+        help=(
+            "Whether embeddings and last MatMul layers should be compressed to a primary precision (usually, INT4)."
+        ),
+    )
+    optional_group.add_argument(
         "--disable-stateful",
         action="store_true",
         help=(
@@ -198,6 +206,7 @@ class OVExportCommand(BaseOptimumCLICommand):
                 and self.args.ratio is None
                 and self.args.group_size is None
                 and self.args.sym is None
+                and self.args.all_layers is None
                 and self.args.model in _DEFAULT_4BIT_CONFIGS
             ):
                 quantization_config = _DEFAULT_4BIT_CONFIGS[self.args.model]
@@ -207,6 +216,7 @@ class OVExportCommand(BaseOptimumCLICommand):
                     "ratio": 1 if is_int8 else (self.args.ratio or 0.8),
                     "sym": self.args.sym or False,
                     "group_size": -1 if is_int8 else self.args.group_size,
+                    "all_layers": None if is_int8 else self.args.all_layers,
                 }
 
             if self.args.weight_format in {"int4_sym_g128", "int4_asym_g128", "int4_sym_g64", "int4_asym_g64"}:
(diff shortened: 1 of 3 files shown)


tests/openvino/test_exporters_cli.py
```diff
@@ -84,14 +83,13 @@ class OVCLIExportTestCase(unittest.TestCase):
...
-    TEST_4BIT_CONFIGURATONS = []
-    for arch in SUPPORTED_4BIT_ARCHITECTURES:
-        for option in SUPPORTED_4BIT_OPTIONS:
-            TEST_4BIT_CONFIGURATONS.append([arch[0], arch[1], option])
+    TEST_4BIT_CONFIGURATONS = [
+        ("text-generation-with-past", "opt125m", "int4_sym_g128", 62, 86),
+        ("text-generation-with-past", "opt125m", "int4_asym_g128", 62, 86),
+        ("text-generation-with-past", "opt125m", "int4_sym_g64", 62, 86),
+        ("text-generation-with-past", "opt125m", "int4_asym_g64", 62, 86),
+        ("text-generation-with-past", "llama_awq", "int4 --ratio 1.0 --sym --group-size 16 --all-layers", 0, 32),
```
nikita-savelyevv, line 91: Had to test `--all-layers` on a different model because `opt125m` fails at compilation when all layers are quantized into INT4

optimum/commands/export/openvino.py
```diff
@@ -119,6 +119,14 @@ def parse_args_openvino(parser: "ArgumentParser"):
             "or ['conceptual_captions','laion/220k-GPT4Vision-captions-from-LIVIS','laion/filtered-wit'] for diffusion models."
         ),
     )
+    optional_group.add_argument(
+        "--all-layers",
+        action="store_true",
+        default=None,
+        help=(
+            "Whether embeddings and last MatMul layers should be compressed to a primary precision (usually, INT4)."
```
helena-intel, line 127: Is there a non-INT4 usecase? I see below that if it is provided, all_layers is set to `None if is_int8 else self.args.all_layers` so it seems like it's ignored for INT8? If it is only for INT4 it would be good to clarify that in the help message. And possibly make it mutually exclusive with INT8.

tests/openvino/test_exporters_cli.py
AlexKoff88: You can try to test it like this to keep the tests light: https://github.com/huggingface/optimum-intel/blob/6faf44562d762ca66bb1db93c9c2703e37cf5020/tests/openvino/test_quantization.py#L191

optimum/commands/export/openvino.py
AlexKoff88: This is for INT4 only.

nikita-savelyevv: Thanks for your comment! By default we compress those layers to INT8, even if number of bits is set to 4. This flag allow to compress those layer to INT4 as well. I've updated the description, hopefully it is more clear now (since currently the only supported primary precision is INT4, I've rephrased it a bit).

tests/openvino/test_exporters_cli.py
nikita-savelyevv: Thanks! I actually had the same idea, but the model here also has the same shortcoming. The reason the test does not fail there is because the compressed model is not being compiled. `llama_awq` is an alias for `HuggingFaceH4/tiny-random-LlamaForCausalLM` so this model is also small one.

AlexKoff88 (approved):

echarlaix (approved):
LGTM, thanks @nikita-savelyevv

PR #714, opened 2024-05-16
Fix llama and gemma modeling patching for openvino export

echarlaix:
Fix compatibility for transformers v4.41.0 

cc @helena-intel @eaidova

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -28,7 +28,6 @@ from optimum.exporters.onnx.model_configs import (
     UNetOnnxConfig,
     VaeDecoderOnnxConfig,
     VaeEncoderOnnxConfig,
-    Wav2Vec2OnnxConfig,
 )
 from optimum.exporters.tasks import TasksManager
 from optimum.utils import DEFAULT_DUMMY_SHAPES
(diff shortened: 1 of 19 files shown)


optimum/exporters/openvino/model_configs.py
```diff
@@ -90,6 +91,21 @@ def init_model_configs():
 register_in_tasks_manager = TasksManager.create_register("openvino", overwrite_existing=True)
 
 
+@register_in_tasks_manager(
```
eaidova, line 94: I do not see any difference with onnx version config. Why do you need to have it explicitly?

echarlaix: Added as `DEFAULT_ONNX_OPSET` now needs to be set to 14 (now that F.scaled_dot_product_attention is used by default for this model), will be removed before merging the PR as it'll be added in optimum directly

optimum/intel/openvino/trainer.py
```diff
@@ -153,8 +153,6 @@
         "{re}.*Embedding.*",
         "{re}.*add___.*",
         "{re}.*layer_norm_.*",
-        "{re}.*matmul_1",
-        "{re}.*__truediv__.*",
     ],
```
echarlaix, line 156: Removed to fix 
```
nncf.errors.ValidationError: No match has been found among the model operations for the following ignored/target scope definitions: - ignored_scope: ['{re}.*__truediv__.*', '{re}.*matmul_1']
```

echarlaix: not related to llama and gemma modeling patching so moving it to https://github.com/huggingface/optimum-intel/pull/724

PR #720, opened 2024-05-20
update codegen config for support codegen2

eaidova:
# What does this PR do?
update codegen config for support codegen2, added support qwen2moe and dbrx models

## Before submitting


optimum/exporters/openvino/model_configs.py
```diff
@@ -738,3 +750,75 @@ def patch_model_for_export(
...
+            self.sequence_length,
+            self.hidden_size // self.num_attention_heads,
+        )
+        return [
+            (
+                self.random_float_tensor(shape, framework=framework, dtype=float_dtype),
+                self.random_float_tensor(shape, framework=framework, dtype=float_dtype),
+            )
+            for _ in range(self.num_layers)
+        ]
```
echarlaix, line 801: why not use `MistralDummyPastKeyValuesGenerator` directly ?
```suggestion
```

echarlaix: (same question for `AquilaDummyPastKeyValuesGenerator`)

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1328,3 +1406,248 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    _dbrx_update_causal_mask = _dbrx_update_causal_mask_legacy
+
+
+class DBRXModelPatcher(DecoderModelPatcher):
+    def __enter__(self):
+        super().__enter__()
+        self._model.transformer._orig_update_causal_mask = self._model.transformer._update_causal_mask
+        self._model.transformer._update_causal_mask = types.MethodType(
+            _dbrx_update_causal_mask, self._model.transformer
+        )
```
echarlaix, line 1637: should this be applied for transformers >= v4.39 as for gemma / mistral https://github.com/huggingface/optimum-intel/blob/7a929e8d6da0dac4fbd8995add32a663e7b9afc5/optimum/exporters/openvino/model_patcher.py#L365 ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1328,3 +1406,248 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        if not is_tracing and torch.any(attention_mask != 1):
+            # Attend to all tokens in fully masked rows in the causal_mask, for example the relevant first rows when
+            # using left padding. This is required by F.scaled_dot_product_attention memory-efficient attention path.
+            # Details: https://github.com/pytorch/pytorch/issues/110213
+            causal_mask = AttentionMaskConverter._unmask_unattended(causal_mask, min_dtype)
+
+    return causal_mask
+
+
+def _dbrx_update_causal_mask_latest(
```
echarlaix, line 1538: could you add a comment to specify where the original code come from (https://github.com/huggingface/optimum-intel/blob/7a929e8d6da0dac4fbd8995add32a663e7b9afc5/optimum/exporters/openvino/model_patcher.py#L294) and specify in the function what differs from the original code

echarlaix (commented):
Thanks @eaidova

optimum/exporters/openvino/model_configs.py
eaidova: Aquila has difference from Mistral because parameter of num_key_value_heads can be not available in normalized config, we need to handle that in pkv generator. For dbrx I staied it during experiments (I had invalid model where key has different size of head_dim, so it can be removed

optimum/exporters/openvino/model_patcher.py
eaidova: dbrx was added only in transformers 4.40, so no need to check lower bound

optimum/exporters/openvino/model_configs.py
eaidova: removed

optimum/exporters/openvino/model_patcher.py
eaidova: added

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1328,3 +1406,259 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+                layer.attn._attn = types.MethodType(codegen_wrapped_scaled_dot_product, layer.attn)
+                layer.attn._orig_attn = orig_self_attn_fwd
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+        for layer in self._model.transformer.h:
+            if hasattr(layer.attn, "_orig_attn"):
+                layer.attn._attn = layer.attn._orig_attn
+
+
```
echarlaix, line 1431:
```suggestion
# adapted from https://github.com/huggingface/transformers/blob/v4.40.2/src/transformers/models/dbrx/modeling_dbrx.py#L763
```

optimum/exporters/openvino/model_patcher.py
echarlaix: could this be added for `_dbrx_experts_forward` and `_baichuan7b_attn_forward` as well ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1328,3 +1406,259 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    )
+    w2_chunked = self.mlp.w2.view(self.mlp.moe_num_experts, self.mlp.ffn_hidden_size, self.mlp.hidden_size).chunk(
+        self.moe_num_experts, dim=0
+    )
+    w1_chunked = [w1.squeeze(dim=0) for w1 in w1_chunked]
+    v1_chunked = [v1.squeeze(dim=0) for v1 in v1_chunked]
+    w2_chunked = [w2.squeeze(dim=0) for w2 in w2_chunked]
+    for expert_idx in range(0, self.moe_num_experts):
+        topk_idx, token_idx = torch.where(expert_mask[expert_idx])
+
```
echarlaix, line 1455: could you add a comment on why 
```
if token_idx.shape[0] == 0:
    continue
```
was removed

optimum/exporters/openvino/model_patcher.py
```diff
@@ -665,6 +673,71 @@ def _baichuan13b_atten_forward(
     return attn_output, attn_weights, past_key_value
 
 
+def _baichuan7b_attn_forward(
```
echarlaix, line 676: not sure where the original code is from but something like 
```suggestion
# Adapted from https://huggingface.co/baichuan-inc/Baichuan-7B/blob/262c8cb58b6d3615c208d9230baa869fddee2adb/modeling_baichuan.py#L181
def _baichuan7b_attn_forward(
```
would be nice

echarlaix (approved):

optimum/exporters/openvino/model_patcher.py
eaidova: added

PR #721, opened 2024-05-21
[OV] Move data-driven quantization after model export for text-generation models

nikita-savelyevv:
# What does this PR do?

In order to apply data-driven weights compression, an instance of `OVModelForCausalLM` is [required](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/quantization.py#L692). It however is not available during quantization applied at model export ([here](https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/convert.py#L89)). 

That's why in this PR some logic is added so that such case is processed separately after model is exported. This results in some save/load overhead, but compared to runtime of data-drive weights compression it should be negligible. Worth to note, data-free compression is still applied during export resulting in no additional overhead.

## Before submitting


optimum/commands/export/openvino.py
```diff
@@ -305,3 +348,19 @@ def run(self):
...
+                quantization_config.tokenizer = quantization_config.tokenizer or str(self.args.output)
+                # TODO: set save_directory=self.args.output once OV is updated to 2024.3
+                quantizer.quantize(ov_config=OVConfig(quantization_config=quantization_config))
+                with tempfile.TemporaryDirectory() as temp_dir:
+                    import shutil
+
+                    model.save_pretrained(temp_dir)
+                    ov_config.save_pretrained(self.args.output)
+                    shutil.copy(f"{temp_dir}/openvino_model.xml", f"{self.args.output}/openvino_model.xml")
+                    shutil.copy(f"{temp_dir}/openvino_model.bin", f"{self.args.output}/openvino_model.bin")
```
nikita-savelyevv, line 366: Had to add this workaround because OpenVINO does not currently support saving into the same location where the model is loaded from (ticket 110054). This is expected to be fixed in OV 2024.3.

nikita-savelyevv:

optimum/commands/export/openvino.py
AlexKoff88: @eaidova, please take a look

optimum/commands/export/openvino.py
```diff
@@ -305,3 +353,21 @@ def run(self):
...
+                )
+                quantizer = OVQuantizer(model)
+                quantization_config.tokenizer = quantization_config.tokenizer or str(self.args.output)
+                # TODO: set save_directory=self.args.output once OV is updated to 2024.3
+                quantizer.quantize(ov_config=OVConfig(quantization_config=quantization_config))
+                with tempfile.TemporaryDirectory() as temp_dir:
+                    import shutil
+
+                    model.save_pretrained(temp_dir)
+                    ov_config.save_pretrained(self.args.output)
```
AlexKoff88, line 371: @nikita-savelyevv, does this workaround with tmp folder mean that we cannot save the model at the same path but can copy files there? Looks a bit strange.

optimum/commands/export/openvino.py
eaidova: maybe we can intorduce model_name parameter for from_pretrained/save_pretrained methods? That will allow having both models in the same dir (in the same time it maybe useful for loading model if IR saved by different tools or renamed). Or we can try disable mmap via ov_config for now (it should help with saving in the same location)

optimum/commands/export/openvino.py
nikita-savelyevv: Yes, I had troubles saving to the same location, but copying to that location works fine

optimum/commands/export/openvino.py
nikita-savelyevv: > @eaidova thank you for your suggestion! I've replaced saving to temporary directory with disabling mmap

For some reason when doing it this way I observe that a significant amount of additional memory is allocated. The amount roughly equals the model size which is rather significant. I guess I'll revert these changes for now.

AlexKoff88 (approved):

tests/openvino/test_exporters_cli.py
```diff
@@ -197,17 +205,19 @@ def test_exporters_cli_hybrid_quantization(self, model_type: str, exp_num_fq: in
...
                 check=True,
+                capture_output=True,
             )
             model_kwargs = {"use_cache": task.endswith("with-past")} if "generation" in task else {}
             model = eval(_HEAD_TO_AUTOMODELS[task.replace("-with-past", "")]).from_pretrained(tmpdir, **model_kwargs)
 
             _, num_int8, num_int4 = get_num_quantized_nodes(model)
             self.assertEqual(expected_int8, num_int8)
             self.assertEqual(expected_int4, num_int4)
+            self.assertTrue("--quant-method awq" not in option or b"Applying AWQ" in result.stdout)
```
echarlaix, line 220: should it be 
```suggestion
            self.assertTrue("--awq" not in option or b"Applying AWQ" in result.stdout)
```

optimum/commands/export/openvino.py
```diff
@@ -305,3 +357,25 @@ def run(self):
...
+                    quantizer.quantize(ov_config=OVConfig(quantization_config=quantization_config))
+                    with tempfile.TemporaryDirectory() as temp_dir:
+                        model.save_pretrained(temp_dir)
+                        ov_config.save_pretrained(self.args.output)
+                        shutil.copy(f"{temp_dir}/openvino_model.xml", f"{self.args.output}/openvino_model.xml")
+                        shutil.copy(f"{temp_dir}/openvino_model.bin", f"{self.args.output}/openvino_model.bin")
+                except Exception as e:
+                    # Delete non-compressed model if compression failed for some reason
+                    shutil.rmtree(str(self.args.output))
+                    raise e
```
echarlaix, line 381: Why not exporting + applying quantization using `OVModelForCausalLM` directly (and not calling `main_export` for this specific case) ?

echarlaix: If possible then it could actually make sense to do this for all models (as it's already the case for SD models)

tests/openvino/test_exporters_cli.py
nikita-savelyevv: Thanks! Fixed

optimum/commands/export/openvino.py
nikita-savelyevv: > Why not exporting + applying quantization using OVModelForCausalLM directly (and not calling main_export for this specific case) ?

Thanks for the suggestion! This is indeed better.

> If possible then it could actually make sense to do this for all models (as it's already the case for SD models)

It would be more convenient from code maintenance side. But compared to calling `main_export()` with provided `quantization_config`, when we export and quantize through `.from_pretrained()` there is an additional model saving step to temporary directory which leads to an additional overhead, especially for large models. I think we should if avoid this if possible.

optimum/commands/export/openvino.py
```diff
@@ -270,25 +304,16 @@ def run(self):
...
-            from ...exporters.openvino.convert import export_tokenizer
-
-            output = Path(self.args.output)
-            tokenizer = getattr(model, "tokenizer", None)
-            if tokenizer is not None:
-                export_tokenizer(tokenizer, output / "tokenizer")
-
-            tokenizer_2 = getattr(model, "tokenizer_2", None)
-            if tokenizer_2 is not None:
-                export_tokenizer(tokenizer_2, output / "tokenizer_2")
```
nikita-savelyevv, line 291: I extended this logic to be applied for text-generation models too, in case converting through `from_pretrained()`. @apaniukov is that correct?

optimum/commands/export/openvino.py
```diff
@@ -270,25 +304,16 @@ def run(self):
...
+            # To quantize a text-generation model with a dataset, an instantiated OVModelForCausalLM is required
+            model = OVModelForCausalLM.from_pretrained(
+                self.args.model, export=True, quantization_config=quantization_config
```
echarlaix, line 314:
```suggestion
                self.args.model, export=True, quantization_config=quantization_config, stateful=not self.args.disable_stateful
```

echarlaix (approved):
Looks good, thanks @nikita-savelyevv

optimum/commands/export/openvino.py
echarlaix: looks good !

optimum/commands/export/openvino.py
nikita-savelyevv: Thanks, done

optimum/commands/export/openvino.py
```diff
@@ -281,14 +312,41 @@ def run(self):
...
+                    self.args.model, trust_remote_code=self.args.trust_remote_code
+                )
+                tokenizer.save_pretrained(self.args.output)
+            except Exception:
+                logger.warning("Could not save tokenizer")
+
+            if tokenizer and not self.args.disable_convert_tokenizer:
+                from ...exporters.openvino.convert import export_tokenizer
+
+                export_tokenizer(tokenizer, self.args.output / "tokenizer")
```
nikita-savelyevv, line 349: Had to add such logic because otherwise tokenizer ends up being saved to a temporary directory.

optimum/commands/export/openvino.py
```diff
@@ -218,6 +249,10 @@ def run(self):
                     "sym": self.args.sym or False,
                     "group_size": -1 if is_int8 else self.args.group_size,
                     "all_layers": None if is_int8 else self.args.all_layers,
+                    "dataset": self.args.dataset,
+                    "num_samples": self.args.num_samples,
+                    "quant_method": "awq" if self.args.awq else None,
```
echarlaix, line 254: not sure if it should be `QuantizationMethod.AWQ` instead of "awq" or if the configuration takes care of this https://github.com/huggingface/optimum-intel/blob/52875b992ac2248e49063adcf8116f1fb39cacd2/optimum/intel/openvino/quantization.py#L822

optimum/commands/export/openvino.py
echarlaix: to simplify the code a bit what do you think about calling directly `maybe_load_preprocessors` https://github.com/huggingface/optimum-intel/blob/8dd2a895016b3a9db7a628294d13ccf1ac25a1e4/optimum/exporters/openvino/__main__.py#L344

and to save the tokenizer we can also use `maybe_save_preprocessors` https://github.com/huggingface/optimum-intel/blob/8dd2a895016b3a9db7a628294d13ccf1ac25a1e4/optimum/exporters/openvino/convert.py#L625

also we could create a function https://github.com/huggingface/optimum-intel/blob/8dd2a895016b3a9db7a628294d13ccf1ac25a1e4/optimum/exporters/openvino/__main__.py#L364-L390 that could be re-used here

optimum/commands/export/openvino.py
nikita-savelyevv: I hesitated to do it this way because it would require to introduce dependency on `transformers` in this file in order to import `QuantizationMethod`. But now I see that `transformers` is a general requirement of `optimum-intel` so it should be fine I guess.

optimum/commands/export/openvino.py
nikita-savelyevv: Thanks! Done

echarlaix (approved):
Looks great, thanks a lot @nikita-savelyevv

PR #723, opened 2024-05-22
dump model before compress cli

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -19,11 +19,12 @@ import logging
 import os
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
+import tempfile
 
 import onnx
 from transformers.utils import is_tf_available, is_torch_available
 
-from openvino.runtime import Model, PartialShape, save_model
+from openvino.runtime import Model, PartialShape, save_model, Core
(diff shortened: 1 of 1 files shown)


optimum/exporters/openvino/convert.py
```diff
@@ -412,11 +414,24 @@ def ts_patched_forward(*args, **kwargs):
 
         if stateful:
             patch_stateful(model.config, ov_model)
+        if ov_config.quantization_config:
```
AlexKoff88, line 417: there is another PR where we probably solve this problem: https://github.com/huggingface/optimum-intel/pull/721/files

nikita-savelyevv: Yes, seems like the changes are independent.

@eaidova Just out of curiosity, do you have numbers for how much memory can be saved with this approach?

AlexKoff88: thanks, @eaidova. @nikita-savelyevv, can you please adopt changes from this PR?

nikita-savelyevv: @AlexKoff88 I don't think this applies to my case because I don't have access to the PyTorch or OpenVINO model objects after the `main_export` call. So I can't delete them.

Or do you mean to copy changes from this PR to my PR? If so, in my opinion these should be added separately.

PR #725, opened 2024-05-23
ipex 2.3 released

jiqing-feng:
Hi @echarlaix . I make some changes to the llama model since the ipex 2.3 is released. The API name has changed, and the assisted decoding cannot support for now.

### DIFF
--- a/.github/workflows/test_ipex.yml
+++ b/.github/workflows/test_ipex.yml
@@ -18,6 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9]
+        transformers-version: [4.38.2, 4.41.2]
         os: [ubuntu-latest]
 
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test_ipex.yml
+++ b/.github/workflows/test_ipex.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9]
-        transformers-version: [4.38.2, 4.41.2]
+        transformers-version: [4.38.0, 4.41.2]
         os: [ubuntu-latest]
 
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test_ipex.yml
+++ b/.github/workflows/test_ipex.yml
@@ -33,6 +33,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install torch torchaudio torchvision --extra-index-url https://download.pytorch.org/whl/cpu
         pip install .[ipex,tests]
+        pip install transformers==${{ matrix.transformers-version }}
     - name: Test with Pytest
       run: |
         pytest tests/ipex/
--- a/.github/workflows/test_ipex.yml
+++ b/.github/workflows/test_ipex.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9]
-        transformers-version: [4.38.0, 4.41.2]
+        transformers-version: [4.39.0, 4.41.2]
         os: [ubuntu-latest]
 
     runs-on: ${{ matrix.os }}
--- a/optimum/exporters/ipex/model_patcher.py
+++ b/optimum/exporters/ipex/model_patcher.py
@@ -33,7 +33,6 @@ from .modeling_utils import (
 
 _IPEX_EXPORTED_ARCH = ("LlamaForCausalLM",)
 _IPEX_EXPORTED_TASK = ("text-generation",)
-_IPEX_EXPORTED_GENERATION_METHODS = ("sample", "greedy_search", "beam_sample", "beam_search")
 
 
 def convert_func(m, func_name, new_function):
--- a/optimum/exporters/ipex/model_patcher.py
+++ b/optimum/exporters/ipex/model_patcher.py
@@ -25,6 +25,7 @@ from transformers.models.llama.modeling_llama import (
 from optimum.intel.utils.import_utils import is_ipex_version
 
 from .modeling_utils import (
+    _IPEX_MINIMUM_VERSION_FOR_PATCHING,
     _IPEXLlamaDecoderLayerRef,
     _llama_attn_forward,
     _llama_layer_norm_forward,
@@ -34,10 +35,6 @@ from .modeling_utils import (
 
 logger = logging.getLogger(__name__)
 
-# Please also update in the setup.py and .github/workflows/test_ipex.yml if you change the transformers version
-_TRANSFORMERS_MIN_VERSION = "4.38.0"
-_TRANSFORMERS_MAX_VERSION = "4.41.2"
-_IPEX_MINIMUM_VERSION_FOR_PATCHING = "2.3.0"
 _IPEX_EXPORTED_ARCH = ("LlamaForCausalLM",)
 _IPEX_EXPORTED_TASK = ("text-generation",)
 _IPEX_EXPORTED_GENERATION_METHODS = ("sample", "greedy_search", "beam_sample", "beam_search")
--- a/optimum/exporters/ipex/model_patcher.py
+++ b/optimum/exporters/ipex/model_patcher.py
@@ -34,6 +34,10 @@ from .modeling_utils import (
 
 logger = logging.getLogger(__name__)
 
+# Please also update in the setup.py and .github/workflows/test_ipex.yml if you change the transformers version
+_TRANSFORMERS_MIN_VERSION = "4.38.0"
+_TRANSFORMERS_MAX_VERSION = "4.41.2"
+_IPEX_MINIMUM_VERSION_FOR_PATCHING = "2.3.0"
 _IPEX_EXPORTED_ARCH = ("LlamaForCausalLM",)
 _IPEX_EXPORTED_TASK = ("text-generation",)
 _IPEX_EXPORTED_GENERATION_METHODS = ("sample", "greedy_search", "beam_sample", "beam_search")
@@ -67,8 +71,10 @@ def patch_op(m, target_m, new_op_name, new_op):
 
 
 def _patch_llama_model(model):
-    if is_ipex_version("<", "2.3.0"):
-        raise ImportError("Only ipex version >= 2.3.0 supports RotaryEmbedding and IndirectAccessKVCacheAttention")
+    if is_ipex_version("<", _IPEX_MINIMUM_VERSION_FOR_PATCHING):
+        raise ImportError(
+            f"Only ipex version >= {_IPEX_MINIMUM_VERSION_FOR_PATCHING} supports RotaryEmbedding and IndirectAccessKVCacheAttention"
+        )
 
     logger.warning(
         "Only greedy search and beam search with do_sample=True/False are verified for the patched model. It may have risk if other generation methods are applied"
--- a/optimum/exporters/ipex/model_patcher.py
+++ b/optimum/exporters/ipex/model_patcher.py
@@ -12,6 +12,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import logging
+
 from transformers.models.llama.modeling_llama import (
     LlamaAttention,
     LlamaDecoderLayer,
@@ -29,6 +31,8 @@ from .modeling_utils import (
     _llama_model_forward,
 )
 
+
+logger = logging.getLogger(__name__)
 
 _IPEX_EXPORTED_ARCH = ("LlamaForCausalLM",)
 _IPEX_EXPORTED_TASK = ("text-generation",)
@@ -64,6 +68,10 @@ def patch_op(m, target_m, new_op_name, new_op):
 def _patch_llama_model(model):
     if is_ipex_version("<", "2.3.0"):
         raise ImportError("Only ipex version >= 2.3.0 supports RotaryEmbedding and IndirectAccessKVCacheAttention")
+
+    logger.warning(
+        "Only greedy search and beam search with do_sample=True/False are verified for the patched model. It may have risk if other generation methods are applied"
+    )
 
     from intel_extension_for_pytorch.llm.modules import IndirectAccessKVCacheAttention, RotaryEmbedding
 
--- a/optimum/exporters/ipex/model_patcher.py
+++ b/optimum/exporters/ipex/model_patcher.py
@@ -62,10 +62,10 @@ def patch_op(m, target_m, new_op_name, new_op):
 
 
 def _patch_llama_model(model):
-    if is_ipex_version("<", "2.5.0"):
-        raise ImportError("Only ipex version > 2.3.0 supports RotaryEmbedding and IndirectAccessKVCache")
+    if is_ipex_version("<", "2.3.0"):
+        raise ImportError("Only ipex version >= 2.3.0 supports RotaryEmbedding and IndirectAccessKVCacheAttention")
 
-    from intel_extension_for_pytorch.llm.modules import IndirectAccessKVCache, RotaryEmbedding
+    from intel_extension_for_pytorch.llm.modules import IndirectAccessKVCacheAttention, RotaryEmbedding
 
     ipex_rope = RotaryEmbedding(
         model.config.max_position_embeddings,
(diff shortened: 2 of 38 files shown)


tests/ipex/test_modeling.py
```diff
@@ -259,17 +253,19 @@ def test_ipex_patching_beam_search(self, test_name, model_arch, use_cache):
...
+            GenerationConfig(max_new_tokens=4, do_sample=True, top_p=1.0, top_k=5, penalty_alpha=0.6),
             GenerationConfig(max_new_tokens=4, do_sample=True, top_p=0.9, top_k=0),
         )
         for text in texts:
             tokens = tokenizer(text, padding=True, return_tensors="pt")
             for generation_config in generation_configs:
                 outputs = model.generate(**tokens, generation_config=generation_config)
-                transformers_outputs = trasnformers_model.generate(**tokens, generation_config=generation_config)
                 self.assertIsInstance(outputs, torch.Tensor)
-                self.assertEqual(outputs, transformers_outputs)
```
echarlaix, line 264: why is this removed ?

tests/ipex/test_modeling.py
```diff
@@ -220,6 +219,10 @@ def test_pipeline(self, model_arch):
         self.assertTrue(all("This is a sample" in item["generated_text"] for item in outputs))
 
     @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    @unittest.skipIf(
+        is_ipex_version(">=", "2.3.0"),
+        reason="CPU IPEXModel does not support assisted decoding when ipex version >= 2.3.0",
+    )
```
echarlaix, line 225: could you explain why assisted decoding is deprecated with v2.3.0?

tests/ipex/test_modeling.py
```diff
@@ -235,21 +238,12 @@ def test_assisted_decoding(self, model_arch):
...
-            {
-                "model_arch": IPEX_PATCHED_SUPPORTED_ARCHITECTURES,
-                "use_cache": [True, False],
-            }
-        )
-    )
-    @unittest.skipIf(is_ipex_version("<", "2.5.0"), reason="Only ipex version > 2.3.0 supports ipex model patching")
-    def test_ipex_patching_beam_search(self, test_name, model_arch, use_cache):
+    @parameterized.expand(IPEX_PATCHED_SUPPORTED_ARCHITECTURES)
+    @unittest.skipIf(is_ipex_version("<", "2.3.0"), reason="Only ipex version >= 2.3.0 supports ipex model patching")
```
echarlaix, line 242: I think we should keep the test for both `use_cache` option, wdyt ?

jiqing-feng:
Hi @echarlaix . Thanks for your review. The core optimization for llama model is the IAKV which will change the shape of kv-cache, so functions in assisted decoding like [crop_past_key_values](https://github.com/huggingface/transformers/blob/main/src/transformers/generation/candidate_generator.py#L353) cannot be used. I also want to skip the test without kv-cache because the llama model is designed for kv-cache optimization. I kept all the base optimization model tests and only skip the 2 tests for llama2 model which used the high optimization.

tests/ipex/test_modeling.py
jiqing-feng: The IPEXModel has some fused ops which will cause float point precision loss, so we cannot guarantee the results could exactly the same. Besides, the generation config set `do_sample=True`.

tests/ipex/test_modeling.py
```diff
@@ -259,16 +257,13 @@ def test_ipex_patching_beam_search(self, test_name, model_arch, use_cache):
             GenerationConfig(max_new_tokens=4, num_beams=4, do_sample=True),
             GenerationConfig(max_new_tokens=4, num_beams=8, do_sample=True),
             GenerationConfig(max_new_tokens=4, num_beams=32, do_sample=True),
-            GenerationConfig(max_new_tokens=4, do_sample=not use_cache, top_p=1.0, top_k=5, penalty_alpha=0.6),
```
jiqing-feng, line 262: The IPEXModel is not supported [_contrastive_search](https://github.com/jiqing-feng/transformers/blob/main/src/transformers/generation/utils.py#L1993) for now, we will try to enable it in the future.

tests/ipex/test_modeling.py
jiqing-feng: Yes, I have some little changes in the llama model to enable without cache forward.

tests/ipex/test_modeling.py
jiqing-feng: See [here](https://github.com/huggingface/optimum-intel/pull/725#issuecomment-2138517113)

optimum/exporters/ipex/modeling_utils.py
```diff
@@ -288,12 +290,15 @@ def forward(
...
-
-        if not self.distributed:
-            hidden_states = self.mlp_linear_add(mlp_gate, residual)
+        if hasattr(self, "linear_silu_mul"):
+            mlp_gate = self.linear_silu_mul(hidden_states)
+            if hasattr(self, "mlp_linear_add"):
+                hidden_states = self.mlp_linear_add(mlp_gate, residual)
+            else:
+                hidden_states = self.mlp.down_proj(mlp_gate)
+                hidden_states = residual + hidden_states
```
echarlaix, line 299: We might need to add tests for different transformers version to ensure compatibility for this patching or otherwise limit the transformers version to the latest version, what do you think ?

optimum/intel/ipex/modeling_base.py
```diff
@@ -95,6 +98,8 @@ def ipex_jit_trace(model, task, use_cache):
     if "past_key_values" in sample_inputs and use_cache:
         # Make sure the model will output past_key_values in generation tasks
         model.config.use_cache = True
+    else:
+        model.config.use_cache = False
```
echarlaix, line 102: should it be :
```suggestion
    model.config.use_cache = past_key_values" in sample_inputs 
```

tests/ipex/test_modeling.py
echarlaix: I see, could we then add a warning to state this is not supported (at least for transformers >= v4.39.0) and then upgrade it in the [setup.py](https://github.com/huggingface/optimum-intel/blob/v1.17.0/setup.py#L65) maybe ?

tests/ipex/test_modeling.py
echarlaix: I'd expect the final output to be exactly the same, how much variations could we expect here ? And how should we evaluate that the export and generation steps are working as expected without being able to compare it with the original model ?

jiqing-feng:
Hi @echarlaix. Reply to [it](https://github.com/huggingface/optimum-intel/pull/725#discussion_r1621209279). I have set `do_sample=False` so you can compare the results between the ipex model and the original model.

jiqing-feng:
Reply to [it](https://github.com/huggingface/optimum-intel/pull/725#discussion_r1621202993). It is not related to the transformer version; it's the current limitation for IAKV. I have added the warning to clarify that only greedy search and beam search are verified for the patched model. Please have a look, thx!

optimum/exporters/ipex/modeling_utils.py
jiqing-feng: Has limit the transformers version.

tests/ipex/test_pipelines.py
```diff
@@ -56,7 +56,6 @@ class PipelinesIntegrationTest(unittest.TestCase):
         "gpt2",
         "gpt_neo",
         "gpt_neox",
-        "llama",
```
echarlaix, line 59: why removing the pipeline test for llama models ?

jiqing-feng: There are limitation for head_size of IAKV which is >=64.

jiqing-feng:
1. Check the generation config and raise an error if it does not support it.
2. Transformers version.
3. Test re-load for patched model (llama).

.github/workflows/test_ipex.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9]
+        transformers-version: [4.38.2, 4.41.2]
```
echarlaix, line 21: we also need to install this specific version of transformers so with something like `pip install transformers==${{ matrix.transformers }}` otherwise the latest version will always be installed

optimum/exporters/ipex/model_patcher.py
```diff
@@ -62,18 +66,22 @@ def patch_op(m, target_m, new_op_name, new_op):
 
 
 def _patch_llama_model(model):
-    if is_ipex_version("<", "2.5.0"):
-        raise ImportError("Only ipex version > 2.3.0 supports RotaryEmbedding and IndirectAccessKVCache")
+    if is_ipex_version("<", "2.3.0"):
```
echarlaix, line 69: could you add a variable for this so that we can use it everywhere (also in tests)
```python
_IPEX_MINIMUM_VERSION_FOR_PATCHING = "2.3.0"
```

optimum/exporters/ipex/modeling_utils.py
```diff
@@ -219,8 +225,10 @@ def _llama_model_forward(
 # Adapted from https://github.com/huggingface/transformers/blob/v4.38.2/src/transformers/models/llama/modeling_llama.py#L694
 class _IPEXLlamaDecoderLayerRef(nn.Module):
     def __init__(self, module, config, distributed=False):
-        if is_ipex_version("<", "2.5.0"):
+        if is_ipex_version("<", "2.3.0"):
             raise ImportError("Only ipex version > 2.3.0 supports Linear2SiluMul and LinearAdd")
+        if is_transformers_version("<", "4.38.2") or is_transformers_version(">", "4.41.2"):
+            raise ImportError("Only transformers versions 4.38.2 ~ 4.41.2 are verified.")
```
echarlaix, line 231: same could you add variables like TRANSFORMERS_MIN_VERSION and TRANSFORMERS_MAX_VERSION, also we can add a comment to specify that should when updated test_ipex.yml and setup.py should also be

setup.py
```diff
@@ -62,7 +62,7 @@
     "neural-compressor": ["neural-compressor>=2.2.0", "onnxruntime<1.15.0", "accelerate"],
     "openvino": ["openvino>=2023.3", "nncf>=2.10.0", "openvino-tokenizers[transformers]"],
     "nncf": ["nncf>=2.10.0"],
-    "ipex": ["intel-extension-for-pytorch", "transformers>=4.36.0,<4.39.0"],
+    "ipex": ["intel-extension-for-pytorch", "transformers>=4.36.0,<=4.41.2"],
```
echarlaix, line 65: in the tests I see the minimum tested version is v4.38.2 so we need to update it here as well 
```suggestion
    "ipex": ["intel-extension-for-pytorch", "transformers>=4.38.2,<=4.41.2"],
```

.github/workflows/test_ipex.yml
echarlaix: also could we use v4.38.0  directly
```suggestion
        transformers-version: [4.38.0, 4.41.2]
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -92,9 +95,7 @@ def ipex_jit_trace(model, task, use_cache):
 
     model.config.return_dict = False
 
-    if "past_key_values" in sample_inputs and use_cache:
-        # Make sure the model will output past_key_values in generation tasks
-        model.config.use_cache = True
+    model.config.use_cache = "past_key_values" in sample_inputs
```
echarlaix, line 98: also in my previous suggestion I missed that this should be done only for generation models so I think we can do something like :
```suggestion
    if "past_key_values" in sample_inputs:
        model.config.use_cache = True
```
what do you think ?

jiqing-feng:
Hi @echarlaix . I have finished all your required changes. Could you please take a look at the failed ipex CI? It is a weird import error.

jiqing-feng:
> Hi @echarlaix . I have finished all your required changes. Could you please take a look at the failed ipex CI? It is a weird import error.

Fixed!

optimum/intel/ipex/modeling_base.py
```diff
@@ -561,6 +590,18 @@ def forward(
 
         return CausalLMOutputWithPast(logits=logits, past_key_values=past_key_values)
 
+    def _prepare_generation_config(
+        self, generation_config: Optional[GenerationConfig], **kwargs: Dict
+    ) -> Tuple[GenerationConfig, Dict]:
+        generation_config, model_kwargs = self.model_cls._prepare_generation_config(self, generation_config, **kwargs)
```
echarlaix, line 596:
```suggestion
        generation_config, model_kwargs = super()._prepare_generation_config(generation_config, **kwargs)
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -561,6 +590,18 @@ def forward(
...
+        generation_method = generation_config.get_generation_mode(kwargs.get("assistant_model", None)).value
```
echarlaix, line 597: assistant_model will not be part of the kwargs when calling `generate` https://github.com/huggingface/transformers/blob/bd5091df8db7cea1a9f94f797fc11487f840ade1/src/transformers/generation/utils.py#L1628 so I think we can remove
 ```suggestion
        generation_method = generation_config.get_generation_mode().value
```

Comment (echarlaix) [reply optimum/intel/ipex/modeling_base.py:597]: also assisted decoding is not supported for patched model is this correct? if yes we can also add a warning when given when calling `generate`

Comment (jiqing-feng) [reply optimum/intel/ipex/modeling_base.py:597]: But the assistant model is required when determine generation mode, see [here](https://github.com/huggingface/transformers/blob/bd5091df8db7cea1a9f94f797fc11487f840ade1/src/transformers/generation/utils.py#L1777-L1778)

File: .github/workflows/test_ipex.yml [hunk 15]
```diff
@@ -18,6 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9]
+        transformers-version: [4.38.0, 4.41.2]
```
echarlaix, line 21:
```suggestion
        transformers-version: [4.39.0, 4.41.2]
```

optimum/exporters/ipex/modeling_utils.py
```diff
@@ -19,9 +19,15 @@
...
 from transformers.modeling_outputs import BaseModelOutputWithPast
-from transformers.models.llama.modeling_llama import repeat_kv
+from transformers.models.llama.modeling_llama import apply_rotary_pos_emb, repeat_kv
 
-from optimum.intel.utils.import_utils import is_ipex_version
+from optimum.intel.utils.import_utils import is_ipex_version, is_transformers_version
+
+
+# Please also update in the setup.py and .github/workflows/test_ipex.yml if you change the transformers version
+_TRANSFORMERS_MIN_VERSION = "4.38.0"
```
echarlaix, line 28:
```suggestion
_TRANSFORMERS_MIN_VERSION = "4.39.0"
```

setup.py
```diff
@@ -62,7 +62,7 @@
     "neural-compressor": ["neural-compressor>=2.2.0", "onnxruntime<1.15.0", "accelerate"],
     "openvino": ["openvino>=2023.3", "nncf>=2.10.0", "openvino-tokenizers[transformers]"],
     "nncf": ["nncf>=2.10.0"],
-    "ipex": ["intel-extension-for-pytorch", "transformers>=4.36.0,<4.39.0"],
+    "ipex": ["intel-extension-for-pytorch", "transformers>=4.38.0,<=4.41.2"],
```
echarlaix, line 65:
```suggestion
    "ipex": ["intel-extension-for-pytorch", "transformers>=4.39.0,<=4.41.2"],
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -561,6 +586,25 @@ def forward(
...
+        generation_config, model_kwargs = super()._prepare_generation_config(generation_config, **kwargs)
+        generation_method = generation_config.get_generation_mode().value
+        if generation_method not in _IPEX_EXPORTED_GENERATION_METHODS:
+            raise ValueError(
+                f"The generation method {generation_method} is not supported for IPEXModelForCausalLM for now, support methods are {_IPEX_EXPORTED_GENERATION_METHODS}"
+            )
+
+        return generation_config, model_kwargs
+
+    def generate(self, **kwargs):
```
echarlaix, line 601:
```suggestion
    def generate(self, *args, **kwargs):
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -561,6 +586,25 @@ def forward(
...
+        if self._is_ipex_exported and kwargs.get("assistant_model", None):
+            raise ValueError(
+                f"Assisted decoding is not supported for patched models for now, support methods are {_IPEX_EXPORTED_GENERATION_METHODS}"
+            )
+        return super().generate(**kwargs)
```
echarlaix, line 606:
```suggestion
        return super().generate(*args, **kwargs)
```

PR #727, opened 2024-05-23
Fix nncf quantization for decoder models

echarlaix:
Added a fix to be able to quantize instances of `OVModelForCausalLM` using the quantizer

cc @nikita-savelyevv

### DIFF
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -42,7 +42,7 @@ from ..utils.import_utils import is_nncf_available
 from ..utils.modeling_utils import MULTI_QUERY_ATTN_MODELS
 from .configuration import _DEFAULT_4BIT_CONFIGS, OVConfig, OVWeightQuantizationConfig, _check_default_4bit_configs
 from .modeling import _TOKENIZER_FOR_DOC, INPUTS_DOCSTRING, MODEL_START_DOCSTRING, OVModel
-from .utils import ONNX_WEIGHTS_NAME, OV_XML_FILE_NAME, STR_TO_OV_TYPE
+from .utils import ONNX_WEIGHTS_NAME, OV_TO_NP_TYPE, OV_XML_FILE_NAME, STR_TO_OV_TYPE
 
 
 if TYPE_CHECKING:
@@ -409,6 +409,7 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
             elif self.use_cache:
                 for input_name in self.key_value_input_names:
                     model_inputs = self.model.input(input_name)
+                    dtype = OV_TO_NP_TYPE[model_inputs.get_element_type().get_type_name()]
                     shape = model_inputs.get_partial_shape()
                     if self.config.model_type == "chatglm":
                         shape[0] = 0
@@ -419,7 +420,7 @@ class OVModelForCausalLM(OVBaseDecoderModel, GenerationMixin):
                             shape[2] = 0
                         else:
                             shape[1] = 0
-                    inputs[input_name] = Tensor(model_inputs.get_element_type(), shape.get_shape())
+                    inputs[input_name] = np.empty([dim.get_length() for dim in shape], dtype=dtype)
         else:
             # past_key_values are not used explicitly, instead they are handled inside the model
             if past_key_values is None:
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -348,11 +348,6 @@ class OVQuantizer(OptimumQuantizer):
                     data_collator=data_collator,
                 )
                 if self.model.export_feature == "text-generation" and self.model.use_cache:
-                    if self.model.stateful:
-                        raise ValueError(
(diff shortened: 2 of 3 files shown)


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -419,7 +420,7 @@ def prepare_inputs(
                             shape[2] = 0
                         else:
                             shape[1] = 0
-                    inputs[input_name] = Tensor(model_inputs.get_element_type(), shape.get_shape())
+                    inputs[input_name] = np.empty([dim.get_length() for dim in shape], dtype=dtype)
```
echarlaix, line 423: Looks like instances of `openvino.runtime.Tensor` cannot be pickled which is needed for the fix added in https://github.com/huggingface/optimum-intel/pull/632 for the `InferRequestWrapper`

AlexKoff88: I hope it will not break the tracing of models. cc'ed @eaidova, @slyalin.

tests/openvino/test_quantization.py
```diff
@@ -75,7 +75,7 @@
 class OVQuantizerTest(unittest.TestCase):
     SUPPORTED_ARCHITECTURES_WITH_EXPECTED_QUANTIZED_MATMULS = (
         (OVModelForSequenceClassification, "bert", 32, 35),
-        # (OVModelForCausalLM, "gpt2", 41, 23),
+        (OVModelForCausalLM, "gpt2", 41, 3),
```
AlexKoff88, line 78: I don't feel like the number int8 constants are correct in this test. @nikita-savelyevv, can you please take a look?

optimum/intel/openvino/quantization.py
```diff
@@ -347,8 +347,12 @@ def _quantize_ovbasemodel(
                     remove_unused_columns=remove_unused_columns,
                     data_collator=data_collator,
                 )
-
                 if self.model.export_feature == "text-generation" and self.model.use_cache:
+                    if self.model.stateful:
+                        raise ValueError(
+                            "Currently, only non-stateful models are supported for quantization, to export such model you can do : `.from_pretrained(export=True, stateful=False)`"
+                        )
+
```
nikita-savelyevv, line 355: Could you please explain why stateful models are not supported? On my side the test for `gpt2` model with removed `stateful=False`, quantizes model without any errors.

optimum/intel/openvino/quantization.py
```diff
@@ -705,15 +710,16 @@ def _prepare_text_generation_dataset(
 
         num_samples = quantization_config.num_samples or 200
 
-        self.model.request = InferRequestWrapper(self.model.model.request, collected_inputs)
```
nikita-savelyevv, line 708: Thanks for the catch! Could you please remove the `#TODO` above?

tests/openvino/test_quantization.py
```diff
@@ -75,7 +75,7 @@
 class OVQuantizerTest(unittest.TestCase):
     SUPPORTED_ARCHITECTURES_WITH_EXPECTED_QUANTIZED_MATMULS = (
         (OVModelForSequenceClassification, "bert", 32, 35),
-        # (OVModelForCausalLM, "gpt2", 41, 23),
+        (OVModelForCausalLM, "gpt2", 31, 3),
```
nikita-savelyevv, line 78: For some reason, `auto` and `ov` tests result in a different number of FQ nodes: 41 and 31 respectively. I'll take a look why.

optimum/intel/openvino/quantization.py
echarlaix: nice catch removed in https://github.com/huggingface/optimum-intel/pull/727/commits/6e2d448bcdea45678f7e72b868e44757fb262bd9

tests/openvino/test_quantization.py
echarlaix: thanks for checking ! I updated it in https://github.com/huggingface/optimum-intel/pull/727/commits/054c66aa596415ec857f11691d394eb85686235e in the meantime

optimum/intel/openvino/quantization.py
echarlaix: I had issues with it when testing but it looks like it's working now, added it back in https://github.com/huggingface/optimum-intel/pull/727/commits/e7f1569f37a84077b1fb4532ff6f40445f812724

tests/openvino/test_quantization.py
echarlaix: yes makes sense, added it in https://github.com/huggingface/optimum-intel/pull/727/commits/d3223d053296d262fb63c49e0ddecea076aa7f05

tests/openvino/test_quantization.py
nikita-savelyevv: @AlexKoff88, when running quantization on OV backend, the number of int8 constants is 22. For PT backend it's only 3 though. I think there are some issues on PT backend, will look into it or create a ticket.

nikita-savelyevv (approved):

AlexKoff88 (approved):

PR #732, opened 2024-05-28
Enable IPEXModel with deepspeed

jiqing-feng:
This PR enables IPEXModel with deepspeed and needs to use the intel_extension_for_pytorch environment to run.


jiqing-feng:
Hi @echarlaix . This PR is ready for review. The main change for IPEXModel is that the new API of `_from_models`. Do you accept it? If so, I will add some tests for this API by your instruction. If not, could you share your idea of how to enable deepspeed by IPEXModel? Thx!

jiqing-feng:
Hi @echarlaix . Can you take a look at my new changes? I integrated the `ipex_jit_trace` in the `__init__` method so users can use the init method to get an exported model.

jiqing-feng:
HI @IlyasMoutawwakil . Can you help take a look at this PR? Thx!

optimum/intel/ipex/modeling_base.py
```diff
@@ -136,13 +136,22 @@ def __init__(
...
-        OptimizedModel.__init__(self, model=model, config=config)
         if is_torch_xpu_available(check_device=True):
             self._device = torch.device("xpu:0")
         elif torch.cuda.is_available():
             self._device = torch.device("cuda:0")
         else:
             self._device = torch.device("cpu")
+
+        # CPU only support jit model for now.
+        if self._device == torch.device("cpu") and not isinstance(model, torch.jit.RecursiveScriptModule):
```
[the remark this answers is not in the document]
echarlaix: also could you add a test for this ?

optimum/exporters/ipex/modeling_utils.py
```diff
@@ -296,18 +295,18 @@ def forward(
...
             )
         super().__init__()
         _setattr_from_module(self, module)
         self.config = config
-        self.distributed = distributed
         from intel_extension_for_pytorch.llm.modules import Linear2SiluMul, LinearAdd
 
-        if not self.distributed:
+        # LinearAllreduce and LinearLayer cannot use fused op LinearAdd
+        if module.down_proj.__class__.__name__ not in ["LinearAllreduce"]:
```
echarlaix, line 309: when does this happen ? could you add a test for this ?

examples/ipex/deepspeed/run_generation_deepspeed.py
```diff
@@ -0,0 +1,154 @@
...
+        data = {"type": "ds_model", "checkpoints": checkpoint_files, "version": 1.0}
+        json.dump(data, open(checkpoints_json, "w"))
+
+
+repo_root = get_repo_root(model_name)
+
+write_checkpoints_json()
+dist.barrier()
+
+model = deepspeed.init_inference(
```
echarlaix, line 136: I don't think this makes sense to add this example to optimum-intel as it's mostly unrelated here (only loading the resulting model with `IPEXModelForCausalLM`) what do you think about moving it [here](https://github.com/intel/intel-extension-for-pytorch/tree/main/examples/cpu/inference/python) instead and add a link to it in optimum-intel ?

optimum/exporters/ipex/modeling_utils.py
jiqing-feng: Yes, but it requires importing deepspeed.

yao-matrix: > when does this happen ? could you add a test for this ?

In "DeepSpeed + transformers or optimum-intel" case, DeepSpeed will shard to model for tensor parallelism, and some of the sharded model's layer will be changed to "LinearAllReduce" to fuse linear and allreduce into one op.  So, in this case, layer name is not enough for us to judge whether we need do linear add fusion, that why we change the logic from op name only check to `op name + layer class` check.

W/ this DeepSpeed TP can enjoy optimum-intel's optimized model performance.

tests/ipex/test_modeling.py
```diff
@@ -147,25 +152,32 @@ def test_compare_to_transformers(self, model_arch):
...
+        # Test re-load model
         with tempfile.TemporaryDirectory() as tmpdirname:
             ipex_model.save_pretrained(tmpdirname)
             loaded_model = self.IPEX_MODEL_CLASS.from_pretrained(tmpdirname)
             loaded_model_outputs = loaded_model(**tokens)
 
+        # Test init method
+        init_model = self.IPEX_MODEL_CLASS(transformers_model)
+        init_model_outputs = init_model(**tokens)
+
```
echarlaix, line 164:
```suggestion
        self.assertIsInstance(init_model_outputs.model, torch.jit.RecursiveScriptModule)
```

tests/ipex/test_modeling.py
```diff
@@ -354,20 +373,27 @@ def test_compare_to_transformers(self, model_arch):
...
         self.assertTrue(torch.equal(outputs.logits, loaded_model_outputs.logits))
+        self.assertTrue(torch.equal(outputs.logits, init_model_outputs.logits))
 
     @parameterized.expand(SUPPORTED_ARCHITECTURES)
     def test_pipeline(self, model_arch):
         model_id = MODEL_NAMES[model_arch]
         model = self.IPEX_MODEL_CLASS.from_pretrained(model_id, export=True)
         preprocessor = AutoFeatureExtractor.from_pretrained(model_id)
-        pipe = pipeline("audio-classification", model=model, feature_extractor=preprocessor)
+        pipe = pipeline("audio-classification", model=model, feature_extractor=preprocessor, framework="pt")
```
echarlaix, line 396: why is this needed ?
```suggestion
        pipe = pipeline("audio-classification", model=model, feature_extractor=preprocessor)
```

optimum/exporters/ipex/modeling_utils.py
echarlaix: thanks a lot for the explanation

echarlaix (approved):
LGTM, thanks for the addition @jiqing-feng !

examples/ipex/deepspeed/run_generation_deepspeed.py
echarlaix: thanks !

jiqing-feng:
Hi @echarlaix . Thanks for your review, I have fixed all your comments.

PR #737, opened 2024-05-29
fix pipeline accelerator default to ipex not ort

### LINKED ISSUE #735: Pipeline accelerator defaults to 'ort' runtime instead of 'ipex'
When pipeline is instantiated if accelerator is not specified defaults to 'ort' despite the documentation says default should be 'ipex'.

rbrugaro:
# What does this PR do?
Pipeline accelerator backend can be selected between ["ipex", "inc" and "openvino"]. The args description mentions 'ipex' as default but default was set to 'ort'. 

Fixes #735 

## Before submitting
- [x ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).


echarlaix (approved):
thanks for the fix @rbrugaro

PR #743, opened 2024-06-03
improve doc around supported tasks and accelertor options

rbrugaro:
# fix documentation
Prior https://github.com/huggingface/optimum-intel/pull/739 removes 'ipex' as default accelerator and 
seems is the only supported one in MAPPING_LOADING_FUNC. removed inc and openvino from options

also, updated list of supported pipeline tasks to match those of optimum/intel/ipex/modeling_base.py

Fixes # 742

## Before submitting
- [x ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).


rbrugaro:
@echarlaix

echarlaix (approved):
Thanks for fixing the docstrings @rbrugaro

PR #744, opened 2024-06-04
multi instances of infer

wgzintel:


optimum/intel/openvino/modeling_base.py
```diff
@@ -457,11 +459,11 @@ def compile(self):
                 cache_dir = Path(self.model_save_dir).joinpath("model_cache")
                 ov_config["CACHE_DIR"] = str(cache_dir)
                 logger.info(f"Setting OpenVINO CACHE_DIR to {str(cache_dir)}")
-            self.request = core.compile_model(self.model, self._device, ov_config)
+            self.compiled_model = core.compile_model(self.model, self._device, ov_config)
```
helena-intel, line 462: `compiled_model` does make more sense, but we have a lot of existing code that expects the requests property here, including integrations in other products, so if this is going to be changed there has to be a deprecation period with a DeprecationWarning for a while

PR #747, opened 2024-06-05
Add setuptools to fix issue with Python 3.12, add Windows to OpenVINO basic test

helena-intel:
- Add Windows to test_openvino_basic.yml.
- Update highest Python version to 3.12 (highest version currently supported)
- Add setuptools to requirements (needed for Python 3.12)
- Remove auto-gptq from tests requirements. This broke the Python 3.12 tests and it seems it is not currently being used


echarlaix (approved):
LGTM, thanks @helena-intel !

PR #748, opened 2024-06-05
Disable future warnings/info messages on import

helena-intel:
Importing an OVModel shows info messages and warnings that are not relevant for optimum-intel end users. That's not an issue in a script, but in notebooks/demos it can be distracting. This PR disables the INFO message from NNCF, and FutureWarnings that result from imports in `__init__.py`

7025d07633b6ca9f8cfb18d483d600fc


echarlaix (approved):
LGTM, thanks @helena-intel !

PR #750, opened 2024-06-06
loading generation config if it is part of model

eaidova:
# What does this PR do?

there is misalignment with behaviour how we convert model and how to load in generation config part. In some cases, generation config parameters can be different from default and not provided in config. I see issue with attempt to run whisper-tiny with pipelines due to difference in generation config, when openvino model loaded and when original pytorch model (due to obtaining generation config from model config instead of using provided with model before conversion)

## Before submitting

### DIFF
--- a/tests/openvino/test_export.py
+++ b/tests/openvino/test_export.py
@@ -21,7 +21,7 @@ from typing import Optional
 import torch
 from parameterized import parameterized
 from sentence_transformers import SentenceTransformer, models
-from transformers import AutoConfig, AutoTokenizer
+from transformers import AutoConfig, AutoTokenizer, GenerationConfig
 from utils_tests import MODEL_NAMES
 
 from optimum.exporters.onnx.constants import SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED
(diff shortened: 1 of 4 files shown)


eaidova:

tests/openvino/test_export.py
```diff
@@ -124,6 +126,51 @@ def _openvino_export(
...
+                self.assertIsInstance(ov_model, OVBaseModel)
+                self.assertTrue(ov_model.can_generate())
+                self.assertTrue(ov_model.generation_config is not None)
+                self.assertIsInstance(ov_model.generation_config, GenerationConfig)
+                self.assertTrue(ov_model.generation_config.top_k == 42)
+
+                # check that generate config remains after repeated saving
+                with TemporaryDirectory() as tmpdirname2:
+                    ov_model.save_pretrained(tmpdirname2)
+                    ov_model = auto_model.from_pretrained(tmpdirname2)
```
echarlaix, line 167:
```suggestion
                    ov_model = auto_model.from_pretrained(tmpdirname2, use_cache=use_cache)
```

PR #751, opened 2024-06-06
Trigger openvino slow tests with openvino-test label

echarlaix:
So that the openvino-test label can trigger the OpenVINO slow tests as discussed @eaidova  @helena-intel

### DIFF
--- a/.github/workflows/test_openvino_basic.yml
+++ b/.github/workflows/test_openvino_basic.yml
@@ -6,12 +6,13 @@ on:
     - cron:  '41 1 * * *'  # run every day at 1:41
   push:
     paths:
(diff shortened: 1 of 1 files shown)


echarlaix:
@helena-intel  @eaidova let me know if that works for you

helena-intel:
Thanks @echarlaix ! Is there a way for a PR author to add this label or can only repo owners do that? I don't see a way for me to do it in the UI.

In case not everyone knows: the basic test yaml file has a workflow_dispatch trigger so PR authors can always run this test manually in their fork too
image

IlyasMoutawwakil (approved):

PR #754, opened 2024-06-07
Fix compatibility with transformers < v4.39.0 release 

echarlaix:
* `StaticCache` was introduced in transformers v4.38.0
* `_prepare_generation_config` was introduced in transformers v4.39.0
* Fix `MixtralModelPatcher` for transformers < v4.37.0


echarlaix:
@eaidova let me know if that works for you

PR #758, opened 2024-06-10
Deprecate transformers v4.36.0

echarlaix:
Also fixes issue for OpenVINO export of arctic architectures patching for transformers < v4.37.0 https://huggingface.co/Snowflake/snowflake-arctic-instruct/blob/main/modeling_arctic.py#L1010 

cc @helena-intel @eaidova

### DIFF
--- a/.github/workflows/test_openvino.yml
+++ b/.github/workflows/test_openvino.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.8", "3.12"]
-        transformers-version: ["4.36.0", "4.41.*"]
+        transformers-version: ["4.37.0", "4.41.*"]
         os: [ubuntu-latest]
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


helena-intel:
If possible, it would be great if we could support transformers versions a bit longer. 4.36.2 was released about 6 months ago. That's not so long ago. For example [LlavaMed 1.5](https://github.com/microsoft/LLaVA-Med) was released three weeks ago and needs transformers 4.36.2

echarlaix:
> If possible, it would be great if we could support transformers versions a bit longer. 4.36.2 was released about 6 months ago. That's not so long ago. For example [LlavaMed 1.5](https://github.com/microsoft/LLaVA-Med) was released three weeks ago and needs transformers 4.36.2

Opened https://github.com/huggingface/optimum-intel/pull/764 to fix the arctic models export in the meantime, I'm ok with delaying this PR even though 6 months sounds reasonable to me, what's your expection in terms of support ?

PR #761, opened 2024-06-11
[OV Optimum] Keep ShapeOf on Parameter / ReadValue in case of added beam_idx -> Gather

jane-intel:
# What does this PR do?
In the case of inserted `beam_idx -> Gather`, we keep `ShapeOf` operations on the original `Parameter` operations (`ReadValue` operations after the model becomes stateful). This helps avoid special handling of such ShapeOf operations in runtime (CVS-143648).

## Before submitting


jane-intel:
We've chosen a different approach.

PR #763, opened 2024-06-12
Update to NNCF 2.11

nikita-savelyevv:
# What does this PR do?

- Update NNCF requirement to version 2.11
- Enable scale estimation weight compression algorithm
- Add option `awq` option to `OVWeightQuantizationConfig` so that AWQ can be enabled by providing `OVWeightQuantizationConfig(awq=True)` besides already present `OVWeightQuantizationConfig(quant_method=QuantizationMethod.AWQ)`. 

## Before submitting


optimum/intel/openvino/configuration.py
```diff
@@ -243,6 +255,11 @@ def post_init(self):
         if self.tokenizer is not None and not isinstance(self.tokenizer, str):
             raise ValueError(f"Tokenizer is expected to be a string, but found {self.tokenizer}")
 
+        if self.quant_method == QuantizationMethod.AWQ:
+            self.quant_method = OVQuantizationMethod.DEFAULT
+            self.awq = True
+            logger.warning('Using quant_method=\"AWQ\" is deprecated. Please use awq=True instead in the future.')
```
nikita-savelyevv, line 261: I decided to change API for applying AWQ via source code, i.e. when we create `OVWeightQuantizationConfig`. 

Despite the fact there is an `QuantizationMethod.AWQ` enum, this approach does not quite fit the whole API of NNCF weight compression because AWQ is just another method, alike to Scale Estimation that can additionally be applied along a regular weight compression, and not some specific standalone method like hybrid quantization for example.

I've added a backward compatible logic to still apply AWQ if it was provided via previous API.

During export via optimum-cli it is already like this, e.g. there is an `--awq` argument.

nikita-savelyevv: cc @AlexKoff88

optimum/commands/export/openvino.py
```diff
@@ -253,8 +263,9 @@ def run(self):
                     "all_layers": None if is_int8 else self.args.all_layers,
                     "dataset": self.args.dataset,
                     "num_samples": self.args.num_samples,
-                    "quant_method": QuantizationMethod.AWQ if self.args.awq else None,
+                    "awq": self.args.awq,
```
AlexKoff88, line 266: I think originally it looked like this but @echarlaix asked to make AWQ option unified with the rest of the HF ecosystem.

AlexKoff88:
I am ok with the rest of the changes but the way how AWQ option looks should be clarified. @echarlaix, can you please comment?

optimum/commands/export/openvino.py
echarlaix: yes I'd me in favor of keeping the `quant_method` attribute to keep compatibility with what's done in transformers, I think we can have a awq argument instead of a quant_method though if easier : `OVWeightQuantizationConfig(awq=True)` even though slight preference to something like `OVWeightQuantizationConfig(quant_method="awq")` so that when (if) new quantization methods are added it'll be easier than to have multiple argument for each of them

echarlaix (commented):
LGTM @nikita-savelyevv

optimum/commands/export/openvino.py
nikita-savelyevv: @echarlaix Thanks for the idea. We indeed could keep both `awq=True` and `quant_method="awq"`.

AlexKoff88 (approved):

optimum/intel/openvino/configuration.py
```diff
@@ -198,6 +214,9 @@ def __init__(
         self.all_layers = all_layers
         self.sensitivity_metric = sensitivity_metric
         self.quant_method = quant_method
+        self.scale_estimation = scale_estimation
+        if awq:
+            self.quant_method = QuantizationMethod.AWQ
```
echarlaix, line 219: I'd prefer to only have one path to enable awq (currently we have both `quant_method` and `awq` arguments), what do you think about something like `OVWeightQuantizationConfig(quant_method="awq")` ?

nikita-savelyevv: @echarlaix Ah, I thought you meant it's fine keeping both. Ok, removed `awq` argument.

echarlaix: thanks a lot @nikita-savelyevv

echarlaix (approved):

echarlaix:
Thanks a lot @nikita-savelyevv will merge once the tests pass

PR #766, opened 2024-06-14
Enable security scanning

mfuntowicz:


echarlaix (approved):
thanks a lot @mfuntowicz

PR #772, opened 2024-06-19
Fix INC WoQ model loading issue

changwangss:
# What does this PR do?

woq quantized model saved following the format like https://huggingface.co/TheBloke/Llama-2-7B-Chat-GPTQ/tree/main since itrex v1.4, the quantization config should be added to model.config when model saving and these messages also used for loading.
there is a tricky when woq quantized model do save.
```
from intel_extension_for_transformers.transformers.llm.quantization.utils import convert_to_quantized_model
from intel_extension_for_transformers.transformers.modeling.modeling_auto import save_low_bit

quantized_model = convert_to_quantized_model(model, quantization_config)
# quantized_model 's Linear is QuantizatiedLinearQbits 
self._quantized_model.save_pretrained = types.MethodType(save_low_bit, self._quantized_model)
quantized_model.save_pretrained()
# quantized_model 's Linear will be changed to WeightOnlyLinear, it is due to we would like to save the same format with GPTQ
```
so if we still want to used the quantizer.quantized_model, we should loading it from local to restore it.

quickly validated command
```
 python run_clm.py --model_name_or_path EleutherAI/gpt-neo-125M --dataset_name wikitext  --dataset_config_name wikitext-2-raw-v1 --apply_quantization --quantization_approach weight_only --verify_loading  --output_dir ./tmp/clm_output
```

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/intel/neural_compressor/quantization.py
+++ b/optimum/intel/neural_compressor/quantization.py
@@ -40,6 +40,7 @@ from transformers import (
 
 from optimum.exporters import TasksManager
 from optimum.exporters.onnx import OnnxConfig
+from optimum.intel.neural_compressor import INCModelForCausalLM
 from optimum.onnxruntime import ORTModel
 from optimum.onnxruntime.modeling_decoder import ORTModelForCausalLM
 from optimum.onnxruntime.modeling_seq2seq import ORTModelForConditionalGeneration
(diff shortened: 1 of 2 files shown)


optimum/intel/neural_compressor/quantization.py
```diff
@@ -309,10 +310,14 @@ def quantize(
...
             )
-
+            quantization_config.remove_redundant_parameters()
+            self._quantized_model.config.quantization_config = quantization_config
             self._quantized_model.quantization_config = quantization_config
             self._quantized_model.save_pretrained = types.MethodType(save_low_bit, self._quantized_model)
             self._quantized_model.save_pretrained(save_directory)
+            # self._quantized_model saved to "output_dir" with GPTQ format
+            # which is replaced by inplace so loading it to restore optimized model
+            self._quantized_model = INCModelForCausalLM.from_pretrained(save_directory)
```
echarlaix, line 320: not sure this is needed as we won't use it later on, to fix the `run_clm` example why not load it there instead ?

optimum/intel/neural_compressor/quantization.py
```diff
@@ -309,10 +310,14 @@ def quantize(
             self._quantized_model = convert_to_quantized_model(
                 self._original_model, quantization_config, device=quantization_config.device
             )
-
+            quantization_config.remove_redundant_parameters()
+            self._quantized_model.config.quantization_config = quantization_config
```
echarlaix, line 314: currently a `quantization_config.json` file is expected to load ITREX model https://github.com/huggingface/optimum-intel/blob/v1.17.2/optimum/intel/neural_compressor/modeling_base.py#L186 why should save it in the main config ?

PR #776, opened 2024-06-20
add support glm4

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -171,26 +171,14 @@ class ChatGLM2DummyPastKeyValuesGenerator(DummyPastKeyValuesGenerator):
 
     def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
         if not self.standart_cache_layout:
-            past_key_shape = (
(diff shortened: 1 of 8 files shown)


optimum/exporters/openvino/model_configs.py
```diff
@@ -167,20 +167,35 @@ def __init__(
...
+                self.multi_query_group_num,
+                self.sequence_length,
+                self.head_dim,
+            )
+            past_value_shape = (
+                self.batch_size,
+                self.multi_query_group_num,
+                self.sequence_length,
+                self.head_dim,
+            )
```
echarlaix, line 198: `past_key_shape` and `past_value_shape` looks the same in both cases why not something like :
```suggestion
        if not self.standart_cache_layout:
            pkv_shape = (
                self.sequence_length,
                self.batch_size,
                self.multi_query_group_num,
                self.head_dim,
            )
        else:
            pkv_shape = (
                self.batch_size,
                self.multi_query_group_num,
                self.sequence_length,
                self.head_dim,
            )
```

echarlaix (approved):
LGTM, thanks @eaidova

PR #777, opened 2024-06-21
Update export function path for neural-compressor2.6

PenghuiCheng:
# What does this PR do?

Update export function path for neural-compressor >= v2.6.

## Before submitting

### DIFF
--- a/optimum/intel/neural_compressor/quantization.py
+++ b/optimum/intel/neural_compressor/quantization.py
@@ -25,7 +25,7 @@ from typing import Callable, Optional, Union
 import torch
 from datasets import Dataset, load_dataset
 from neural_compressor.config import PostTrainingQuantConfig
-from neural_compressor.experimental.export import torch_to_int8_onnx
+from neural_compressor.utils.export import torch_to_int8_onnx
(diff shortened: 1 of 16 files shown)


optimum/intel/neural_compressor/quantization.py
```diff
@@ -25,7 +25,7 @@
 import torch
 from datasets import Dataset, load_dataset
 from neural_compressor.config import PostTrainingQuantConfig
-from neural_compressor.experimental.export import torch_to_int8_onnx
+from neural_compressor.utils.export import torch_to_int8_onnx
```
echarlaix, line 28: To continue supporting previous versions of `neural-compressor` why not check its version with [`is_neural_compressor_version`](https://github.com/huggingface/optimum-intel/blob/ef23a5d44d001578e90ac2626530394178808cc8/optimum/intel/utils/import_utils.py#L309C5-L309C33)

echarlaix (approved):
thanks for the fix @PenghuiCheng

optimum/intel/neural_compressor/quantization.py
PenghuiCheng: Yes, you are right, fixed

PR #778, opened 2024-06-24
[OV]: load and convert llms in original precision

eaidova:
# What does this PR do?

allow loading bfloat16 and float16 models in original precision for conversion. It significantly reduces memory consumption and loading time during model conversion for large models

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -301,9 +301,11 @@ def main_export(
         and task.startswith("text-generation")
         and getattr(config, "torch_dtype", torch.float32) in [torch.float16, torch.bfloat16]
     ):
-        if is_openvino_version(">=", "2024.2") and config.torch_dtype == torch.float16:
+        if ov_config is not None and ov_config.dtype in {"fp16", "fp32"}:
+            dtype = torch.float16 if ov_config.dtype == "fp16" else torch.float32
+        elif is_openvino_version(">=", "2024.2") and config.torch_dtype == torch.float16:
             dtype = torch.float16
-        if is_openvino_version(">=", "2024.3") and config.torch_dtype == torch.bfloat16:
+        elif is_openvino_version(">=", "2024.3") and config.torch_dtype == torch.bfloat16:
             dtype = torch.bfloat16
 
     if dtype is not None:
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -26,7 +26,7 @@ from optimum.exporters import TasksManager
 from optimum.exporters.onnx.base import OnnxConfig
 from optimum.exporters.onnx.constants import SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED
 from optimum.exporters.openvino.convert import export_from_model
-from optimum.intel.utils.import_utils import is_openvino_tokenizers_available, is_transformers_version
+from optimum.intel.utils.import_utils import is_openvino_tokenizers_available, is_transformers_version, is_openvino_version
 from optimum.utils.save_utils import maybe_load_preprocessors
 
 from .utils import clear_class_registry
@@ -230,6 +230,7 @@ def main_export(
 
     do_gptq_patching = False
     custom_architecture = False
+    patch_16bit = False
     loading_kwargs = {}
     if library_name == "transformers":
         config = AutoConfig.from_pretrained(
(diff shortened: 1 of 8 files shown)


optimum/exporters/openvino/__main__.py
```diff
@@ -84,6 +88,7 @@ def main_export(
     stateful: bool = True,
     convert_tokenizer: bool = False,
     library_name: Optional[str] = None,
+    model_loading_kwargs: Optional[Dict[str, Any]] = None,
```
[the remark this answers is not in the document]
eaidova: I thought about possibliity to force some other arguments in future e.g. attention_realization for switching between sdpa and eager manually, but it maybe too advanced, so for me having type argument is ok.

optimum/exporters/openvino/__main__.py
```diff
@@ -361,6 +391,7 @@ class StoreAttr(object):
         preprocessors=preprocessors,
         device=device,
         trust_remote_code=trust_remote_code,
+        patch_16bit_model=patch_16bit,
```
echarlaix, line 394: could we instead update and use `ov_config.dtype` instead of adding `patch_16bit_model` ?

echarlaix (approved):
Looks great, thanks a lot @eaidova

optimum/exporters/openvino/__main__.py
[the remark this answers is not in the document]
echarlaix: thanks @eaidova !

optimum/exporters/openvino/__main__.py
eaidova: layers patching is time consuming process we would like to avoid doing that for cases where it is not really neseccary. We need it only for pytorch model wegiths represented in fp16 and bf16 format for avoid issues tracing on cpu (torchscript does not support some ops running on cpu in this format). ov_config.dtype currently used for output model type and signal to compress weights to fp16 after conversion, mixing them in 1 parameter may lead to some issues:

* inability to convert fp32 pytorch models to fp16 openvino model (now it is possible)
* compressing model weights twice (having model in fp16 on torch side and then repeat compression for openvino)

eaidova:
> Looks great, thanks a lot @eaidova

@echarlaix  thanks, we still investigating impact on models accuracy and quantization on our side. Could you please do not merge these changes, until we do not have whole picture?

optimum/exporters/openvino/__main__.py
```diff
@@ -257,7 +266,28 @@ def main_export(
...
             trust_remote_code = False
+    dtype = loading_kwargs.get("torch_dtype")
+    if isinstance(dtype, str):
+        dtype = config.torch_dtype if dtype == "auto" else getattr(torch, dtype)
+
+    if (
+        dtype is None
+        and framework == "pt"
+        and not do_gptq_patching
+        and task.startswith("text-generation")
```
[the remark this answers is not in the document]
eaidova: in future we will propagate for other model types. text-generation models (especially LLMs) in most cases suffer for ineffective weights conversion when we load weights saved in float16/bfloat16 and implicit convert them to float32 when call from_pretrained method. The idea behind these changes is allowing load model as is for avoid weights conversion and also speedup some operations with weights like linear or matmul converting them directly to ov without passing tracing.

optimum/exporters/openvino/__main__.py
```diff
@@ -213,6 +218,9 @@ def main_export(
         model_name_or_path, subfolder=subfolder, revision=revision, cache_dir=cache_dir, token=token
     )
 
+    if framework == "pt":
+        import torch
+
```
IlyasMoutawwakil, line 223: I would prefer if imports are mostly at the top of the file, conditioned by the package's availability.

eaidova: I can not agree, it will lead to import some packages that can be not required during model export (e.g. import tf, if you export pytorch model) and leads to some failure that will block model export however these packages actually do not required to export model

optimum/exporters/openvino/__main__.py
```diff
@@ -100,6 +104,7 @@ def main_export(
     stateful: bool = True,
     convert_tokenizer: bool = False,
     library_name: Optional[str] = None,
+    model_loading_kwargs: Optional[Dict[str, Any]] = None,
```
IlyasMoutawwakil, line 107: why add a new argument instead of using `model_kwargs` which I believe is passed to the model's `from_pretrained`

optimum/exporters/openvino/__main__.py
IlyasMoutawwakil: yes for tensorflow it's understandable, but torch is a required dependency in optimum either ways, and used at multiple stages of the export.

eaidova: ok, moved

IlyasMoutawwakil (approved):

eaidova:
@IlyasMoutawwakil could you please merge?

PR #780, opened 2024-06-25
auto mapping class

jiqing-feng:
Hi @echarlaix. This PR enables auto mapping for model class, which mean we can use `IPEXModel` to load any supported task, for example `IPEXModel.from_pretrained("meta-llama/Llama-2-7b-chat-hf")` will map the `IPEXModelForCausalLM` class to export the model.

The usage is the same as `AutoModel` from transformers.

### DIFF
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -23,7 +23,6 @@ from typing import Dict, Optional, Tuple, Union
 
 import intel_extension_for_pytorch as ipex
 import torch
-import transformers
 from huggingface_hub import hf_hub_download
 from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
 from intel_extension_for_pytorch.cpu._auto_kernel_selection import _enable_tpp
@@ -41,6 +40,7 @@ from transformers import (
     GenerationConfig,
     GenerationMixin,
     PretrainedConfig,
+    is_torch_xpu_available,
 )
 from transformers.dynamic_module_utils import get_class_from_dynamic_module
 from transformers.modeling_outputs import CausalLMOutputWithPast, ModelOutput
@@ -138,7 +138,7 @@ class IPEXModel(OptimizedModel):
         warmup: bool = True,
         **kwargs,
     ):
-        if transformers.is_torch_xpu_available(check_device=True):
+        if is_torch_xpu_available(check_device=True):
             self._device = torch.device("xpu:0")
         elif torch.cuda.is_available():
             self._device = torch.device("cuda:0")
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -55,6 +55,7 @@ from ...exporters.ipex.model_patcher import _IPEX_EXPORTED_TASK, _IPEX_MINIMUM_V
 from ..generation.modeling import prepare_jit_inputs
 from ..utils.import_utils import is_ipex_version, is_torch_version, is_transformers_version
 from ..utils.modeling_utils import MULTI_QUERY_ATTN_MODELS, patch_decoder_attention_mask, recursive_to_device
+from .utils import _HEAD_TO_AUTOMODELS
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


optimum/intel/ipex/modeling_base.py
```diff
@@ -226,7 +227,13 @@ def _from_pretrained(
             if is_torch_version("<", "2.1.0"):
                 raise ImportError("`torch>=2.0.0` is needed to trace your model")
 
-            task = cls.export_feature
+            task = TasksManager.infer_task_from_model(model_id, subfolder=subfolder, revision=revision)
```
echarlaix, line 230: This will not work for local models so I don't think this is something that we should enable

jiqing-feng:
Hi @echarlaix , you were right; it's difficult to get the task from the local model. I think we should let users assign tasks in the API because it is hard to remember all IPEXModel classes, such as `IPEXModelForCausalLM`.  With my changes, users can easily use it by `IPEXModel.from_pretrained("gpt2", task="text-generation")`. 

It is more friendly for users, WDYT?

IlyasMoutawwakil:
The transformers `AutoModel` doesn't map task classes, but rather `feature-extraction` models. For example using `AutoModel` with a `bert-base-uncased` model will return the core `BertModel` (with no task head) https://github.com/huggingface/transformers/blob/dc76e9fa7f0d19ff7cfc33bd3a22acd7df167fce/src/transformers/models/bert/modeling_bert.py#L956.
same thing for llama for example it will return the `LlamaModel` class https://github.com/huggingface/transformers/blob/dc76e9fa7f0d19ff7cfc33bd3a22acd7df167fce/src/transformers/models/llama/modeling_llama.py#L869.
```python
model = AutoModel.from_pretrained("facebook/opt-125m")
print(model.__class__) # <class 'transformers.models.opt.modeling_opt.OPTModel'>
```
This feature will result in a big API difference and confusion around the `feature-extraction` task.

jiqing-feng:
> The transformers `AutoModel` doesn't map task classes, but rather `feature-extraction` models. For example using `AutoModel` with a `bert-base-uncased` model will return the core `BertModel` (with no task head) https://github.com/huggingface/transformers/blob/dc76e9fa7f0d19ff7cfc33bd3a22acd7df167fce/src/transformers/models/bert/modeling_bert.py#L956. same thing for llama for example it will return the `LlamaModel` class https://github.com/huggingface/transformers/blob/dc76e9fa7f0d19ff7cfc33bd3a22acd7df167fce/src/transformers/models/llama/modeling_llama.py#L869.
> 
> ```python
> model = AutoModel.from_pretrained("facebook/opt-125m")
> print(model.__class__) # <class 'transformers.models.opt.modeling_opt.OPTModel'>
> ```
> 
> This feature will result in a big API difference and confusion around the `feature-extraction` task.

Thanks @IlyasMoutawwakil ! I realized it and have changed the API to support assigned task instead of mapping task by the model. Do you mind take a review? Thx!

IlyasMoutawwakil:
My last comment is about the new changes.

> This feature will result in a big API difference and confusion around the feature-extraction task and its specific auto class. (`AutoModel` vs `IPEXModel`)

The current API is based on the simplicity of going from the transformers `AutoModelForxxx` to `IPEXModelForxxx` or `OVModelForxxxx`, which allows them to go from one backend to another and use the one that gives the best hardware compatibility and/or utilization.

Also the `pipeline` function already allow for the "single entry point" API.

jiqing-feng:
> My last comment is about the new changes.
> 
> > This feature will result in a big API difference and confusion around the feature-extraction task and its specific auto class. (`AutoModel` vs `IPEXModel`)
> 
> The current API is based on the simplicity of going from the transformers `AutoModelForxxx` to `IPEXModelForxxx` or `OVModelForxxxx`, which allows them to go from one backend to another and use the one that gives the best hardware compatibility and/or utilization.
> 
> Also the `pipeline` function already allow for the "single entry point" API.

Sorry for not making it clear; my change has no impact on IPEXModel except you pass the parameter `task`.

echarlaix:
> Sorry for not making it clear; my change has no impact on IPEXModel except you pass the parameter `task`. I will close it if you think exposing the parameter `task` is not acceptable, but I want to make sure I have clarified it. Thx :)

Thanks a lot @jiqing-feng, closing it

PR #783, opened 2024-06-26
update llama and gemma patching for resolving bf16 execution issues with sin/cos

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


tests/openvino/test_modeling.py
```diff
@@ -963,6 +966,7 @@ def test_beam_search(self, model_arch):
         )
 
         gen_configs = [
+            beam_sample_gen_config,
```
echarlaix, line 969: should be removed

```suggestion
```

tests/openvino/test_modeling.py
```diff
@@ -933,15 +933,18 @@ def test_beam_search(self, model_arch):
...
             max_new_tokens=10,
             min_new_tokens=10,
             num_beams=4,
             do_sample=True,
             eos_token_id=None,
-            top_k=1,
         )
 
+        if model_arch == "minicpm":
+            beam_sample_gen_config.top_k = 1
```
echarlaix, line 946: why do we need this for minicpm ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -497,50 +497,65 @@ def _llama_gemma_update_causal_mask_latest(
...
+                ).float()
+                emb = torch.cat((sinusoid_inp, sinusoid_inp), dim=-1)
+                return torch.cat((torch.sin(emb), torch.cos(emb)), dim=1)
+
+            base = self._model.model.layers[0].self_attn.rotary_emb.base
+            dim = self._model.model.layers[0].self_attn.rotary_emb.dim
+            self._model.model.register_buffer(
+                "embed_positions",
+                create_sinusoidal_positions(max_positions, dim, base),
+            )
```
echarlaix, line 541: why is this needed ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -497,50 +497,65 @@ def _llama_gemma_update_causal_mask_latest(
...
-        super().__exit__(exc_type, exc_value, traceback)
-        if hasattr(self._model.model, "_orig_update_causal_mask"):
-            self._model.model._update_causal_mask = self._model.model._orig_update_causal_mask
+def llama_gemma_rotary_emb_forward(self, x, position_ids, seq_len=None):
+    _seq_len = torch.max(position_ids) + 1 if seq_len is None else seq_len
+    if _seq_len > self.embed_positions.shape[0]:
+        if seq_len is None:
+            return self._orig_forward(self, x, position_ids)
+        else:
+            return self._orig_forward(self, x, position_ids, seq_len)
```
echarlaix, line 506: where does the `seq_len` argument comes from ?  

https://github.com/huggingface/transformers/blob/v4.42.3/src/transformers/models/llama/modeling_llama.py#L107

```suggestion
        return self._orig_forward(x, position_ids)
```

echarlaix (commented):
Thanks @eaidova

optimum/exporters/openvino/model_patcher.py
eaidova: from gemma https://github.com/huggingface/transformers/blob/v4.42.3/src/transformers/models/gemma/modeling_gemma.py#L104

tests/openvino/test_modeling.py
eaidova: model provides instable results without it

optimum/exporters/openvino/model_patcher.py
echarlaix: This will fail for llama models then no ? I think we need to have a test for this

optimum/exporters/openvino/model_patcher.py
```diff
@@ -497,50 +497,65 @@ def _llama_gemma_update_causal_mask_latest(
...
-                rotary_emb = layer.self_attn.rotary_emb
-                layer.self_attn.rotary_emb.inv_freq = 1.0 / (
-                    rotary_emb.base ** (torch.arange(0, rotary_emb.dim, 2, dtype=torch.int64).float() / rotary_emb.dim)
-                )
-
-    def __exit__(self, exc_type, exc_value, traceback):
-        super().__exit__(exc_type, exc_value, traceback)
-        if hasattr(self._model.model, "_orig_update_causal_mask"):
-            self._model.model._update_causal_mask = self._model.model._orig_update_causal_mask
+def llama_gemma_rotary_emb_forward(self, x, position_ids, seq_len=None):
```
echarlaix, line 500: could you add a link to where the original code comes from ?

optimum/exporters/openvino/model_patcher.py
echarlaix: Thanks for the explanation. Why modifying `self._model.model` here ?

optimum/exporters/openvino/model_patcher.py
eaidova: not sure that I understand you... if llama does not have seq_len argument, then it will be not provided to forward and it is equal to None as default argument, so passing seq_len to original forward will never happens for it

eaidova: if you speak what happens if current sequence len > current cached (max position embeddings 2048 for llama2, 4096 for llama3 if I right remember), it is expected that model will fail due to tracing limitations and the same for onnx and other models as well. For pytorch if model is not trained to extend max context, it will continue generation but as the rule it will generate some kind of trash  (random tokens), so anyway after that results will be nonvalid.

We are working for supporting models with context extension, but not ready to commit on that right now

echarlaix, line 506:
```suggestion
            return self._orig_forward(x, position_ids)
        else:
            return self._orig_forward(x, position_ids, seq_len)
```

optimum/exporters/openvino/model_patcher.py
eaidova: for avoiding constant duplication, I'll check if we can avoid this

eaidova: reworked

optimum/exporters/openvino/model_patcher.py
eaidova: added

optimum/exporters/openvino/model_patcher.py
```diff
@@ -497,50 +497,66 @@ def _llama_gemma_update_causal_mask_latest(
...
+                
+                sinusoid_inp = torch.einsum(
+                    "i , j -> i j", torch.arange(num_pos, dtype=torch.int64).float(), inv_freq
+                ).float()
+                emb = torch.cat((sinusoid_inp, sinusoid_inp), dim=-1)
+                return torch.cat((torch.sin(emb), torch.cos(emb)), dim=1)
+
+            base = self._model.model.layers[0].self_attn.rotary_emb.base
+            dim = self._model.model.layers[0].self_attn.rotary_emb.dim
+            embed_positions = create_sinusoidal_positions(max_positions, dim, base),
```
eaidova, line 542:
```suggestion
            embed_positions = create_sinusoidal_positions(max_positions, dim, base)
```

echarlaix (approved):
LGTM, thanks @eaidova !

optimum/exporters/openvino/model_patcher.py
```diff
@@ -497,50 +497,66 @@ def _llama_gemma_update_causal_mask_latest(
...
 
+            max_positions = self._model.config.max_position_embeddings
+
+            # cos/sin for rotary position embeddings also having issues with bf16 and efficiency due to calculation on each step
+            # use precomputed
+            def create_sinusoidal_positions(num_pos: int, dim: int, base: int = 10000) -> torch.Tensor:
+                # adopted from https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L101 
+                
+                inv_freq = 1.0 / (base ** (torch.arange(0, dim, 2, dtype=torch.int64) / dim))
+
```
echarlaix, line 533:
```suggestion
                # adopted from https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L101 
                inv_freq = 1.0 / (base ** (torch.arange(0, dim, 2, dtype=torch.int64) / dim))

```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -497,50 +497,65 @@ def _llama_gemma_update_causal_mask_latest(
...
+            max_positions = self._model.config.max_position_embeddings
 
-    def __exit__(self, exc_type, exc_value, traceback):
-        super().__exit__(exc_type, exc_value, traceback)
-        if hasattr(self._model.model, "_orig_update_causal_mask"):
-            self._model.model._update_causal_mask = self._model.model._orig_update_causal_mask
+            # cos/sin for rotary position embeddings also having issues with bf16 and efficiency due to calculation on each step
+            # use precomputed
+            def create_sinusoidal_positions(num_pos: int, dim: int, base: int = 10000) -> torch.Tensor:
+                # adopted from https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L101
```
echarlaix, line 530:
```suggestion
                # adapted from https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L101
```

echarlaix, line 530:
```suggestion
                # adopted from https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L101
```

echarlaix:
We can merge once 
`tests/openvino/test_exporters_cli.py::OVCLIExportTestCase::test_exporters_cli_int4_4_text_generation_with_past`
is fixed, seems like there is a mismatch coming from the channel vs group size

eaidova:
> We can merge once `tests/openvino/test_exporters_cli.py::OVCLIExportTestCase::test_exporters_cli_int4_4_text_generation_with_past` is fixed, seems like there is a mismatch coming from the group size

test fixed

tests/openvino/test_exporters_cli.py
```diff
@@ -90,20 +90,26 @@ class OVCLIExportTestCase(unittest.TestCase):
         ("text-generation-with-past", "opt125m", "int4_asym_g128", 4, 144),
         ("text-generation-with-past", "opt125m", "int4_sym_g64", 4, 144),
         ("text-generation-with-past", "opt125m", "int4_asym_g64", 4, 144),
-        ("text-generation-with-past", "llama_awq", "int4 --ratio 1.0 --sym --group-size 16 --all-layers", 0, 32),
+        (
+            "text-generation-with-past",
+            "llama_awq",
+            "int4 --ratio 1.0 --sym --group-size 8 --all-layers",
+            0,
+            16 if is_transformers_version("<", "4.39.0") else 17,
```
echarlaix, line 98: to fix the test looks like we need :
```suggestion
            32 if is_transformers_version("<", "4.39.0") else 34,
```

PR #784, opened 2024-06-26
Fix inference for batched inputs for llama

echarlaix:
Fix inference for batched inputs for fp32 model coming from `min_dtype = torch.finfo(torch.float16).min`

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -441,8 +441,7 @@ def _llama_gemma_update_causal_mask_latest(
     # difference with original modeling
     # using minimum from dtype with larger bandwith (floa32) may lead to overflow
     # during execution on platforms with default lower precision (bfloat16, float16)
-    min_dtype = torch.finfo(torch.float16).min
-
+    min_dtype = torch.finfo(dtype).min
     sequence_length = input_tensor.shape[1]
     if using_static_cache:
         target_length = past_key_values.get_max_length()
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -1181,10 +1181,10 @@ def _phi3_self_attn_sdpa_forward(
             use_cache=use_cache,
         )
 
-    # TO DO: remove llama imports when transformers with phi3 support will be released
-    try:
-        from transformers.models.phi3.modelling_phi3 import apply_rotary_pos_emb, repeat_kv
-    except ImportError:
+    # TODO: remove llama imports when transformers >= v4.41.0
+    if is_transformers_version(">=", "4.41.0"):
+        from transformers.models.phi3.modeling_phi3 import apply_rotary_pos_emb, repeat_kv
+    else:
         from transformers.models.llama.modeling_llama import apply_rotary_pos_emb, repeat_kv
 
     bsz, q_len, _ = hidden_states.size()
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -1180,11 +1180,7 @@ def _phi3_self_attn_sdpa_forward(
             use_cache=use_cache,
         )
 
-    # TODO: remove llama imports when transformers >= v4.41.0
-    if is_transformers_version(">=", "4.41.0"):
-        from transformers.models.phi3.modeling_phi3 import apply_rotary_pos_emb, repeat_kv
-    else:
-        from transformers.models.llama.modeling_llama import apply_rotary_pos_emb, repeat_kv
+    from transformers.models.phi3.modeling_phi3 import apply_rotary_pos_emb, repeat_kv
 
     bsz, q_len, _ = hidden_states.size()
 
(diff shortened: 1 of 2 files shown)


optimum/exporters/openvino/model_patcher.py
```diff
@@ -441,8 +441,7 @@ def _llama_gemma_update_causal_mask_latest(
     # difference with original modeling
     # using minimum from dtype with larger bandwith (floa32) may lead to overflow
     # during execution on platforms with default lower precision (bfloat16, float16)
-    min_dtype = torch.finfo(torch.float16).min
-
+    min_dtype = torch.finfo(dtype).min
```
eaidova, line 444: @echarlaix, what exactly issue you observe?

this fix breaks inference on platforms that support bf16 as default inference precision,  as fp32 min value can not be fit into this type and lead to overflow

echarlaix: I'm having a mismatch between the original transformers model and the exported model in the generated outputs for padded outputs such as  : 

```
tokens = tokenizer(["This is an input", "This is"], return_tensors="pt", padding=True)
```

I updated `test_compare_to_transformers` in case you'd like to reproduce

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1181,11 +1180,7 @@ def _phi3_self_attn_sdpa_forward(
             use_cache=use_cache,
         )
 
-    # TO DO: remove llama imports when transformers with phi3 support will be released
```
echarlaix, line 1184: also I think we could remove `_phi3_self_attn_sdpa_forward` completly https://github.com/huggingface/transformers/blob/ccdabc5642bf84849af93f591e207dc625c8e1e1/src/transformers/models/phi3/modeling_phi3.py#L729
wdyt @eaidova ?

optimum/exporters/openvino/model_patcher.py
echarlaix: there is also an issue for beam sample generation for llama models https://github.com/huggingface/optimum-intel/actions/runs/9672360994/job/26684537476 but looks unrelated

optimum/exporters/openvino/model_patcher.py
eaidova: your changes will not allow to run this model with trust_remote_code = True, because the goal of these changes is resolve buggy implementation https://huggingface.co/microsoft/Phi-3-mini-4k-instruct/blob/main/modeling_phi3.py

eaidova: the problem that phi3 has SDPA disabled in transformers https://github.com/huggingface/transformers/blob/ccdabc5642bf84849af93f591e207dc625c8e1e1/src/transformers/models/phi3/modeling_phi3.py#L923, so this code is never reached on our side even if we explicitly set attention implementation sdpa via lfrom_pretrained arguments

echarlaix: ok that's clear let's keep it then, added https://github.com/huggingface/optimum-intel/pull/784/commits/280543b757a38ef7076d08f5ccac5677d4fbb9dc to enable `trust_remote_code=True` (in which case the transformers version is not guaranteed)

optimum/exporters/openvino/model_patcher.py
slyalin: @echarlaix, this is critical for us. I do not see any clear evidence why this particular thing breaks the tests you mentioned. Could it be a side effect of some unknown different bug?

@dmitry-gorokhov, FYI, this is an attempt to rollback the fix for platforms that infer in fp16 and not accurately converting not fitting finite floats to real -inf in the target precision. It should be fixed in the plugin by saturating the result instead of using -inf. Looks like this is reasonable, as the change in the inference precision is an internal detail of how plugin works.

eaidova: @echarlaix that looks like small thresholds issue, because the maximum absolute different i test with provided input is 0.00011798739433288574, it is not significantly larger that 1e-4 used as threshold in test, besides argsort for tokens is also preserved, so difference between logits should not be visible on selection of generated tokens

echarlaix: I'd have expected this to be even more visible for models with a higher number of layers, but can revert if needed for bf16/fp16 models. Concerning the argsort for tokens that should be preserved, I don't think we can make this conclusion as difference in outputs could be higher for real (larger) models, needs a bit of testing but closing it for now

eaidova: can we postpone these changes until we make our internal decision how to proceed on runtime side? The changing of min_dtype is the only reason for making this patching and ti was done due to bf16/fp16 runtime issue, it was proposed as workaround to unblock inference on some platforms. Without that, there is no need in this patch at all.

We also expect some additional changes may improve situation. it is allowing to load model in original precision without previous conversion to fp32 before conversion (significantly reduce required memory for conversion and loading time for models saved in non fp32 format) - https://github.com/huggingface/optimum-intel/pull/778

echarlaix: Yes works for me!

PR #786, opened 2024-06-27
Patch fusion linear for bert and vit

jiqing-feng:
Hi @echarlaix . I added 2 more supported patched models: Bert and Vit. I only use the linear fusion module to optimize these 2 models; it will bring at least a 10% speed-up on SPR with little code addition.

Would you please take a review on these changes? Thx!


optimum/exporters/ipex/model_patcher.py
```diff
@@ -73,13 +93,10 @@ def _patch_llama_model(model):
...
-    return model
-
-
-def _patch_model(model):
     if isinstance(model, LlamaForCausalLM):
         model = _patch_llama_model(model)
+    elif isinstance(model, BertForQuestionAnswering):
+        model = _patch_bert_model(model)
+    elif isinstance(model, ViTForImageClassification):
+        model = _patch_vit_model(model)
```
IlyasMoutawwakil, line 101: why only patching these specific tasks ? the conversion should be appliable to any `model.config.model_type==bert` (or vit)

jiqing-feng: You were right, I am changing on it :) Thx!

optimum/intel/ipex/modeling_base.py
```diff
@@ -60,7 +64,11 @@
 logger = logging.getLogger(__name__)
 
 
-_IPEX_SUPPORT_MODEL_TYPES = ("llama",)
+_IPEX_SUPPORT_MODEL = {
+    "llama": ("text-generation",),
+    "bert": ("question-answering",),
+    "vit": ("image-classification",),
+}
```
IlyasMoutawwakil, line 71: same remark, a model type based conversion seems more adequate as the task variants of models only differ on a "model head" level, with the same core architecture.

optimum/exporters/ipex/model_patcher.py
jiqing-feng: fixed

optimum/intel/ipex/modeling_base.py

optimum/exporters/ipex/model_patcher.py
```diff
@@ -73,13 +92,10 @@ def _patch_llama_model(model):
...
-    convert_functions(model, LlamaModel, "forward", _llama_model_forward)
-    convert_functions(model, LlamaRMSNorm, "forward", _llama_layer_norm_forward)
-    convert_class(model, LlamaDecoderLayer, _IPEXLlamaDecoderLayer, model.config)
-    return model
-
-
-def _patch_model(model):
     if isinstance(model, LlamaForCausalLM):
         model = _patch_llama_model(model)
+    elif isinstance(model, BertForQuestionAnswering):
```
echarlaix, line 97: why not check the `model_type` from the config instead https://huggingface.co/google-bert/bert-base-uncased/blob/main/config.json#L14 ?  (so that it's for all bert architecture not only for QA)

optimum/intel/ipex/modeling_base.py
```diff
@@ -70,17 +74,22 @@ def _is_patched_with_ipex(model, task):
...
-            model.config.model_type in _IPEX_SUPPORT_MODEL_TYPES
-            and task in _IPEX_EXPORTED_TASK
-            and model.config.hidden_size >= 64
-        )
+        return False
+
+    return model.config.model_type in _IPEX_SUPPORT_MODEL_TYPES
+
+
+def prepare_inputs_for_ipex_model(model, task, use_cache):
```
echarlaix, line 88:
```suggestion
def _prepare_inputs_for_ipex_model(model, task, use_cache):
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -60,7 +64,7 @@
 logger = logging.getLogger(__name__)
 
 
-_IPEX_SUPPORT_MODEL_TYPES = ("llama",)
+_IPEX_SUPPORT_MODEL_TYPES = ("llama", "bert", "vit")
```
echarlaix, line 67: could you add tests for both architectures ?

optimum/exporters/ipex/modeling_utils.py
```diff
@@ -139,14 +150,9 @@ def _llama_model_forward(
 # Adapted from https://github.com/huggingface/transformers/blob/v4.38.2/src/transformers/models/llama/modeling_llama.py#L321
 class _IPEXLlamaAttention(nn.Module):
     def __init__(self, module, config) -> None:
-        if is_ipex_version("<", _IPEX_MINIMUM_VERSION_FOR_PATCHING):
-            raise ImportError(
-                f"Only ipex version > {_IPEX_MINIMUM_VERSION_FOR_PATCHING} supports IndirectAccessKVCacheAttention, LinearAdd, RotaryEmbedding"
-            )
```
echarlaix, line 145: why moving ?

optimum/exporters/ipex/modeling_utils.py
```diff
@@ -27,9 +27,20 @@
 
 _IPEX_MINIMUM_VERSION_FOR_PATCHING = "2.3.0"
 
+if is_ipex_version("<", _IPEX_MINIMUM_VERSION_FOR_PATCHING):
+    raise ImportError(f"Only ipex version > {_IPEX_MINIMUM_VERSION_FOR_PATCHING} supports patching")
```
jiqing-feng, line 31: Reply to https://github.com/huggingface/optimum-intel/pull/786#discussion_r1661330118, I moved the logic at the top of this file.

optimum/exporters/ipex/modeling_utils.py
jiqing-feng: See https://github.com/huggingface/optimum-intel/pull/786#discussion_r1661698887

optimum/intel/ipex/modeling_base.py
jiqing-feng: The Bert and Vit models are already in the tests, and they will automatically go into the patching path. Please give me more instructions, like what kind of tests you need. Thanks!

optimum/exporters/ipex/modeling_utils.py
echarlaix: I'm not sure we want to restrict support to IPEX >= v2.3.0 for all models (even the ones that are not patched) ?

jiqing-feng:
Hi @echarlaix , I will get some patched models and upload them for testing. Could you take a look at the failed tests? You will find that we could pass the test when running individually but fail when running together. That's mainly because of the file system state, do you have any suggestions for it? I searched online and found that `@pytest.fixture` may works, but have no clue on how to use it. If you think it is acceptable, you can give me some instructions and I will fix it. If you have better idea, please let me know. Thx!

jiqing-feng:
Hi @echarlaix. I have changed the Ipex version check to make sure it will not have an impact on no patching path and also avoid duplicate codes. I also uploaded 2 models: [patched_tiny_random_bert_for_question_answering](https://huggingface.co/Jiqing/patched_tiny_random_bert_for_question_answering) and [patched_tiny_random_vit_for_image_classification](https://huggingface.co/Jiqing/patched_tiny_random_vit_for_image_classification) for test patching models. Could you please take a review? Thx!

echarlaix (approved):
LGTM, thanks a lot @jiqing-feng

tests/ipex/test_modeling.py
```diff
@@ -188,6 +188,21 @@ def test_pipeline(self, model_arch):
         self.assertGreaterEqual(outputs["score"], 0.0)
         self.assertIsInstance(outputs["answer"], str)
 
+    @unittest.skipIf(is_ipex_version("<", "2.3.0"), reason="Only ipex version > 2.3.0 supports ipex model patching")
+    def test_patched_model(self):
+        ipex_model = IPEXModelForQuestionAnswering.from_pretrained(
+            "Jiqing/patched_tiny_random_bert_for_question_answering"
```
echarlaix, line 194: thanks a lot for adding this test !

jiqing-feng:
Hi @echarlaix . For the failed tests, the real issue comes from [enable_tpp](https://github.com/jiqing-feng/optimum-intel/blob/fusion/optimum/intel/ipex/modeling_base.py#L103). It will change some environment variables and trigger the traced model check [here](https://github.com/jiqing-feng/optimum-intel/blob/fusion/optimum/intel/ipex/modeling_base.py#L78), so the tests will fail when running together. 

I have fixed them by change another way to check if the traced model has been patched or not. Please take a review, thx!

PR #787, opened 2024-06-27
OV configurations alignment

KodiaqQ:
- Updated OV configurations to align with **openvino.genai** [nncf_utils.py](https://github.com/openvinotoolkit/openvino.genai/blob/master/llm_bench/python/utils/nncf_utils.py).


KodiaqQ:
cc @AlexKoff88, @MaximProshin

optimum/intel/openvino/configuration.py
```diff
@@ -46,18 +46,48 @@
...
-    "stabilityai/stablelm-3b-4e1t": {"bits": 4, "sym": True, "group_size": 64, "ratio": 0.8},
+    "stabilityai/stablelm-3b-4e1t": {
+        "bits": 4,
+        "sym": True,
+        "group_size": 64,
+        "ratio": 0.8,
+        "dataset": "wikitext2",
+        "awq": True,
+    },
     "stablelm-epoch-3b-preview": {"bits": 4, "sym": True, "group_size": 64, "ratio": 0.8},
```
eaidova, line 60: can this line be removed, it looks like it is accidentally copied with wrong naming format?

KodiaqQ: Removed. Thanks!

eaidova (approved):

AlexKoff88 (approved):

MaximProshin:
@KodiaqQ , I still see some misalignment with genai. In genai we have 28 models, but only 26 here. Can you explain it?

KodiaqQ:
> @KodiaqQ , I still see some misalignment with genai. In genai we have 28 models, but only 26 here. Can you explain it?

As was noticed, the `stablelm-epoch-3b-preview` and `llama-7b` configurations are not applicable anymore, so I removed them.
Maybe @AlexKoff88 can confirm that as we discussed.

optimum/intel/openvino/configuration.py
```diff
@@ -46,18 +46,47 @@
...
     "meta-llama/Llama-2-13b-chat": {"bits": 4, "sym": True, "group_size": 64, "ratio": 0.8},
-    "stabilityai/stablelm-3b-4e1t": {"bits": 4, "sym": True, "group_size": 64, "ratio": 0.8},
-    "stablelm-epoch-3b-preview": {"bits": 4, "sym": True, "group_size": 64, "ratio": 0.8},
-    "stabilityai/stablelm-zephyr-3b": {"bits": 4, "sym": False, "group_size": 128, "ratio": 1.0},
+    "stabilityai/stablelm-3b-4e1t": {
+        "bits": 4,
+        "sym": True,
+        "group_size": 64,
+        "ratio": 0.8,
+        "dataset": "wikitext2",
```
helena-intel, line 57: Is there a chance that using a specific dataset in the default configs causes better accuracy when evaluating on wikitext, but worse for actual use-cases?

KodiaqQ: I think that this is possible. 
But this is only alignment with the default state from here [openvino.genai](https://github.com/openvinotoolkit/openvino.genai/blob/master/llm_bench/python/utils/nncf_utils.py), configurations aren't new.
cc @AlexKoff88

MaximProshin: Let's skip it here. This task is not to find better configs but rather align optimum-intel configs with ones are currently used in the OV validation which will also switch to optimum-intel. Without it, we can't switch.

MaximProshin:
@echarlaix , can you please help to merge it?

echarlaix (approved):

PR #805, opened 2024-07-04
Increase default 4-bit compression ratio from 0.8 to 1.0

nikita-savelyevv:
This PR aligns default 4-bit weight compression parameters between optimum-intel and opengino.genai repositories. The default parameters are: `bits=4, sym=False, ratio=1.0, group_size=128`. This will be applied when there is no custom compression recipe for the given model id.

Corresponding PR to openvino.genai: https://github.com/openvinotoolkit/openvino.genai/pull/577

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -103,6 +103,14 @@ _DEFAULT_4BIT_CONFIGS = {
     },
 }
 
+_DEFAULT_4BIT_CONFIG = {
+    "bits": 4,
+    "ratio": 1.0,
+    "sym": False,
+    "group_size": 128,
+    "all_layers": None,
+}
+
 
 class OVQuantizationMethod(str, Enum):
     DEFAULT = "default"
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -221,7 +221,7 @@ class OVExportCommand(BaseOptimumCLICommand):
 
     def run(self):
         from ...exporters.openvino.__main__ import infer_task, main_export, maybe_convert_tokenizers
-        from ...intel.openvino.configuration import _DEFAULT_4BIT_CONFIGS, OVConfig
+        from ...intel.openvino.configuration import _DEFAULT_4BIT_CONFIG, _DEFAULT_4BIT_CONFIGS, OVConfig
 
         def _get_default_int4_config(model_id_or_path, library_name):
             if model_id_or_path in _DEFAULT_4BIT_CONFIGS:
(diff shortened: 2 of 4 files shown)


nikita-savelyevv:
@AlexKoff88 proposed to raise default compression ratio from 0.8 to 1.0

cc @MaximProshin @eaidova

optimum/intel/openvino/configuration.py
```diff
@@ -103,6 +103,14 @@
     },
 }
 
+_DEFAULT_4BIT_CONFIG = {
+    "bits": 4,
+    "ratio": 1.0,
+    "sym": False,
+    "group_size": 128,
```
eaidova, line 110: I'm not sure about this value, because some models can has smaller head_dim

eaidova (approved):

optimum/intel/openvino/configuration.py
nikita-savelyevv: Yes, `group_size` was previously `None`, but I decided to make it more explicit and set it to 128. Otherwise, the decision will be left to be made by NNCF. If `group_size=None` is provided, NNCF will set it to 128 too. But this can potentially change in the future in theory resulting in some bad chain reaction. This is was my motivation.

> some models can has smaller head_dim

This is true. But as I understand, it will fail even if `None` is provided for the reason described above.

MaximProshin (approved):

echarlaix (approved):
LGTM, thanks @nikita-savelyevv

echarlaix:
To fix the code style check you can do the following : 
```
pip install .[quality]
make style
```

nikita-savelyevv:
> To fix the code style check you can do the following :
> 
> ```
> pip install .[quality]
> make style
> ```

Thanks! Actually, for quite some time I had troubles with `make style` command. For some reason it changes many files in the repo, even those that I didn't change. Possibly, I need to change some local configs. Do you have any suggestions?

echarlaix:
> Thanks! Actually, for quite some time I had troubles with `make style` command. For some reason it changes many files in the repo, even those that I didn't change. Possibly, I need to change some local configs. Do you have any suggestions?

Could it be a mismatch in the black or ruff version ? I think it could make sense to pin a specific version on our side, will open a PR for this

echarlaix:
Will merge once the tests pass, I think only `TEST_4BIT_CONFIGURATONS` needs to be udpated  https://github.com/huggingface/optimum-intel/blob/48cc82aa361beb58cd690775b400702a3de1421b/tests/openvino/test_exporters_cli.py#L88

PR #808, opened 2024-07-05
Update ruff version

echarlaix:


IlyasMoutawwakil (approved):
is there a specific reason why we're not using the latest version ?

echarlaix:
> is there a specific reason why we're not using the latest version ?

using the same ruff version as in transformers 
https://github.com/huggingface/transformers/blob/v4.42.3/setup.py#L160

Issue #811, opened 2024-07-08
Issue while exporting internlm2-chat-7b  model to OpenVINO format


lumurillo:
When trying to export the [internlm2-chat-7b](https://huggingface.co/internlm/internlm2-chat-7b) model to the OpenVINO format, got an error.

The environment used was the following:
* OS: Windows 11
* Python: 3.11
* OpenVINO:  2024.3.0.dev20240705
* optimum: 1.21.1
* optimum-intel: 1.19.0.dev0+eeb1df0

The error log is the following:

```
optimum-cli export openvino --weight-format int4 --sym --trust-remote-code --convert-tokenizer --model internlm/internlm2-chat-7b internlm2-chat-7b-int4 --convert-tokenizer option is deprecated. Tokenizer will be converted by default.
Framework not specified. Using pt to export the model.
model-00001-of-00008.safetensors: 100%|███████████████████████████████████████| 1.95G/1.95G [01:47<00:00, 14.3MB/s]
model-00002-of-00008.safetensors: 100%|███████████████████████████████████████| 1.95G/1.95G [01:32<00:00, 20.9MB/s]
model-00003-of-00008.safetensors: 100%|███████████████████████████████████████| 1.98G/1.98G [01:56<00:00, 17.0MB/s]
model-00004-of-00008.safetensors: 100%|███████████████████████████████████████| 1.95G/1.95G [02:25<00:00, 13.4MB/s]
model-00005-of-00008.safetensors: 100%|███████████████████████████████████████| 1.98G/1.98G [01:53<00:00, 17.4MB/s]
model-00006-of-00008.safetensors: 100%|███████████████████████████████████████| 1.95G/1.95G [01:51<00:00, 17.4MB/s]
model-00007-of-00008.safetensors: 100%|███████████████████████████████████████| 1.98G/1.98G [01:52<00:00, 17.5MB/s]
model-00008-of-00008.safetensors: 100%|███████████████████████████████████████| 1.75G/1.75G [01:32<00:00, 19.0MB/s]
Downloading shards: 100%|███████████████████████████████████████████████████████████| 8/8 [14:59<00:00, 112.39s/it]
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████| 8/8 [00:22<00:00,  2.75s/it]
generation_config.json: 100%|█████████████████████████████████████████████████████████████| 123/123 [00:00<?, ?B/s]
[...]

PR #814, opened 2024-07-10
fix model patching for internlm2

eaidova:
# What does this PR do?

Fixes https://github.com/huggingface/optimum-intel/pull/811

## Before submitting


echarlaix (approved):
Thanks for the fix @eaidova

PR #815, opened 2024-07-10
Set default 4-bit compression ratio to 1.0

echarlaix:
Following https://github.com/huggingface/optimum-intel/pull/805

cc @nikita-savelyevv

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -277,7 +277,7 @@ class OVExportCommand(BaseOptimumCLICommand):
             else:
                 quantization_config = {
                     "bits": 8 if is_int8 else 4,
-                    "ratio": 1 if is_int8 else (self.args.ratio or 1.0),
+                    "ratio": 1 if is_int8 else (self.args.ratio or _DEFAULT_4BIT_CONFIG["ratio"]),
                     "sym": self.args.sym or False,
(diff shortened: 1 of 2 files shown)


eaidova (approved):

optimum/commands/export/openvino.py
```diff
@@ -277,7 +277,7 @@ def _get_default_int4_config(model_id_or_path, library_name):
             else:
                 quantization_config = {
                     "bits": 8 if is_int8 else 4,
-                    "ratio": 1 if is_int8 else (self.args.ratio or 0.8),
+                    "ratio": 1 if is_int8 else (self.args.ratio or 1.0),
```
nikita-savelyevv, line 280: Could we import default ratio from [_DEFAULT_4BIT_CONFIG](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py#L107)?

nikita-savelyevv (commented):
Thank you for noticing this!

optimum/commands/export/openvino.py
echarlaix: Makes sense, added in https://github.com/huggingface/optimum-intel/pull/815/commits/4989382a3b7fd89db8c86a43edf7f14f6e36c4d0

nikita-savelyevv (approved):

PR #820, opened 2024-07-12
fix loading models from cache

eaidova:
# What does this PR do?
After these changes https://github.com/openvinotoolkit/openvino/pull/24781 openvino will be not able to use symlinks as paths for reading models. symlinks are produced by HF Hub API during saving models in cache. This PR should resolve problem with reading models downloaded from hub

Fixes # (issue)

## Before submitting


echarlaix (approved):
LGTM, thanks!

PR #821, opened 2024-07-12
Dockerfile update

rbrugaro:
-bump old versions of ipex and torch
-fix syntax that was causing error/warning during build with DOCKER_BUILDKIT=1
-switch ipex installation to install from package instead of source (quicker)
-switch from conda to conda-forge


docker/Dockerfile.intel
```diff
@@ -25,46 +24,39 @@ RUN --mount=type=cache,id=apt-dev,target=/var/cache/apt \
...
-    ~/miniconda.sh -b -p /opt/conda && \
-    rm ~/miniconda.sh && \
-    /opt/conda/bin/conda install -y python=${PYTHON_VERSION} conda-build pyyaml numpy ipython mkl mkl-include ninja cython typing pybind11 Pillow && \
-    /opt/conda/bin/conda clean -ya
+RUN curl -fsSL -v -o ~/miniforge3.sh -O  https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh  && \
+    chmod +x ~/miniforge3.sh && \
+    ~/miniforge3.sh -b -p /opt/miniforge3 && \
+    rm ~/miniforge3.sh && \
+    /opt/miniforge3/bin/conda install -y python=${PYTHON_VERSION} conda-build pyyaml numpy ipython mkl mkl-include ninja cython typing pybind11 Pillow && \
+    /opt/miniforge3/bin/conda clean -ya
```
IlyasMoutawwakil, line 39: can we drop the need for conda at all and simply use pip ?

IlyasMoutawwakil (commented):
thanks for the update !
I think we can further minimize the dockerfile by dropping conda and the multi-stage build, since the dockerfile is quite simple and doesn't do any compilation/building.
That way it can also serve as a minimal and simple recipe for ppl who have access to an ubuntu VM without docker or conda.

rbrugaro:
@IlyasMoutawwakil I made the changes, please have a look!

IlyasMoutawwakil:
Thank you @rbrugaro ! LGTM 
@echarlaix would it make sense to add a build sanity test for the docker image ?

IlyasMoutawwakil (approved):

echarlaix:
> Thank you @rbrugaro ! LGTM @echarlaix would it make sense to add a build sanity test for the docker image ?

Yes definitely !

rbrugaro:
thanks @IlyasMoutawwakil  @echarlaix. I can add a build sanity test. Should this test only verify the build completes and docker run is successful or needs to verify some functionality in the container too?
optimum intel is currently not installed in the Dockerfile that way users can install it with the specific backends and extras they may want while keeping the base image lightweight.  
Your input is appreciated!

IlyasMoutawwakil:
@rbrugaro I think just the verification that the build completes and that docker run is successful are enough.

PR #823, opened 2024-07-15
support assisted decoding in ipex 2.4

jiqing-feng:
Support assisted decoding if ipex >= 2.5


optimum/intel/ipex/modeling_base.py
```diff
@@ -86,10 +87,37 @@ def _is_patched_with_ipex(model, task):
...
+            [
+                (
+                    torch.zeros(1, past_length, past_length, 1, dtype=torch.long).contiguous(),
+                    torch.zeros(max_position, batch_size, num_attn, d_k, dtype=dtype).contiguous(),
+                    torch.zeros(max_position, batch_size, num_attn, d_k, dtype=dtype).contiguous(),
+                    beam_idx_tmp,
+                )
+                for i in range(num_layers)
+            ]
+        )
```
echarlaix, line 117: why not have our own `LlamaDummyPastKeyValuesGenerator` https://github.com/huggingface/optimum/blob/v1.21.2/optimum/utils/input_generators.py#L1122 ?

optimum/intel/ipex/modeling_base.py
```diff
@@ -86,10 +87,37 @@ def _is_patched_with_ipex(model, task):
...
+        num_attn = sample_inputs["past_key_values"][0][0].shape[1]
+        d_k = sample_inputs["past_key_values"][0][0].shape[-1]
+        dtype = sample_inputs["past_key_values"][0][0].dtype
+
+        num_layers = len(sample_inputs["past_key_values"])
+        beam_idx_tmp = torch.zeros((max_position, batch_size), dtype=torch.long).contiguous()
+        past_key_values = tuple(
+            [
+                (
+                    torch.zeros(1, past_length, past_length, 1, dtype=torch.long).contiguous(),
```
echarlaix, line 110: what does this value correspond to ?

optimum/intel/ipex/modeling_base.py
```diff
@@ -693,3 +723,17 @@ def _ipex_reorder_cache(
...
+        new_past_key_values = []
+        for i in range(len(past_key_values)):
+            pkv = []
+            pkv.append(past_key_values[i][0][:, :max_length, :max_length, :])
+            pkv += [past_key_values[i][_] for _ in range(1, 4)]
+            new_past_key_values.append(tuple(pkv))
+        new_past_key_values = tuple(new_past_key_values)
+        return new_past_key_values
+    else:
+        _crop_past_key_values(model, past_key_values, max_length)
```
echarlaix, line 739:
```suggestion
    return _crop_past_key_values(model, past_key_values, max_length)
```

optimum/intel/ipex/modeling_base.py
jiqing-feng: Done.

optimum/intel/ipex/modeling_base.py
jiqing-feng: It's meaningless, this is for keeping the original past length to match transformers functions like [_get_initial_cache_position](https://github.com/huggingface/transformers/blob/main/src/transformers/generation/utils.py#L1401)

jiqing-feng:
Hi @echarlaix . I have fixed all your comments, please re-run the tests. Thx!

optimum/intel/ipex/modeling_base.py
```diff
@@ -472,6 +486,8 @@ def __init__(
 
         if self._is_ipex_exported:
             self._reorder_cache = _ipex_reorder_cache
+            transformers.generation.candidate_generator._crop_past_key_values = _ipex_crop_past_key_values
+            transformers.generation.utils._crop_past_key_values = _ipex_crop_past_key_values
```
echarlaix, line 490: I don't think we should patch a transformers function here

jiqing-feng: I patched the function before `generate` and un-patched it after `generate`. Please review my new changes, thx.

jiqing-feng:
Hi @IlyasMoutawwakil , could you please review this PR? Thx!

optimum/intel/ipex/modeling_base.py
```diff
@@ -609,11 +623,22 @@ def _prepare_generation_config(
...
+        if self._is_ipex_exported and kwargs.get("assistant_model", None):
+            transformers.generation.candidate_generator._crop_past_key_values = _ipex_crop_past_key_values
+            transformers.generation.utils._crop_past_key_values = _ipex_crop_past_key_values
+            patch_functions = True
+        result = super().generate(*args, **kwargs)
+        # Un-patch functions
+        if patch_functions:
+            transformers.generation.candidate_generator._crop_past_key_values = _crop_past_key_values
+            transformers.generation.utils._crop_past_key_values = _crop_past_key_values
+        return result
```
IlyasMoutawwakil, line 641: I think if a user performs generation with error handling, and the generation at `result = super().generate(*args, **kwargs)` fails, transformers will continue to be patched. Let's add some error handling to make sure we don't leave the lib patched and forward the error.

optimum/intel/ipex/modeling_base.py
```diff
@@ -693,3 +718,16 @@ def _ipex_reorder_cache(
...
         )
+
+
+def _ipex_crop_past_key_values(model, past_key_values, max_length):
+    if isinstance(model, IPEXModel):
+        new_past_key_values = []
+        for i in range(len(past_key_values)):
+            pkv = []
+            pkv.append(past_key_values[i][0][:, :max_length, :max_length, :])
+            pkv += [past_key_values[i][_] for _ in range(1, 4)]
```
IlyasMoutawwakil, line 729: why cropping the first one only ?

IlyasMoutawwakil (changes requested):
Can you please add an explanation in the PR of why this patching is necessary ?
Do assistant models interfere with torchscript ?

optimum/intel/ipex/modeling_base.py
jiqing-feng: Because the the real past_key_values are saved in past_key_values[i][1] and past_key_values[i][2] with a fixed_length (pre allocate all memories). So the real length are saved in past_key_values[i][0], the OP will select past key values by past_key_values[i][1][:past_key_values[i][0].shape[2]] and past_key_values[i][2][:past_key_values[i][0].shape[2]]. For more details, please see [IAKV](https://intel.github.io/intel-extension-for-pytorch/cpu/latest/tutorials/llm.html#indirect-access-kv-cache).

optimum/intel/ipex/modeling_base.py
jiqing-feng: Hi @IlyasMoutawwakil . Thanks for your advice. Would you mind giving me more details about it? Can `try cache` handle all cases even we don't know what kind of error will happen?

optimum/intel/ipex/modeling_base.py
jiqing-feng: I use try and except to handle this case, would you please take a review? Thx!

jiqing-feng:
> Can you please add an explanation in the PR of why this patching is necessary ? Do assistant models interfere with torchscript ?



> Can you please add an explanation in the PR of why this patching is necessary ? Do assistant models interfere with torchscript ?

Yes. In assisted decoding, both the target model and the assistant model need to call `crop_pask_key_values`. The patched IPEXModel has a different format of past_key_values, so we need our own crop functions to apply IPEX KV-cache.

optimum/intel/ipex/modeling_base.py
jiqing-feng: I use try and except to handle this case, would you please take a review? Thx!

optimum/intel/ipex/modeling_base.py
```diff
@@ -609,11 +623,30 @@ def _prepare_generation_config(
...
-        return super().generate(*args, **kwargs)
+        # Patch functions to support IAKV cache
+        if self._is_ipex_exported:
+            _patch_crop_past_key_values()
+        try:
+            result = super().generate(*args, **kwargs)
+        except Exception as error:
+            _unpatch_crop_past_key_values()
+            assert False, f"IPEXModelForCausalLM generation failed due to {error}"
+        _unpatch_crop_past_key_values()
```
IlyasMoutawwakil, line 638: forwarding the exception make more sense
```suggestion
        try:
            result = super().generate(*args, **kwargs)
        except Exception as e:
            _unpatch_crop_past_key_values()
            raise e
        else:
            _unpatch_crop_past_key_values()
```

IlyasMoutawwakil:
> Yes. In assisted decoding, both the target model and the assistant model need to call crop_pask_key_values. The patched IPEXModel has a different format of past_key_values, so we need our own crop functions to apply IPEX KV-cache.
Do you want me to write it as a comment in the codes?

Thanks for the explanation ! One last code change and it's good for me 🤗

jiqing-feng:
> > Yes. In assisted decoding, both the target model and the assistant model need to call crop_pask_key_values. The patched IPEXModel has a different format of past_key_values, so we need our own crop functions to apply IPEX KV-cache.
> > Do you want me to write it as a comment in the codes?
> 
> Thanks for the explanation ! One last code change and it's good for me 🤗

I have fixed your required changes. The `_unpatch_crop_past_key_values()` can be safely added at the end the generation because it will revert changes if we do patch and keep the same if we didn't do patch.

IlyasMoutawwakil (approved):

jiqing-feng:
Hi @echarlaix . I have enabled assisted decoding tests in ipex 2.4. Please take a review, thx~

optimum/intel/ipex/modeling_base.py
```diff
@@ -103,11 +132,9 @@ def ipex_jit_trace(model, task, use_cache):
...
 
     model.config.return_dict = False
-
-    if "past_key_values" in sample_inputs:
-        model.config.use_cache = use_cache
-        if not use_cache:
-            sample_inputs.pop("past_key_values")
+    model.config.use_cache = use_cache
+    if "past_key_values" in sample_inputs and not use_cache:
+        sample_inputs.pop("past_key_values")
```
echarlaix, line 137: (minor) since we removed `get_dummy_input` could this be removed ?

optimum/intel/ipex/modeling_base.py
```diff
@@ -652,11 +679,22 @@ def _prepare_generation_config(
...
+            _patch_crop_past_key_values()
+
+        result = super().generate(*args, **kwargs)
+
+        return result
+
+
+def _patch_crop_past_key_values():
+    transformers.generation.candidate_generator._crop_past_key_values = _ipex_crop_past_key_values
+    transformers.generation.utils._crop_past_key_values = _ipex_crop_past_key_values
```
echarlaix, line 697: could this be unpatch after export ?

echarlaix (approved):

optimum/exporters/ipex/model_config.py
```diff
@@ -0,0 +1,98 @@
...
+    FalconOnnxConfig,
+    GPT2OnnxConfig,
+    LlamaOnnxConfig,
+)
+from optimum.utils import DEFAULT_DUMMY_SHAPES
+from optimum.utils.input_generators import DummyPastKeyValuesGenerator, DummyTextInputGenerator
+from optimum.utils.normalized_config import NormalizedTextConfig
+
+
+DEFAULT_DUMMY_SHAPES["batch_size"] = 1
```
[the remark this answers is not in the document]
jiqing-feng: IPEX OP only support the following input:
1. (input_ids = 1; kv-cache >= 0; batch_size >=1)
2. (input_ids >= 1; kv-cache = 0; batch_size >=1)
3. (input_ids > 1; kv-cache > 0; batch_size = 1)

We currently use the 3rd input types.

The 3rd input type is the real case in assisted decoding. If we use the original `DEFAULT_DUMMY_SHAPES`, the dummy input will be the 4th input type (input_ids > 1; kv-cache > 0; batch_size > 1) which will cause core dump of IPEX OP. The 4th input type will not appear in any cases because assisted decoding only supports batch_size=1 in transformers. 

Besides, I have tested the 3rd input type will not have any impact on the performance of IPEXModel greedy_search and beam_search and other decoding methods.

optimum/intel/ipex/modeling_base.py
jiqing-feng: As we discussed in Teams, this is the only way to enable all assisted decoding cases:
1. transformers target model + ipex draft model
2. ipex target model + transformers draft model
3. ipex target model + ipex draft model

Cause the `_crop_past_key_values` function is the same level as the model, we cannot do un-patch inside the generate function because it will run after generate, see [here](https://github.com/huggingface/transformers/blob/main/src/transformers/generation/candidate_generator.py#L209-L227). 

I have checked the model type inside the `_ipex_crop_past_key_values`. It only has impact on IPEX model, transformers model will go into the original function, so there is no risk even we don't unpatch.

jiqing-feng:
Hi @echarlaix . I can do the unpatch for target model only, please look the new changes

PR #824, opened 2024-07-15
Update default NNCF configurations

nikita-malininn:
# What does this PR do?

- Updates default NNCF configuration according to experiments.
- Fixes wrong AWQ parameter usage.

### DIFF
--- a/tests/openvino/test_quantization.py
+++ b/tests/openvino/test_quantization.py
@@ -61,7 +61,13 @@ from optimum.intel import (
     OVWeightQuantizationConfig,
     OVDynamicQuantizationConfig,
 )
-from optimum.intel.openvino.configuration import OVQuantizationMethod, OVQuantizationConfigBase, _DEFAULT_4BIT_CONFIGS
+from optimum.intel.openvino.configuration import (
+    OVQuantizationMethod,
+    OVQuantizationConfigBase,
+    _DEFAULT_4BIT_CONFIGS,
+    _DEFAULT_4BIT_CONFIG,
+)
+from copy import deepcopy
 
 from optimum.intel.openvino.quantization import InferRequestWrapper
 from optimum.intel.utils.import_utils import is_openvino_version, is_transformers_version
@@ -820,6 +826,13 @@ class OVQuantizationConfigTest(unittest.TestCase):
         (dict(bits=8, fast_bias_correction=True, weight_only=False), OVQuantizationConfig, None),
     )
 
+    def get_default_configurations() -> dict:
+        default_configurations = deepcopy(_DEFAULT_4BIT_CONFIGS)
+        default_configurations.update({"default": _DEFAULT_4BIT_CONFIG})
+        return default_configurations
+
+    DEFAULT_CONFIGURATIONS = get_default_configurations()
+
     @parameterized.expand(QUANTIZATION_CONFIGS)
     def test_config_serialization(self, quantization_config: OVQuantizationConfigBase):
         ov_config = OVConfig(quantization_config=quantization_config)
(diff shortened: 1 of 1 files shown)


KodiaqQ:
@nikita-savelyevv, @eaidova, review, please.

echarlaix (approved):

KodiaqQ:
@echarlaix, please do not merge, until the test wouldn't be added (after discussion with @eaidova).

optimum/intel/openvino/configuration.py
```diff
@@ -33,18 +33,17 @@
...
-    "bigscience/bloomz-7b1": {"bits": 4, "sym": False, "group_size": 32, "ratio": 0.6},
     "togethercomputer/RedPajama-INCITE-7B-Instruct": {"bits": 4, "sym": False, "group_size": 128},
     "HuggingFaceH4/zephyr-7b-beta": {
         "bits": 4,
         "sym": True,
         "group_size": 128,
         "ratio": 0.8,
         "dataset": "wikitext2",
-        "awq": True,
+        "quant_method": "awq",
```
nikita-savelyevv, line 46:
```suggestion
        "quant_method": OVQuantizationMethod.AWQ,
```

Here and below

optimum/intel/openvino/configuration.py
```diff
@@ -33,18 +33,17 @@
 logger = logging.getLogger(__name__)
 
 _DEFAULT_4BIT_CONFIGS = {
-    "databricks/dolly-v2-3b": {"bits": 4, "sym": False, "group_size": 128, "ratio": 0.8},
+    "databricks/dolly-v2-3b": {"bits": 4, "sym": False, "group_size": 128, "scale_estimation": True},
     "EleutherAI/gpt-j-6b": {"bits": 4, "sym": False, "group_size": 64},
     "facebook/opt-6.7b": {"bits": 4, "sym": False, "group_size": 64, "ratio": 0.8},
-    "bigscience/bloomz-7b1": {"bits": 4, "sym": False, "group_size": 32, "ratio": 0.6},
```
echarlaix, line 39: Is it removed because the configuration is no more adapted ?

nikita-savelyevv (commented):
Thanks!

optimum/intel/openvino/configuration.py
KodiaqQ: Yes. Since an asymmetric scheme is the default, it would be enough for this model and custom configuration is no longer needed.

optimum/intel/openvino/configuration.py
KodiaqQ: This requires to replace the `OVQuantizationMethod` definition before `_DEFAULT_4BIT_CONFIGS`. Do you agree with that change?

nikita-savelyevv: > This requires to replace the `OVQuantizationMethod` definition before `_DEFAULT_4BIT_CONFIGS`. Do you agree with that change?

Yes, sure

KodiaqQ:
@echarlaix, @nikita-savelyevv, review again, please.

tests/openvino/test_quantization.py
```diff
@@ -849,6 +849,14 @@ def test_config_from_dict(self, quantization_config: dict, config_type: type, wa
             if hasattr(ov_config.quantization_config, k):
                 self.assertEqual(getattr(ov_config.quantization_config, k), v)
 
+    @parameterized.expand(_DEFAULT_4BIT_CONFIGS)
```
nikita-savelyevv, line 852: Could you also please include `_DEFAULT_4BIT_CONFIG` here?

KodiaqQ: I'll do that, but the test wouldn't look acceptable then, since `_DEFAULT_4BIT_CONFIGS` is a dictionary with the `model_id` as key and `_DEFAULT_4BIT_CONFIG` is not.

nikita-savelyevv (approved):

eaidova (approved):

tests/openvino/test_quantization.py
eaidova: this will not work?
@prametrized(_DEFAULT_4BIT_CONFIGS.update({"default": _DEFAULT_4BIT_CONFIG))

KodiaqQ: Unfortunately, no. `.update` method doesn't return any value.

KodiaqQ:
@echarlaix, could you review this again, please?

PR #825, opened 2024-07-15
Infer if the model needs to be exported

echarlaix:
As https://github.com/huggingface/optimum-intel/pull/722/ this PR remove the need to specify if the model needs to be exported or not with `export`

```diff
  from optimum.intel import OVModelForCausalLM

- model = OVModelForCausalLM.from_pretrained("gpt2", export=True)
+ model = OVModelForCausalLM.from_pretrained("gpt2")
```


optimum/intel/openvino/modeling_base.py
```diff
@@ -279,6 +268,59 @@ def _from_pretrained(
...
+
+        ov_files = _find_files_matching_pattern(
+            model_id,
+            pattern=r"(.*)?openvino(.*)?\_model.xml",
+            subfolder=subfolder,
+            use_auth_token=token,
+            revision=revision,
+        )
+        _export = len(ov_files) == 0
+        if _export ^ export:
```
eaidova, line 305: but what if I want force model conversion e.g. I already have converted model with previous openvino version and want to reconvert with new one and store both pt and ov model in the same directory?

eaidova: also how it works for pipelined models, when you need several parts to launch model? (e.g. for seq2seq model you have only openvino_decoder_model, but openvino_encoder_model missed, will it trigger model conversion, I guess not...

AlexKoff88: > but what if I want force model conversion e.g. I already have converted model with previous openvino version and want to reconvert with new one and store both pt and ov model in the same directory?

you should force it with `export=True` in the `.from_pretrained()` API. To store both pt and ov at the same location you need to use AutoModeFor... and OVModelFor... and call `.save_pretrained()` for both.

eaidova: > you should force it with export=True in the .from_pretrained() API. To store both pt and ov at the same location you need to use AutoModeFor... and OVModelFor... and call .save_pretrained() for both.

yes, it should work this way in my understanding too, but looking on condition below, looks like not, because  in code below _export value (that True only if there is no converted models) provided in function for loading model. So force export=True will be ignored

echarlaix: fixed in https://github.com/huggingface/optimum-intel/pull/825/commits/2d4fdbd6f6a406fd11ae20909cb39ec6ec618bd3

PR #830, opened 2024-07-19
allow passing trust remote code for tokenizer init in weight compression

eaidova:
# What does this PR do?

for data-aware weights compression, there is need in tokenizer initialization that may requires to execute remote code


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -311,7 +311,7 @@ def _from_transformers(
             stateful=None,
             load_in_8bit=load_in_8bit,
             quantization_config=quantization_config,
-            **kwargs,
+            trust_remote_code=trust_remote_code**kwargs,
```
eaidova, line 314:
```suggestion
            trust_remote_code=trust_remote_code,
            **kwargs,
```

nikita-savelyevv (approved):

optimum/intel/openvino/quantization.py
```diff
@@ -671,7 +671,9 @@ def _remove_unused_columns(self, dataset: "Dataset"):
     def _prepare_builtin_dataset(self, quantization_config: OVWeightQuantizationConfig):
         from optimum.gptq.data import get_dataset, prepare_dataset
 
-        tokenizer = AutoTokenizer.from_pretrained(quantization_config.tokenizer)
+        tokenizer = AutoTokenizer.from_pretrained(
+            quantization_config.tokenizer, trust_remote_code=quantization_config.trust_remote_code
```
echarlaix, line 675: Instead of adding a new argument why not use directly `model.preprocessors` that could be loaded with https://github.com/huggingface/optimum/blob/5ece6e88156de6a0de848ad6c1af381c81cfbcc3/optimum/utils/save_utils.py#L27

eaidova: but this anyway requires to know trust_remote_code value. Issue what I found, is using optimum-cli for buichuan-7b failed with default int4 configuration, because model and its tokenizer can not be loaded without --trust-remote-code, this method just hiding attempt to load, it can not overcome trust_remote_code

echarlaix: When loading the model `trust_remote_code` will be specified in any case so shouldn't be an issue to use the preprocessors attribute instead of loading the tokenizer here no ? I'm just not sure we want to add a `trust_remote_code` in the quantization config as it's currently just used to load the tokenizer

eaidova: the problem that weight compression can be run when model already converted (e.g. convert model to fp32 and save on disk, then load converted model with quantization config) at this moment we do not have info about remote_code. Do you recommend to accept instance of tokenizer object instead of str? Can it be solution, @nikita-savelyevv @AlexKoff88?

AlexKoff88: We will not be able to serialize quantization_config if `quantization_config.tokenizer` is an object and not a string.

echarlaix: > the problem that weight compression can be run when model already converted (e.g. convert model to fp32 and save on disk, then load converted model with quantization config) at this moment we do not have info about remote_code.

I see, but in this case shouldn't the tokenizer have been saved as well after model conversion ?

nikita-savelyevv: @echarlaix We've had some discussion on this with @eaidova . We've concluded that it is possible that the tokenizer is not present in the same folder as the model, when the model was exported and saved via optimum-intel Python API. For example:
```
model = OVModelForCausalLM.from_pretrained(model_id, export=True)
model.save_pretrained("my_model")
```

Then, if the user decides to compress the saved model with

```
model = OVModelForCausalLM.from_pretrained("my_model", quantization_config=quantization_config|)
```

he will have to pass `tokenizer_id` to `OVWeightQuantizationConfig`, because it was not saved to the disk. With the changes in this PR, the user will also be able to pass `trust_remote_code=True` to quantization config for the tokenizer to be loaded properly.

KodiaqQ:
Absent of this fix in the main branch blocks the calibration. For example, for the `THUDM/chatglm3-6b`, it is not possible to compress the model with **optimum-intel**.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -782,6 +783,9 @@ def _from_pretrained(
 
         if isinstance(quantization_config, dict) and quantization_config == {"bits": 4}:
             quantization_config = _DEFAULT_4BIT_CONFIGS.get(config.name_or_path, _DEFAULT_4BIT_CONFIG)
+            if quantization_config.get("dataset") is not None:
```
IlyasMoutawwakil, line 786:
```suggestion
            if quantization_config.get("dataset", None) is not None:
```

optimum/commands/export/openvino.py
```diff
@@ -275,6 +275,9 @@ def _get_default_int4_config(model_id_or_path, library_name):
                     "scale_estimation": self.args.scale_estimation,
                 }
 
+            if quantization_config.get("dataset") is not None:
```
IlyasMoutawwakil, line 278:
```suggestion
            if quantization_config.get("dataset", None) is not None:
```

AlexKoff88:
@IlyasMoutawwakil, can we merge this as it unlocks model enabling on our side?

AlexKoff88 (approved):

PR #836, opened 2024-07-22
Add dataset for configurations with scale

KodiaqQ:
# What does this PR do?

- Fixes default configurations that contain `scale_estimation`. This algorithm requires a `dataset`.
- Updated test.


KodiaqQ:
@eaidova, @nikita-savelyevv, review, please.

nikita-savelyevv (approved):
Thanks!

eaidova (approved):

KodiaqQ:
> I see that at [openvinotoolkit/openvino.genai#660](https://github.com/openvinotoolkit/openvino.genai/pull/660) you use a specific dataset size. Here, the default value of 128 will be used. Should `num_samples` be also specified here explicitly?

Maybe it should be unified as well. How the correct format of the configuration with the sampled dataset should look?

optimum/intel/openvino/configuration.py
```diff
@@ -40,7 +40,13 @@ class OVQuantizationMethod(str, Enum):
 
 
 _DEFAULT_4BIT_CONFIGS = {
-    "databricks/dolly-v2-3b": {"bits": 4, "sym": False, "group_size": 128, "scale_estimation": True},
+    "databricks/dolly-v2-3b": {
```
AlexKoff88, line 43: We will raise the bar on the HW requirements to do that kind of optimization as data-aware methods require more RAM and time to produce the result.

nikita-savelyevv:
> > I see that at [openvinotoolkit/openvino.genai#660](https://github.com/openvinotoolkit/openvino.genai/pull/660) you use a specific dataset size. Here, the default value of 128 will be used. Should `num_samples` be also specified here explicitly?
> 
> Maybe it should be unified as well. How the correct format of the configuration with the sampled dataset should look?

It should be specified with `num_samples` parameter [here](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py#L257). 

With this, there is another question. In optimum-intel `optimum.gptq` is used for calibration data collection, [here](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/quantization.py#L671). If you've used `openvino.genai` in your experiments on finding optimal configurations with data-aware compression, then I believe the calibration data collection logic is different. Are there guarantees that such differences won't lead to compression configuration not being optimal anymore?

KodiaqQ:
> It should be specified with `num_samples` parameter [here](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py#L257).
> 
> With this, there is another question. In optimum-intel `optimum.gptq` is used for calibration data collection, [here](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/quantization.py#L671). If you've used `openvino.genai` in your experiments on finding optimal configurations with data-aware compression, then I believe the calibration data collection logic is different. Are there guarantees that such differences won't lead to compression configuration not being optimal anymore?

This number is about dataset size, not the calibration size as I can see.
Since the logic of the calibration differs, we need to align `optimum` with `genai` somehow ASAP. May I ask, why it was not done properly before? What should we do now?

nikita-savelyevv:
> This number is about dataset size, not the calibration size as I can see.

`num_samples` corresponds to `subset_size` parameter of `nncf.compress_weights()`, the call is [here](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/quantization.py#L811). I'm not sure how dataset size configuration in openvino.genai is done, but I hope this helps.

> Since the logic of the calibration differs, we need to align optimum with genai somehow ASAP. May I ask, why it was not done properly before? 

I'm not aware whether compression results were compared between openvino.genai and optimum-intel before or not. And I also assume there was no effort for that. 

> What should we do now?

Ideally, we should confirm that models compressed with optimum-intel reach the same metrics as with openvino.genai. I guess that's what we should've done initially when updating configurations here. This would also have helped with finding issues in specifying configurations.

KodiaqQ:
> Ideally, we should confirm that models compressed with optimum-intel reach the same metrics as with openvino.genai. I guess that's what we should've done initially when updating configurations here. This would also have helped with finding issues in specifying configurations.

Filled CVS-147470 and set this PR into Draft since there is no guarantee that tools (and their results) are aligned.

optimum/intel/openvino/configuration.py
KodiaqQ: What should we do then?

AlexKoff88: This model is relatively small and may be it is ok to use data-aware configuration for it. Can you please share RAM and optimization time estimations for the config that you are pushing? If they are moderate we could probably keep your config.

KodiaqQ: Okay. I'll re-calibrate this model along with the `4BIT_DEFAULT` configuration as the reference. Do I need to measure same numbers for other models in this PR?

KodiaqQ: Here are graphs for the **dolly-v2-3b** model on the **main**:
dolly-v2-3b_main
and **nm/fix_defaults** branches:
dolly-v2-3b_branch
@AlexKoff88, @nikita-savelyevv

AlexKoff88: It seems like 32GB is no longer enough to convert 3B model in this case which disrupts UX. @nikita-savelyevv, @alexsu52, can you please comment if we have plans to improve memory consumption in the nearest future?

nikita-savelyevv: @AlexKoff88 There is some work planned in this direction but for now it is how it is. Also, the time increases about 10x, which is probably more critical.

KodiaqQ: FYI, there are other options that exist. The closer one is to reducing the ratio to 0.7:

Config | Word PPL | Diff vs. FP
-- | -- | --
FP16, INT4_ASYM, GS   128, R 0.8 | 17,171 | 0,059685714
FP16, INT4_ASYM, GS 128, R 0.7 | 16,705 | 0,030933426
FP16, INT4_ASYM, GS 128, R 1.0, Scale | 16,845 | 0,039567696

AlexKoff88: Smaller group size can also improve accuracy.

PR #839, opened 2024-07-25
[OpenVINO] Set Left Padding For Text Generation Task

apaniukov:
# What does this PR do?
Set the left padding side during tokenizer conversion for text generation tasks.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -706,6 +706,7 @@ def export_tokenizer(
     tokenizer,
     output: Union[str, Path],
     suffix: Optional[str] = "",
+    task: Optional[str] = None,
 ):
     # avoid circular imports
     from optimum.intel.openvino import OV_DETOKENIZER_NAME, OV_TOKENIZER_NAME
@@ -721,6 +722,11 @@ def export_tokenizer(
 
     if output.exists():
         tokenizer = maybe_convert_tokenizer_to_fast(tokenizer, output)
+
+    if task is not None and task.startswith("text-generation"):
+        logger.info(f"Set padding side to left for `{task}` task.")
+        tokenizer.padding_side = "left"
+        tokenizer.truncation_side = "left"
 
     try:
         converted = convert_tokenizer(tokenizer, with_detokenizer=True)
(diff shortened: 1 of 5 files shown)


optimum/exporters/openvino/convert.py
```diff
@@ -722,6 +724,15 @@ def export_tokenizer(
     if output.exists():
         tokenizer = maybe_convert_tokenizer_to_fast(tokenizer, output)
 
+    if (
+        task is not None
+        and task.startswith("text-generation")
```
AlexKoff88, line 729: Is this only related to exact "text-generation" as there are few variants of it? how about `text2text-generation`?

IlyasMoutawwakil: quoting a response from a transfromers maintainer about left padding for text2text-generation: 

> It can be in the following corner case (all of these must apply):
> - you are doing batch decoding
> - you are passing the start of the decoding ids, in addition to the inputs to the encoder
> - the passed decoding ids don’t have the same length, they must be left-padded
> 
> (if the encoder inputs don’t have the same length, they must be right-padded :eyes: )

> By default no decoder inputs are passed, and thus no left padding is ever needed :slightly_smiling_face:

so overall it seems never needed.

tests/openvino/test_exporters_cli.py
```diff
@@ -171,6 +175,9 @@ def test_exporters_cli_tokenizers(self, task: str, model_type: str):
             if number_of_tokenizers == 1:
                 self.assertTrue("Detokenizer is not supported, convert tokenizer only." in output, output)
 
+            if task.startswith("text-generation") and compare_versions("openvino-tokenizers", ">=", "2024.3.0.0"):
+                self.assertTrue("Set tokenizer padding side to left" in output, output)
```
IlyasMoutawwakil, line 179: can you please use `self.assertIn`, it has better verbosity when the test fails

IlyasMoutawwakil:
can you also run the code styling

apaniukov:
> can you also run the code styling

Done

apaniukov:
There is one issue, but it is in a different part of the library:
```
ruff check . --fix
optimum/intel/pipelines/pipeline_base.py:298:54: F811 Redefinition of unused `PreTrainedTokenizerFast` from line 32
Found 1 error.
make: *** [Makefile:29: style] Error 1
```
Should I fix it too, or it is better to do it in other PR?

IlyasMoutawwakil (approved):

AlexKoff88 (approved):

PR #844, opened 2024-07-25
Limit PyTorch version

helena-intel:
`optimum-cli export openvino -m tiiuae/falcon-7b --weight-format int4 falcon-7b-ov-int4` results in an error with PyTorch 2.4 on Windows: The specified module could not be found. Error loading "[...]fbgem.dll" or one of its dependencies.

IPEX installs a lower version of PyTorch already so this change should be safe while we look into this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 import os
 import re
 import subprocess
+import sys
 
 from setuptools import find_namespace_packages, setup
 
@@ -27,7 +28,7 @@ except Exception as error:
     assert False, "Error: Could not open '%s' due %s\n" % (filepath, error)
 
 INSTALL_REQUIRE = [
-    "torch>=1.11",
+    "torch>=1.11,<2.4" if sys.platform.startswith("win") else "torch>=1.11",
     "transformers>=4.36.0,<4.43.0",
     "optimum@git+https://github.com/huggingface/optimum.git",
     "datasets>=1.4.0",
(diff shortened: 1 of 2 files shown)


eaidova:
@helena-intel  I suppose issue not into specific pytorch version, I found this thread related to your reported issue  https://discuss.pytorch.org/t/failed-to-import-pytorch-fbgemm-dll-or-one-of-its-dependencies-is-missing/201969/2 and it was reported in may, when torch 2.3 released. I suppose the recommendation for fixing that should be not limit pytorch version, but update VS C++ package

helena-intel:
> 
@helena-intel I suppose issue notiwith specific pytorch version, I found this thread related to your reported issue https://discuss.pytorch.org/t/failed-to-import-pytorch-fbgemm-dll-or-one-of-its-dependencies-is-missing/201969/2 and it was reported in may, when torch 2.3 released. I suppose the recommendation for fixing that should be not limit pytorch version, but update VS C++ package

That is annoying. Many of our Windows customers will run into this. But for me the issue was fixed after installing PyTorch 2.3, otherwise I would not have made the PR. I see that most of the comments from that issue are also from after the PyTorch 2.4 release so it seems possible there's a relationship.

setup.py
```diff
@@ -27,7 +28,7 @@
     assert False, "Error: Could not open '%s' due %s\n" % (filepath, error)
 
 INSTALL_REQUIRE = [
-    "torch>=1.11",
+    "torch>=1.11,<2.4" if sys.platform.startswith("win") else "torch>=1.11",
```
echarlaix, line 31: Would prefer to have a warning requiring to downgrade torch to 2.3 or update VS C++  for the impacted architectures instead of pinning torch, would that work for you @helena-intel ?

PR #851, opened 2024-07-28
Update tests in accordance to NNCF 2.12

nikita-savelyevv:
# What does this PR do?

Changes:
- Tweak some reference values
- Update model ids for `test_exporters_cli_int4_with_local_model_and_default_config`
- Add SDPA layer to ignored scope in `OVTrainer` tests for `bert` model. This layer was not quantized before, so this does not reduce the number of FQ nodes.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -114,7 +114,6 @@ _DEFAULT_4BIT_CONFIGS = {
     "microsoft/phi-2": {"bits": 4, "sym": False, "group_size": 128, "ratio": 0.9},
 }
 
-
 _DEFAULT_4BIT_CONFIG = {
     "bits": 4,
     "ratio": 1.0,
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -13,13 +13,14 @@
 #  limitations under the License.
 import copy
 import inspect
+import json
 import logging
 from dataclasses import dataclass
 from enum import Enum
+from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
 import torch
-from transformers import PretrainedConfig
 from transformers.utils.quantization_config import QuantizationConfigMixin
 
 from optimum.configuration_utils import BaseConfig
@@ -121,6 +122,32 @@ _DEFAULT_4BIT_CONFIG = {
     "group_size": 128,
     "all_layers": None,
 }
+
+
+def _check_default_4bit_configs(model_id_or_path: str):
+    if model_id_or_path in _DEFAULT_4BIT_CONFIGS:
+        return _DEFAULT_4BIT_CONFIGS[model_id_or_path]
+
+    config_path = Path(model_id_or_path) / "config.json"
+    if config_path.exists():
+        with config_path.open("r") as config_f:
+            config = json.load(config_f)
+            original_model_name = config.get("_name_or_path", "")
+        if original_model_name in _DEFAULT_4BIT_CONFIGS:
+            return _DEFAULT_4BIT_CONFIGS[original_model_name]
+
+    return None
+
+
+def get_default_int4_config(model_id_or_path: str):
+    """
+    Args:
+        model_id_or_path (`str`):
+            id of the model or path to it.
+    Returns:
+        Default int4 config for the given model or generic default int4 config.
+    """
+    return _check_default_4bit_configs(model_id_or_path) or _DEFAULT_4BIT_CONFIG
 
 
 @dataclass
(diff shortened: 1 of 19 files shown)


tests/openvino/test_exporters_cli.py
```diff
@@ -86,31 +86,31 @@ class OVCLIExportTestCase(unittest.TestCase):
     )
 
     TEST_4BIT_CONFIGURATONS = [
-        ("text-generation-with-past", "opt125m", "int4_sym_g128", 4, 144),
+        ("text-generation-with-past", "opt125m", "int4_sym_g128", 4, 72),
```
nikita-savelyevv, line 89: Starting form NNCF 2.12, symmetric compression of weights is performed without zero point. This halves the number of reference int4 constants, because for each symmetrically compressed weight there used to be 2 low precision constants: one for the weight and one for the zero point.

cc @l-bat

tests/openvino/test_quantization.py
```diff
@@ -82,7 +82,7 @@
 class OVQuantizerTest(unittest.TestCase):
     SUPPORTED_ARCHITECTURES_TORCH_MODEL = (
         (OVModelForSequenceClassification, "bert", 32 if is_transformers_version("<", "4.41.0") else 22, 35),
-        (OVModelForCausalLM, "gpt2", 41 if is_transformers_version("<", "4.42.0") else 21, 3),
+        (OVModelForCausalLM, "gpt2", 41 if is_transformers_version("<", "4.42.0") else 21, 23),
```
nikita-savelyevv, line 85: The number of int8 constants has increased due to improvements in https://github.com/openvinotoolkit/nncf/pull/2713

cc @AlexanderDokuchaev

optimum/intel/openvino/configuration.py
```diff
@@ -39,10 +39,20 @@ class OVQuantizationMethod(str, Enum):
...
+    "group_size": 128,
+    "all_layers": None,
+}
+
+
 _DEFAULT_4BIT_CONFIGS = {
     "databricks/dolly-v2-3b": {"bits": 4, "sym": False, "group_size": 128, "ratio": 0.8},
     "EleutherAI/gpt-j-6b": {"bits": 4, "sym": False, "group_size": 64},
     "facebook/opt-6.7b": {"bits": 4, "sym": False, "group_size": 64, "ratio": 0.8},
+    "bigscience/bloomz-7b1": dict(_DEFAULT_4BIT_CONFIG, **{"group_size": 32}),
```
nikita-savelyevv, line 55: Fix after removal of this config in #824. I've updated the default config with `group_size=32`, otherwise compression with `group_size=128` and `ratio=1.0` fails.

cc @KodiaqQ

nikita-savelyevv:
Keeping in draft until release

optimum/intel/openvino/configuration.py
KodiaqQ: What exactly is it fix? Test or real model failure?

nikita-savelyevv: @KodiaqQ, both

KodiaqQ: Please provide an example of a failing model with this configuration. My experiments show that **bloomz-7b1** successfully quantized with the `group_size=128`. The test is not the case.

nikita-savelyevv: @KodiaqQ that's right. I've updated the test case to rely on different model ids. 

As I understand this test is to check that the default config for a specific model is used, and it is not critical that a tiny model is `falcon-40b`, but config for `falcon-7b` is used. I've selected this pair of models because `falcon-7b` has `"group_size"=64` in its default config and tiny `falcon-40b` is one of the few models that actually can be compressed with such group size.

cc @eaidova

echarlaix (approved):
LGTM thanks @nikita-savelyevv, do you know the approximate date for the next nncf release ? Also thanks for fixing `test_exporters_cli_int4_with_local_model_and_default_config`, could this be added in a second PR that we can merge today ? Let me know if you'd like me to do it instead

nikita-savelyevv:
@echarlaix The release should actually happen today, so I think we can wait a bit and merge this in one go

echarlaix:
> @echarlaix The release should actually happen today, so I think we can wait a bit and merge this in one go

Works for me !

echarlaix:
Can you update `TEST_4BIT_CONFIGURATONS` https://github.com/huggingface/optimum-intel/blob/7a224c2419240d5fb58f2f75c2e29f179ed6da28/tests/openvino/test_exporters_cli.py#L88 (fixing conflicts with https://github.com/huggingface/optimum-intel/pull/855)

nikita-savelyevv:
@echarlaix Done. The release is out :)

echarlaix:
@nikita-savelyevv can you take a look at the failing tests before we can merge ?

optimum/intel/openvino/trainer.py
```diff
@@ -153,6 +153,7 @@
         "{re}.*Embedding.*",
         "{re}.*add___.*",
         "{re}.*layer_norm_.*",
+        "{re}.*scaled_dot_product_attention_0",
```
nikita-savelyevv, line 156: In https://github.com/openvinotoolkit/nncf/pull/2761, support for SDPA was added. However, it does not work properly for `create_compressed_model` API. Adding to ignored scope for now. This does not change the resulting number of FQ nodes, because SDPA was not quantized before too. Ticket: 148687.

cc @AlexanderDokuchaev

PR #856, opened 2024-07-31
Support Transformers 4.43

IlyasMoutawwakil:
# What does this PR do?

Adds support of transfromers 4.43 as in https://github.com/huggingface/optimum/pull/1971, mainly:
- whisper introducing a new forward input argument cache_position.
- `_reorder_cache` removed from some causal models to use `Cache.reorder_cache` instead.
- missing `model.dtype` used by transfromers pipelines to cast some features (pixels, audio samples, etc)

The rest is changes in the testing suite to adapt to the modeling changes in the library.
Also fixes the failing OV training tests on main.

## Before submitting


IlyasMoutawwakil:
I think it's ready.
ITREX still needs to fix `qbits` modeling for WOQ. OV looks good to me (compared to CI on main).

IlyasMoutawwakil:
There's a qwen test that fails with transfromers 4.37 even tho I did add the `transformers_stream_generator` package 
https://github.com/huggingface/optimum-intel/actions/runs/10213421527/job/28258759384?pr=856#step:5:2163

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -915,415 +928,19 @@ def _from_pretrained(
...
+            def __init__(self):
+                self.conv1 = self.Conv(stride=(1,))
+                self.conv2 = self.Conv(stride=(2,))
 
-        # 8. If we return all segments, the predicted output sequences are put under `"sequences"`.
-        if return_segments:
-            return {"sequences": sequences, "segments": current_segments}
+            class Conv:
+                def __init__(self, stride):
+                    self.stride = stride
```
echarlaix, line 943: why is this needed ?

echarlaix (approved):
Looks great, thanks a lot @IlyasMoutawwakil !

optimum/intel/openvino/modeling_seq2seq.py
IlyasMoutawwakil: this bypasses the need to override the generate method to compute `input_stride` manually
there's a comment right afterwards that explains it

IlyasMoutawwakil: the problem is actually in transformers modeling because there's actually no need to compute the input stride, it's always the same ; the stride of those two convolutions is hard coded in the whisper modeling
https://github.com/huggingface/transformers/blob/main/src/transformers/models/whisper/modeling_whisper.py#L1010-L1011

helena-intel (approved):
Thank you! Also really appreciate keeping transformers 4.36 support.

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -915,415 +935,19 @@ def _from_pretrained(
...
 
-        return sequences
+    # a dummy model attribute that's used in the generate method to compute the input stride
+    # input_stride = self.model.encoder.conv1.stride[0] * self.model.encoder.conv2.stride[0]
+    model = DummyWhisperModel()
```
IlyasMoutawwakil, line 953: I found a way to keep support of whisper by using `WhisperGenerationMixin` from 4.37 and above and this simple class otherwise:
```python
    class WhisperGenerationMixin:
        generate = WhisperForConditionalGeneration.generate
```
which also works thanks to the dummy model fix 🤗

IlyasMoutawwakil:
need to resolve conflicts from merged PRs

tests/openvino/test_exporters_cli.py
```diff
@@ -253,10 +253,12 @@ def test_exporters_cli_int4(self, task: str, model_type: str, option: str, expec
 
     def test_exporters_cli_int4_with_local_model_and_default_config(self):
         with TemporaryDirectory() as tmpdir:
-            pt_model = AutoModelForCausalLM.from_pretrained(MODEL_NAMES["falcon-40b"])
-            # overload for matching with default configuration
-            pt_model.config._name_or_path = "tiiuae/falcon-7b-instruct"
+            model_id = "bigscience/bloomz-560m"
```
echarlaix, line 256: I think we can keep using `MODEL_NAMES["falcon-40b"]` which should be much smaller no ?

optimum/intel/openvino/configuration.py
```diff
@@ -90,8 +90,8 @@ class OVQuantizationMethod(str, Enum):
     "bigscience/bloomz-560m": {
         "bits": 4,
         "sym": True,
-        "group_size": 64,
-        "ratio": 0.8,
+        "group_size": 128,
+        "ratio": 1.0,
```
echarlaix, line 94:
```suggestion
        "group_size": 64,
        "ratio": 0.8,
```

optimum/intel/openvino/modeling_base.py
```diff
@@ -103,6 +104,13 @@ def __init__(
         if enable_compilation:
             self.compile()
 
+    @property
+    def device(self) -> torch.device:
+        """
+        `torch.device`: The device on which the module is (for torch compatibility).
+        """
+        return torch.device(self._device.lower())
```
echarlaix, line 112: This won't work when device is set to "gpu" / "npu"

echarlaix (approved):

tests/openvino/test_exporters_cli.py
IlyasMoutawwakil: done

optimum/intel/openvino/modeling_base.py
IlyasMoutawwakil: I guess it makes sense to always return `torch.device("cpu")` then ? since this property is mostly used by `generate` method to check input devices and transformers `pipeline` to put processed inputs on the model's device.

IlyasMoutawwakil:
ready to merge

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -832,26 +850,25 @@ class OVModelForSpeechSeq2Seq(OVModelForSeq2SeqLM):
...
+        attention_mask=None,
         head_mask=None,
         decoder_head_mask=None,
         cross_attn_head_mask=None,
         use_cache=None,
         encoder_outputs=None,
         **kwargs,
-    ) -> Dict:
-        if decoder_attention_mask is None:
-            decoder_attention_mask = torch.ones_like(input_ids).to(input_ids.device)
```
echarlaix, line 847: could this be an issue for model that were exported with `decoder_attention_mask` ?

IlyasMoutawwakil: done

optimum/intel/openvino/utils.py
```diff
@@ -21,6 +21,7 @@
 from typing import Tuple, Type, Union
 
 import numpy as np
+import torch
```
eaidova, line 24: should we be more careful importing torch here? I suppose it may be not available

eaidova (approved):

optimum/intel/openvino/utils.py
IlyasMoutawwakil: `torch` is a requirement in optimum https://github.com/huggingface/optimum/blob/main/setup.py#L19

tests/openvino/test_modeling.py
```diff
@@ -943,7 +943,7 @@ def test_beam_search(self, model_arch):
             eos_token_id=None,
         )
 
-        if model_arch == "minicpm":
+        if model_arch in ["minicpm", "internlm2"]:
```
IlyasMoutawwakil, line 946: is this okay ? idk why the generation loop is not deterministic with `internlm2`

eaidova: yes, it is due to random initialization of pt model

IlyasMoutawwakil:
looks good, the failing tests in windows are mostly os errors, I can try fixing them in another PR

PR #859, opened 2024-08-06
Added custom OV configs for VAE encoder and decoder

nikita-savelyevv:
# What does this PR do?

Introduce a way to provide custom OV configs for VAE Decoder and Encoder models in SD pipeline. This is useful to specify inference precision of these models. Example usage:
```
pipeline = OVStableDiffusionPipeline.from_pretrained(
    model_path,
    device="GPU",
    vae_decoder_ov_config={"INFERENCE_PRECISION_HINT": "f16"},
    vae_encoder_ov_config={"INFERENCE_PRECISION_HINT": "f16"},
)
```

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -98,6 +98,8 @@ class OVStableDiffusionPipelineBase(OVBaseModel, OVTextualInversionLoaderMixin):
         ov_config: Optional[Dict[str, str]] = None,
         model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
         quantization_config: Optional[Union[OVWeightQuantizationConfig, Dict]] = None,
+        vae_decoder_ov_config: Optional[Dict[str, str]] = None,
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -116,15 +118,23 @@ def __init__(
         else:
             self._model_save_dir = model_save_dir
 
-        self.vae_decoder = OVModelVaeDecoder(vae_decoder, self)
+        default_vae_ov_config = deepcopy(self.ov_config)
+        if "GPU" in self._device:
+            default_vae_ov_config.update({"INFERENCE_PRECISION_HINT": "f32"})
```
eaidova, line 123: I guess it can be done simplier, we can just check that inference precision hint set in config before and do not update it if it was provided

nikita-savelyevv: Thanks for the suggestion! Updated

eaidova (approved):

AlexKoff88 (approved):

IlyasMoutawwakil (approved):

PR #861, opened 2024-08-07
Added a warning for missing --weight-format argument

nikita-savelyevv:
**Reason for change**

When the following command is executed
```
optimum-cli export openvino -m TinyLlama/TinyLlama-1.1B-Chat-v1.0 --sym ./tiny-llama
```
the model will be exported with compression to int8_asym, despite `--sym` argument was provided. This is because when `--weight-format` argument is not provided, the default compression is applied, and by default int8_asym is applied.

This behavior has lead to some confusion. Because int8_sym was expected to be applied in such case.

**Changes**

Add a warning stating that compression parameters will not be applied because precision was not specified with `--weight-format` argument.

## Before submitting

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -192,9 +192,9 @@ def parse_args_openvino(parser: "ArgumentParser"):
 
 def no_compression_parameter_provided(args):
     return all(
-        map(
-            lambda x: x is None,
-            (
+        (
+            it is None
+            for it in (
                 args.ratio,
                 args.group_size,
                 args.sym,
@@ -203,7 +203,7 @@ def no_compression_parameter_provided(args):
                 args.num_samples,
                 args.awq,
                 args.sensitivity_metric,
-            ),
+            )
         )
     )
 
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -190,6 +190,24 @@ def parse_args_openvino(parser: "ArgumentParser"):
     )
 
 
+def no_compression_parameter_provided(args):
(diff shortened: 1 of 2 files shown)


optimum/commands/export/openvino.py
```diff
@@ -230,23 +248,17 @@ def run(self):
 
         if self.args.weight_format is None:
             ov_config = None
+            if not no_compression_parameter_provided(self.args):
+                logger.warning(
+                    "The provided compression parameters will not affect conversion because of the missing --weight-format argument."
+                )
```
nikita-savelyevv, line 254: To be honest, this warning is not very visible because it gets lost with other info/warning messages related to conversion.

@AlexKoff88 What do you think about raising an exception in this case? My view is that if `--weight-format` was omitted, then no compression parameters should be accepted.

optimum/exporters/openvino/convert.py
```diff
@@ -635,9 +635,9 @@ def export_from_model(
             if is_nncf_available():
                 from ...intel.openvino.configuration import OVConfig
 
-                ov_config = OVConfig(quantization_config={"bits": 8})
+                ov_config = OVConfig(quantization_config={"bits": 8, "sym": False})
 
-                logger.info("The model weights will be quantized to int8.")
+                logger.info("The model weights will be quantized to int8_asym.")
```
nikita-savelyevv, line 640: Just to be more apparent

nikita-savelyevv:
cc @l-bat

AlexKoff88 (approved):

AlexKoff88 (approved):

PR #864, opened 2024-08-08
Fix data collision in caching in InferRequestWrapper

sbalandi:
# What does this PR do?

`InferRequestWrapper` collects inputs data and use dict `tensor_cache` to avoid extra data copying. Value of each item of inputs is converted to hash as `hash(input_tensor.tobytes())`, hash set as key of `tensor_cache` and `deepcopy` of value set as value. If inputs has several items with same tensor value, but different structure, for example `{'input0': tensor([[11]]), 'input1': tensor([11])}`, they would have same hash and one of this value would be collected with wrong shape. This pr fixes it.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -132,9 +132,10 @@ class InferRequestWrapper:
             data_hash = hash(data.tobytes())
 
             # Avoid data copying if tensor contains data encountered earlier
-            if data_hash not in self.tensor_cache:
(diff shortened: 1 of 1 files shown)


sbalandi:
@eaidova  please, review

eaidova (approved):

AlexKoff88 (approved):

IlyasMoutawwakil (approved):

PR #865, opened 2024-08-11
Support SentenceTransformers models inference in optimum

aleksandr-mokrov:
# What does this PR do?
Support SentenceTransformers models inference in optimum
You can use SentenceTransformers for feature extraction like in original pipeline (https://huggingface.co/sentence-transformers/all-mpnet-base-v2#usage-sentence-transformers):

```python
from optimum.intel.openvino import OVSentenceTransformer

model_id="sentence-transformers/all-mpnet-base-v2"
model = OVSentenceTransformer.from_pretrained(model_id, export=True)
sentences = ["This is an example sentence", "Each sentence is converted"]
embeddings = model.encode(sentences)
```

## Before submitting


optimum/intel/openvino/modeling.py
```diff
@@ -371,23 +373,27 @@ class OVModelForFeatureExtraction(OVModel):
 
     def __init__(self, model=None, config=None, **kwargs):
         super().__init__(model, config, **kwargs)
+        
+        from sentence_transformers import SentenceTransformer
```
eaidova, line 377: I suppose original model class should not depends from sentence transformers. Can we create separated model class OVModelForSentenceTransformers that will have this functionality. We should keep default model class from models exported with library_name="transformers" and have separated one for exported with "sentence_transformers". Both model variants are valid and should work, now you override to have only sentence transformers.

eaidova, line 377: it is also good to add checks that sentence transformers available here and check it before imports https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/utils/import_utils.py

jdchmiel:
I used this PR and sample code to achieve 7x latency improvement vs sentencetransformers! Will this be merged soon or the alternate path of a new "OVModelForSentenceTransformers" be created?  I can take a stab at that but don't want to step on toes or duplicate work, but I also do not want to wait for something that nobody has plans or has prioritized to do!

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,96 @@
...
+from .modeling import MODEL_START_DOCSTRING, OVModel
+
+
+@add_start_docstrings(
+    """
+    OpenVINO Model for feature extraction tasks for Sentence Transformers.
+    """,
+    MODEL_START_DOCSTRING,
+)
+class OVModelForSTFeatureExtraction(OVModel):
```
eaidova, line 19: model class naming is unclear, maybe better name it as `OVModelForSentenceTransformer`?

optimum/intel/openvino/modeling.py
```diff
@@ -451,6 +443,11 @@ def _from_transformers(
         else:
             ov_config = OVConfig(dtype="fp32")
 
+        if is_sentence_transformers_available():
+            library_name = "sentence_transformers"
```
eaidova, line 447: I do not think that it is good idea to switch model export library by default if lib available. Model can be not supported by sentence transformers. It is better to have this logic in specific class and keep export part in base model without changes.

optimum/intel/openvino/modeling.py
```diff
@@ -464,17 +461,33 @@ def _from_transformers(
...
         config.save_pretrained(save_dir_path)
-        return cls._from_pretrained(
+        model = cls._from_pretrained(
             model_id=save_dir_path,
             config=config,
             load_in_8bit=load_in_8bit,
             quantization_config=quantization_config,
             **kwargs,
         )
+        if len(model.model.outputs) == 2:  # Sentence Transormers output
```
eaidova, line 475: I think it is also good to check names as model can have 2 output due to different reasons

optimum/intel/openvino/modeling.py
```diff
@@ -464,17 +461,33 @@ def _from_transformers(
...
+                model=model.model,
+                config=config,
+                model_save_dir=save_dir_path,
+                quantization_config=quantization_config,
+                model_id=model_id,
+                **kwargs,
+            )
+            return model
+        else:
+            if is_sentence_transformers_available():
```
eaidova, line 488: I do not think that this warning is relevant as it is expected to get transformers-based model instance (from user understanding it is desire behaviour if this model class selected). I think it is better provide opposite warning if we understand that sentence-transformers loaded in this class and recommend to use dedicated class explicitly

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,178 @@
...
+            "token": None,
+            "trust_remote_code": False,
+            "revision": None,
+            "local_files_only": False,
+            "model_max_length": 384,
+        }
+        tokenizer = AutoTokenizer.from_pretrained(
+            self.model_id,
+            **tokenizer_args,
+        )
```
IlyasMoutawwakil, line 146: loading a the tokenizer with each call to `tokenize` doesn't make sense, can you move this to `__init__` (so that we can use the model args ?)

optimum/intel/openvino/modeling.py
```diff
@@ -468,13 +468,30 @@ def _from_transformers(
...
+            from .modeling_sentence_transformers import OVModelForSentenceTransformer
+
+            model = OVModelForSentenceTransformer(
+                model=model.model,
+                config=config,
+                model_save_dir=save_dir_path,
+                quantization_config=quantization_config,
+                model_id=model_id,
+                **kwargs,
+            )
```
IlyasMoutawwakil, line 493: we should test this part as well

optimum/intel/openvino/modeling_sentence_transformers.py
aleksandr-mokrov: loading the tokenizer was moved

optimum/intel/openvino/modeling.py
aleksandr-mokrov: It was mistake in the pipeline. This logic was moved to _from_pretrained, test was added.

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,182 @@
...
+            self.orig_model_id_or_path,
+            **self.tokenizer_args,
+        )
+
+    def _save_pretrained(self, save_directory: Union[str, Path]):
+        super()._save_pretrained(save_directory)
+        self.tokenizer.save_pretrained(save_directory)
+
+    def forward(self, inputs: Dict[str, torch.Tensor]):
+        self.compile()
```
IlyasMoutawwakil, line 53: why is this being called at each forward pass ?

eaidova: that is standard formatting for all ov model classes, compile() hass check inside and run compilation only if self.request is None

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,182 @@
...
+        self._text_length = MethodType(SentenceTransformer._text_length, self)
+        self.default_prompt_name = None
+        self.truncate_dim = None
+        self.orig_model_id_or_path = orig_model_id_or_path
+        self.tokenizer_args = {
+            "token": None,
+            "trust_remote_code": False,
+            "revision": None,
+            "local_files_only": False,
+            "model_max_length": 384,
```
IlyasMoutawwakil, line 41: why 384 ?

aleksandr-mokrov: From defaults, but it's not really needed. Removed.

eaidova:
@echarlaix  @IlyasMoutawwakil  could you please help to retrigger ci?

tests/openvino/test_modeling_sentence_transformers.py
```diff
@@ -0,0 +1,62 @@
...
+        "bert",
+        "mpnet",
+    )
+
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    def test_compare_to_transformers(self, model_arch):
+        model_id = MODEL_NAMES[model_arch]
+        set_seed(SEED)
+        ov_model = OVModelForSentenceTransformer.from_pretrained(
+            model_id, from_transformers=True, ov_config=F32_CONFIG
```
echarlaix, line 50: here we can use `export` directly 
```suggestion
            model_id, export=True, ov_config=F32_CONFIG
```

tests/openvino/test_modeling.py
```diff
@@ -601,6 +602,20 @@ def test_pipeline(self, model_arch):
         del model
         gc.collect()
 
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    def test_sentence_tenaformers_pipeline(self, model_arch):
```
echarlaix, line 606: small typo 
```suggestion
    def test_sentence_transformers_pipeline(self, model_arch):
```

also we can also remove pipeline for clarity as it's used for other tests to refer to the transformers pipelines 

```suggestion
    def test_sentence_transformers(self, model_arch):
```

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,180 @@
...
+from .modeling import MODEL_START_DOCSTRING, OVModel
+
+
+@add_start_docstrings(
+    """
+    OpenVINO Model for feature extraction tasks for Sentence Transformers.
+    """,
+    MODEL_START_DOCSTRING,
+)
+class OVModelForSentenceTransformer(OVModel):
```
echarlaix, line 25: For transformers we have `AutoModelForXxx` -> `OVModelForXxx`
For diffusers we have  `XxxPipeline` -> `OVXxxPipeline`
For sentence transformers what do you think about `SentenceTransformer` -> `OVSentenceTransformer` ?


```suggestion
class OVSentenceTransformer(OVModel):
```
also cc @eaidova

optimum/intel/openvino/modeling.py
```diff
@@ -412,6 +413,120 @@ def forward(
...
+            from .modeling_sentence_transformers import OVModelForSentenceTransformer
+
+            return OVModelForSentenceTransformer(
+                model=model,
+                config=config,
+                model_save_dir=model_cache_path.parent,
+                quantization_config=quantization_config,
+                orig_model_id_or_path=model_id,
+                **kwargs,
+            )
```
echarlaix, line 520: Would be in favor of raising an error instead here (stating that `OVModelForSentenceTransformer` should be used to load this model instead)

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,180 @@
...
+    export_feature = "feature-extraction"
+
+    def __init__(self, model=None, config=None, orig_model_id_or_path=None, **kwargs):
```
echarlaix, line 28: if used to load the tokenizer why not load it in _from_pretrained directly :

```suggestion
    def __init__(self, model=None, config=None, tokenizer=None, **kwargs):
```

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,180 @@
...
+        self.tokenizer_args = {
+            "token": None,
+            "trust_remote_code": False,
+            "revision": None,
+            "local_files_only": False,
+        }
+        self.tokenizer = AutoTokenizer.from_pretrained(
+            self.orig_model_id_or_path,
+            **self.tokenizer_args,
+        )
```
echarlaix, line 45: would be nice to use the arguments provided in https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_base.py#L224-L234

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,180 @@
...
+            task=task or cls.export_feature,
+            subfolder=subfolder,
+            revision=revision,
+            cache_dir=cache_dir,
+            token=token,
+            local_files_only=local_files_only,
+            force_download=force_download,
+            trust_remote_code=trust_remote_code,
+            ov_config=ov_config,
+            library_name="sentence_transformers",
```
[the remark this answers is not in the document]
aleksandr-mokrov: If `library_name` is not specified we use transformer pipeline for compatibility: https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/__main__.py#L239.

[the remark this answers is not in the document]
echarlaix: > If library_name is not specified we use transformer pipeline for compatibility: https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/__main__.py#L239.

Yes, I think we could remove this now that we have a `OVModelForSentenceTransformer` class to load the resulting models, even though we need to be careful as this is a breaking change, cc @eaidova to also have your opinion on this

optimum/intel/openvino/modeling_sentence_transformers.py
aleksandr-mokrov: fixed

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,180 @@
...
+            config=config,
+            load_in_8bit=load_in_8bit,
+            quantization_config=quantization_config,
+            orig_model_id_or_path=model_id,
+            **kwargs,
+        )
+
+        model.model_id = model_id
+
+        return model
```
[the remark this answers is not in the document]
aleksandr-mokrov: fixed

optimum/intel/openvino/modeling_sentence_transformers.py
aleksandr-mokrov: fixed

optimum/intel/openvino/modeling.py
aleksandr-mokrov: done

optimum/intel/openvino/modeling_sentence_transformers.py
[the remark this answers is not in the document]
eaidova: can we remove this behvaiour later (on next release)? there are embeddings models integrations in some libs that expects models in transformers format only and it will requires some time to switch them to support models converted as sentence transformers (or at least updating recommendation to explicitly set --library for this case)

andrei-kochin:
@echarlaix could you please trigger the CI for this?

optimum/intel/openvino/modeling.py
```diff
@@ -412,6 +413,110 @@ def forward(
...
+            kwargs["generation_config"] = generation_config
+        except Exception:
+            pass
+
+        if {"token_embeddings", "sentence_embedding"}.issubset(
+            {name for output in model.outputs for name in output.names}
+        ):  # Sentence Transormers outputs
+            raise ValueError(
+                "This model is Sentence Tranfromers converted model. Please use OVSentenceTransformer explicitly for this model."
+            )
```
echarlaix, line 510: Could this be moved in the `__init__` (before anything else to avoid any potential model compilation) so that we don't overwrite `_from_pretrained` if this is the only difference

optimum/intel/__init__.py
```diff
@@ -180,6 +181,21 @@
     _import_structure["neural_compressor"].append("INCStableDiffusionPipeline")
 
 
+try:
+    if not (is_openvino_available() and is_sentence_transformers_available()):
+        raise OptionalDependencyNotAvailable()
+except OptionalDependencyNotAvailable:
+    _import_structure["utils.dummy_openvino_and_sentence_transformers_objects"] = [
```
echarlaix, line 188: `dummy_openvino_and_sentence_transformers_objects.py` missing

optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,179 @@
...
+            "revision": revision,
+            "local_files_only": local_files_only,
+        }
+        if tokenizer_kwargs:
+            kwargs["tokenizer_args"].update(tokenizer_kwargs)
+
+        tokenizer = AutoTokenizer.from_pretrained(
+            model_id,
+            **tokenizer_args,
+        )
```
echarlaix, line 136: shouldn't this be in `_from_pretrained` instead ? (tokenizer should be loaded when loading the model not just only when the model is also converted to the OpenVINO format), suggesting https://github.com/huggingface/optimum-intel/pull/888/commits/ec5199921297e27c6ceebf389845d07a2bea8875 let me know what you think

tests/openvino/test_modeling.py
```diff
@@ -601,6 +602,19 @@ def test_pipeline(self, model_arch):
...
+        """
+        Check if we call OVModelForFeatureExtraction passing saved ir-model with outputs
+        from Sentence Transformers then an appropriate exception raises.
+        """
+        model_id = MODEL_NAMES[model_arch]
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            save_dir = str(tmp_dir)
+            OVSentenceTransformer.from_pretrained(model_id, export=True).save_pretrained(save_dir)
+            with pytest.raises(ValueError):
+                OVModelForFeatureExtraction.from_pretrained(save_dir)
```
echarlaix, line 616:
```suggestion
            with self.assertRaises(Exception) as context:
                OVModelForFeatureExtraction.from_pretrained(save_dir)
            self.assertIn("Please use OVSentenceTransformer", str(context.exception))
```

optimum/intel/utils/dummy_openvino_and_sentence_transformers_objects.py
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .import_utils import DummyObject, requires_backends
+
+
+class OVSentenceTransformer(metaclass=DummyObject):
+    _backends = ["openvino"]
```
echarlaix, line 19:
```suggestion
    _backends = ["openvino", "sentence_transformers"]
```

tests/openvino/test_modeling_sentence_transformers.py
```diff
@@ -0,0 +1,60 @@
...
+    SUPPORTED_ARCHITECTURES = (
+        "bert",
+        "mpnet",
+    )
+
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    def test_compare_to_transformers(self, model_arch):
+        model_id = MODEL_NAMES[model_arch]
+        set_seed(SEED)
+        ov_model = OVSentenceTransformer.from_pretrained(model_id, export=True, ov_config=F32_CONFIG)
```
echarlaix, line 49: Also could we add a test to make sure the model is correctly saved and can be correctly loaded as done in https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_modeling.py#L118 or https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_modeling.py#L1280

echarlaix:
To fix the code style check you can do the following : 
```
pip install .[quality]
make style
```

tests/openvino/test_modeling_sentence_transformers.py
aleksandr-mokrov: A test was added

optimum/intel/openvino/modeling.py
```diff
@@ -369,6 +369,13 @@ class OVModelForFeatureExtraction(OVModel):
     auto_model_class = AutoModel
 
     def __init__(self, model=None, config=None, **kwargs):
+        if {"token_embeddings", "sentence_embedding"}.issubset(
+            {name for output in model.outputs for name in output.names}
+        ):  # Sentence Transormers outputs
+            raise ValueError(
+                "This model is Sentence Tranfromers converted model. Please use OVSentenceTransformer explicitly for this model."
```
echarlaix, line 376:
```suggestion
                "This model is a Sentence Transformers model. Please use `OVSentenceTransformer` to load this model."
```

aleksandr-mokrov: It requires editing the test `test_modeling.py::OVModelForFeatureExtractionIntegrationTest`. Fixed

echarlaix: thanks for the fix !

echarlaix (approved):

echarlaix:
failing tests unrelated, looks like an incompatible openvino-tokenizers version for openvino nightly like fixed in https://github.com/huggingface/optimum-intel/pull/885 cc @eaidova

eaidova:
> failing tests unrelated, looks like an incompatible openvino-tokenizers version for openvino nightly like fixed in #885 cc @eaidova

 yes, it is bug in openvino nightly package itself, already fixed and will be resolved in next openvino nightly

PR #867, opened 2024-08-13
Implement weight compression dataset preparation from list of strings

nikita-savelyevv:
# What does this PR do?

Provide ability to compress model with a list of strings dataset as it is claimed in the docstring.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -351,7 +351,7 @@ class OVQuantizer(OptimumQuantizer):
                     "quantization. Will rely on `calibration_dataset`."
                 )
 
-            if calibration_dataset is None and isinstance(quantization_config.dataset, str):
+            if calibration_dataset is None and quantization_config.dataset is not None:
                 from optimum.intel import OVModelForCausalLM
 
                 if isinstance(self.model, OVModelForCausalLM):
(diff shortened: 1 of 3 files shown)


optimum/intel/openvino/quantization.py
```diff
@@ -676,7 +676,12 @@ def _prepare_builtin_dataset(self, quantization_config: OVWeightQuantizationConf
             quantization_config.tokenizer, trust_remote_code=quantization_config.trust_remote_code
         )
         nsamples = quantization_config.num_samples if quantization_config.num_samples else 128
-        calibration_dataset = get_dataset(quantization_config.dataset, tokenizer, seqlen=32, nsamples=nsamples)
+        if isinstance(quantization_config.dataset, str):
+            calibration_dataset = get_dataset(quantization_config.dataset, tokenizer, seqlen=32, nsamples=nsamples)
+        else:
```
AlexKoff88, line 681: @nikita-savelyevv, can you please invest more time here. First, we need to check that this is a List of strings. Second, I don't like the name of the function where this logic is located: `_prepare_builtin_dataset`

nikita-savelyevv: > First, we need to check that this is a List of strings.

Added the check

> Second, I don't like the name of the function where this logic is located: _prepare_builtin_dataset

Renamed to `_prepare_causal_lm_dataset`. What do you think?

AlexKoff88: It could be `_prepare_text_dataset` but I am also ok with your name.

AlexKoff88 (approved):

PR #868, opened 2024-08-15
enable IPEXModelForSeq2SeqLM by torch.compile

jiqing-feng:
Enable IPEXModelForSeq2SeqLM by torch.compile. Need to add tests.

It could bring google-t5/t5-base at least 10% speed-up and [bart-large-cnn](https://huggingface.co/facebook/bart-large-cnn) at least 30% speed-up in torch 2.5.0.dev20240731+cpu


jiqing-feng:
Hi @echarlaix . Would you please help check if the integration is acceptable? If you think the integration is okay, I will add some tests for it.

optimum/intel/ipex/modeling_base.py
```diff
@@ -113,21 +114,35 @@ def ipex_jit_trace(model, task, use_cache):
...
+    if task == "text2text-generation":
+        if is_torch_version("<", "2.5.0"):
+            warnings.warn(
+                "The Seq2Seq model will keep eager mode, please upgrade your pytorch version for better performance with graph mode."
+            )
+        else:
+            torch._inductor.cpp_wrapper = True
+            model.forward = torch.compile(model.forward, dynamic=True)
+            # compile
+            warnings.warn("Run a few epochs for torch.compile by your use case.")
```
echarlaix, line 127: So for seq2seq models we would just use torch.compile without applying any IPEX optimizations afterwards ? Also not sure we want to support both TorchScript and torch.compile, would be better to start deprecating TorchScript and use torch.compile for all models

yao-matrix: Yes, Ella, we are starting to deprecating TorchScript and use torch.compile. We are trending to go the "torch.compile for graph compilation + ipex for peak perf custom op" way to make optimum-intel get peak performance for transformers models on Intel platforms(cpu, xpu). And we'll start from see2seq model and enable more tasks along the way per pytorch & ipex readiness.

optimum/intel/ipex/modeling_base.py
```diff
@@ -113,21 +114,35 @@ def ipex_jit_trace(model, task, use_cache):
...
-        )
-        trace_model = torch.jit.freeze(trace_model)
-        trace_model(**sample_inputs)
-        trace_model(**sample_inputs)
 
-    return trace_model
+    if task == "text2text-generation":
+        if is_torch_version("<", "2.5.0"):
+            warnings.warn(
+                "The Seq2Seq model will keep eager mode, please upgrade your pytorch version for better performance with graph mode."
```
IlyasMoutawwakil, line 121: maybe include the minimal version for pytorch in the message

optimum/intel/ipex/modeling_base.py
echarlaix: Do you know when we will be able to use torch.compile for all tasks ? I'd rather wait for us to be able to use torch.compile for all tasks and not have different behavior for different tasks. Also currently there is no IPEX optimization applied

jiqing-feng: Hi Ella, I have tested torch.compile and found it still has many issues. But we cannot block every things by it. I know enabling different tasks with 2 different optimization ways (torch.compile and jit) is not friendly to users because they need to keep 2 knowledge, but most users do not care what kind of optimization you take as long as the performance is great to them. 

In addition, we cannot give up the text2text-generation task only because of the little inconvenience.

PR #872, opened 2024-08-20
Add INT4 config matching based on model folder name and model short id

nikita-savelyevv:
**Changes**

Consider a case when HF model is downloaded via model repo cloning, e.g.
```
git clone https://huggingface.co/openlm-research/open_llama_3b
```
Sometimes `config.json` for a model won't contain a valid `_name_or_path` field. In such case we won't be able to match a default int4 config for it.

This PR adds an additional option to match the config: by the model folder name. 

Because of the fact that such processing might result in duplicates, a test is added to preventively catch this.

**Related tickets**
149069

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -189,7 +189,7 @@ def _check_default_4bit_configs(model_id_or_path: str):
             return _DEFAULT_4BIT_CONFIGS[original_model_name]
 
     for model_id, config in _DEFAULT_4BIT_CONFIGS.items():
-        short_id = model_id.split("/")[1]
+        short_id = model_id.split("/")[-1]
         if model_path.name == short_id:
             return config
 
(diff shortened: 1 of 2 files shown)


nikita-savelyevv:
cc @AlexKoff88 @eaidova @echarlaix @skuros

optimum/intel/openvino/configuration.py
```diff
@@ -179,14 +179,20 @@ def _check_default_4bit_configs(model_id_or_path: str):
...
+    config_path = model_path / "config.json"
     if config_path.exists():
         with config_path.open("r") as config_f:
             config = json.load(config_f)
             original_model_name = config.get("_name_or_path", "")
         if original_model_name in _DEFAULT_4BIT_CONFIGS:
             return _DEFAULT_4BIT_CONFIGS[original_model_name]
 
+    for model_id, config in _DEFAULT_4BIT_CONFIGS.items():
+        short_id = model_id.split("/")[1]
```
eaidova, line 192: I think better to use -1. previously some old models on hub were stored without organization (e.g. gpt2), they probably still available via redirection.

eaidova (approved):

optimum/intel/openvino/configuration.py
nikita-savelyevv: Done

AlexKoff88 (approved):

IlyasMoutawwakil (approved):

eaidova:
@IlyasMoutawwakil could you please merge?

PR #873, opened 2024-08-20
ov compile only mode

eaidova:
# What does this PR do?
this PR introduces inference mode for OpenVINO models using only device execution graph representation (compiled_model).
OpenVINO uses 2 models representation:
* Graph representation (Intermediate Representation) - device-agnostic, used for applying optimizations using openvino transformations and nncf, provides access to internal operations and modifications (changing input\output precision and reshaping), nonexecutable (does not have inference capabilities  need to run compile_model for moving to Compiled model format)
* Exeсtable representation (Compiled Model) - device-specific representation used for inference (obtained as result of IR compilation on specific device).

Both models sharing weights, but may have some difference (e.g. when device apply some additional own optimizations and does not preserve mapping between IR weights and compiled model, some weights and shapes can be also calculated on the first inference step - e.g. decompression of int4 constants) 

Besides that openvino support caching compiled model for speedup next loading on the same device. In this case we do not need to use Graph model representation as compiled model will be restored from cache. Experiments shows that it may reduce peak memory consumption:

default scenario for phi3-mini with caching
image-2024-07-12-01-35-42-179

 phi3-mini with caching and compile_only=True
image-2024-07-12-01-34-10-049

The idea of this PR to allow users get benifit of usage cached compiled model moving in special mode (I named it `compile_only`, as we do not permit other possible actions that possible to do with model (quantization, device changing, reshaping) except compilation) with acceptance possible limitations. Motivation of these changes that default scenario is very flexible and more suitable for experiments, while in deployment scenario (when all experiments already done before) memory constrains can be more strict.

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/stateful.py
+++ b/optimum/exporters/openvino/stateful.py
@@ -24,9 +24,11 @@ from optimum.exporters import TasksManager
 from optimum.intel.utils.import_utils import _openvino_version, is_openvino_version, is_transformers_version
 
 
-def model_has_state(ov_model: ov.Model):
-    # TODO: Provide a better way based on the variables availability, but OV Python API doesn't expose required methods
-    return len(ov_model.get_sinks()) > 0
+def model_has_state(ov_model: ov.Model, is_compiled=False):
+    if not is_compiled:
+        # TODO: Provide a better way based on the variables availability, but OV Python API doesn't expose required methods
+        return len(ov_model.get_sinks()) > 0
+    return len(ov_model.query_state()) > 0
 
 
 def model_has_input_output_name(ov_model: ov.Model, name: str):
--- a/optimum/exporters/openvino/stateful.py
+++ b/optimum/exporters/openvino/stateful.py
@@ -24,11 +24,11 @@ from optimum.exporters import TasksManager
 from optimum.intel.utils.import_utils import _openvino_version, is_openvino_version, is_transformers_version
 
 
-def model_has_state(ov_model: ov.Model, is_compiled=False):
-    if not is_compiled:
-        # TODO: Provide a better way based on the variables availability, but OV Python API doesn't expose required methods
-        return len(ov_model.get_sinks()) > 0
-    return len(ov_model.query_state()) > 0
+def model_has_state(ov_model: ov.Model):
+    if isinstance(ov_model, ov.runtime.CompiledModel):
+        return len(ov_model.query_state()) > 0
+    # TODO: Provide a better way based on the variables availability, but OV Python API doesn't expose required methods
+    return len(ov_model.get_sinks()) > 0
 
 
 def model_has_input_output_name(ov_model: ov.Model, name: str):
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -134,9 +134,8 @@ class OVModel(OVBaseModel):
         Use the specified `device` for inference. For example: "cpu" or "gpu". `device` can
         be in upper or lower case. To speed up first inference, call `.compile()` after `.to()`.
         """
-        if self.compile_only:
-            logger.warning("`to()` does not supported in `compile_only` mode")
-            return self
+        if self.compile_only and isinstance(device, str):
+            raise ValueError("`to()` is not supported in `compile_only` mode, please intialize model without this option")
 
         if isinstance(device, str):
             self._device = device.upper()
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -440,8 +440,10 @@ class OVModelForFeatureExtraction(OVModel):
         **kwargs,
     ):
         if compile_only:
-            logger.warning("`compile_only` mode will be disabled because it does not support model export."
-                            "Please provide openvino model obtained using optimum-cli or saved on disk using `save_pretrained`")
+            logger.warning(
+                "`compile_only` mode will be disabled because it does not support model export."
+                "Please provide openvino model obtained using optimum-cli or saved on disk using `save_pretrained`"
+            )
             compile_only = False
 
         save_dir = TemporaryDirectory()
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -134,7 +134,7 @@ class OVModel(OVBaseModel):
         Use the specified `device` for inference. For example: "cpu" or "gpu". `device` can
         be in upper or lower case. To speed up first inference, call `.compile()` after `.to()`.
         """
-        if self.compile_only and isinstance(device, str):
+        if self._compile_only and isinstance(device, str):
             raise ValueError(
                 "`to()` is not supported with `compile_only` mode, please intialize model without this option"
             )
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -135,7 +135,9 @@ class OVModel(OVBaseModel):
         be in upper or lower case. To speed up first inference, call `.compile()` after `.to()`.
         """
         if self.compile_only and isinstance(device, str):
-            raise ValueError("`to()` is not supported in `compile_only` mode, please intialize model without this option")
+            raise ValueError(
+                "`to()` is not supported with `compile_only` mode, please intialize model without this option"
+            )
 
         if isinstance(device, str):
             self._device = device.upper()
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -134,6 +134,10 @@ class OVModel(OVBaseModel):
         Use the specified `device` for inference. For example: "cpu" or "gpu". `device` can
         be in upper or lower case. To speed up first inference, call `.compile()` after `.to()`.
         """
+        if self.compile_only:
+            logger.warning("`to()` does not supported in `compile_only` mode")
+            return self
+
         if isinstance(device, str):
             self._device = device.upper()
             self.request = None
@@ -435,6 +439,11 @@ class OVModelForFeatureExtraction(OVModel):
         quantization_config: Union[OVWeightQuantizationConfig, Dict] = None,
         **kwargs,
     ):
+        if compile_only:
+            logger.warning("`compile_only` mode will be disabled because it does not support model export."
+                            "Please provide openvino model obtained using optimum-cli or saved on disk using `save_pretrained`")
+            compile_only = False
+
         save_dir = TemporaryDirectory()
         save_dir_path = Path(save_dir.name)
         # This attribute is needed to keep one reference on the temporary directory, since garbage collecting
@@ -469,6 +478,7 @@ class OVModelForFeatureExtraction(OVModel):
             config=config,
             load_in_8bit=load_in_8bit,
             quantization_config=quantization_config,
+            compile_only=compile_only,
             **kwargs,
         )
 
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -440,6 +440,7 @@ class OVModelForFeatureExtraction(OVModel):
         quantization_config: Union[OVWeightQuantizationConfig, Dict] = None,
         **kwargs,
     ):
+        compile_only = kwargs.get("compile_only", False)
         if compile_only:
             logger.warning(
                 "`compile_only` mode will be disabled because it does not support model export."
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -23,7 +23,7 @@ import openvino
 import torch
 from huggingface_hub import hf_hub_download
 from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
-from openvino import Core, convert_model
+from openvino import Core, convert_model, CompiledModel
 from openvino._offline_transformations import apply_moc_transformations, compress_model_transformation
 from transformers import GenerationConfig, PretrainedConfig
 from transformers.file_utils import add_start_docstrings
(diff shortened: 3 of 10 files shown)


optimum/intel/openvino/modeling_base.py
```diff
@@ -186,6 +193,42 @@ def fix_op_names_duplicates(model: openvino.runtime.Model):
...
+            ov_config["CACHE_DIR"] = str(cache_dir)
+            logger.info(f"Setting OpenVINO CACHE_DIR to {str(cache_dir)}")
+        if file_name.suffix == ".onnx":
+            compiled_model = core.compile_model(
+                file_name, device.upper() if device is not None else device, config=ov_config
+            )
+        else:
+            compiled_model = core.compile_model(
+                file_name, device.upper() if device is not None else device, config=ov_config
+            )
```
IlyasMoutawwakil, line 226: why the condition if it's the same behavior ?

optimum/exporters/openvino/stateful.py
```diff
@@ -24,9 +24,11 @@
...
 
 
-def model_has_state(ov_model: ov.Model):
-    # TODO: Provide a better way based on the variables availability, but OV Python API doesn't expose required methods
-    return len(ov_model.get_sinks()) > 0
+def model_has_state(ov_model: ov.Model, is_compiled=False):
+    if not is_compiled:
+        # TODO: Provide a better way based on the variables availability, but OV Python API doesn't expose required methods
+        return len(ov_model.get_sinks()) > 0
+    return len(ov_model.query_state()) > 0
```
IlyasMoutawwakil, line 31: can't we check if it's compiled with
```python
isinstance(model, ov.CompiledModel)
```

optimum/intel/openvino/modeling_base.py
eaidova: removed

optimum/exporters/openvino/stateful.py
eaidova: used

eaidova:
> Can you add a description of what this PR does. The `compile_only` flag is confusing, what's the exact use case and how does it solve it ?

@IlyasMoutawwakil  I belive this description should be enough?

optimum/intel/openvino/modeling.py
```diff
@@ -135,6 +135,10 @@ def to(self, device: str):
         Use the specified `device` for inference. For example: "cpu" or "gpu". `device` can
         be in upper or lower case. To speed up first inference, call `.compile()` after `.to()`.
         """
+        if self.compile_only:
+            logger.warning("`to()` does not supported in `compile_only` mode")
```
IlyasMoutawwakil, line 139:
```suggestion
            logger.warning("`to()` is not supported with `compile_only` mode")
```

optimum/intel/openvino/modeling_base.py
```diff
@@ -70,9 +70,16 @@ def __init__(
...
         enable_compilation = kwargs.get("compile", True)
 
-        if self.is_dynamic:
+        if self.compile_only and not enable_compilation:
+            raise ValueError(
+                "`compile_only` mode does not support disabling compilation."
+                "Please provide `compile=True` if you want to use `compile_only=True` or set `compile_only=False`"
+            )
+
+        if self.is_dynamic and not self.compile_only:
```
IlyasMoutawwakil, line 82: if I understand correctly, `compile only` also doesn't allow re-compilation, so the argument `compile=` doesn't make much sense when used with `compile_only`, whether it's true (re-compilation) or false (disable compilation) ?

optimum/intel/openvino/modeling_base.py
```diff
@@ -195,6 +234,12 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
             save_directory (`str` or `Path`):
                 The directory where to save the model files.
         """
+
+        if self.compile_only:
+            logger.warning(
+                "`compile_only` does not support model saving on disk, you already should have preconverted model"
+            )
+            return
```
IlyasMoutawwakil, line 242: not a fan of logging a warning and returning nothing when the user explicitly calls `save_pretrained`

optimum/intel/openvino/modeling_base.py
```diff
@@ -434,6 +491,14 @@ def _from_transformers(
         # would end-up removing the directory containing the underlying OpenVINO model
         cls._model_save_dir_tempdirectory_instance = save_dir
 
+        compile_only = kwargs.pop("compile_only", False)
+        if compile_only:
+            logger.warning(
+                "`compile_only` mode will be disabled because it does not support model export."
+                "Please provide openvino model obtained using optimum-cli or saved on disk using `save_pretrained`"
+            )
+            compile_only = False
```
IlyasMoutawwakil, line 500: same thing, dismissing an explicit argument with a warning doesn't make sense to me

optimum/intel/openvino/modeling_base.py
eaidova: what will be prefered by you, raise error? There is no entry that can be loaded back after call save_pretrained, that is why I do not provide opportunity to save for avoiding next loading errors. Besides that this mode assumes that user already have saved on disk model before, so it just will create copy of already saved model

optimum/intel/openvino/modeling_base.py
eaidova: yes, recompilation is not supported, but we need to keep compile flag for regular case.

optimum/intel/openvino/modeling_base_seq2seq.py
```diff
@@ -188,16 +189,41 @@ def _from_pretrained(
...
+                    model_save_dir,
+                )
+                if use_cache:
+                    decoder_with_past = cls._compile_model(
+                        os.path.join(model_id, decoder_with_past_file_name),
+                        kwargs.get("device", "CPU"),
+                        kwargs.get("ov_config"),
+                        True,
+                        model_save_dir,
+                    )
```
echarlaix, line 226: Here what about skipping loading and update the request in the `.__init__()` instead ?

optimum/intel/openvino/modeling_base.py
```diff
@@ -287,9 +332,21 @@ def _from_pretrained(
             local_files_only=local_files_only,
         )
 
+        compile_only = kwargs.get("compile_only", False)
+
         quantization_config = cls._prepare_weight_quantization_config(quantization_config, load_in_8bit)
 
-        model = cls.load_model(model_cache_path, quantization_config=quantization_config)
+        model = None
+        if not compile_only:
```
echarlaix, line 340: From my understanding the only thing that `compile_only` does is to skip the `read_model` part is this correct ? To me the argument name `compile_only` is not super clear, but not sure I have a better alternative

optimum/intel/openvino/modeling_base_seq2seq.py
eaidova: I think it will be messy... initially request was used for infer request object and some models (e.g.decoder in seq2seq and cualalm models because request gives opportunity to get raw openvino tensors without coping, it gives better memory management for nonCPU devices), others appears on compiled model ability to create infer request inside due to this misalignment  - some models will have model filed others not.

Besides that if we will keep only infer request for decoders and still plan introduce concurrency mode, we need to keep object that produces infer requests (compiled model) to be able create additional infer request even in compile only mode

optimum/intel/openvino/modeling_base.py
eaidova: maybe inference_only? it is the only functional that works in this mode :)

optimum/intel/openvino/modeling_base_seq2seq.py
eaidova: if you mean that this should be moved to `__init__`, then I think it will requires to handle all paths as arguments of `__init__`

optimum/intel/openvino/modeling_base.py
eaidova: for all methods or only for saving?

optimum/intel/openvino/modeling_base.py
```diff
@@ -70,9 +70,16 @@ def __init__(
         self.is_dynamic = dynamic_shapes
         self.ov_config = {} if ov_config is None else {**ov_config}
         self.preprocessors = kwargs.get("preprocessors", [])
+        self.compile_only = kwargs.get("compile_only", False)
```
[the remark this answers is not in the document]
eaidova: not fully understand your comment if it is anyway provided as argument why we should perform any actions to recognize it?

optimum/intel/openvino/modeling.py
```diff
@@ -439,6 +443,14 @@ def _from_transformers(
                 raise ValueError("You cannot use both `use_auth_token` and `token` arguments at the same time.")
             token = use_auth_token
 
+        compile_only = kwargs.pop("compile_only", False)
+        if compile_only:
+            logger.warning(
+                "`compile_only` mode will be disabled because it does not support model export."
+                "Please provide openvino model obtained using optimum-cli or saved on disk using `save_pretrained`"
+            )
```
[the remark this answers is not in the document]
eaidova: please check my comment above, in case if export flag ommited in future, user may not know there is already converted model or not, to not make changes in behaviour after your changes, I stayed warning here

optimum/intel/openvino/modeling_base.py
echarlaix: Yes I see your point even though might lead users to think the model can be used other than for inference (like training) not sure.  Was thinking about something to highlight that the model cannot be modified, but for now I think we can keep `compile_only` that'll work

optimum/intel/openvino/modeling_base.py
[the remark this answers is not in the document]
echarlaix: Why adding an extra argument if not needed (`compile_only` can be inferred from the model) ? Also will avoid any potential issue resulting from incompatible arguments like `OVModelForXxx(model=compiled_model, compile_only=False)`

optimum/intel/openvino/modeling_base.py
echarlaix: Would make sense for all methods to avoid any confusion

AlexKoff88 (approved):

optimum/intel/openvino/modeling_base.py
eaidova: done

optimum/intel/openvino/modeling_base.py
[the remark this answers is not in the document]
eaidova: done

optimum/intel/openvino/modeling_base.py
```diff
@@ -73,9 +73,19 @@ def __init__(
...
+        if self.compile_only and not enable_compilation:
+            raise ValueError(
+                "`compile_only` mode does not support disabling compilation."
+                "Please provide `compile=True` if you want to use `compile_only=True` or set `compile_only=False`"
+            )
+
+        if self.compile_only and not isinstance(model, CompiledModel):
+            raise ValueError("`compile_only` expect that already compiled model will be provided")
+
+        if self.is_dynamic and not self.compile_only:
```
echarlaix, line 88: Shouldn't we set `self.is_dynamic` to False if `self._compile_only=True` ?

something like 
```suggestion
        if self.is_dynamic:
```

and 

```
self.is_dynamic = dynamic_shapes and not self.compile_only
```

optimum/intel/openvino/modeling_base.py
```diff
@@ -198,6 +238,11 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
             save_directory (`str` or `Path`):
                 The directory where to save the model files.
         """
+
+        if self.compile_only:
+            raise ValueError(
+                "`save_pretrained()` is not supported with `compile_only` mode, please intialize model without this option"
```
echarlaix, line 244:
```suggestion
                "`save_pretrained()` is not supported with `compile_only=True` mode, to save your model please initialize your model with compile_only=False"
```

optimum/intel/openvino/modeling_base.py
```diff
@@ -614,6 +679,10 @@ def half(self):
         """
         Converts all the model weights to FP16
         """
+        if self.compile_only:
+            raise ValueError(
+                "`reshape()` is not supported with `compile_only` mode, please intialize model without this option"
```
echarlaix, line 684:
```suggestion
                "`half()` is not supported with `compile_only=True` mode, to use this option please initialize your model with compile_only=False"
```

optimum/intel/openvino/modeling_base_seq2seq.py
```diff
@@ -210,11 +233,25 @@ def _from_pretrained(
...
-                decoder_with_past = cls.load_model(file_names["decoder_with_past"], quantization_config)
-
+            if not compile_only:
+                encoder = cls.load_model(file_names["encoder"], quantization_config)
+                decoder = cls.load_model(file_names["decoder"], quantization_config)
+                if use_cache:
+                    decoder_with_past = cls.load_model(file_names["decoder_with_past"], quantization_config)
+            else:
+                encoder = cls._compile_model(
+                    file_names["encoder"], kwargs.get("device", "CPU"), kwargs.get("ov_config"), True, model_save_dir
```
echarlaix, line 243:
```suggestion
                    file_names["encoder"], kwargs.get("device", "CPU"), kwargs.get("ov_config"), model_save_dir
```

optimum/intel/openvino/modeling_base_seq2seq.py
```diff
@@ -210,11 +233,25 @@ def _from_pretrained(
...
+                )
+                decoder = cls._compile_model(
+                    file_names["decoder"], kwargs.get("device", "CPU"), kwargs.get("ov_config"), True, model_save_dir
```
echarlaix, line 246:
```suggestion
                    file_names["decoder"], kwargs.get("device", "CPU"), kwargs.get("ov_config"), model_save_dir
```

eaidova:
@echarlaix could you please take a look one more time, thanks!

echarlaix (approved):

optimum/intel/openvino/modeling.py
```diff
@@ -134,6 +134,11 @@ def to(self, device: str):
         Use the specified `device` for inference. For example: "cpu" or "gpu". `device` can
         be in upper or lower case. To speed up first inference, call `.compile()` after `.to()`.
         """
+        if self.compile_only and isinstance(device, str):
```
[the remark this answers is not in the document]
eaidova: renamed

PR #878, opened 2024-08-22
Apply weight compression after model save to reduce peak RAM during export

nikita-savelyevv:
# What does this PR do?

Currently during OV model export with compression, compression is applied right after the model is converted. Because of this, we can't deallocated the memory required for the conversion before the compression start. 

This PR introduces an intermediate full precision model saving step. Model compression is applied after this step allowing to save some memory. For INT8 compression the improvement is estimated to be at least 20%. Please see the figures below for "before" and "after" for INT8 compression of llama-2-7b-hf model.

Memory figures were obtained with https://github.com/openvinotoolkit/nncf/blob/develop/tools/memory_monitor.py

**Memory improvement is only for data-free compression via optimum-cli**. For export via Python API, i.e. `from_pretrained(..., export=True)` the memory is already handled efficiently with intermediate saving. The same for data-aware compression via optimum-cli.

| Before         | After   |
|-------------------|-------|
| system_memory_usage_from-zero | system_memory_usage_from-zero |

| Model                 | Peak Mem. Before (MiB) | Peak Mem. After (MiB) | Time Before (sec.) | Time After (sec.) |
|-----------------------|-------------------------|------------------------|--------------------|-------------------|
| opt-350m, INT8        | 1789                    | 1461 (-19%)                  | 20                 | 20 (+0%)               |
| opt-350m, INT4        | 1734                    | 1466 (-15%)                | 28                 | 29 (+3%)               |
| tiny-llama-1.1b, INT8 | 4370                    | 3888 (-11%)                  | 33                 | 32 (-3%)               |
| tiny-llama-1.1b, INT4 | 4005                    | 3798 (-5%)                  | 63                 | 62 (-2%)               |
| open-llama-3b, INT8   | 11203                   | 7822 (-30%)                  | 48                 | 55 (+15%)               |
| open-llama-3b, INT4   | 9750                    | 7822 (-20%)                  | 150                | 158 (+5%)              |
| llama2-7b-hf, INT8    | 21108                   | 14551 (-31%)                 | 80                 | 94 (+18%)               |
| llama2-7b-hf, INT4    | 18039                   | 14531 (-19%)                 | 276                | 292 (+6%)              |

For some unknown reason the results for tiny-llama-1.1b are unexpected. There is no yet an explanation for that. For other models there is noticeable memory saving with the cost of longer conversion due to intermediate model saving.

**Related tickets**
147935

## Before submitting


nikita-savelyevv:
@eaidova @AlexKoff88 please take a look

eaidova (approved):

optimum/exporters/openvino/__main__.py
```diff
@@ -425,6 +429,50 @@ class StoreAttr(object):
...
+                    quantization_config = {"bits": 8, "sym": False}
+                    logger.info("The model weights will be quantized to int8_asym.")
+                else:
+                    logger.warning(
+                        "The model will be converted with no weights quantization. Quantization of the weights to int8 "
+                        "requires nncf. Please install it with `pip install nncf`"
+                    )
+                    break
+        if quantization_config is None:
+            continue
```
IlyasMoutawwakil, line 455:
```suggestion
        
        if ov_config is None:
            num_parameters = 0
            for op in submodel.get_ops():
                if op.get_type_name() == "Constant" and op.get_element_type() in [Type.f16, Type.f32, Type.bf16]:
                    num_parameters += reduce(operator.mul, op.shape, 1)
            if num_parameters >= _MAX_UNCOMPRESSED_SIZE:
                if is_nncf_available():
                    quantization_config = {"bits": 8, "sym": False}
                    logger.info("The model weights will be quantized to int8_asym.")
                else:
                    logger.warning(
                        "The model will be converted with no weights quantization. Quantization of the weights to int8 "
                        "requires nncf. Please install it with `pip install nncf`"
                    )
                    break
        else:
            quantization_config = ov_config.quantization_config
```

optimum/exporters/openvino/__main__.py
```diff
@@ -425,6 +429,52 @@ class StoreAttr(object):
...
+
+        if not is_nncf_available():
+            raise ImportError("Quantization of the weights requires nncf, please install it with `pip install nncf`")
+
+        from optimum.intel.openvino.quantization import _weight_only_quantization
+
+        _weight_only_quantization(submodel, quantization_config)
+
+        compressed_submodel_path = submodel_path.parent / f"{submodel_path.stem}_compressed.xml"
+        save_model(submodel, compressed_submodel_path, compress_to_fp16=ov_config and ov_config.dtype == "fp16")
```
echarlaix, line 467: Here the model weights will be quantized to 8/4-bit so `ov_config.dtype` will never be equal to "fp16" in this case no ?

echarlaix (approved):

optimum/exporters/openvino/__main__.py
nikita-savelyevv: Good catch. Yes.

IlyasMoutawwakil (approved):

eaidova:
@IlyasMoutawwakil  @echarlaix  could you please merge?

PR #882, opened 2024-08-28
Introduce support for `mxfp4` data type for OV weight compression

nikita-savelyevv:
# What does this PR do?

- Added a new possible value for `--weight-format` CLI argument: `mxfp4`.
- Added `weight_format` parameter to `OVWeightQuantizationConfig` with three possible values: `int4`, `int8` and `mxfp4`. The latter is used for compression with mxfp4 `E2M1` mode.
- Added a couple of tests for compression with the new data type. Had to add quite a bit of logic to extend test infrastructure to work with data types other than int4/int8.
- Deprecate `"int4_sym_g128", "int4_asym_g128", "int4_sym_g64", "int4_asym_g64"` values for `--weight-format` CLI argument, as planned before.

### Example Usage

Optimum CLI:
```
optimum-cli export openvino -m facebook/opt-125m --task text-generation-with-past --weight-format mxfp4 ./tmp
```

Optimum Python API:
```
    model_id="facebook/opt-125m",
    export=True,
    quantization_config=OVWeightQuantizationConfig(bits=4, weight_format="mxfp4"),
)
```

## Before submitting

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -70,9 +70,9 @@ def parse_args_openvino(parser: "ArgumentParser"):
     optional_group.add_argument(
         "--weight-format",
         type=str,
-        choices=["fp32", "fp16", "int8", "int4", "int4_sym_g128", "int4_asym_g128", "int4_sym_g64", "int4_asym_g64"],
+        choices=["fp32", "fp16", "int8", "int4", "mxfp4_e2m1", "int4_sym_g128", "int4_asym_g128", "int4_sym_g64", "int4_asym_g64"],
         default=None,
-        help="he weight format of the exported model.",
+        help="The weight format of the exported model.",
     )
     optional_group.add_argument(
         "--library",
@@ -255,12 +255,11 @@ class OVExportCommand(BaseOptimumCLICommand):
         elif self.args.weight_format in {"fp16", "fp32"}:
             ov_config = OVConfig(dtype=self.args.weight_format)
         else:
-            is_int8 = self.args.weight_format == "int8"
-
-            # For int4 quantization if no parameter is provided, then use the default config if exist
-            if no_compression_parameter_provided(self.args) and not is_int8:
+            # For int4 quantization if no parameter is provided, then use the default config if exists
+            if no_compression_parameter_provided(self.args) and self.args.weight_format == "int4":
                 quantization_config = get_default_int4_config(self.args.model)
             else:
+                is_int8 = self.args.weight_format == "int8"
                 quantization_config = {
                     "bits": 8 if is_int8 else 4,
                     "ratio": 1 if is_int8 else (self.args.ratio or _DEFAULT_4BIT_CONFIG["ratio"]),
@@ -272,6 +271,7 @@ class OVExportCommand(BaseOptimumCLICommand):
                     "quant_method": "awq" if self.args.awq else "default",
                     "sensitivity_metric": self.args.sensitivity_metric,
                     "scale_estimation": self.args.scale_estimation,
+                    "dtype": "int" if "int" in self.args.weight_format else self.args.weight_format,
                 }
 
             if quantization_config.get("dataset", None) is not None:
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -70,7 +70,7 @@ def parse_args_openvino(parser: "ArgumentParser"):
     optional_group.add_argument(
         "--weight-format",
         type=str,
-        choices=["fp32", "fp16", "int8", "int4", "mxfp4_e2m1"],
+        choices=["fp32", "fp16", "int8", "int4", "mxfp4"],
         default=None,
         help="The weight format of the exported model.",
     )
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -70,7 +70,17 @@ def parse_args_openvino(parser: "ArgumentParser"):
     optional_group.add_argument(
         "--weight-format",
         type=str,
-        choices=["fp32", "fp16", "int8", "int4", "mxfp4_e2m1", "int4_sym_g128", "int4_asym_g128", "int4_sym_g64", "int4_asym_g64"],
+        choices=[
+            "fp32",
+            "fp16",
+            "int8",
+            "int4",
+            "mxfp4_e2m1",
+            "int4_sym_g128",
+            "int4_asym_g128",
+            "int4_sym_g64",
+            "int4_asym_g64",
+        ],
         default=None,
         help="The weight format of the exported model.",
     )
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -70,17 +70,7 @@ def parse_args_openvino(parser: "ArgumentParser"):
     optional_group.add_argument(
         "--weight-format",
         type=str,
-        choices=[
-            "fp32",
-            "fp16",
-            "int8",
-            "int4",
-            "mxfp4_e2m1",
-            "int4_sym_g128",
-            "int4_asym_g128",
-            "int4_sym_g64",
-            "int4_asym_g64",
-        ],
+        choices=["fp32", "fp16", "int8", "int4", "mxfp4_e2m1"],
         default=None,
         help="The weight format of the exported model.",
     )
@@ -281,18 +271,11 @@ class OVExportCommand(BaseOptimumCLICommand):
                     "quant_method": "awq" if self.args.awq else "default",
                     "sensitivity_metric": self.args.sensitivity_metric,
                     "scale_estimation": self.args.scale_estimation,
-                    "dtype": "int" if "int" in self.args.weight_format else self.args.weight_format,
+                    "weight_format": self.args.weight_format,
                 }
 
             if quantization_config.get("dataset", None) is not None:
                 quantization_config["trust_remote_code"] = self.args.trust_remote_code
-
-            if self.args.weight_format in {"int4_sym_g128", "int4_asym_g128", "int4_sym_g64", "int4_asym_g64"}:
-                logger.warning(
-                    f"--weight-format {self.args.weight_format} is deprecated, possible choices are fp32, fp16, int8, int4"
-                )
-                quantization_config["sym"] = "asym" not in self.args.weight_format
-                quantization_config["group_size"] = 128 if "128" in self.args.weight_format else 64
             ov_config = OVConfig(quantization_config=quantization_config)
 
         quantization_config = ov_config.quantization_config if ov_config else None
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -390,8 +390,6 @@ class OVWeightQuantizationConfig(OVQuantizationConfigBase):
                 raise ValueError("`all_layers` parameters is not supported for 8-bit quantization")
             if self.sensitivity_metric:
                 raise ValueError("`sensitivity_metric` parameters is not supported for 8-bit quantization")
-            if self.dataset:
-                raise ValueError("`dataset` parameters is not supported for 8-bit quantization")
             if self.quant_method == OVQuantizationMethod.AWQ:
                 raise ValueError("AWQ algorithm is not support for 8-bit quantization")
             if self.scale_estimation:
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -386,6 +386,16 @@ class OVWeightQuantizationConfig(OVQuantizationConfigBase):
                 raise ValueError(
                     f"For 8-bit quantization, `group_size` is expected to be set to -1, but was set to {self.group_size}"
                 )
+            if self.all_layers:
+                raise ValueError("`all_layers` parameters is not supported for 8-bit quantization")
+            if self.sensitivity_metric:
+                raise ValueError("`sensitivity_metric` parameters is not supported for 8-bit quantization")
+            if self.dataset:
+                raise ValueError("`dataset` parameters is not supported for 8-bit quantization")
+            if self.quant_method == OVQuantizationMethod.AWQ:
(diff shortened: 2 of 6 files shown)


nikita-savelyevv:
@AlexKoff88 @MaximProshin please take a look

optimum/commands/export/openvino.py
```diff
@@ -272,6 +281,7 @@ def run(self):
                     "quant_method": "awq" if self.args.awq else "default",
                     "sensitivity_metric": self.args.sensitivity_metric,
                     "scale_estimation": self.args.scale_estimation,
+                    "dtype": "int" if "int" in self.args.weight_format else self.args.weight_format,
```
IlyasMoutawwakil, line 284: i think this naming is confusing here as dtype usually takes values like float16, int8. Also "`dtype`" only makes sense when both weights and activations have the same format. Storing `weight_format` and using it for dispatching makes more sense for me. WDYT ?

nikita-savelyevv: @IlyasMoutawwakil Thanks, I agree your points. I also had an idea to name it `weight_format`, but decided against it.

If we name it `weight_format`, we can either store values in the form of `int8` (similarly to CLI), or in the format as in the PR right now, i.e. `int`. The problem I see with the first approach is that there will be a redundancy as we also have a `bits` field. With the second approach there will be a misaligned between CLI and OV*QuantizationConfig as the same variable name corresponds to values in different format.

That's why I decided to introduce a separate field. Initially if was `data_type`, but I renamed it to `dtype`. Maybe this is not the best option.

I guess there is some drawback to any approach. Which case do you think is less critical?

nikita-savelyevv: Ideally, I would remove `bits` and `sym` in favor of `weight_format`, because it is more flexible. And also `sym` is not relevant to `mxfp4_e2m1` type. For activation quantization config we can also add `activation_format` field.

But this will require quite a refactoring. Also, support for `mxfp4_e2m1` is kind of in an experimental stage right now and I we don't have to fully commit to it yet.

IlyasMoutawwakil: I think the redundancy is alright in this case (without removing `bits` and `sym`).

optimum/intel/openvino/configuration.py
```diff
@@ -386,6 +390,13 @@ def post_init(self):
         if self.tokenizer is not None and not isinstance(self.tokenizer, str):
             raise ValueError(f"Tokenizer is expected to be a string, but found {self.tokenizer}")
 
+        if self.weight_format is None:
+            self.weight_format = "int4" if self.bits == 4 else "int8"
+        if self.weight_format not in ["int4", "int8", "mxfp4_e2m1"]:
+            raise ValueError(
+                f"Data type must be on of the following: ['int4', 'int8', 'mxfp4_e2m1'], but found: {self.weight_format}."
```
IlyasMoutawwakil, line 397:
```suggestion
                f"`weight_format` must be one of the following: ['int4', 'int8', 'mxfp4_e2m1'], but found: {self.weight_format}."
```

optimum/commands/export/openvino.py
```diff
@@ -70,9 +70,9 @@ def parse_args_openvino(parser: "ArgumentParser"):
     optional_group.add_argument(
         "--weight-format",
         type=str,
-        choices=["fp32", "fp16", "int8", "int4", "int4_sym_g128", "int4_asym_g128", "int4_sym_g64", "int4_asym_g64"],
+        choices=["fp32", "fp16", "int8", "int4", "mxfp4_e2m1"],
```
IlyasMoutawwakil, line 73: if we're gonna do this, why not just call it fp4 instead of its full OCP MX name.

nikita-savelyevv: After some discussion we decided to name specifically `"mxfp4_e2m1"`. This should coney the meaning that the format is the same as in MX specification.

optimum/intel/openvino/configuration.py
nikita-savelyevv: Thanks, fixed

optimum/intel/openvino/configuration.py
```diff
@@ -386,6 +390,13 @@ def post_init(self):
         if self.tokenizer is not None and not isinstance(self.tokenizer, str):
             raise ValueError(f"Tokenizer is expected to be a string, but found {self.tokenizer}")
 
+        if self.weight_format is None:
+            self.weight_format = "int4" if self.bits == 4 else "int8"
+        if self.weight_format not in ["int4", "int8", "mxfp4_e2m1"]:
+            raise ValueError(
+                f"Weight format must be one of the following: ['int4', 'int8', 'mxfp4_e2m1'], but found: {self.weight_format}."
+            )
```
echarlaix, line 398: We already have part of this information with the bit argument so this might be a bit redundant, why not having an argument to make this distinction instead (int4 vs fp4) of introducing `weight_format` ?

echarlaix: also what can/cannot be applied with mxfp4, is there any incompatibility with other arguments that we should look for here ?

optimum/commands/export/openvino.py
echarlaix: > if we're gonna do this, why not just call it fp4 instead of its full OCP MX name.

agreed why not use fp4 or mxfp4 and specify any additional information in the documentation ?

echarlaix (commented):

optimum/commands/export/openvino.py
nikita-savelyevv: @AlexKoff88 @MaximProshin Could you please comment on that?

optimum/intel/openvino/configuration.py
nikita-savelyevv: > We already have part of this information with the bit argument so this might be a bit redundant

I agree about redundancy, we had a discussion about this with @IlyasMoutawwakil above https://github.com/huggingface/optimum-intel/pull/882#discussion_r1735912690 . Is it critical in your opinion? Another option is to have `weight_format` store just `int` instead of `int4` or `int8`.

> why not having an argument to make this distinction instead (int4 vs fp4) of introducing `weight_format`?

We have other potential non-int data types for compression, for example nf4. A simple flag will do for now, but this approach will not be extendable in the future.

optimum/commands/export/openvino.py
AlexKoff88: Too many concerns about naming. How about `mxfp4`? This name reflects the format to some degree but there is still an uncertainty in the mantissa/exponent sizes. I think it is ok for now as we support only one format in OpenVINO and have no plans to extend it in the near term.

optimum/intel/openvino/configuration.py
```diff
@@ -382,10 +386,34 @@ def post_init(self):
                 raise ValueError(
                     f"For 8-bit quantization, `group_size` is expected to be set to -1, but was set to {self.group_size}"
                 )
+            if self.all_layers:
+                raise ValueError("`all_layers` parameters is not supported for 8-bit quantization")
```
echarlaix, line 390:
```suggestion
                raise ValueError("The `all_layers` parameter is not supported for 8-bit quantization")
```

optimum/intel/openvino/configuration.py
```diff
@@ -382,10 +386,34 @@ def post_init(self):
...
+            if self.sensitivity_metric:
+                raise ValueError("`sensitivity_metric` parameters is not supported for 8-bit quantization")
```
echarlaix, line 392:
```suggestion
                raise ValueError("The `sensitivity_metric` parameter is not supported for 8-bit quantization")
```

optimum/intel/openvino/configuration.py
```diff
@@ -382,10 +386,34 @@ def post_init(self):
...
+            if self.quant_method == OVQuantizationMethod.AWQ:
+                raise ValueError("AWQ algorithm is not support for 8-bit quantization")
```
echarlaix, line 394:
```suggestion
                raise ValueError("The AWQ algorithm is not supported for 8-bit quantization and got `quant_method="awq"` please update accordingly")
```

optimum/intel/openvino/configuration.py
```diff
@@ -382,10 +386,34 @@ def post_init(self):
...
+            if self.scale_estimation:
+                raise ValueError("Scale Estimation algorithm is not support for 8-bit quantization.")
```
echarlaix, line 396:
```suggestion
                raise ValueError("The Scale Estimation algorithm is not supported for 8-bit quantization and got `scale_estimation=True`, please set `scale_estimation=False`.")
```

optimum/intel/openvino/configuration.py
```diff
@@ -382,10 +386,34 @@ def post_init(self):
...
+            )
+        if self.weight_format == "mxfp4":
+            if self.bits != 4:
+                raise ValueError(
+                    "When applying weight compression with 'mxfp4' weight format the `bits` parameters must be set to 4"
+                )
+            if self.quant_method == OVQuantizationMethod.AWQ:
+                raise ValueError("AWQ algorithm is not support for 'mxfp4' weight format")
+            if self.scale_estimation:
+                raise ValueError("Scale Estimation algorithm is not support for 'mxfp4' weight format")
```
echarlaix, line 415:
```suggestion
                raise ValueError("The Scale Estimation algorithm is not supported for 'mxfp4' weight format")
```

optimum/intel/openvino/configuration.py
```diff
@@ -382,10 +386,34 @@ def post_init(self):
...
+            raise ValueError(
+                f"Weight format must be one of the following: ['int4', 'int8', 'mxfp4'], but found: {self.weight_format}."
+            )
+        if self.weight_format == "mxfp4":
+            if self.bits != 4:
+                raise ValueError(
+                    "When applying weight compression with 'mxfp4' weight format the `bits` parameters must be set to 4"
+                )
+            if self.quant_method == OVQuantizationMethod.AWQ:
+                raise ValueError("AWQ algorithm is not support for 'mxfp4' weight format")
```
echarlaix, line 413:
```suggestion
                raise ValueError("The AWQ algorithm is not supported for 'mxfp4' weight format")
```

optimum/intel/openvino/configuration.py
```diff
@@ -382,10 +386,34 @@ def post_init(self):
...
+        if self.weight_format is None:
+            self.weight_format = "int4" if self.bits == 4 else "int8"
+        if self.weight_format not in ["int4", "int8", "mxfp4"]:
+            raise ValueError(
+                f"Weight format must be one of the following: ['int4', 'int8', 'mxfp4'], but found: {self.weight_format}."
+            )
+        if self.weight_format == "mxfp4":
+            if self.bits != 4:
+                raise ValueError(
+                    "When applying weight compression with 'mxfp4' weight format the `bits` parameters must be set to 4"
```
echarlaix, line 410:
```suggestion
                    f"When applying weight compression with 'mxfp4' weight format the `bits` parameters must be set to 4, but found: {self.bits}"
```

optimum/intel/openvino/configuration.py
```diff
@@ -473,8 +501,13 @@ def __init__(
         self.compression = kwargs.get(
             "compression", None
         )  # A field for backward-compatability of training-time compression parameters
-        bits = self.quantization_config.bits if self.quantization_config else None
-        self.dtype = "int" + str(bits) if isinstance(bits, int) else dtype
+        if self.quantization_config is not None:
+            if isinstance(self.quantization_config, OVWeightQuantizationConfig):
+                self.dtype = self.quantization_config.weight_format
+            else:
+                self.dtype = "int8"
```
echarlaix, line 508: should it be :
```suggestion
                self.dtype = "int" + str(bits) if isinstance(bits, int)
```

optimum/intel/openvino/configuration.py
echarlaix: > Nevertheless, I've now added corresponding checks to config post_init method. Also added similar checks for INT8 compression.

Thanks a lot for adding !



> I agree about redundancy, we had a discussion about this with @IlyasMoutawwakil above https://github.com/huggingface/optimum-intel/pull/882#discussion_r1735912690 . Is it critical in your opinion? Another option is to have weight_format store just int instead of int4 or int8.

Works for me!

echarlaix (approved):
Left couple of minor comment

optimum/intel/openvino/configuration.py
nikita-savelyevv: For activation quantization currently only 8-bits option is supported. But using `bits` explicitly is fine too.

nikita-savelyevv:
> Left couple of minor comment

Thanks! Updated.

optimum/intel/openvino/configuration.py
echarlaix: OK I missed this part was for dynamic quantization only, then my bad your first suggestion was actually the most adapted. Could make sense to introduce `w_bit` / `a_bit` values in the future to clarify things when saving the configuration, as for dynamic quantization the weights could be quantized to 4-bit with `OVDynamicQuantizationConfig(bits=4, ...)` in which case `w_bit=4` and `a_bit=8`

optimum/intel/openvino/configuration.py
```diff
@@ -473,8 +505,13 @@ def __init__(
         self.compression = kwargs.get(
             "compression", None
         )  # A field for backward-compatability of training-time compression parameters
-        bits = self.quantization_config.bits if self.quantization_config else None
-        self.dtype = "int" + str(bits) if isinstance(bits, int) else dtype
+        if self.quantization_config is not None:
+            if isinstance(self.quantization_config, OVWeightQuantizationConfig):
+                self.dtype = self.quantization_config.weight_format
+            else:
+                self.dtype = f"int{quantization_config.bits}"
```
echarlaix, line 512:
```suggestion
                self.dtype = "int8"
```

echarlaix: suggestion from https://github.com/huggingface/optimum-intel/pull/882#discussion_r1740986547 shouldn't have be applied in the first place

PR #883, opened 2024-08-29
OVModelForVisionCausalLM

eaidova:
# What does this PR do?

Enables conversion and inference for multimodality llm like llava, llava-next, falcon-vl, pixtral, internvl
Example of usage:

```
from PIL import Image
import requests

model_id = "llava-hf/llava-v1.6-mistral-7b-hf"
model = OVModelForVisualCausalLM.from_pretrained(model_id)
image_file = "http://images.cocodataset.org/val2017/000000039769.jpg"

conversation = [
    {
        "role": "user",
        "content": [
            {"type": "text", "text": "What are these?"},
            {"type": "image"},
        ],
    },
]
prompt = processor.apply_chat_template(conversation, add_generation_prompt=True)
raw_image = Image.open(requests.get(image_file, stream=True).raw)
inputs = processor(images=raw_image, text=prompt, return_tensors="pt")

output = model.generate(**inputs, max_new_tokens=20, do_sample=False)
print(processor.decode(output[0], skip_special_tokens=True))
```

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -489,7 +489,6 @@ def export_models(
         )
 
     for i, model_name in enumerate(models_and_export_configs.keys()):
-        logger.warning(model_name)
         submodel, sub_export_config = models_and_export_configs[model_name]
         output_name = output_names[i] if output_names is not None else Path(model_name + ".xml")
         output_path = output_dir / output_name
@@ -568,9 +567,8 @@ def export_from_model(
 
         logger.info(f"Automatic task detection to: {task}.")
 
-    stateful = (
-        stateful
-        and ensure_export_task_support_stateful(task)
+    stateful = stateful and (
+        ensure_export_task_support_stateful(task)
         or ensure_model_type_support_stateful(getattr(getattr(model, "config", {}), "model_type", ""))
     )
     # TODO: support onnx_config.py in the model repo
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -99,6 +99,7 @@ def export(
     ov_config: Optional["OVConfig"] = None,
     stateful: bool = True,
     patch_16bit_model: bool = False,
+    library_name: Optional[str] = None,
 ) -> Tuple[List[str], List[str]]:
     """
     Exports a Pytorch or TensorFlow model to an OpenVINO Intermediate Representation.
@@ -151,6 +152,7 @@ def export(
             model_kwargs=model_kwargs,
             stateful=stateful,
             patch_16bit_model=patch_16bit_model,
+            library_name=library_name,
         )
 
     elif is_tf_available() and issubclass(type(model), TFPreTrainedModel):
@@ -161,7 +163,7 @@ def export(
             raise RuntimeError("`tf2onnx` does not support export on CUDA device.")
         if input_shapes is not None:
             logger.info("`input_shapes` argument is not supported by the Tensorflow ONNX export and will be ignored.")
-        return export_tensorflow(model, config, opset, output, ov_config=ov_config)
+        return export_tensorflow(model, config, opset, output, ov_config=ov_config, library_name=library_name)
 
     else:
         raise RuntimeError(
@@ -175,6 +177,7 @@ def export_tensorflow(
     opset: int,
     output: Path,
     ov_config: Optional["OVConfig"] = None,
+    library_name: Optional[str] = None,
 ):
     """
     Export the TensorFlow model to OpenVINO format.
@@ -194,7 +197,7 @@ def export_tensorflow(
     input_names, output_names = export_tensorflow_onnx(model, config, opset, onnx_path)
     ov_model = convert_model(str(onnx_path))
 
-    library_name = _infer_library_from_model_or_model_class(model=model)
+    library_name = _infer_library_from_model_or_model_class(model=model, library_name=library_name)
 
     _save_model(
         ov_model,
@@ -214,6 +217,7 @@ def export_pytorch_via_onnx(
     input_shapes: Optional[Dict] = None,
     model_kwargs: Optional[Dict[str, Any]] = None,
     ov_config: Optional["OVConfig"] = None,
+    library_name: Optional[str] = None,
 ):
     """
     Exports a PyTorch model to an OpenVINO Intermediate Representation via ONNX export.
@@ -255,7 +259,7 @@ def export_pytorch_via_onnx(
     torch.onnx.export = orig_torch_onnx_export
     ov_model = convert_model(str(onnx_output))
 
-    library_name = _infer_library_from_model_or_model_class(model=model)
+    library_name = _infer_library_from_model_or_model_class(model=model, library_name=library_name)
 
     _save_model(
         ov_model,
@@ -277,6 +281,7 @@ def export_pytorch(
     ov_config: Optional["OVConfig"] = None,
     stateful: bool = False,
     patch_16bit_model: bool = False,
+    library_name: Optional[str] = None,
 ) -> Tuple[List[str], List[str]]:
     """
     Exports a PyTorch model to an OpenVINO Intermediate Representation.
@@ -409,6 +414,7 @@ def export_pytorch(
                 input_shapes,
                 model_kwargs,
                 ov_config=ov_config,
+                library_name=library_name,
             )
 
         ov_model.validate_nodes_and_infer_types()  # TODO: remove as unnecessary validation?
@@ -426,7 +432,7 @@ def export_pytorch(
         if stateful:
             patch_stateful(model.config, ov_model)
 
-        library_name = _infer_library_from_model_or_model_class(model=model)
+        library_name = _infer_library_from_model_or_model_class(model=model, library_name=library_name)
 
         _save_model(
             ov_model,
@@ -453,6 +459,7 @@ def export_models(
     ov_config: Optional["OVConfig"] = None,
     stateful: bool = True,
     patch_16bit_model: bool = False,
+    library_name: Optional[str] = None,
 ) -> Tuple[List[List[str]], List[List[str]]]:
     """
     Export the models to OpenVINO IR format
@@ -505,6 +512,7 @@ def export_models(
                 ov_config=ov_config,
                 stateful=stateful[i] if isinstance(stateful, (list, tuple)) else stateful,
                 patch_16bit_model=patch_16bit_model,
+                library_name=library_name,
             )
         )
 
@@ -701,6 +709,7 @@ def export_from_model(
         opset=opset,
         model_kwargs=model_kwargs,
         patch_16bit_model=patch_16bit_model,
+        library_name=library_name,
     )
 
     return files_subpaths
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -360,7 +360,6 @@ def export_pytorch(
                     input_dict = dict(zip(keys, tuple_input))
                     kwargs[input_name] = input_dict
                 outputs = patched_forward(*args, **kwargs)
-                print(outputs.keys())
                 return tuple([value if not isinstance(value, list) else tuple(value) for value in outputs.values()])
 
             patcher.patched_forward = ts_patched_forward
@@ -799,7 +798,6 @@ def _get_llava_submodels_and_export_configs(model: Union["PreTrainedModel", "TFP
     vision_tower.config.output_hidden_states = True
     vision_export_config_constructor = TasksManager.get_exporter_config_constructor(model=vision_tower, exporter="openvino", task="feature-extraction", library_name=library_name)
     vision_export_config = vision_export_config_constructor(vision_tower.config, int_dtype=int_dtype, float_dtype=float_dtype, preprocessors=preprocessors, legacy=False)
-    logger.warning(vision_export_config)
     lm_export_config = LMInputEmbedsConfigHelper(decoder["model"][1])
     models_for_export["language_model"] = (decoder["model"][0], lm_export_config)
     InputEmbedOpenvVINOConfig.NORMALIZED_CONFIG_CLASS = lm_export_config.orig_export_config.NORMALIZED_CONFIG_CLASS
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -607,7 +607,7 @@ def export_from_model(
         )
 
     logging.disable(logging.INFO)
-    export_config, models_and_export_configs, stateful_for_parts = _get_submodels_and_export_configs(
+    export_config, models_and_export_configs, stateful_submodels = _get_submodels_and_export_configs(
         model=model,
         task=task,
         monolith=False,
@@ -697,7 +697,7 @@ def export_from_model(
         input_shapes=input_shapes,
         device=device,
         ov_config=ov_config,
-        stateful=stateful_for_parts,
+        stateful=stateful_submodels,
         opset=opset,
         model_kwargs=model_kwargs,
         patch_16bit_model=patch_16bit_model,
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -31,7 +31,7 @@ from optimum.exporters.onnx.base import OnnxConfig
 from optimum.exporters.onnx.convert import check_dummy_inputs_are_allowed
 from optimum.exporters.onnx.convert import export_pytorch as export_pytorch_to_onnx
 from optimum.exporters.onnx.convert import export_tensorflow as export_tensorflow_onnx
-from optimum.exporters.utils import _get_submodels_and_export_configs
+from optimum.exporters.utils import _get_submodels_and_export_configs as _default_get_submodels_and_export_configs, get_decoder_models_for_export
 from optimum.intel.utils.import_utils import (
     _nncf_version,
     _open_clip_version,
@@ -57,6 +57,7 @@ from .utils import (
     clear_class_registry,
     remove_none_from_dummy_inputs,
 )
+from .model_configs import InputEmbedOpenvVINOConfig, LMInputEmbedsConfigHelper, LLavaMultimodalProjectorOpenVINOConfig
 
 
 logger = logging.getLogger(__name__)
@@ -316,8 +317,9 @@ def export_pytorch(
         #       this function is one of the _internal_ steps in a bigger model conversion pipeline.
 
     with torch.no_grad():
-        model.config.torchscript = False
-        model.config.return_dict = True
+        if hasattr(model, "config"):
+            model.config.torchscript = False
+            model.config.return_dict = True
         model.eval()
 
         # Check if we need to override certain configuration item
@@ -358,6 +360,7 @@ def export_pytorch(
                     input_dict = dict(zip(keys, tuple_input))
                     kwargs[input_name] = input_dict
                 outputs = patched_forward(*args, **kwargs)
+                print(outputs.keys())
                 return tuple([value if not isinstance(value, list) else tuple(value) for value in outputs.values()])
 
             patcher.patched_forward = ts_patched_forward
@@ -496,7 +499,7 @@ def export_models(
                 input_shapes=input_shapes,
                 model_kwargs=model_kwargs,
                 ov_config=ov_config,
-                stateful=stateful,
+                stateful=stateful[i] if isinstance(stateful, (list, tuple)) else stateful,
                 patch_16bit_model=patch_16bit_model,
             )
         )
@@ -560,7 +563,7 @@ def export_from_model(
 
         logger.info(f"Automatic task detection to: {task}.")
 
-    stateful = stateful and ensure_export_task_support_stateful(task)
+    stateful = stateful and ensure_export_task_support_stateful(task) or getattr(getattr(model, "config", {}), "model_type", None) == "llava"
 
     # TODO: support onnx_config.py in the model repo
     if custom_architecture and custom_export_configs is None:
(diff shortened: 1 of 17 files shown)


eaidova:
@echarlaix could you please take a look? Thanks!

P.S. I'm still working on extending models coverage, but I think it makes sense to start looking on general API now

optimum/exporters/openvino/model_configs.py
```diff
@@ -1018,3 +1035,448 @@ def patch_model_for_export(
...
+
+        if behavior == LlavaConfigBehavior.VISION_EMBEDDINGS:
+            vision_embedding = model.vision_tower
+            vision_embedding.config.output_hidden_states = True
+            vision_embedding.vision_model.config.output_hidden_states = True
+            return vision_embedding
+
+        if behavior == LlavaConfigBehavior.TEXT_EMBEDDINGS:
+            text_embedding = model.get_input_embeddings()
+            text_embedding.Llavaedding
```
eaidova, line 1321:
```suggestion
           return text_embedding
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -1018,3 +1035,448 @@ def patch_model_for_export(
...
+
+        if behavior == LlavaConfigBehavior.VISION_EMBEDDINGS:
+            vision_embedding = model.vision_tower
+            vision_embedding.config.output_hidden_states = True
+            vision_embedding.vision_model.config.output_hidden_states = True
+            return vision_embedding
+
+        if behavior == LlavaConfigBehavior.TEXT_EMBEDDINGS:
+            text_embedding = model.get_input_embeddings()
+           return text_embedding
```
eaidova, line 1321:
```suggestion
           text_embedding.config = model.config.text_config
           return text_embedding
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -1018,3 +1035,449 @@ def patch_model_for_export(
...
+        if behavior == LlavaConfigBehavior.VISION_EMBEDDINGS:
+            vision_embedding = model.vision_tower
+            vision_embedding.config.output_hidden_states = True
+            vision_embedding.vision_model.config.output_hidden_states = True
+            return vision_embedding
+
+        if behavior == LlavaConfigBehavior.TEXT_EMBEDDINGS:
+            text_embedding = model.get_input_embeddings()
+           text_embedding.config = model.config.text_config
+           return text_embedding
```
eaidova, line 1322:
```suggestion
            text_embedding = model.get_input_embeddings()
            text_embedding.config = model.config.text_config
            return text_embedding
```

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -0,0 +1,1205 @@
...
+        local_files_only: bool = False,
+        task: Optional[str] = None,
+        use_cache: bool = True,
+        trust_remote_code: bool = False,
+        load_in_8bit: Optional[bool] = None,
+        quantization_config: Union[OVWeightQuantizationConfig, Dict] = None,
+        **kwargs,
+    ):
+        """
+        Export a vanilla Transformers model into an ONNX model using `transformers.onnx.export_onnx`.
```
AlexKoff88, line 532: Do we really export to the ONNX? The docstrings should be revised accordingly.

AlexKoff88 (approved):

optimum/intel/openvino/modeling_visual_language.py
eaidova: I accidentally copied that from https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_base.py#L534
I see the same doc string in other model classes for openvino (seq2seq, stable diffusion support e.t.c.), @echarlaix  maybe it is time to revise them in all models too?

optimum/exporters/openvino/convert.py
```diff
@@ -687,7 +697,7 @@ def export_from_model(
         input_shapes=input_shapes,
         device=device,
         ov_config=ov_config,
-        stateful=stateful,
+        stateful=stateful_for_parts,
```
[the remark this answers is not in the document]
eaidova: done

optimum/exporters/openvino/convert.py
```diff
@@ -778,3 +788,84 @@ def _add_version_info_to_model(model: Model, library_name: Optional[str] = None)
...
+    models_for_export = {}
+    stateful_parts = []
+
+    model_type = model.config.model_type.replace("_", "-")
+
+    if model_type == "internvl-chat" and preprocessors is not None:
+        model.config.img_context_token_id = preprocessors[0].convert_tokens_to_ids("<IMG_CONTEXT>")
+
+    if hasattr(model, "image_newline"):
+        model.config.image_newline = model.image_newline.tolist()
```
echarlaix, line 812: why is this needed ?

optimum/exporters/openvino/convert.py
```diff
@@ -778,3 +788,84 @@ def _add_version_info_to_model(model: Model, library_name: Optional[str] = None)
...
+    main_config = main_config_cls(
+        model.config, int_dtype=int_dtype, float_dtype=float_dtype, preprocessors=preprocessors
+    )
+    for behavior in main_config.SUPPORTED_BEHAVIORS:
+        model_id = f"{behavior}_model"
+        model_part_config = main_config.with_behavior(behavior)
+        model_part = main_config.get_model_for_behaviour(model, behavior)
+        models_for_export[model_id] = (model_part, model_part_config)
+        stateful_parts.append(stateful if getattr(model_part_config, "use_past", False) else False)
+    return main_config, models_for_export, stateful_parts
```
echarlaix, line 825: `stateful_parts` can be inferred from `models_for_export` so can't we get this value from there instead of returning `stateful_parts` here and in `_get_submodels_and_export_configs` to pass it to `export_models` afterwards ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2652,3 +2652,53 @@ def __init__(
...
+        model.forward = model.extract_feature
+
+        super().__init__(config, model, model_kwargs)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+        self._model.forward = self._model.__orig_forward
+
+
+def llava_vision_embed_forward(self, pixel_values):
```
echarlaix, line 2674: can you add the link to the original code + specify what has been modified so that it's later easier for maintenance ?

optimum/intel/utils/modeling_utils.py
```diff
@@ -267,6 +267,8 @@ def _infer_library_from_model_or_model_class(
 ):
     if model.__module__.startswith("open_clip"):
         library_name = "open_clip"
+    elif model.__module__.startswith("torch"):
+        library_name = "transformers"
```
echarlaix, line 271: why this ?

optimum/exporters/openvino/model_configs.py
```diff
@@ -1194,3 +1210,372 @@ def patch_model_for_export(
...
+
+    def generate_dummy_inputs(self, framework: str = "pt", **kwargs) -> Dict:
+        if self._behavior == LlavaConfigBehavior.VISION_EMBEDDINGS and self._config.model_type == "pixtral":
+            kwargs["batch_size"] = 1
+        return super().generate_dummy_inputs(framework, **kwargs)
+
+
+@register_in_tasks_manager("llava-next", *["image-text-to-text"], library_name="transformers")
+class LlavaNextOpenVINOConfig(LlavaOpenVINOConfig):
+    pass
```
echarlaix, line 1429: we could add the minimum transformers version needed with MIN_TRANSFORMERS_VERSION

optimum/exporters/openvino/model_configs.py
```diff
@@ -1194,3 +1210,372 @@ def patch_model_for_export(
...
+            return self.__class__(
+                self._orig_config,
+                task=self.task,
+                int_dtype=self.int_dtype,
+                float_dtype=self.float_dtype,
+                behavior=behavior,
+                preprocessors=self._preprocessors,
+            )
+
+    def get_model_for_behaviour(self, model, behavior: Union[str, LlavaConfigBehavior]):
```
echarlaix, line 1560:
```suggestion
    def get_model_for_behavior(self, model, behavior: Union[str, LlavaConfigBehavior]):
```

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -0,0 +1,1185 @@
...
+        self.request.wait()
+        logits = self.request.get_tensor("logits").data
+        logits = torch.from_numpy(logits).to(self.device)
+        past_key_values = ((),)
+        self._past_length += inputs["inputs_embeds"].shape[1]
+
+        return CausalLMOutputWithPast(logits=logits, past_key_values=past_key_values)
+
+
+class OVModelPart:
```
echarlaix, line 169: could we re-use part of https://github.com/huggingface/optimum-intel/blob/9cc0ee41b5e9ef114afeaeb11ba715d4165b7d74/optimum/intel/openvino/modeling_diffusion.py#L634 and inherit from it (base class could be moved in https://github.com/huggingface/optimum-intel/blob/9cc0ee41b5e9ef114afeaeb11ba715d4165b7d74/optimum/intel/openvino/modeling_base.py)

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -0,0 +1,1185 @@
...
+        if self.text_emb_request is None:
+            logger.info(f"Compiling the Text embeddings model to {self._device} ...")
+            self.text_emb_request = core.compile_model(self.text_emb_model, self._device, self.ov_config)
+
+    def to(self, device: str):
+        if isinstance(device, str):
+            self._device = device.upper()
+            self.clear_requests()
+
+        return self
```
echarlaix, line 68: could make sense to modify `OVModel.to` to not have to overwrite it https://github.com/huggingface/optimum-intel/blob/9cc0ee41b5e9ef114afeaeb11ba715d4165b7d74/optimum/intel/openvino/modeling.py#L129 (just adding a `clear_requests` method), what do you think ?

echarlaix (commented):
Looking good, thanks a lot @eaidova

optimum/exporters/openvino/convert.py
eaidova: this is tranable parameter filled by random values during model weights init, we can not capture this during model export as it is not a part of submodels inference forward pass, that is why for transferring it from pt to ov I need to use config for that. If you have better suggestion, I can try to implement this

optimum/exporters/openvino/model_configs.py
eaidova: I think it can be complicated e.g. pixtral model is also has model_type llava, but it is required 4.45 to create such model, while usual llava added in 4.39 if I'm not wrong

optimum/exporters/openvino/convert.py
eaidova: I think it maybe nontrivial and more complex if we move that on other level, where some info about model type can be already lost. e.g. seq2seq models also split on parts and have decoder with past that currently should remain stateless until ov runtime will be ready for this IR format https://github.com/huggingface/optimum-intel/pull/779 (yes, we still working on that and I belive it will happens soon). also with exploring more models, it maybe changed too, e.g. there is blip2-t5-xl model that uses encoder_decoder architecture for language model with the same working approach that also will require some changes in this logic in future

optimum/intel/utils/modeling_utils.py
eaidova: model.get_input_embeddings() returns instance of torch.nn.Embeddings class, if it is possible skip this check for such models, I think we can avoid this workaround

optimum/intel/openvino/modeling_visual_language.py
eaidova: not sure about inheritage, becuase it adds some extra functional (e.g. _compile vs compile, model export as completely independent models), but agree for moving

optimum/exporters/openvino/model_configs.py

optimum/intel/openvino/modeling_visual_language.py
eaidova: done

optimum/intel/openvino/modeling_visual_language.py
eaidova: done

optimum/exporters/openvino/stateful.py
```diff
@@ -213,14 +224,22 @@ def patch_stateful(config: PretrainedConfig, ov_model: ov.Model):
     if not key_value_input_names or not key_value_output_names:
         return
 
+    main_input_name = "input_ids" if model_has_input_output_name(ov_model, "input_ids") else "inputs_embeds"
```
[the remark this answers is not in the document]
eaidova: done

optimum/intel/utils/modeling_utils.py
eaidova: added opportunity to propagate library_name from whole model to subcomponets

optimum/exporters/openvino/convert.py
eaidova: done

optimum/exporters/openvino/model_patcher.py
eaidova: done

optimum/exporters/openvino/model_configs.py

optimum/exporters/openvino/convert.py
echarlaix: Makes sense, let's keep it that way then!

> https://github.com/huggingface/optimum-intel/pull/779 (yes, we still working on that and I belive it will happens soon)

that's very cool !

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -0,0 +1,1192 @@
...
+                Whether or not to only look at local files (i.e., do not try to download the model).
+        """
+        if use_auth_token is not None:
+            warnings.warn(
+                "The `use_auth_token` argument is deprecated and will be removed soon. Please use the `token` argument instead.",
+                FutureWarning,
+            )
+            if token is not None:
+                raise ValueError("You cannot use both `use_auth_token` and `token` arguments at the same time.")
+            token = use_auth_token
```
echarlaix, line 380: shouldn't be needed I think as already check in https://github.com/huggingface/optimum-intel/blob/6a8718b76a5fc5e6796483746eded64e4007e94d/optimum/intel/openvino/modeling_base.py#L394

```suggestion
```

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -0,0 +1,1192 @@
...
+    ):
+        if use_auth_token is not None:
+            warnings.warn(
+                "The `use_auth_token` argument is deprecated and will be removed soon. Please use the `token` argument instead.",
+                FutureWarning,
+            )
+            if token is not None:
+                raise ValueError("You cannot use both `use_auth_token` and `token` arguments at the same time.")
+            token = use_auth_token
+
```
echarlaix, line 541:
```suggestion
```

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -0,0 +1,1192 @@
...
+            text_embeddings=text_embeddings,
+            vision_embeddings=vision_embeddings,
+            config=config,
+            model_save_dir=model_save_dir,
+            quantization_config=quantization_config,
+            **kwargs,
+        )
+
+    @classmethod
+    def _from_transformers(
```
echarlaix, line 515: what is the main difference with https://github.com/huggingface/optimum-intel/blob/6a8718b76a5fc5e6796483746eded64e4007e94d/optimum/intel/openvino/modeling_decoder.py#L262 ? If only the "with-past" suffix maybe we can change it there instead wdyt ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -0,0 +1,1192 @@
...
+                    kwargs.get("ov_config"),
+                    model_save_dir,
+                )
+                for part in model_cls.additional_parts:
+                    kwargs[part] = model_cls._compile_model(
+                        file_names[part],
+                        kwargs.get("device", "CPU"),
+                        kwargs.get("ov_config"),
+                        model_save_dir,
+                    )
```
echarlaix, line 490: could the code be simplified with something like 

```suggestion
        model_file_names = {
            "language_model": "openvino_language_model.xml",
            "text_embeddings":  "openvino_text_embeddings_model.xml",
            "vision_embeddings":  "openvino_vision_embeddings_model.xml",
        }
    
        for part in model_cls.additional_parts:
            model_file_names[part] = f"openvino_{part}_model.xml"

        model_cls = MODEL_TYPE_TO_CLS_MAPPING[config.model_type]

        quantization_config = model_cls._prepare_weight_quantization_config(quantization_config, load_in_8bit)
        compile_only = kwargs.get("compile_only", False)

        if os.path.isdir(model_id):
            model_save_dir = Path(model_id)
            file_names = {k:os.path.join(model_id, model_file_names[k]) for k in model_file_names}
        else:
            file_names = model_file_names.copy()
            for name, file_name in model_file_names.items():
                model_cache_path = hf_hub_download(
                    repo_id=model_id,
                    filename=file_name,
                    token=token,
                    revision=revision,
                    cache_dir=cache_dir,
                    force_download=force_download,
                    local_files_only=local_files_only,
                )
                file_names[name] = model_cache_path

            model_save_dir = Path(model_cache_path).parent

        if not compile_only:
            language_model = model_cls.load_model(file_names["language_model"], quantization_config)
            text_embeddings = model_cls.load_model(file_names["text_embeddings"], quantization_config)
            vision_embeddings = model_cls.load_model(file_names["vision_emnbeddings"], quantization_config)
            for part in model_cls.additional_parts:
                kwargs[part] = model_cls.load_model(file_names[part], quantization_config)
        else:
            language_model = model_cls._compile_model(
                file_names["language_model"],
                kwargs.get("device", "CPU"),
                kwargs.get("ov_config"),
                model_save_dir,
            )
            text_embeddings = model_cls._compile_model(
                file_names["text_embeddings"],
                kwargs.get("device", "CPU"),
                kwargs.get("ov_config"),
                model_save_dir,
            )
            vision_embeddings = model_cls._compile_model(
                file_names["vision_embeddings"],
                kwargs.get("device", "CPU"),
                kwargs.get("ov_config"),
                model_save_dir,
            )
            for part in model_cls.additional_parts:
                kwargs[part] = model_cls._compile_model(
                    file_names[part],
                    kwargs.get("device", "CPU"),
                    kwargs.get("ov_config"),
                    model_save_dir,
                )
```

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -0,0 +1,1192 @@
...
+                        patch_feats.append(self.vision_embeddings(np.expand_dims(patch_value, 0)).last_hidden_state)
+                    patch_feats = np.concatenate(patch_feats, axis=1)
+                else:
+                    patch_feats = self.vision_embeddings(patch).last_hidden_state
+                image_features.append(patch_feats)
+            image_features = np.concatenate(image_features, 0)
+
+        return image_features
+
+    def merge_vision_text_embeddings(
```
echarlaix, line 750: could you add a link to the original code for maintenance ? https://github.com/huggingface/transformers/blob/d7950bff82b18c823193d17d72188c5e46d06c83/src/transformers/models/llava/modeling_llava.py#L297C9-L297C45

echarlaix (approved):
Huge work, thanks @eaidova, left couple of minor comments

optimum/intel/openvino/modeling_visual_language.py
eaidova: do you mean I should inherit from OVModelForCausalLM to get this behaviour? I think it may lead to issues with quantization as it may requires slightly different approach for doing that (difference in input data collection, different approaches for vision embeddings and text model and so on)

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -0,0 +1,1192 @@
...
+        if self._compile_only:
+            raise ValueError(
+                "`to()` is not supported with `compile_only` mode, please intialize model without this option"
+            )
+
+        if isinstance(device, str):
+            self._device = device.upper()
+            self.clear_requests()
+
+        return self
```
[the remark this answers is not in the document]
eaidova: removed

optimum/intel/openvino/modeling_visual_language.py
eaidova: removed

optimum/intel/openvino/modeling_visual_language.py
eaidova: removed

optimum/intel/openvino/modeling_visual_language.py
eaidova: simplified

optimum/intel/openvino/modeling_visual_language.py
eaidova: added

optimum/intel/openvino/modeling_visual_language.py
echarlaix: makes sense, let's keep it that way then!

echarlaix:
> P.S. I'm still working on extending models coverage, but I think it makes sense to start looking on general API now

Very nice, let's merge this PR and we can always extend support in following PRs, does it work for you @eaidova ?

eaidova:
> > P.S. I'm still working on extending models coverage, but I think it makes sense to start looking on general API now
> 
> Very nice, let's merge this PR and we can always extend support in following PRs, does it work for you @eaidova ?

@echarlaix , yes, I agree, could you please merge?

PR #886, opened 2024-09-03
Deprecate `compression_option` and `compression_ratio` OV export parameters

nikita-savelyevv:
# What does this PR do?

`compression_option` and `compression_ratio` parameters are no longer used and to be deprecated since v1.17

## Before submitting


echarlaix (approved):
Thanks @nikita-savelyevv !

PR #889, opened 2024-09-08
task specific openvino diffusion pipelines

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -16,6 +16,7 @@ import importlib
 import logging
 import os
 import shutil
+from collections import OrderedDict
 from copy import deepcopy
 from pathlib import Path
 from tempfile import TemporaryDirectory, gettempdir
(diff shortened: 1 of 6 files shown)


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -77,8 +86,6 @@
 class OVStableDiffusionPipelineBase(OVBaseModel, OVTextualInversionLoaderMixin):
     auto_model_class = StableDiffusionPipeline
     config_name = "model_index.json"
-    export_feature = "text-to-image"
-    _library_name = "diffusers"
```
echarlaix, line 81: shouldn't it be removed no ?

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -636,24 +642,31 @@ def __init__(
         parent_model: OVBaseModel,
         ov_config: Optional[Dict[str, str]] = None,
         model_name: str = "text_encoder",
+        model_dir: str = None,
```
echarlaix, line 645: is it set anywhere and if no what about removing it (potentially from `OVModelPart`)

echarlaix (approved):

PR #890, opened 2024-09-10
Add a warning about OVTrainer deprecation

nikita-savelyevv:
# What does this PR do?

OVTrainer was introduced for quantized/sparse fine-tuning of relatively small models, but most SotA transformers nowadays are much larger and finetuning them on client side is not feasible anymore. For such models weight compression is mainly applied now. So OVTrainer has not been popular for quite some time and at the same time has some testing burden connected with it. We've decided to deprecate it.

In this PR a warning about deprecation in the next release is added.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/trainer.py
+++ b/optimum/intel/openvino/trainer.py
@@ -213,6 +213,10 @@ class OVTrainer(Trainer):
         ov_config: Optional[OVConfig] = None,
         task: Optional[str] = None,
     ):
+        logger.warning(
+            "Quantization aware training capabilities are deprecated and will be removed in optimum-intel v1.19.0. "
+        )
+
         self.neftune_noise_alpha = None
 
         super().__init__(
--- a/optimum/intel/openvino/trainer.py
+++ b/optimum/intel/openvino/trainer.py
@@ -214,7 +214,7 @@ class OVTrainer(Trainer):
         task: Optional[str] = None,
     ):
         logger.warning(
-            "Quantization aware training capabilities are deprecated and will be removed in optimum-intel v1.19.0. "
+            "OVTrainer is deprecated and will be removed in optimum-intel v1.22.0. "
         )
 
         self.neftune_noise_alpha = None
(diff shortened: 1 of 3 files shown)


nikita-savelyevv:

cc @vuiseng9

optimum/intel/openvino/trainer.py
```diff
@@ -213,6 +213,10 @@ def __init__(
         ov_config: Optional[OVConfig] = None,
         task: Optional[str] = None,
     ):
+        logger.warning(
+            "Quantization aware training capabilities are deprecated and will be removed in optimum-intel v1.19.0. "
```
AlexKoff88, line 217: I am not sure if we agreed on a certain version to remove it. @echarlaix , I would like to know your opinion.

echarlaix: If the reason to deprecate the `OVTrainer` is that it isn't much used, why only remove the QAT part and not the trainer itself ? From my point of view it's the same in terms of maintenance (at least on the optimum-intel side).  In any case we need to have this warning before deprecating it officially for at least couple of minor release, we are now at v1.19.0 so we can push it to 1.22.0 wdyt ?

nikita-savelyevv: @echarlaix @AlexKoff88, I've updated the deprecation message. It was poor phrasing on my side, indeed it should've said OVTrainer instead of QAT

AlexKoff88: > If the reason to deprecate the `OVTrainer` is that it isn't much used, why only remove the QAT part and not the trainer itself ? From my point of view it's the same in terms of maintenance (at least on the optimum-intel side). In any case we need to have this warning before deprecating it officially for at least couple of minor release, we are now at v1.19.0 so we can push it to 1.22.0 wdyt ?

Thanks, Ella. I don't mind.

echarlaix: Thanks ! Also do you mind adding a  warning in the [documentation](https://huggingface.co/docs/optimum/main/en/intel/openvino/optimization#training-time) directly https://github.com/huggingface/optimum-intel/blob/08e14924b307c990bfbc3fae4071ca8e1e35310b/docs/source/openvino/optimization.mdx?plain=1#L45

and we can also remove this code snippet https://github.com/huggingface/optimum-intel?tab=readme-ov-file#quantization-aware-training

nikita-savelyevv: @echarlaix, thanks! Done.

PR #895, opened 2024-09-11
Remove the need to enforce transformers library for OVModelForFeatureExtraction

echarlaix:
Already set to `transformers_library` by default https://github.com/huggingface/optimum-intel/blob/08e14924b307c990bfbc3fae4071ca8e1e35310b/optimum/intel/openvino/modeling_base.py#L505 since https://github.com/huggingface/optimum-intel/pull/865


IlyasMoutawwakil (approved):
LGTM, thanks !

PR #897, opened 2024-09-13
fix config saving for openvino model

eaidova:
# What does this PR do?
in same cases optimum cli failed to export models due to inability to save config because it may contains some extra arguments, I added possibility to avoid this issue (example of such model https://huggingface.co/openai/whisper-tiny)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/utils.py
+++ b/optimum/exporters/openvino/utils.py
@@ -14,6 +14,7 @@
 
 import inspect
 from collections import namedtuple
+from pathlib import Path
 from typing import Any, Dict, List, Tuple, Union
 
 from transformers.utils import is_torch_available
@@ -158,3 +159,10 @@ def clear_class_registry():
     torch._C._jit_clear_class_registry()
     torch.jit._recursive.concrete_type_store = torch.jit._recursive.ConcreteTypeStore()
     torch.jit._state._clear_class_state()
+
+
+def save_config(save_dir, config):
+    save_dir = Path(save_dir)
+    save_dir.mkdir(exist_ok=True)
+    output_config_file = Path(save_dir / "config.json")
+    config.to_json_file(output_config_file, use_diff=True)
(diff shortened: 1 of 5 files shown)


optimum/exporters/openvino/utils.py
```diff
@@ -158,3 +159,10 @@ def clear_class_registry():
     torch._C._jit_clear_class_registry()
     torch.jit._recursive.concrete_type_store = torch.jit._recursive.ConcreteTypeStore()
     torch.jit._state._clear_class_state()
+
+
+def save_config(save_dir, config):
```
AlexKoff88, line 164: How about always calling your `save_config` instead of `model.config.save_pretrained`?

optimum/intel/openvino/modeling_base.py
```diff
@@ -581,8 +589,10 @@ def _from_transformers(
             ov_config=ov_config,
             library_name=cls._library_name,
         )
-
-        config.save_pretrained(save_dir_path)
+        try:
```
AlexKoff88, line 592: I guess the same is relevant here.

optimum/exporters/openvino/utils.py
eaidova: done

AlexKoff88 (approved):

echarlaix:
The generation parameters included in the model's config will be moved to the generation_config in https://github.com/huggingface/optimum-intel/pull/902 which should fix this issue

eaidova:
> The generation parameters included in the model's config will be moved to the generation_config in #902 which should fix this issue

@echarlaix  thanks for confirmation and better solution! Could you please provide info when it is planned to be ready to merge?

echarlaix:
> @echarlaix thanks for confirmation and better solution! Could you please provide info when it is planned to be ready to merge?

Couple of tests failing still requiring fixes, but if the config part is ready so we can merge this part if needed on your side, let me know!

PR #908, opened 2024-09-23
Patch model for common usage

jiqing-feng:
The ipex patching is a common optimization for all tasks, so we can remove the task limitation.


jiqing-feng:
Hi @echarlaix @IlyasMoutawwakil . Please take a review at this PR, it removes the limitation for patching.

PR #909, opened 2024-09-24
Reset OV inference request after model quantization

nikita-savelyevv:
# What does this PR do?

Reset inference request after model is changed by quantization. Otherwise, after quantization call the non-quantized model is inferred.

## Before submitting


AlexKoff88 (approved):

echarlaix (approved):
Thanks for the fix !

PR #911, opened 2024-09-25
Fix non-stateful OVModelForCausalLM inference on SPR

nikita-savelyevv:
# What does this PR do?

Create empty KV-cache input using ov.Tensor instead of np.ndarray to support the case of BF16 input dtype.

**Related tickets**
153482

## Before submitting


eaidova (approved):

nikita-savelyevv:
The failed test is irrelevant (it fails sometimes)

AlexKoff88 (approved):

AlexKoff88:
@nikita-savelyevv, I see one test failed. Is it expected as we are moving to new OpenVINO and NNCF versions?

AlexKoff88 (approved):

echarlaix (approved):

PR #912, opened 2024-09-25
Add GPTQ support

nikita-savelyevv:
# What does this PR do?

Add support for NNCF weights compression with GPTQ. 

Usage:
```
quantization_config = OVWeightQuantizationConfig(bits=4, dataset="wikitext2", gptq=True)
ov_config = OVConfig(quantization_config=quantization_config)
model = OVModelFromCausalLM.from_pretrained(model_id, export=True, ov_config=ov_config)
```

```
optimum-cli export openvino --task text-generation-with-past -m <model_id> --weight-format int4 --dataset wikitext2 --gptq ./model
```

## Before submitting

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -158,6 +158,16 @@ def parse_args_openvino(parser: "ArgumentParser"):
         ),
     )
     optional_group.add_argument(
+        "--gptq",
+        action="store_true",
+        default=None,
+        help=(
+            "Indicates whether to apply GPTQ algorithm that optimizes compressed weights in a layer-wise fashion to "
+            "minimize the difference between activations of a compressed and original layer. Please note, that "
+            "applying GPTQ takes additional memory and time."
+        ),
+    )
+    optional_group.add_argument(
         "--sensitivity-metric",
         type=str,
         default=None,
@@ -203,6 +213,8 @@ def no_compression_parameter_provided(args):
                 args.dataset,
                 args.num_samples,
                 args.awq,
+                args.scale_estimation,
+                args.gptq,
                 args.sensitivity_metric,
             )
         )
@@ -272,6 +284,7 @@ class OVExportCommand(BaseOptimumCLICommand):
                     "quant_method": "awq" if self.args.awq else "default",
                     "sensitivity_metric": self.args.sensitivity_metric,
                     "scale_estimation": self.args.scale_estimation,
+                    "gptq": self.args.gptq,
                     "weight_format": self.args.weight_format,
                 }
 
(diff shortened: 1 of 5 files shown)


optimum/commands/export/openvino.py
```diff
@@ -157,6 +157,16 @@ def parse_args_openvino(parser: "ArgumentParser"):
...
         ),
     )
+    optional_group.add_argument(
+        "--gptq",
+        action="store_true",
+        default=None,
+        help=(
+            "Indicates whether to apply GPTQ algorithm that optimizes compressed weights in a layer-wise fashion to "
+            "minimize the difference between activations of a compressed and original layer. Please note, that "
+            "applying GPTQ takes additional memory and time."
```
alexsu52, line 167:
```suggestion
            "Indicates whether to apply GPTQ algorithm that compress weights layer-by-layer  to, "
            "minimizing the difference between the original and compressed layers. Please note, that "
            "applying GPTQ takes additional memory and time."
```

optimum/commands/export/openvino.py
```diff
@@ -157,6 +157,16 @@ def parse_args_openvino(parser: "ArgumentParser"):
...
+        ),
+    )
```
IlyasMoutawwakil, line 169: doesn't it make more sense to add GPTQ as a `quant_method` like AWQ instead of creating a new field for it ?

optimum/commands/export/openvino.py
```diff
@@ -272,6 +284,7 @@ def run(self):
                     "quant_method": "awq" if self.args.awq else "default",
                     "sensitivity_metric": self.args.sensitivity_metric,
                     "scale_estimation": self.args.scale_estimation,
+                    "gptq": self.args.gptq,
```
IlyasMoutawwakil, line 287: why not use the quant_method field that already takes awq as a value ?

nikita-savelyevv: Because GPTQ and AWQ are non-exclusive algorithms and can be applied simultaneously. The same for Scale Estimation.

AlexKoff88: I would say three methods can be applied sequentially in one run. This is what we are already practicing and getting better accuracy.

AlexKoff88 (approved):

nikita-savelyevv:
The failed test is irrelevant (it fails sometimes)

IlyasMoutawwakil (approved):

alexsu52 (approved):

nikita-savelyevv:
The PR is ready for merge

optimum/commands/export/openvino.py
echarlaix: clear misunderstanding from my side as I didn't know awq/gptq should / could be combined to get better accuracy. One possibility could be to create a dedicated `quant_method` to combine these different methods @AlexKoff88 what do you think?

AlexKoff88: Actually, there is also the third one "Scale Estimation" so that we saw that all three together can provide the best results. I understand that it is not that easy to understand this from the user side. But for now, we are ok with this flexibility as we basically provide the optimization recipes.

PR #916, opened 2024-09-26
sd3 pipeline support

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


IlyasMoutawwakil:
The changes to be made in this PR after the refactorization are very simple and mostly will make it more readable:

- no need to copy paste modeling from diffusers anymore, you just subclass (OVDiffusionPipeline, StabeldDiffusion3Pipeline) and it should be good in most cases.
- adding tests is as simple as adding the architecture to the task specific tests, I see you haven't added tests yet tho...

I also suggest you take a look at the original refactorization work in https://github.com/huggingface/optimum/pull/2021, I explain there issues with the previous approach and why the refactorization is not only that but also fixing many issues with the previous design (reproducibility of results & numeric consistency with diffusers), which weren't tested before or simply ignored.

optimum/exporters/openvino/convert.py
```diff
@@ -885,3 +899,116 @@ def _get_submodels_and_export_configs(
...
+            model=text_encoder_3,
+            exporter=exporter,
+            library_name="diffusers",
+            task="feature-extraction",
+            model_type="clip-text-with-projection",
+        )
+        export_config = export_config_constructor(text_encoder_3.config, int_dtype=int_dtype, float_dtype=float_dtype)
+        models_for_export["text_encoder_3"] = (text_encoder_3, export_config)
+
+    return None, models_for_export
```
IlyasMoutawwakil, line 1014: very convoluted, we could easily add this in main optimum and make the developer and super-user experience better.

eaidova: it is what I would like to avoid, I do not want to make support onnx

optimum/intel/utils/dummy_openvino_and_diffusers_objects.py
```diff
@@ -145,3 +145,14 @@ def __init__(self, *args, **kwargs):
...
+
+class OVStableDiffusion3Img2ImgPipeline(metaclass=DummyObject):
+    _backends = ["openvino", "diffusers"]
+
+    def __init__(self, *args, **kwargs):
+        requires_backends(self, ["openvino", "diffusers"])
+
+    @classmethod
+    def from_pretrained(cls, *args, **kwargs):
+        requires_backends(cls, ["openvino", "diffusers"])
```
IlyasMoutawwakil, line 158: should add the other tasks as well

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -504,21 +559,24 @@ def to(self, *args, device: Optional[str] = None, dtype: Optional[torch.dtype] =
...
+        model = self.unet.model if self.unet is not None else self.transformer.model
+        width = model.inputs[0].get_partial_shape()[3]
         if width.is_dynamic:
             return -1
         return width.get_length() * self.vae_scale_factor
 
     @property
     def batch_size(self) -> int:
-        batch_size = self.unet.model.inputs[0].get_partial_shape()[0]
+        model = self.unet.model if self.unet is not None else self.transformer
```
IlyasMoutawwakil, line 578:
```suggestion
        model = self.unet.model if self.unet is not None else self.transformer.model
```

IlyasMoutawwakil (approved):
LGTM, can you run styling and add tests please.

optimum/intel/utils/dummy_openvino_and_diffusers_objects.py
eaidova: added

tests/openvino/utils_tests.py
```diff
@@ -118,6 +118,7 @@
     "stable-diffusion-openvino": "hf-internal-testing/tiny-stable-diffusion-openvino",
     "stable-diffusion-xl": "echarlaix/tiny-random-stable-diffusion-xl",
     "stable-diffusion-xl-refiner": "echarlaix/tiny-random-stable-diffusion-xl-refiner",
+    "stable-diffusion-3": "yujiepan/stable-diffusion-3-tiny-random",
```
IlyasMoutawwakil, line 121: in the exporters and exporters cli as well:
https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_exporters_cli.py#L90-L94
https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_export.py#L67-L71

IlyasMoutawwakil:
- it seems sd3 still doesn't support callbacks (idk if it ever will as it doesn't have a unet), need to be skipped.
- it's also not as numerically consistent with diffusers, which is interesting, we can increase atol and rtol to 1e-3 and 1e-1 for this architecture for now, but worth investigating, especially if each exported component is numerically consistent.
- and finally I guess `output_type="latent"` is different for this model, again because it uses a transformer instead of unet https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_diffusion.py#L419-L422

IlyasMoutawwakil (changes requested):
will be merged once tests are updated and passed

tests/openvino/utils_tests.py
```diff
@@ -118,7 +118,7 @@
     "stable-diffusion-openvino": "hf-internal-testing/tiny-stable-diffusion-openvino",
     "stable-diffusion-xl": "echarlaix/tiny-random-stable-diffusion-xl",
     "stable-diffusion-xl-refiner": "echarlaix/tiny-random-stable-diffusion-xl-refiner",
-    "stable-diffusion-3": "yujiepan/stable-diffusion-3-tiny-random",
+    "stable-diffusion-3": "katuni4ka/tiny-random-stabdle-diffusion-3",
```
IlyasMoutawwakil, line 121: did you mean
```suggestion
    "stable-diffusion-3": "katuni4ka/tiny-random-stable-diffusion-3",
```

tests/openvino/utils_tests.py
eaidova: added

IlyasMoutawwakil:
@eaidova some tests are still failing

optimum/exporters/openvino/model_configs.py
```diff
@@ -1570,3 +1576,183 @@ def patch_model_for_export(
...
+            latent_image_ids[..., 2] = latent_image_ids[..., 2] + np.arange(img_ids_width)[None, :]
+
+            latent_image_id_height, latent_image_id_width, latent_image_id_channels = latent_image_ids.shape
+
+            latent_image_ids = np.tile(latent_image_ids[None, :], (self.batch_size, 1, 1, 1))
+            latent_image_ids = latent_image_ids.reshape(
+                self.batch_size, latent_image_id_height * latent_image_id_width, latent_image_id_channels
+            )
+            latent_image_ids.astype(DTYPE_MAPPER.np[float_dtype])
+            return latent_image_ids
```
IlyasMoutawwakil, line 1717: why not generate the random float with the final shape directly (self.batch_size, latent_image_id_height * latent_image_id_width, 3)

optimum/exporters/openvino/model_configs.py
```diff
@@ -1570,3 +1576,183 @@ def patch_model_for_export(
...
+        "decoder_input_ids",
+        "decoder_attention_mask",
+        "encoder_outputs",
+        "encoder_hidden_states",
+        "txt_ids",
+    )
+
+    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+        if input_name == "txt_ids":
+            return self.constant_tensor([self.batch_size, self.sequence_length, 3], 0, DTYPE_MAPPER.pt(float_dtype))
```
IlyasMoutawwakil, line 1731: there seems to be an issue with `constant_tensor` where it doesn't take a string dtype and dispatch it like the other methods, can you use the torch method directly so that we can fix this in optimum without it impacting optimum-intel.

optimum/exporters/openvino/model_configs.py
eaidova: because this info used as indices for iteration inside rope tensor, random values may raise IndexError for this case, so it is better to have determenistic result for this case

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2705,3 +2705,39 @@ def __init__(
...
+        stacked_out = torch.stack([cos_out, -sin_out, sin_out, cos_out], dim=-1)
+        out = stacked_out.view(batch_size, -1, dim // 2, 2, 2)
+        return out.float()
+
+    n_axes = ids.shape[-1]
+    emb = torch.cat(
+        [rope(ids[..., i], self.axes_dim[i], self.theta) for i in range(n_axes)],
+        dim=-3,
+    )
+    return emb.unsqueeze(1)
```
IlyasMoutawwakil, line 2731: would like an explanation of what fails with the original one

optimum/exporters/openvino/model_configs.py
IlyasMoutawwakil: the same is true for position_ids, random values in a small range (0, 1) wouldn't be an issue.

eaidova: I'll try but not sure as desired type is float, but values should be int

optimum/exporters/openvino/model_patcher.py
eaidova: it is just openvino limitation we do not support this case for einsum on runtime side, there is plan to support it, but for now do not wait next release package I prefer to eliminate it.

eaidova:
@IlyasMoutawwakil  now all relevant tests are fixed, please take a look one more time

optimum/exporters/openvino/model_configs.py
```diff
@@ -1570,3 +1576,173 @@ def patch_model_for_export(
...
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        random_batch_size_range: Optional[Tuple[int, int]] = None,
+        **kwargs,
+    ):
+        self.task = task
+        if random_batch_size_range:
+            low, high = random_batch_size_range
+            self.batch_size = random.randint(low, high)
+        else:
+            self.batch_size = batch_size
```
IlyasMoutawwakil, line 1597: export with random batch size is almost never used, I don't see the need

optimum/exporters/openvino/model_configs.py
```diff
@@ -1570,3 +1576,173 @@ def patch_model_for_export(
...
+            inputs["hidden_states"] = hidden_states
+        return inputs
+
+
+@register_in_tasks_manager("t5-encoder-model", *["feature-extraction"], library_name="diffusers")
+class T5EncoderOpenVINOConfig(CLIPTextOnnxConfig):
+    def patch_model_for_export(
+        self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ) -> ModelPatcher:
+        return ModelPatcher(self, model, model_kwargs=model_kwargs)
```
IlyasMoutawwakil, line 1649: just a `pass` would be enough here no ?

optimum/intel/__init__.py
```diff
@@ -267,6 +275,9 @@
             OVPipelineForImage2Image,
             OVPipelineForInpainting,
             OVPipelineForText2Image,
+            OVStableDiffusion3Img2ImgPipeline,
+            OVStableDiffusion3InpaintPipeline,
+            OVStableDiffusion3Pipeline,
```
IlyasMoutawwakil, line 280: OVFlux is missing here

tests/openvino/test_export.py
```diff
@@ -70,6 +72,9 @@ class ExportModelTest(unittest.TestCase):
         "latent-consistency": OVLatentConsistencyModelPipeline,
     }
 
+    if is_transformers_version(">=", "4.45"):
```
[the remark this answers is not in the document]
eaidova: problem in tokenizers version, models saved using transformers 4.45 and tokenizers 0.20 can not be loaded with lower version transformers

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -510,21 +564,26 @@ def to(self, *args, device: Optional[str] = None, dtype: Optional[torch.dtype] =
 
     @property
     def height(self) -> int:
-        height = self.unet.model.inputs[0].get_partial_shape()[2]
+        # flux transformer does not preserve info about height/width, they are knwon in vae_decoder
+        model = self.unet.model if self.unet is not None else self.vae.decoder.model
+        height = model.inputs[0].get_partial_shape()[2]
```
IlyasMoutawwakil, line 569: the vae decoder seems to have this information while always present, let's use it by default.

optimum/exporters/openvino/model_configs.py
eaidova: it is artefact from time where clip has hardcoded sdpa replacement to eager for avoiding wrong attention impelementation

optimum/exporters/openvino/model_configs.py
eaidova: removed

optimum/intel/__init__.py
eaidova: added

optimum/intel/openvino/modeling_diffusion.py
eaidova: changed

tests/openvino/test_diffusion.py
```diff
@@ -126,18 +132,19 @@ def test_compare_to_diffusers_pipeline(self, model_arch: str):
         height, width, batch_size = 128, 128, 1
         inputs = self.generate_inputs(height=height, width=width, batch_size=batch_size)
 
-        ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
-        diffusers_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
+        ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch], text_encoder_3=None)
+        diffusers_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch], text_encoder_3=None)
 
         for output_type in ["latent", "np", "pt"]:
+            print(output_type)
```
IlyasMoutawwakil, line 139:
```suggestion
```

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -829,24 +987,26 @@ def forward(
...
+            model_outputs["pooler_output"] = torch.from_numpy(ov_outputs[1])
+        if self.hidden_states_output_names and "last_hidden_state" not in model_outputs:
+            model_outputs["last_hidden_state"] = torch.from_numpy(ov_outputs[self.hidden_states_output_names[-1]])
+        if (
+            self.hidden_states_output_names
+            and output_hidden_states
+            or getattr(self.config, "output_hidden_states", False)
+        ):
+            hidden_states = [torch.from_numpy(ov_outputs[out_name]) for out_name in self.hidden_states_output_names]
+            model_outputs["hidden_states"] = hidden_states
```
IlyasMoutawwakil, line 1006: not sure this complexity is needed, the original code does the same thing by seperating the `torch.from_numpy` and collecting the hidden states in a list if requested, the only change you'll have to do is to accounf for the missing `self.config.num_hidden_layers` by using `self.config.num_layers` otherwise for t5-encoder https://huggingface.co/stabilityai/stable-diffusion-3-medium-diffusers/blob/main/text_encoder_3/config.json#L21

eaidova: problem that in some cases torch.from_numpy doing coping of tensors that is not good from memory usage point of view. Additonally, in torch hidden states outputs provided as list, while openvino has then flattenized, so it may lead to additional issues and unexpected keys in ModelOutput (some pipelines uses indexing for accessing outputs and indexation will not the same like in original output format). Also, there is possible issue that output with name last_hidden_state will not in outputs (due to renaming output ports binded to the same tensor it exists in output twice as last_hidden_stat and the last output from hidden_states list, in ov such tensors will have the same name and in some cases it will be hidden_states.n instead of last_hidden_state

eaidova: so I tried to make output closer to original one for avoid issues in future, but if you do not agree, please say something. that is not problem for me to use your suggested way

IlyasMoutawwakil: > please say something

I already did. but I have no issue with it if you're satisfied with it, it does the job.

tests/openvino/test_quantization.py
```diff
@@ -300,11 +302,18 @@ class OVWeightCompressionTest(unittest.TestCase):
...
         (OVStableDiffusionPipeline, "stable-diffusion", 72, 195),
         (OVStableDiffusionXLPipeline, "stable-diffusion-xl", 84, 331),
         (OVLatentConsistencyModelPipeline, "latent-consistency", 50, 135),
-    )
+    ]
+
+    if is_transformers_version(">=", "4.45.0"):
+        SUPPORTED_ARCHITECTURES_WITH_HYBRID_QUANTIZATION.extend(
+            [
+                (OVStableDiffusion3Pipeline, "stable-diffusion-3", 9, 65),
```
IlyasMoutawwakil, line 314: not adding flux here ?

tests/openvino/test_exporters_cli.py
```diff
@@ -84,13 +90,18 @@ class OVCLIExportTestCase(unittest.TestCase):
...
-    SUPPORTED_SD_HYBRID_ARCHITECTURES = (
+    SUPPORTED_SD_HYBRID_ARCHITECTURES = [
         ("stable-diffusion", 72, 195),
         ("stable-diffusion-xl", 84, 331),
         ("latent-consistency", 50, 135),
-    )
+    ]
+
+    if is_transformers_version(">=", "4.45"):
+        SUPPORTED_SD_HYBRID_ARCHITECTURES.append(("stable-diffusion-3", 9, 65))
```
IlyasMoutawwakil, line 104: not sure again but maybe forgot flux

tests/openvino/test_diffusion.py
```diff
@@ -648,12 +720,13 @@ def test_compare_to_diffusers_pipeline(self, model_arch: str):
         inputs = self.generate_inputs(height=height, width=width, batch_size=batch_size)
 
         for output_type in ["latent", "np", "pt"]:
+            print(output_type)
```
IlyasMoutawwakil, line 722:
```suggestion
```

tests/openvino/test_diffusion.py
```diff
@@ -586,7 +648,7 @@ def test_num_images_per_prompt(self, model_arch: str):
                         outputs = pipeline(**inputs, num_images_per_prompt=num_images_per_prompt).images
                         self.assertEqual(outputs.shape, (batch_size * num_images_per_prompt, height, width, 3))
 
-    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    @parameterized.expand(["stable-diffusion", "stable-diffusion-xl"])
```
IlyasMoutawwakil, line 650:
```suggestion
    @parameterized.expand(SUPPORTED_ARCHITECTURES)
```

tests/openvino/test_diffusion.py
```diff
@@ -427,16 +485,17 @@ def test_compare_to_diffusers_pipeline(self, model_arch: str):
         height, width, batch_size = 128, 128, 1
         inputs = self.generate_inputs(height=height, width=width, batch_size=batch_size)
 
-        diffusers_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
-        ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
+        diffusers_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch], text_encoder_3=None)
+        ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch], text_encoder_3=None)
```
IlyasMoutawwakil, line 488: why `text_encoder_3=None` ?

tests/openvino/test_diffusion.py
```diff
@@ -369,7 +417,7 @@ def test_num_images_per_prompt(self, model_arch: str):
                         outputs = pipeline(**inputs, num_images_per_prompt=num_images_per_prompt).images
                         self.assertEqual(outputs.shape, (batch_size * num_images_per_prompt, height, width, 3))
 
-    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    @parameterized.expand(["stable-diffusion", "stable-diffusion-xl", "latent-consistency"])
```
IlyasMoutawwakil, line 419:
```suggestion
    @parameterized.expand(SUPPORTED_ARCHITECTURES)
```

tests/openvino/test_diffusion.py
```diff
@@ -126,18 +132,18 @@ def test_compare_to_diffusers_pipeline(self, model_arch: str):
         height, width, batch_size = 128, 128, 1
         inputs = self.generate_inputs(height=height, width=width, batch_size=batch_size)
 
-        ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
-        diffusers_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
+        ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch], text_encoder_3=None)
+        diffusers_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch], text_encoder_3=None)
```
IlyasMoutawwakil, line 136: same here, i don't see a need for `text_encoder_3=None`

tests/openvino/test_quantization.py
eaidova: yes, hybrid quantization is not applicable for flux, it does not have convolutons.

tests/openvino/test_diffusion.py
eaidova: there is a bug with inference small t5 model on openvino side, on cpu it leads to segfault, while large model works. it is under investigation

tests/openvino/test_diffusion.py
eaidova: this test is not applicable for flux and sd3 as callback interface is deperecated in 0.31.0, these pipelines do not support it, so parametrization parameters correct

tests/openvino/test_diffusion.py
eaidova: see comment above

optimum/intel/openvino/modeling_diffusion.py
eaidova: > I already did.

you did that before I shared my opinion, in my perception, code review is dialog and if it requires some discussion I need to receive feedback on my answer too, because I need to know what do you think about my point of view and I need to change something after my explanation or not. That will be very helpful as it is normal that some decision may changed during discussion.

IlyasMoutawwakil: it's too much wasted time

> problem that in some cases torch.from_numpy doing coping of tensors that is not good from memory usage point of view if we will not care about necessarily or unnecessarily of wrapping specific output.

not true, check documentation https://pytorch.org/docs/stable/generated/torch.from_numpy.html, and this suggestion does the same, just in different places.

> Additonally, in torch hidden states outputs provided as grouped list, while openvino has them flattened, so it may lead to additional issues and unexpected keys in ModelOutput (some pipelines uses indexing for accessing outputs and indexation will not the same like in original output format).

I believe that's why we have model configs with inputs/outputs specifications, we also control the order.

> Also, there is possible issue that output with name last_hidden_state will not in outputs (due to renaming output ports binded to the same tensor, when it exists in output twice as last_hidden_stat and the last output from hidden_states list, in ov such tensors will have the same name and in some cases it will be hidden_states.n instead of last_hidden_state. as usage last_hidden_state is important for some pipelines, I'm trying to restore that in original format

I don't get this part, I really think last_hidden_state will be there since in order, if it's there it comes before the hidden_states list, it's the hidden_states.n that might be missing in this case but Idk this sounds like an issue that requires a minimal reproducible code, otherwise I haven't seen it before.

IlyasMoutawwakil (approved):

optimum/intel/openvino/modeling_diffusion.py
eaidova: I do not want to spend your time in long additional explanation how openvino works, all things that you said does not take into account product realization and its specific, so please trust me as an engineer that I also have some experiments and know more about potential pitfalls, not just hypothetical and theoretical assumptions

IlyasMoutawwakil:
reference for text_encoder_3 inference test skipping : #965

PR #917, opened 2024-09-27
ignore gpt neo for better transformer only for transformers>=4.45

eaidova:
# What does this PR do?

we still need to apply bettertransformer for old transformers for gpt_neo where sdpa attention is not available yet 

## Before submitting


echarlaix (approved):
Thanks for the fix!

PR #922, opened 2024-10-02
add nncf and ov nightly for  slow tests

echarlaix:


.github/workflows/test_openvino_basic.yml
```diff
@@ -22,13 +22,27 @@ jobs:
...
         include:
           - python-version: "3.12"
             os: "ubuntu-22.04"
             transformers-version: "4.36.0"
+            openvino: "openvino"
+            nncf: "nncf"
+          - python-version: "3.12"
+            os: "ubuntu-22.04"
+            transformers-version: "4.45.*"
+            openvino: "--pre -U openvino --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly"
```
eaidova, line 39: do we need somehow add openvino tokenizers version too? (they may be incompatible if we have openvino nightly and release toknizers, as the result tokenizers related tests will always fail)

echarlaix: we currently don't install `openvino-tokenizers` but can be added

echarlaix: added in https://github.com/huggingface/optimum-intel/pull/922/commits/4bb97dd1d8c47f1fa270d42825675e639d36ff1e

AlexKoff88 (approved):

AlexKoff88:
Thanks, @echarlaix.

eaidova (approved):

.github/workflows/test_openvino_basic.yml
```diff
@@ -22,13 +22,27 @@ jobs:
...
+          - python-version: "3.12"
+            os: "ubuntu-22.04"
+            transformers-version: "4.45.*"
+            openvino: "--pre -U openvino openvino-tokenizers --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly"
+            nncf: "nncf"
+          - python-version: "3.12"
+            os: "ubuntu-22.04"
+            transformers-version: "4.45.*"
+            openvino: "openvino openvino-tokenizers"
+            nncf: "git+https://github.com/openvinotoolkit/nncf.git"
```
nikita-savelyevv, line 45: As I understand, the following three version configurations will be tested:
1. OV release and NNCF release
2. OV nightly and NNCF release
3. OV release and NNCF develop

I think there should also be configuration for "OV nightly and NNCF develop" just to be 100% sure. If running time is a factor, option 2 and 3 can be replaced by this one.

cc @AlexKoff88

echarlaix: sure added in https://github.com/huggingface/optimum-intel/pull/922/commits/53acdebc1773063f5674b45e8f4922fea88961ce

PR #924, opened 2024-10-04
fix win

eaidova:
# What does this PR do?

Fixes # https://github.com/openvinotoolkit/openvino/issues/26844

force deleting openvino model and garbage collector during model export to prevent locking files in tmp dir due to mmap.
Additionally fixed installation from source on windows (in setup.py, subprocess.run raises FileNotFound error if can not find git) and refactoring weights compression part moving model size calculation during export and read model only if weights compression should be applied (based on model size or explicitly specified configuration, prevents OOM and invalid IR errors if nncf not installed)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -14,9 +14,10 @@
 
 import logging
 import os
+import shutil
 import warnings
 from pathlib import Path
-from tempfile import TemporaryDirectory, gettempdir
+from tempfile import TemporaryDirectory, gettempdir, mkdtemp
 from typing import Dict, Optional, Union
 
 import openvino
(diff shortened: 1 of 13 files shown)


optimum/intel/openvino/modeling_base.py
```diff
@@ -737,3 +737,9 @@ def _incompatible_inputs_warning(self, inputs: Dict):
                 return f"Got unexpected inputs: `{input_name}` set to {type(inputs[input_name])} while expected to be {dtype}."
 
         return None
+
+    def __del__(self):
+        if hasattr(self, "request"):
+            del self.request
+        if hasattr(self, "model"):
+            del self.model
```
nikita-savelyevv, line 745: Is this required or just to be safe?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -404,6 +404,14 @@ def _make_stateful(self):
         patch_stateful(self.config, self.model)
         self.stateful = True
 
+    def __del__(self):
+        if hasattr(self, "request"):
+            del self.request
+        if hasattr(self, "_original_model"):
+            del self._original_model
+        if hasattr(self, "model"):
+            del self.model
```
nikita-savelyevv, line 413: Same question

nikita-savelyevv (approved):
:+1:

optimum/intel/openvino/modeling_base.py
eaidova: I added for experiment to check that process exit issues will be resolved if we remove model and request in destructor. checking that attribute exists added for safety (if somebody already remove them before)

eaidova:
I still getting segmentation fault (probably caused double free memory corruption in my experiments with resolving tmp dir destruction, so I moved changes required for compression into separated PR to unblock people that waiting this issue resolution https://github.com/huggingface/optimum-intel/pull/925/

PR #925, opened 2024-10-04
fix issue with imposibility remove uncompressed model in tmp after compression

eaidova:
# What does this PR do?

Fixes # https://github.com/openvinotoolkit/openvino/issues/26844

Reworked conversion part and weight compression after it for avoiding tmp directory lock on windows and unnecessary reading model.

Added handler for tmp dir that allow ignore permission errors (natively added in python3.10, for forward-porting to <3.10, we can remove it after removing python3.9 support)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -24,10 +24,6 @@ from transformers import AutoConfig, AutoTokenizer, PreTrainedTokenizerBase
 from transformers.utils import is_torch_available
 
 from openvino.runtime import Core, save_model
-
-
-# create core globally before openvino tokenizers import necessary to prevent errors with failed extension loading
-core = Core()
 from optimum.exporters import TasksManager
 from optimum.exporters.onnx.base import OnnxConfig
 from optimum.exporters.onnx.constants import SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED
@@ -57,6 +53,8 @@ if is_torch_available():
 
 logger = logging.getLogger(__name__)
 
+# create core globally before openvino tokenizers import necessary to prevent errors with failed extension loading
+core = Core()
 
 
 def infer_task(
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -14,7 +14,6 @@
 
 import gc
 import logging
-
 import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Union
@@ -24,7 +23,8 @@ from requests.exceptions import ConnectionError as RequestsConnectionError
 from transformers import AutoConfig, AutoTokenizer, PreTrainedTokenizerBase
 from transformers.utils import is_torch_available
 
-from openvino.runtime import Core, Type, save_model
+from openvino.runtime import Core, save_model
+
 
 # create core globally before openvino tokenizers import necessary to prevent errors with failed extension loading
 core = Core()
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -14,9 +14,8 @@
 
 import gc
 import logging
-import operator
+
 import warnings
-from functools import reduce
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Union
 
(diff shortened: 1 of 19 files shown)


optimum/exporters/openvino/__main__.py
```diff
@@ -436,6 +431,8 @@ class StoreAttr(object):
         if quantization_config is None:
             continue
 
+        submodel = core.read_model(submodel_path)
```
nikita-savelyevv, line 434: Shouldn't this be moved after the following if? Otherwise model will be read even if nncf is not installed

optimum/exporters/openvino/convert.py
```diff
@@ -190,13 +196,17 @@ def export_tensorflow(
...
+        model_size = calculate_model_size(ov_model)
+
     _save_model(
         ov_model,
         output.parent / output,
         ov_config=ov_config,
         library_name=library_name,
     )
-    return input_names, output_names, True
+    return (input_names, output_names, True), model_size
```
AlexKoff88, line 209: I wonder if the `export()` method is used outside of the Optimum code. If so, this change will break backward compatibility.

AlexKoff88 (approved):

optimum/exporters/openvino/convert.py
eaidova: by default return_model_size is False, I changed to add it into output only if this parameter selected to True, so it should not break backward compatibility now

optimum/intel/openvino/utils.py
```diff
@@ -228,3 +232,275 @@ def model_has_dynamic_inputs(model):
...
+        except OSError as err:
+            onexc(os.path.islink, path, err)
+            # can't continue even if onexc hook returns
+            return
+        return _rmtree_unsafe(path, onexc)
+
+
+# copied https://github.com/python/cpython/blob/3.12/Lib/tempfile.py
+# to add behaviour that available only for python3.10+ for older python version
+class TemporaryDirectory(OrigTemporaryDirectory):
```
nikita-savelyevv, line 427: @eaidova , could you please also replace `tempfile.TemporaryDirectory` with this custom one in (all?) openvino tests?

For example, I believe the test below fails because of this.
```
FAILED tests/openvino/test_quantization.py::OVQuantizerQATest::test_automodel_static_quantization_0_hf_internal_testing_tiny_random_BertForQuestionAnswering - PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp_q_68_5j\\openvino_model.bin'
```

eaidova: ok, added

optimum/exporters/openvino/__main__.py
```diff
@@ -444,12 +439,14 @@ class StoreAttr(object):
 
         from optimum.intel.openvino.quantization import _weight_only_quantization
 
+        submodel = core.read_model(submodel_path)
```
IlyasMoutawwakil, line 442: if the model is gonna be read either ways, why nesting `return_model_sizes` and modify the behavior of all the export functions ? please explain the necessity for making export functions return model sizes as well, because it doesn't make sense for me.

eaidova: the porpose of this read model is prepare model for quantization, if model not planned to be compressed, there is no need in this step.
There are several cases for this:
1. nncf is not available
2. quantization disabled by export config (e.g. load_in_8bit=False in  from_pretrained or --weight-format fp16/fp32 in optimum-cli
3. default decision about necessary to compress model based on model size, we have rule by default apply int8 weight compression only in case if model size > 1B parameters and any additional quantization configs not provided (it is also the reason why we need calculate model size and avoid extra reading model if at this moment we already know that weight compression is not applicable)

nikita-savelyevv: I also agree that avoiding an extra model read in case the model is not going to be quantized anyway is a good thing.

nikita-savelyevv:
The PR has been stale for a while. @IlyasMoutawwakil @echarlaix do you have any additional concerns regarding this PR?

optimum/exporters/openvino/__main__.py
echarlaix: Introducing `return_model_sizes` introduces some complexity so would prefer to avoid if possible, what do you think about moving this logic (applying quantization on each submodels depending on multiple criteria) from `main_export` to `_save_model` (or `export_pytorch` / `export_pytorch_via_onnx` / `export_tensorflow`) ? Also the `ov_config` could be updated to know whether we should check the model's size (in case where quantization wasn't specified) or if the model should or shouldn't be quantized (`load_in_8bit` set to `True` of `False` for example) wdyt @eaidova ?

echarlaix (approved):

optimum/exporters/openvino/__main__.py
eaidova: no, this is also thing what we wanted to avoid, when we save ov model, pytorch model still in memory and this memory can not be deallocated until we do not return into place where it was loaded (export_main) that lead to extra memory consumption

eaidova:
@echarlaix  @IlyasMoutawwakil  could you please merge these changes? issue with inability to remove model from tmp dir is blocking several customer requests, thanks

PR #931, opened 2024-10-09
Add $ to the end of filename regex patterns

tomaarsen:
# What does this PR do?
* Add $ to the end of filename regex patterns
* add an unrelated test for a model with modeling files in a subfolder and configuration files in the root. This passes on main, but failed on 1.19.0. As expected, if `subfolder="openvino"`, then the OV files can be seen, if `subfolder=""`, then they are not.

## Before submitting

---

- Tom Aarsen


echarlaix (approved):
Looks good thanks @tomaarsen !

PR #933, opened 2024-10-09
Allow loading local models in a subfolder (with config in root)

tomaarsen:
# What does this PR do?
* Allow loading local models in a subfolder, also when the configuration is in root

Note: This PR depends on https://github.com/huggingface/optimum/pull/2044 to be merged & released, and would require increasing the `optimum` version in the dependencies. I'm not sure how the `optimum` vs `optimum-...` dependencies are handled, so I'm not sure if this is a viable PR.

You can consider this a sister-PR to https://github.com/huggingface/optimum/pull/2044 - the changes were also inspired from it. 

## What didn't work before?

```python
import tempfile
from pathlib import Path

from huggingface_hub import HfApi
from optimum.intel import OVModelForFeatureExtraction

model_id = "sentence-transformers-testing/stsb-bert-tiny-openvino"

# hub model
ov_model = OVModelForFeatureExtraction.from_pretrained(model_id, subfolder="openvino", file_name="openvino_model.xml")
print("Hub model:", ov_model)

# local model
api = HfApi()
with tempfile.TemporaryDirectory() as tmpdirname:
    local_dir = Path(tmpdirname) / "model"
    api.snapshot_download(repo_id=model_id, local_dir=local_dir)

    ov_model = OVModelForFeatureExtraction.from_pretrained(local_dir, subfolder="openvino", file_name="openvino_model.xml")
    print("Remote model:", ov_model)

```
The second model from this snippet would not be loaded, because the `subfolder` was not being inserted between the model path and the filename. https://github.com/huggingface/optimum/pull/2044 fixes the same issue but for ONNX.

## Before submitting

cc @echarlaix 

- Tom Aarsen

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -512,7 +512,7 @@ class OVBaseModel(OptimizedModel):
         # locates a file in a local folder and repo, downloads and cache it if necessary.
         model_path = Path(model_path)
         if model_path.is_dir():
-            model_cache_path = model_path / file_name
+            model_cache_path = model_path / subfolder / file_name
         else:
             file_name = Path(file_name)
             if file_name.suffix != ".onnx":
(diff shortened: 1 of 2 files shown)


echarlaix (approved):
LGTM, thanks @tomaarsen

PR #939, opened 2024-10-10
Add tests for latest transformers release

echarlaix:


tests/openvino/test_modeling.py
```diff
@@ -820,6 +819,10 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
             "mistral-nemo",
         )
 
+    # custom modeling defined in https://huggingface.co/katuni4ka/tiny-random-codegen2 differs from transformers after v4.45 resulting in unadapted patching
+    if is_transformers_version("<", "4.45.0"):
+        SUPPORTED_ARCHITECTURES += ("codegen2",)
```
echarlaix, line 824: cc @eaidova to keep the tests we would need to update the model

PR #941, opened 2024-10-11
allow to use SDPA in clip models

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -24,6 +24,8 @@ from optimum.exporters.onnx.config import OnnxConfig, TextDecoderOnnxConfig, Tex
 from optimum.exporters.onnx.model_configs import (
     CLIPOnnxConfig,
     CLIPTextOnnxConfig,
+    CLIPTextWithProjectionOnnxConfig,
+    CLIPVisionModelOnnxConfig,
     CodeGenOnnxConfig,
     FalconOnnxConfig,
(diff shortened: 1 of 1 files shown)


AlexKoff88 (approved):

echarlaix (commented):
Thanks @eaidova ! Opened https://github.com/huggingface/optimum/pull/2066 to remove `CLIPModelPatcher` direclty in optimum (shouldn't be needed for onnx), would that works for you ?

eaidova:
> Thanks @eaidova ! Opened [huggingface/optimum#2066](https://github.com/huggingface/optimum/pull/2066) to remove `CLIPModelPatcher` direclty in optimum (shouldn't be needed for onnx), would that works for you ?

@echarlaix  thanks, but are you sure that it is not need? as I understand @IlyasMoutawwakil met integration issue for onnx export that is why it was added

echarlaix:
> @echarlaix thanks, but are you sure that it is not need? as I understand @IlyasMoutawwakil met integration issue for onnx export that is why it was added

Actually might be needed for diffusers model, let's merge this PR for now and we can always revert changes if updated in optimum

echarlaix (approved):

IlyasMoutawwakil:
the specific onnx error was https://github.com/pytorch/pytorch/issues/135615 and it was fixed in latest pytorch https://github.com/pytorch/pytorch/issues/135615

PR #945, opened 2024-10-11
refine class IPEXPagedCache's update method

kaixuanliu:


kaixuanliu:
with bs=1, num_beams=4, input_token=128 and output_token=128, after this optimization, 1st token latency improved from 250.07ms to 109.86 ms; 2nd+ token latency keep the same(36.87 ms)

PR #953, opened 2024-10-16
Install torchvision CPU in OpenVINO notebook tests

helena-intel:
Currently torch CPU version is installed, but torchvision CUDA version (as dependency of timm, which is a dependency of optimum-intel[tests]). This leads to issues. This tiny PR fixes that.
Could also be fixed by using --extra-index-url everywhere, but this method is a bit safer and clearer.


echarlaix (approved):
Thanks @helena-intel !

PR #955, opened 2024-10-18
Added notebook to showcase quantization of Sentence Transformers model

AlexKoff88:


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -765,6 +765,10 @@ def ov_config(self) -> OVConfig:
     def dtype(self) -> torch.dtype:
         return OV_TO_PT_TYPE[self.ov_config.get("dtype", "f32")]
 
+    def modules(self):
+        return {}
+
+
```
AlexKoff88, line 771:
```suggestion
```

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -842,6 +842,7 @@ def forward(
         return ModelOutput(**model_outputs)
 
 
+
```
AlexKoff88, line 845:
```suggestion
```

notebooks/openvino/sentence_transformer_quantization.ipynb
```diff
@@ -0,0 +1,607 @@
...
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from transformers import Pipeline\n",
+    "import torch.nn.functional as F\n",
+    "import torch\n",
+    "\n",
+    "\n",
+    "# copied from the model card\n",
```
eaidova, line 217: can you please add link on model card from which copied?

notebooks/openvino/sentence_transformer_quantization.ipynb
```diff
@@ -0,0 +1,607 @@
...
+   "source": [
+    "%pip install optimum[openvino]\n",
+    "%pip install evaluate"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Quantize staticly model to 8-bit with NNCF via Optimum-Intel API"
```
eaidova, line 24: possibly adding some description about how to prepare dataset and how to define quantization config should be helpful.

eaidova (approved):

notebooks/openvino/sentence_transformer_quantization.ipynb
AlexKoff88: Copied from the one that is used in the notebook. Put this into the comment

notebooks/openvino/sentence_transformer_quantization.ipynb
AlexKoff88: Done

notebooks/openvino/sentence_transformer_quantization.ipynb
```diff
@@ -0,0 +1,623 @@
...
+   ],
+   "source": [
+    "default_acc = metric.compute(predictions=result[\"default\"], references=result[\"reference\"])\n",
+    "quantized = metric.compute(predictions=result[\"quantized\"], references=result[\"reference\"])\n",
+    "\n",
+    "print(\"vanilla model: pearson=\", default_acc[\"pearson\"])\n",
+    "print(\"quantized model: pearson=\", quantized[\"pearson\"])\n",
+    "print(\n",
+    "    \"The quantized model achieves \",\n",
+    "    round(quantized[\"pearson\"] / default_acc[\"pearson\"], 2) * 100,\n",
```
helena-intel, line 373:
```suggestion
    "    round((quantized["pearson"] / default_acc["pearson"]) * 100,2),\n",
```
Just a suggestion, but IMO it's better to round after converting to percentage.

notebooks/openvino/sentence_transformer_quantization.ipynb
```diff
@@ -0,0 +1,623 @@
...
+      "Using framework PyTorch: 2.4.1+cpu\n",
+      "Overriding 1 configuration item(s)\n",
+      "\t- use_cache -> False\n",
+      "Compiling the model to CPU ...\n",
+      "Compiling the model to CPU ...\n"
+     ]
+    }
+   ],
+   "source": [
+    "model = OVModelForFeatureExtraction.from_pretrained(MODEL_ID)\n",
```
helena-intel, line 282: This model is already loaded, but I like having standalone snippets, so I think no problem to load it again, but to prevent exporting it again (and for equivalence with the int8 model) it could be loaded from the model save dir "all-MiniLM-L6-v2".

notebooks/openvino/sentence_transformer_quantization.ipynb
```diff
@@ -0,0 +1,623 @@
...
+    "import datasets\n",
+    "from transformers import AutoTokenizer\n",
+    "from optimum.intel import OVModelForFeatureExtraction, OVQuantizer, OVQuantizationConfig, OVConfig\n",
+    "\n",
+    "MODEL_ID = \"sentence-transformers/all-MiniLM-L6-v2\"\n",
+    "\n",
+    "model = OVModelForFeatureExtraction.from_pretrained(MODEL_ID)\n",
+    "model.save_pretrained(\"all-MiniLM-L6-v2\")\n",
+    "\n",
+    "tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)\n",
```
helena-intel, line 182:
```suggestion
    "model.save_pretrained(\"all-MiniLM-L6-v2\")\n",
    "\n",
    "tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)\n",
    "tokenizer.save_pretrained(\"all-MiniLM-L6-v2\")\n",
```
Not necessary to run the notebook, but if you want to load this model from disk you always need the tokenizer so IMO it is nice to save them together.

notebooks/openvino/sentence_transformer_quantization.ipynb
```diff
@@ -0,0 +1,623 @@
...
+    "    round(quantized[\"pearson\"] / default_acc[\"pearson\"], 2) * 100,\n",
+    "    \"% accuracy of the fp32 model\",\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Compare performance of the baselein and INT8 models"
```
helena-intel, line 382:
```suggestion
    "## Compare performance of the baseline and INT8 models"
```

notebooks/openvino/sentence_transformer_quantization.ipynb
```diff
@@ -0,0 +1,623 @@
...
+      "[ INFO ]    Average:       9.77 ms\n",
+      "[ INFO ]    Min:           9.54 ms\n",
+      "[ INFO ]    Max:           11.35 ms\n",
+      "[ INFO ] Throughput:   100.56 FPS\n"
+     ]
+    }
+   ],
+   "source": [
+    "# FP32 baseline model\n",
+    "!benchmark_app -m all-MiniLM-L6-v2/openvino_model.xml -shape \"input_ids[1,384],attention_mask[1,384],token_type_ids[1,384]\" -api sync -niter 200"
```
helena-intel, line 493: This reshapes the model to static shapes, which has great improvement for especially INT8. But most people will not use static shapes in practice, and padding/truncating to 384 is not always desired. IMO it is fairer to compare performance by looping over a dataset (e.g. modifying the evaluate function to add timings) but then there is not as much of a performance difference. If we keep benchmark_app, it would be good to at least explain the static shapes. (Using data_shape instead of shape in benchmark_app does not reshape the model, but you still use the same shape length everywhere, so still not a standard use case)

helena-intel (approved):
Thanks @AlexKoff88 this is a great example.

notebooks/openvino/sentence_transformer_quantization.ipynb
AlexKoff88: Thanks, Helena. I would not agree in this specific case as the tokenizer truncates data anyway so it is about static shape. But I can add information about it.

l-bat:
Why is the `squad` dataset needed?
```python
DATASET_NAME = "squad"
dataset = datasets.load_dataset(DATASET_NAME)
```

AlexKoff88:
> Why is the `squad` dataset needed?
> 
> ```python
> DATASET_NAME = "squad"
> dataset = datasets.load_dataset(DATASET_NAME)
> ```

Thanks @l-bat. Fixed

echarlaix (approved):
Looks great, thanks for the addition @AlexKoff88. Could also be added to https://github.com/huggingface/optimum-intel/blob/v1.20.0/notebooks/openvino/README.md

AlexKoff88:
> Looks great, thanks for the addition @AlexKoff88. Could also be added to https://github.com/huggingface/optimum-intel/blob/v1.20.0/notebooks/openvino/README.md

will do in the follow-up PR

PR #967, opened 2024-10-24
fix device selection for compilation language model in vlm and model saving

eaidova:
# What does this PR do?

lefotovers for OVModelForVisulCausalLM
* device changing in runtime with model.to()
* compile only mode
* downloading preconverted model from hub

all provided functionality converted by tests

## Before submitting


optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -285,7 +296,7 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
         dst_file_names = [
             "openvino_language_model.xml",
             "openvino_text_embeddings_model.xml",
-            "openvino_vision_embeddings.xml",
+            "openvino_vision_embeddings_model.xml",
```
echarlaix, line 299: could be all added to https://github.com/huggingface/optimum-intel/blob/cead516b3b8c38c1a0b8b8937a861edccdfab839/optimum/intel/openvino/utils.py#L53 to avoid potential issues

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -54,14 +54,25 @@ def __init__(
...
+                self.request = self._compile_model(
+                    self.model, self._device, self.ov_config, self.model_save_dir
+                ).create_infer_request()
         self._compile_text_emb()
 
     def _compile_text_emb(self):
         if self.text_emb_request is None:
             logger.info(f"Compiling the Text embeddings model to {self._device} ...")
-            self.text_emb_request = core.compile_model(self.text_emb_model, self._device, self.ov_config)
+            if self._compile_only:
```
echarlaix, line 69: could you add a test for the `compile_only=True` case ? also we would need a test to make sure we don't break compatibility with a model that was already exported https://github.com/huggingface/optimum-intel/blob/v1.20.1/tests/openvino/test_modeling.py#L129

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -54,14 +54,25 @@ def __init__(
...
         if self.request is None:
-            logger.info(f"Compiling the Language model to {self._device} ...")
-            self.request = core.compile_model(self.model, self._device, self.ov_config).create_infer_request()
+            if self._compile_only:
+                self.request = self.model.create_infer_request()
+            else:
+                logger.info(f"Compiling the Language model to {self._device} ...")
+                self.request = self._compile_model(
+                    self.model, self._device, self.ov_config, self.model_save_dir
+                ).create_infer_request()
```
echarlaix, line 63: why can't we call https://github.com/huggingface/optimum-intel/blob/cead516b3b8c38c1a0b8b8937a861edccdfab839/optimum/intel/openvino/modeling_decoder.py#L396-L401 here ?

echarlaix (commented):
Thanks for the fixes!

optimum/intel/openvino/modeling_visual_language.py
echarlaix: not related to this PR but I find having `compile_model` / `compile` / `_compile` / `_compile_text_emb` makes it super difficult to follow everything that we are doing here, not sure how we can make this clearer

optimum/intel/openvino/modeling_visual_language.py
eaidova: we have no such models yet but ok I'll add

optimum/intel/openvino/modeling_visual_language.py
eaidova: moved

optimum/intel/openvino/modeling_visual_language.py
eaidova: added tests fr downloading from hub, compile_nly and also added tests for export

optimum/intel/openvino/modeling_visual_language.py
eaidova: for the first comment, I reused compile for decoder, for the second one, I do not know what I can suggest for now. The problem that we have 2 requests in model (for llm part and separated for embeddings. Adding one more level and separated model may be solution for having own compilation method, but it is also I think may make model usage complicated as there is no possibility to use models separately

tests/openvino/test_modeling.py
```diff
@@ -1932,13 +2002,50 @@ def test_compare_to_transformers(self, model_arch):
             self.assertIsInstance(getattr(ov_model, additional_part), MODEL_PARTS_CLS_MAPPING[additional_part])
         self.assertIsInstance(ov_model.config, PretrainedConfig)
         inputs = ov_model.preprocess_inputs(**preprocessors, text=prompt, image=self.IMAGE.resize((600, 600)))
-        # pytorch minicpmv and internvl are not designed to be used via forward
+        transformers_inputs = copy.deepcopy(inputs)
+        ov_model.to("AUTO")
+        self.assertTrue("AUTO" in ov_model._device)
```
AlexKoff88, line 2007: I wonder why you don't do the exact match here?

tests/openvino/test_modeling.py
```diff
@@ -2079,6 +2186,19 @@ def get_preprocessors(self, model_arch):
             preprocessors = {"processor": processor, "tokenizer": None}
         return preprocessors
 
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    def test_model_can_be_loaded_after_saving(self, model_arch):
```
AlexKoff88, line 2190: What caused the need for this check. Can it be the case for Diffusion models as well?

AlexKoff88 (approved):

tests/openvino/test_modeling.py
```diff
@@ -1932,13 +2002,50 @@ def test_compare_to_transformers(self, model_arch):
...
+        ov_model.clear_requests()
+        self.assertTrue("CPU" in ov_model._device)
+        self.assertTrue("CPU" in ov_model.vision_embeddings._device)
+        self.assertTrue(ov_model.vision_embeddings.request is None)
+        self.assertTrue("CPU" in ov_model.language_model._device)
+        self.assertTrue(ov_model.language_model.request is None)
+        self.assertTrue(ov_model.language_model.text_emb_request is None)
+        for additional_part in ov_model.additional_parts:
+            self.assertTrue("CPU" in getattr(ov_model, additional_part)._device)
+            self.assertTrue(getattr(ov_model, additional_part).request is None)
```
eaidova, line 2036:
```suggestion
        self.asserEqual(ov_model._device, "AUTO")
        self.assertEqualov_model.vision_embeddings._device, "AUTO")
        self.assertTrue(ov_model.vision_embeddings.request is None)
        self.assertEqual(ov_model.language_model._device, "AUTO")
        self.assertTrue(ov_model.language_model.request is None)
        self.assertTrue(ov_model.language_model.text_emb_request is None)
        for additional_part in ov_model.additional_parts:
            self.assertEqual(getattr(ov_model, additional_part)._device, "AUTO")
            self.assertTrue(getattr(ov_model, additional_part).request is None)
        ov_model.to("CPU")
        ov_model.compile()
        self.assertEqual(ov_model._device, "CPU")
        self.assertEqual(ov_model.vision_embeddings._device, "CPU")
        self.assertTrue(ov_model.vision_embeddings.request is not None)
        self.assertEqual(ov_model.language_model._device, "CPU")
        self.assertTrue(ov_model.language_model.request is not None)
        self.assertTrue(ov_model.language_model.text_emb_request is not None)
        for additional_part in ov_model.additional_parts:
            self.assertEqual(getattr(ov_model, additional_part)._device, "CPU")
            self.assertTrue(getattr(ov_model, additional_part).request is not None)
        ov_model.clear_requests()
        self.assertEqual(ov_model._device, "CPU")
        self.assertEqual(ov_model.vision_embeddings._device, "CPU")
        self.assertTrue(ov_model.vision_embeddings.request is None)
        self.assertEqual(ov_model.language_model._device, "CPU")
        self.assertTrue(ov_model.language_model.request is None)
        self.assertTrue(ov_model.language_model.text_emb_request is None)
        for additional_part in ov_model.additional_parts:
            self.assertEqual(getattr(ov_model, additional_part)._device, "CPU")
            self.assertTrue(getattr(ov_model, additional_part).request is None)
```

tests/openvino/test_modeling.py
eaidova: previously we had typo in model name for saving via save_pretrained is why I added this test. I suppose we have the same tests for diffusers (at least recently I added similar test for checking having safety_checker after loading)

tests/openvino/test_modeling.py
eaidova: really why? :) changed

tests/openvino/test_modeling.py
```diff
@@ -1932,13 +2002,50 @@ def test_compare_to_transformers(self, model_arch):
             self.assertIsInstance(getattr(ov_model, additional_part), MODEL_PARTS_CLS_MAPPING[additional_part])
         self.assertIsInstance(ov_model.config, PretrainedConfig)
         inputs = ov_model.preprocess_inputs(**preprocessors, text=prompt, image=self.IMAGE.resize((600, 600)))
-        # pytorch minicpmv and internvl are not designed to be used via forward
+        transformers_inputs = copy.deepcopy(inputs)
+        ov_model.to("AUTO")
+        self.asserEqual(ov_model._device, "AUTO")
```
eaidova, line 2007:
```suggestion
        self.assertEqual(ov_model._device, "AUTO")
```

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -421,14 +441,20 @@ def _from_pretrained(
...
+            "vision_embeddings": OV_VISION_EMBEDDINGS_MODEL_NAME,
+            "vision_embeddings_bin": OV_VISION_EMBEDDINGS_MODEL_NAME.replace(".xml", ".bin"),
         }
 
         model_cls = MODEL_TYPE_TO_CLS_MAPPING[config.model_type]
         for part in model_cls.additional_parts:
             model_file_names[part] = f"openvino_{part}_model.xml"
+            model_file_names[part + "_bin"] = f"openvino_{part}_model.bin"
+        model_cls = MODEL_TYPE_TO_CLS_MAPPING[config.model_type]
+        quantization_config = model_cls._prepare_weight_quantization_config(quantization_config, load_in_8bit)
```
nikita-savelyevv, line 457:
```suggestion
            model_file_names[part + "_bin"] = f"openvino_{part}_model.bin"
        model_cls = MODEL_TYPE_TO_CLS_MAPPING[config.model_type]
```

tests/openvino/test_modeling.py
```diff
@@ -223,6 +231,68 @@ def test_load_from_hub_and_save_decoder_model(self, use_cache):
...
+        is_transformers_version("<", "4.45"),
+        "model tokenizer exported with tokenizers 0.20 is not compatible with old transformers",
+    )
+    def test_load_from_hub_and_save_visual_language_model(self):
+        model_id = self.OV_VLM_MODEL_ID
+        processor = get_preprocessor(model_id)
+        prompt = "<image>\n What is shown in this image?"
+        image = Image.open(
+            requests.get(
+                "http://images.cocodataset.org/val2017/000000039769.jpg",
```
nikita-savelyevv, line 244: I see this url is used multiple times. What do you think about creating some global variable for it?

tests/openvino/test_modeling.py
```diff
@@ -223,6 +231,68 @@ def test_load_from_hub_and_save_decoder_model(self, use_cache):
...
+                stream=True,
+            ).raw
+        )
+        loaded_model = OVModelForVisualCausalLM.from_pretrained(model_id)
+        self.assertIsInstance(loaded_model.config, PretrainedConfig)
+        self.assertIsInstance(loaded_model, MODEL_TYPE_TO_CLS_MAPPING[loaded_model.config.model_type])
```
nikita-savelyevv, line 250:
```suggestion
        loaded_model = OVModelForVisualCausalLM.from_pretrained(model_id)
        self.assertIsInstance(loaded_model, MODEL_TYPE_TO_CLS_MAPPING[loaded_model.config.model_type])
```

Is checked below one more time

tests/openvino/test_modeling.py
```diff
@@ -223,6 +231,68 @@ def test_load_from_hub_and_save_decoder_model(self, use_cache):
...
+            self.assertTrue(hasattr(loaded_model, additional_part))
+            self.assertIsInstance(getattr(loaded_model, additional_part), MODEL_PARTS_CLS_MAPPING[additional_part])
+        self.assertIsInstance(loaded_model.config, PretrainedConfig)
+        # Test that PERFORMANCE_HINT is set to LATENCY by default
+        self.assertEqual(loaded_model.ov_config.get("PERFORMANCE_HINT"), "LATENCY")
+        self.assertEqual(
+            loaded_model.language_model.request.get_compiled_model().get_property("PERFORMANCE_HINT"), "LATENCY"
+        )
+        self.assertEqual(loaded_model.language_model.text_emb_request.get_property("PERFORMANCE_HINT"), "LATENCY")
+        self.assertEqual(loaded_model.vision_embeddings.request.get_property("PERFORMANCE_HINT"), "LATENCY")
```
nikita-savelyevv, line 263: Should performance hint for model parts also be checked?

tests/openvino/test_modeling.py
```diff
@@ -223,6 +231,68 @@ def test_load_from_hub_and_save_decoder_model(self, use_cache):
...
+        with TemporaryDirectory() as tmpdirname:
+            loaded_model.save_pretrained(tmpdirname)
+            folder_contents = os.listdir(tmpdirname)
+            for xml_file_name in [
+                OV_LANGUAGE_MODEL_NAME,
+                OV_TEXT_EMBEDDINGS_MODEL_NAME,
+                OV_VISION_EMBEDDINGS_MODEL_NAME,
+            ]:
+                self.assertTrue(xml_file_name in folder_contents)
+                self.assertTrue(xml_file_name.replace(".xml", ".bin") in folder_contents)
```
nikita-savelyevv, line 277: Should model parts also be checked?

tests/openvino/test_modeling.py
```diff
@@ -223,6 +231,68 @@ def test_load_from_hub_and_save_decoder_model(self, use_cache):
...
+            model = OVModelForVisualCausalLM.from_pretrained(tmpdirname)
+            compile_only_model = OVModelForVisualCausalLM.from_pretrained(tmpdirname, compile_only=True)
+            self.assertIsInstance(compile_only_model.language_model.model, ov.runtime.CompiledModel)
+            self.assertIsInstance(compile_only_model.language_model.request, ov.runtime.InferRequest)
+            self.assertIsInstance(compile_only_model.language_model.text_emb_model, ov.runtime.CompiledModel)
+            self.assertIsInstance(compile_only_model.language_model.text_emb_request, ov.runtime.CompiledModel)
+            self.assertIsInstance(compile_only_model.vision_embeddings.model, ov.runtime.CompiledModel)
+            self.assertIsInstance(compile_only_model.vision_embeddings.request, ov.runtime.CompiledModel)
```
nikita-savelyevv, line 285: Should model parts also be checked?

tests/openvino/test_modeling.py
```diff
@@ -1932,13 +2002,50 @@ def test_compare_to_transformers(self, model_arch):
...
+        ov_model.to("AUTO")
+        self.assertEqual(ov_model._device, "AUTO")
+        self.assertEqual(ov_model.vision_embeddings._device, "AUTO")
+        self.assertTrue(ov_model.vision_embeddings.request is None)
+        self.assertEqual(ov_model.language_model._device, "AUTO")
+        self.assertTrue(ov_model.language_model.request is None)
+        self.assertTrue(ov_model.language_model.text_emb_request is None)
+        for additional_part in ov_model.additional_parts:
+            self.assertEqual(getattr(ov_model, additional_part)._device, "AUTO")
+            self.assertTrue(getattr(ov_model, additional_part).request is None)
```
nikita-savelyevv, line 2015: There are quite a few repeating checks like this added in the PR. What do you think about replacing them with for-loops? Possibly introducing some helper property for `OVModelForVisualCausalLM` like `.submodels` (or something like this) will be helpful. I think it also will be helpful inside some of the `OVModelForVisualCausalLM` methods, like `.half()` for example.

tests/openvino/test_modeling.py
eaidova: reworked

tests/openvino/test_modeling.py
eaidova: reworked

tests/openvino/test_modeling.py
eaidova: reworked

tests/openvino/test_modeling.py
eaidova: added components for model wrappers and submodels for ov.Model objects, please take a look

tests/openvino/test_modeling.py
eaidova: moved to utils TEST_IMAGE_URL

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -601,17 +652,14 @@ def half(self):
...
-                apply_moc_transformations(model, cf=False)
-                compress_model_transformation(model)
+        for _, submodel in self.submodels.items():
+            apply_moc_transformations(submodel, cf=False)
+            compress_model_transformation(submodel)
+        return self
+
+    def to(self, device):
+        self.language_model.to(device)
+        super().to(device)
```
nikita-savelyevv, line 662: Missed this during previous review round. Is it on purpose that only language model device is changed?

tests/openvino/test_modeling.py
```diff
@@ -223,6 +229,77 @@ def test_load_from_hub_and_save_decoder_model(self, use_cache):
...
+
+        for component_name, component in loaded_model.components.items():
+            if component_name == "language_model":
+                self.assertIsInstance(component.model, ov.Model)
+                self.assertEqual(component.request.get_compiled_model().get_property("PERFORMANCE_HINT"), "LATENCY")
+                self.assertIsInstance(component.text_emb_model, ov.Model)
+                self.assertEqual(component.text_emb_request.get_property("PERFORMANCE_HINT"), "LATENCY")
+            else:
+                self.assertIsInstance(component.model, ov.Model)
+                self.assertEqual(component.request.get_property("PERFORMANCE_HINT"), "LATENCY")
```
nikita-savelyevv, line 262:
```suggestion
        for component_name, component in loaded_model.components.items():
            self.assertIsInstance(component.model, ov.Model)
            if component_name == "language_model":
                self.assertEqual(component.request.get_compiled_model().get_property("PERFORMANCE_HINT"), "LATENCY")
                self.assertIsInstance(component.text_emb_model, ov.Model)
                self.assertEqual(component.text_emb_request.get_property("PERFORMANCE_HINT"), "LATENCY")
            else:
                self.assertEqual(component.request.get_property("PERFORMANCE_HINT"), "LATENCY")
```

tests/openvino/test_modeling.py
nikita-savelyevv: Looks good!

nikita-savelyevv (approved):

optimum/intel/openvino/modeling_visual_language.py
eaidova: language model required special behaviour for setting the device as it contains 2 models and in some cases, potentially, it is possible to use it independently, other parts inherit device from base model and their update converted bysuper().to (change device in base model and call clear_requests

AlexKoff88: `device_map`?

AlexKoff88: Anyway, I would recommend doing this in the follow-up PR.

PR #968, opened 2024-10-24
fix conversion for text embeddings for fp16 models

eaidova:
# What does this PR do?

text embedding part in vlm models if original model loaded in bf16/fp16 without that changes exported in onnx due to overriding input signature by openvino pytorch frontend

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -389,7 +389,10 @@ def export_pytorch(
                 if patch_16bit_model:
                     from openvino.frontend.pytorch.patch_model import __make_16bit_traceable
 
+                    # frontend may riases confusing warnings about modules already patched if model splitted on several parts
+                    logging.disable(logging.WARNING)
                     __make_16bit_traceable(model)
+                    logging.disable(logging.NOTSET)
                 check_dummy_inputs_are_allowed(model, dummy_inputs)
                 input_info = _get_input_info(model, config, dummy_inputs)
                 ov_model = convert_model(
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -389,10 +389,7 @@ def export_pytorch(
                 if patch_16bit_model:
                     from openvino.frontend.pytorch.patch_model import __make_16bit_traceable
 
-                    # frontend may riases confusing warnings about modules already patched if model splitted on several parts
-                    logging.disable(logging.WARNING)
                     __make_16bit_traceable(model)
-                    logging.disable(logging.NOTSET)
                 check_dummy_inputs_are_allowed(model, dummy_inputs)
                 input_info = _get_input_info(model, config, dummy_inputs)
                 ov_model = convert_model(
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -2765,6 +2765,7 @@ class FluxTransfromerModelPatcher(ModelPatcher):
             self._model.pos_embed.forward = self._model.pos_embed._orig_forward
 
 
+<<<<<<< HEAD
 def _minicpmv_resampler_forward(self, image_feature, pos_embed, key_padding_mask):
     bs = image_feature.shape[0]
     image_feature = self.kv_proj(image_feature)  # B * L * D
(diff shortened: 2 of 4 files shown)


AlexKoff88 (approved):

optimum/exporters/openvino/convert.py
```diff
@@ -389,7 +389,10 @@ def ts_patched_forward(*args, **kwargs):
                 if patch_16bit_model:
                     from openvino.frontend.pytorch.patch_model import __make_16bit_traceable
 
+                    # frontend may riases confusing warnings about modules already patched if model splitted on several parts
```
nikita-savelyevv, line 392:
```suggestion
                    # frontend may raise confusing warnings about modules already patched if model is split into several parts
```

optimum/exporters/openvino/convert.py
```diff
@@ -389,7 +389,10 @@ def ts_patched_forward(*args, **kwargs):
...
+                    logging.disable(logging.WARNING)
                     __make_16bit_traceable(model)
+                    logging.disable(logging.NOTSET)
```
nikita-savelyevv, line 395: What if some levels were already disabled? After this they'll become enabled back.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2991,3 +2991,25 @@ def __init__(
...
+
+
+class InputEmbeddingPatcher(ModelPatcher):
+    def __init__(
+        self,
+        config: "OnnxConfig",
+        model: Union["PreTrainedModel", "TFPreTrainedModel"],
+        model_kwargs: Dict[str, Any],
+    ):
+        # making 16bit tracable overrides embeedings input signature these changes required to prevent this issue
```
nikita-savelyevv, line 3003: Could you please move the comment to where the patcher is called? Also, I don't quite understand what specifically in the original signature leads to problems, so rephrasing might help.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2991,3 +2991,25 @@ def __init__(
     def __exit__(self, exc_type, exc_value, traceback):
         super().__exit__(exc_type, exc_value, traceback)
         self._model.forward = self._model.__orig_forward
+
+
+class InputEmbeddingPatcher(ModelPatcher):
```
nikita-savelyevv, line 2996: This seems like a quite general patcher. Its name probably should not reflect that it is currently applied to input embeddings, but that it patches forward method to have a single input.

optimum/exporters/openvino/convert.py
nikita-savelyevv, line 395: Is it possible that with this not only irrelevant warnings will be hidden, but the useful ones too? Is there a way to filter specifically those somehow?

optimum/exporters/openvino/model_patcher.py
eaidova: do you mean on model configs level? in the moment patch applying, we can not identify which one of them used, only in config or in patcher class itself

optimum/exporters/openvino/model_patcher.py
eaidova: I do not think that it can be generalized to other modules, because it requires to know original signature in  other modules forward it can be different, so current patcher implementation will not fit to these purpoceses. I do not want to make it generic too, because there is opportunity to fix it on openvino frontend level

optimum/exporters/openvino/convert.py
eaidova: this function does not produce any additional warnings

optimum/exporters/openvino/model_patcher.py
nikita-savelyevv: I mean here https://github.com/eaidova/optimum-intel/blob/ea/fix_16bit_embeddings/optimum/exporters/openvino/model_configs.py#L1271

optimum/exporters/openvino/convert.py
eaidova: good question on which I can not find answer. Also looks like warning was replaced by debug info, so in 2024.5 there is no need to have disabling https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/src/openvino/frontend/pytorch/patch_model.py#L79

optimum/exporters/openvino/model_patcher.py
eaidova: moved

tests/openvino/utils_tests.py
```diff
@@ -84,7 +84,7 @@
     "marian": "sshleifer/tiny-marian-en-de",
     "mbart": "hf-internal-testing/tiny-random-mbart",
     "minicpm": "katuni4ka/tiny-random-minicpm",
-    "minicpmv": "katuni4ka/tiny-random-minicpmv-2_6",
+    "minicpmv": "/home/ea/work/my_optimum_intel/optimum-intel/tiny-random-minicpmv-2_6",
```
AlexKoff88, line 87: @eaidova, can you please fix this?

eaidova: oops... yes, sure :)

eaidova, line 87:
```suggestion
    "minicpmv": "katuni4ka/tiny-random-minicpmv-2_6",
```

PR #970, opened 2024-10-24
fix switching between legacy and new processing for llava

eaidova:
# What does this PR do?

currently the second token is always processed as legacy due to limitation with providing info between iterations that lead on perf issues on some platforms
Migration on usage new preprocessing significantly speedup inference on iGPU (>1s  for the second token to 200ms for llava-1.5-7b-hf in fp32)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -697,6 +697,33 @@ class OVModelForVisualCausalLM(OVBaseModel, GenerationMixin):
 
 
 class _OVLlavaForCausalLM(OVModelForVisualCausalLM):
+    def __init__(
+        self,
+        language_model: ov.Model,
+        text_embeddings: ov.Model,
+        vision_embeddings: ov.Model,
+        config: PretrainedConfig = None,
+        device: str = "CPU",
+        dynamic_shapes: bool = True,
+        ov_config: Optional[Dict[str, str]] = None,
+        model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
+        quantization_config: Union[OVWeightQuantizationConfig, Dict] = None,
+        **kwargs,
+    ):
+        super().__init__(
+            language_model=language_model,
+            text_embeddings=text_embeddings,
+            vision_embeddings=vision_embeddings,
+            config=config,
+            device=device,
+            dynamic_shapes=dynamic_shapes,
+            ov_config=ov_config,
+            model_save_dir=model_save_dir,
+            quantization_config=quantization_config,
+            **kwargs,
+        )
+        self._legacy_processing = not hasattr(self.config, "image_seq_length")
+
     def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
         if input_ids is not None and input_ids.shape[1] == 1:
             return None
@@ -731,10 +758,8 @@ class _OVLlavaForCausalLM(OVModelForVisualCausalLM):
         image_features = torch.from_numpy(vision_embeds) if isinstance(vision_embeds, np.ndarray) else vision_embeds
         inputs_embeds = torch.from_numpy(inputs_embeds) if isinstance(inputs_embeds, np.ndarray) else inputs_embeds
         if legacy_processing is None:
-            legacy_processing = (
-                not hasattr(self.config, "image_seq_length")
-                or ((input_ids == self.config.image_token_index).sum(1).max() < self.config.image_seq_length)
-                or (input_ids.shape[-1] == 1)
+            legacy_processing = not (hasattr(self.config, "image_seq_length") and (input_ids.shape[-1] == 1)) or (
+                (input_ids == self.config.image_token_index).sum(1).max() < self.config.image_seq_length
             )
 
         if legacy_processing:
@@ -815,11 +840,15 @@ class _OVLlavaForCausalLM(OVModelForVisualCausalLM):
     def get_multimodal_embeddings(
         self, input_ids, pixel_values=None, attention_mask=None, position_ids=None, past_key_values=None, **kwargs
     ):
-        legacy_processing = (
-            not hasattr(self.config, "image_seq_length")
-            or ((input_ids == self.config.image_token_index).sum(1).max() < self.config.image_seq_length)
-            or (input_ids.shape[-1] == 1 and pixel_values is not None)
-        )
+        legacy_processing = self._legacy_processing
+        inputs_embeds = self.get_text_embeddings(input_ids, **kwargs)
+
+        if pixel_values is not None and not legacy_processing and past_key_values is None:
+            legacy_processing = (input_ids == self.config.image_token_index).sum(
+                1
+            ).max() < self.config.image_seq_length
+            self._legacy_processing = legacy_processing
+
         inputs_embeds, attention_mask, position_ids = super().get_multimodal_embeddings(
             input_ids, pixel_values, attention_mask, position_ids, legacy_processing=legacy_processing, **kwargs
         )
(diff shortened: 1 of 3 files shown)


AlexKoff88 (approved):

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -780,16 +804,19 @@ def merge_vision_text_embeddings(
     def get_multimodal_embeddings(
         self, input_ids, pixel_values=None, attention_mask=None, position_ids=None, past_key_values=None, **kwargs
     ):
-        legacy_processing = (
-            not hasattr(self.config, "image_seq_length")
-            or ((input_ids == self.config.image_token_index).sum(1).max() < self.config.image_seq_length)
-            or (input_ids.shape[-1] == 1 and pixel_values is not None)
-        )
+        inputs_embeds = self.get_text_embeddings(input_ids, **kwargs)
```
nikita-savelyevv, line 807: This line has no effect

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -780,16 +804,19 @@ def merge_vision_text_embeddings(
...
+
+        if pixel_values is not None and self._support_new_processing and past_key_values is None:
+            legacy_processing = (input_ids == self.config.image_token_index).sum(
+                1
+            ).max() < self.config.image_seq_length
+            self._legacy_processing = legacy_processing
```
nikita-savelyevv, line 813: Setting instance-level variable depending on a model input does not seem right. Can we somehow avoid it?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -903,11 +930,12 @@ def get_multimodal_embeddings(
...
-        legacy_processing = (
-            not hasattr(self.config, "image_seq_length")
-            or ((input_ids == self.config.image_token_index).sum(1).max() < self.config.image_seq_length)
-            or (input_ids.shape[-1] == 1 and pixel_values is not None)
-        )
+        if pixel_values is not None and self._support_new_processing and past_key_values is None:
+            legacy_processing = (input_ids == self.config.image_token_index).sum(
+                1
+            ).max() < self.config.image_seq_length
+            self._legacy_processing = legacy_processing
```
nikita-savelyevv, line 937: Same concern as above

tests/openvino/test_modeling.py
```diff
@@ -1934,6 +1934,39 @@ def test_compare_to_transformers(self, model_arch):
...
+        # previous run was with legacy processing, one more run with features concatenation on preprocessing level
+        if (
+            model_arch in ["llava", "llava-next"]
+            and is_transformers_version(">=", "4.45")
+            and (processor.patch_size is None or processor.vision_feature_select_strategy is None)
+        ):
+            processor.patch_size = ov_model.config.vision_config.patch_size
+            processor.vision_feature_select_strategy = ov_model.config.vision_feature_select_strategy
+            if model_arch == "llava":
+                # testing model for llava does ot have specified image_seq_length and it is different from default
```
nikita-savelyevv, line 1946: This sentence does not seem right

tests/openvino/test_modeling.py
```diff
@@ -1934,6 +1934,39 @@ def test_compare_to_transformers(self, model_arch):
...
+            set_seed(SEED)
+            with torch.no_grad():
+                transformers_outputs = transformers_model(**inputs)
+            set_seed(SEED)
+            ov_outputs = ov_model(**inputs)
+            self.assertTrue(torch.allclose(ov_outputs.logits, transformers_outputs.logits, atol=1e-4))
+            set_seed(SEED)
+            ov_outputs = ov_model.generate(**inputs, generation_config=gen_config)
+            set_seed(SEED)
+            transformers_outputs = transformers_model.generate(**inputs, generation_config=gen_config)
```
nikita-savelyevv, line 1965: Shouldn't this be called inside `with torch.no_grad():` as above?

tests/openvino/test_modeling.py
```diff
@@ -1934,6 +1934,39 @@ def test_compare_to_transformers(self, model_arch):
             f"generation config : {gen_config}, transformers output {transformers_outputs}, ov_model output {ov_outputs}",
         )
 
+        # previous run was with legacy processing, one more run with features concatenation on preprocessing level
+        if (
+            model_arch in ["llava", "llava-next"]
```
nikita-savelyevv, line 1939:
```suggestion
            model_arch in ["llava", "llava_next"]
```

tests/openvino/test_modeling.py
```diff
@@ -1934,6 +1934,39 @@ def test_compare_to_transformers(self, model_arch):
             f"generation config : {gen_config}, transformers output {transformers_outputs}, ov_model output {ov_outputs}",
         )
 
+        # previous run was with legacy processing, one more run with features concatenation on preprocessing level
+        if (
```
nikita-savelyevv, line 1938: Above there are quite a lot of manipulations done to `ov_model` and `transformers_model`. Won't they interfere with the "new" inference here? Maybe better to have two separate test methods for legacy and non-legacy inference?

optimum/intel/openvino/modeling_visual_language.py
eaidova: Let me refresh how model behaves for the second token, if it keeps pixel_values even in generation cycle they are not used anymore, then yes, there is no need to use variable, if not  unfortunatly there is no way to avoid this

tests/openvino/test_modeling.py
eaidova: it means that testing model broken and does not contain required information. Model uses reduced shapes comparing with original llava, but default config does not follow recalculation based on model parameters for this parameter and hardcode value from original model.

I think better to replace model with fix

tests/openvino/test_modeling.py
eaidova: problem that we can not trust warnings in transformers :) in 4.45 transformers raises warning about legacy processing says that it will be removed in 4.46 (not the first time when some decisions about changing interface delayed e.g. using Cache class instead tuples for past_key_values promises to remove tuples support in 4.43 and still here), but it still here and by default preprocessing initialized as legacy if additional parameters are not provided, so I have no info when  new processing will be by default and when legacy gone, that is why I'm trying to be ready to this moment

eaidova: moved in separated test

tests/openvino/test_modeling.py
eaidova: fixed

tests/openvino/test_modeling.py
eaidova: yes, fixed

optimum/intel/openvino/modeling_visual_language.py
eaidova: I reworked legacy part that do not use query_state anymore and does not require share legacy processing variable between generation iterations

optimum/intel/openvino/modeling_visual_language.py
eaidova: yes, thatnks, removed

tests/openvino/test_modeling.py
eaidova: test model confg updated to store correct value

tests/openvino/test_modeling.py
```diff
@@ -1983,12 +1983,67 @@ def test_compare_to_transformers(self, model_arch):
...
-
         del transformers_model
         del ov_model
 
         gc.collect()
 
+    @unittest.skipIf(
+        is_transformers_version("<", "4.45.0"), reason="New preprocessing available only in transformers >= 4.45"
+    )
+    @parameterized.expand(["llava", "llava_next"])
```
eaidova, line 1994:
```suggestion
    @parameterized.expand(["llava", "llava_next"])
    @unittest.skipIf(
        is_transformers_version("<", "4.45.0"), reason="New preprocessing available only in transformers >= 4.45"
    )
```

PR #928, opened 2024-10-08
support minicpmv

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #972, opened 2024-10-24
add minicpmv support

eaidova:
# What does this PR do?
support of minicpmv
```
from PIL import Image
import requests

model_id = "openbmb/MiniCPM-V-2_6"

prompt= "<|im_start|>user\n(<image>./</image>)\nWhat is unusual on this image?<|im_end|>\n<|im_start|>assistant\n"

model = OVModelForVisualCausalLM.from_pretrained(model_dir, trust_remote_code=True)

inputs = processor([prompt], [image], return_tensors="pt")

result  = model.generate(**inputs, max_new_tokens=20)

```

changes were already reviewed as part of https://github.com/huggingface/optimum-intel/pull/928 (I just moved all things relevant to others models support in separated PRs)

## Before submitting

### DIFF
--- a/tests/openvino/test_modeling.py
+++ b/tests/openvino/test_modeling.py
@@ -50,6 +50,7 @@ from transformers import (
     AutoModelForSpeechSeq2Seq,
     AutoModelForTokenClassification,
     AutoModelForVision2Seq,
+    AutoProcessor,
     AutoTokenizer,
     GenerationConfig,
     Pix2StructForConditionalGeneration,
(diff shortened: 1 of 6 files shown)


tests/openvino/test_modeling.py
```diff
@@ -1891,28 +1894,64 @@ def get_transformer_model_class(self, model_arch):
             from transformers import LlavaNextForConditionalGeneration
 
             return LlavaNextForConditionalGeneration
-        return None
+        return AutoModelForCausalLM
+
+    def gen_inputs(self, model_arch, base_text_prompt, image=None):
```
nikita-savelyevv, line 1899: Do you think it would be possible to expose a method like this as a part of an API for working with `OVModelForVisualCausalLM`s? 

It would simplify the demo examples and usage in general. Also, it will come helpful during dataset collection for data-aware compression.

AlexKoff88: I was also thinking that it could help for demos, optimization and validation purposes. No problem if we do it in the follow-up PR.

AlexKoff88 (approved):

nikita-savelyevv (approved):

PR #975, opened 2024-10-27
Clean and clear CI

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting


.github/workflows/test_openvino.yml
```diff
@@ -20,40 +18,59 @@ jobs:
...
-        python-version: ["3.8", "3.12"]
+        test_pattern:
+          [
+            "*modeling.py",
+            "*modeling_diffusion.py",
+            "*modeling_sentence_transformers.py",
+            "*quantization*",
+            "*training*",
+            "*export*",
+          ]
```
IlyasMoutawwakil, line 29: this one gives a maximum test time of 20 minutes, the alternative:
```
        test_pattern:
          [
            "test_modeling*",
            "test_quantization*",
            "test_training*",
            "test_export*",
          ]
```
takes 40 minutes, and the original `tests/openvino/*` takes 1 hour 10 minutes

.github/workflows/build_pr_documentation.yml
```diff
@@ -18,20 +19,18 @@ jobs:
...
+      - uses: actions/checkout@v4
         with:
-          repository: 'huggingface/optimum-intel'
+          repository: "huggingface/optimum-intel"
           path: optimum-intel
 
       - name: Setup environment
         run: |
-          python -m venv venv-doc
-          source venv-doc/bin/activate
```
echarlaix, line 34: This was needed when `ubuntu-latest` was upgraded to `ubuntu-24.04` couple of weeks ago (now reverted) which uses python 3.12. If we want to remove it, we need to replace `ubuntu-latest` with `ubuntu-22.04` to avoid future issues

.github/workflows/test_generation.yml
```diff
@@ -17,22 +18,25 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9]
```
echarlaix, line 20: makes sense, we should also officially drop python 3.8 in https://github.com/huggingface/optimum-intel/blob/v1.20.1/setup.py#L91

.github/workflows/test_openvino.yml
```diff
@@ -20,40 +18,58 @@ jobs:
...
+
+      - name: Install dependencies (nightly)
         run: |
-          pytest tests/openvino/ --ignore tests/openvino/test_modeling_basic.py --durations=0
-      - name: Test basic
+          pip install --upgrade --pre openvino openvino-tokenizers --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
+
+      - name: Test with Pytest (nightly)
         run: |
-          pip uninstall -y nncf
```
echarlaix, line 53: was added to make sure openvino models can be run even if nncf is not installed, since then nncf was added to the openvino extra but would prefer to keep it for now (as technically nncf shouldn't be a required dependency for openvino inference)

echarlaix (approved):
Looks great thanks @IlyasMoutawwakil !

.github/workflows/test_openvino.yml
IlyasMoutawwakil: made sure it's uninstalled for modeling tests

.github/workflows/test_generation.yml
IlyasMoutawwakil: dropped from setup.py

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -409,16 +410,19 @@ def _from_pretrained(
...
                 submodels[name] = kwargs.pop(name)
             elif config.get(name, (None, None))[0] is not None:
-                library_name, library_classes = config.get(name)
-                library = importlib.import_module(library_name)
-                class_obj = getattr(library, library_classes)
+                module_name, module_class = config.get(name)
+                if hasattr(pipelines, module_name):
+                    module = getattr(pipelines, module_name)
+                else:
+                    module = importlib.import_module(module_name)
```
IlyasMoutawwakil, line 424: in diffusers sometimes they import entire modules from `pipelines` https://github.com/huggingface/diffusers/blob/13e8fdecda91e27e40b15fa8a8f456ade773e6eb/src/diffusers/loaders/single_file.py#L65C7-L71C21

setup.py
```diff
@@ -55,18 +55,15 @@
...
 
 EXTRAS_REQUIRE = {
-    "neural-compressor": ["neural-compressor[pt]>3.0", "accelerate", "transformers<4.46"],
-    "openvino": [
-        "openvino==2024.4.1.dev20240926",
-        "nncf>=2.11.0",
-        "openvino-tokenizers[transformers]==2024.4.1.0.dev20240926",
-    ],
     "nncf": ["nncf>=2.11.0"],
+    "openvino": ["nncf>=2.11.0", "openvino==2024.4.1.dev20240926", "openvino-tokenizers==2024.4.1.0.dev20240926"],
```
IlyasMoutawwakil, line 65: is there a reason why these versions are pinned ?

eaidova: yes, it is done because issues with openvino 2024.4.0, 2024.4.1 contains fixing for it, but this package is not released yet, prerelease packages can not be installed in pip without --pre (this approach does not work with setup.py) or with strict version. We have IR to unpin it when release package will be released

setup.py
```diff
@@ -55,18 +55,15 @@
     "tiktoken",
     "sentence-transformers",
     "open_clip_torch>=2.26.1",
+    "peft",
```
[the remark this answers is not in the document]
eaidova: probably good to align lowerbound with diffusers
https://github.com/huggingface/diffusers/blob/main/setup.py#L118

.github/workflows/test_openvino.yml
```diff
@@ -20,40 +18,55 @@ jobs:
...
+
+      - if: ${{ matrix.transformers-version != 'latest' }}
+        name: Downgrade Transformers and Accelerate
+        run: |
+          pip install transformers==${{ matrix.transformers-version }} accelerate==0.*
+
+      - if: ${{ matrix.test-pattern == '*modeling*' }}
+        name: Uninstall NNCF
+        run: |
+          pip uninstall -y nncf
```
AlexKoff88, line 56: Any reason we uninstall NNCF here? We can help to resolve if there are any issues caused by the NNCF presence.

eaidova (approved):

.github/workflows/test_openvino.yml
```diff
@@ -20,40 +18,55 @@ jobs:
...
           pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
-          pip install .[openvino,openvino-tokenizers,tests,diffusers] onnxruntime
+          pip install .[openvino,openvino-tokenizers,diffusers,tests] transformers[testing]
+
+      - if: ${{ matrix.transformers-version != 'latest' }}
+        name: Downgrade Transformers and Accelerate
+        run: |
+          pip install transformers==${{ matrix.transformers-version }} accelerate==0.*
+
+      - if: ${{ matrix.test-pattern == '*modeling*' }}
```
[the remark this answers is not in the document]
IlyasMoutawwakil: > could you please tell a little bit more about idea behind precommit changes? I correctly understand that you split precommit on several independent workflow?

Yes, the number of tests results in ~1h20min wait time for openvino per-commit tests. So the idea is to distribute it.

> I'm wondering is there really need to remove nncf during modeling testing here

My understanding from https://github.com/huggingface/optimum-intel/pull/975/files#r1822437311 is that modeling tests should pass without nncf (they do), by uninstalling it, we're making sure that's the case. The modeling tests that use nncf are marked as slow and they are tested separately.

.github/workflows/test_openvino.yml
IlyasMoutawwakil: No issue, just to make sure NNCF is not a dep for OV modeling (I guess this was the case at some point)
Also https://github.com/huggingface/optimum-intel/pull/975#discussion_r1827733818

.github/workflows/test_openvino_slow.yml
```diff
@@ -0,0 +1,76 @@
...
+jobs:
+  build:
+    if: ${{ (github.event_name == 'workflow_dispatch') || (github.event_name == 'schedule') ||  (github.event_name == 'push') || contains( github.event.pull_request.labels.*.name, 'openvino-test') }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-22.04", "windows-2019"]
+        openvino-version: ["stable", "nightly"]
+        transformers-version: ["4.36.0", "latest"]
+        nncf: ["nncf", "git+https://github.com/openvinotoolkit/nncf.git"]
```
nikita-savelyevv, line 30: To be honest, I don't think all these 16 combinations are strictly necessary. 

It also so happens that I've just started working on introducing a separate OV workflow to run not only slow OV tests, but all OV tests: https://github.com/huggingface/optimum-intel/pull/985. This is because during slow tests almost none quantization-related tests are actually run which is especially critical for ensuring nncf develop compatibility.

So after this PR is merged, if there are no objections I would remove ov-nightly and nncf-develop from slow tests in my PR and introduce a separate workflow to test ov-nightly and nncf-develop versions on full OV scope on a single OS and transformers version.

nikita-savelyevv (commented):
These are useful changes, thank you!

.github/workflows/test_openvino_basic.yml
```diff
@@ -1,83 +0,0 @@
...
-          python -m pip install --upgrade pip
-          # Install PyTorch CPU to prevent unnecessary downloading/installing of CUDA packages
-          pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
-          # Install openvino manually to prevent dependency conflicts when .[openvino] pins
-          # optimum or transformers to a specific version
-          pip install ${{ matrix.openvino }}
-          pip install .[tests]
-
-      - name: Pip freeze
-        run: pip freeze
```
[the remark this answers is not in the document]
IlyasMoutawwakil: added

.github/workflows/test_openvino_slow.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Install dependencies
+        run: |
+          pip install --upgrade pip
+          pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
+          pip install .[tests] transformers[testing]
+
+      - if: ${{ matrix.openvino-version == 'stable' }}
+        name: Install stable OpenVINO
+        run: |
+          pip install openvino openvino-tokenizers
```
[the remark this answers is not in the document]
IlyasMoutawwakil: okay

nikita-savelyevv (approved):

AlexKoff88 (approved):

PR #977, opened 2024-10-29
phi3 vision

eaidova:
# What does this PR do?

```
from PIL import Image 
import requests 
from transformers import AutoProcessor, TextStreamer

model_id = "microsoft/Phi-3.5-vision-instruct"

model = OVModelForVisualCausalLM.from_pretrained(
  model_id, 
  trust_remote_code=True,
)

processor = AutoProcessor.from_pretrained(model_id, 
  trust_remote_code=True, 
) 

messages = [
    {"role": "user", "content": "<|image_1|>\nWhat is unusual on this picture?"},
]
url = "https://github.com/openvinotoolkit/openvino_notebooks/assets/29454499/d5fbbd1a-d484-415c-88cb-9986625b7b11"

prompt = processor.tokenizer.apply_chat_template(
  messages, 
  tokenize=False, 
  add_generation_prompt=True
)

inputs = processor(prompt, [image], return_tensors="pt")

generation_args = { 
    "max_new_tokens": 50, 
    "temperature": 0.0, 
    "do_sample": False,
    "streamer": TextStreamer(processor.tokenizer, skip_prompt=True, skip_special_tokens=True)
} 

generate_ids = model.generate(**inputs, 
  eos_token_id=processor.tokenizer.eos_token_id, 
  **generation_args
)

# remove input tokens 
generate_ids = generate_ids[:, inputs['input_ids'].shape[1]:]
response = processor.batch_decode(generate_ids, 
  skip_special_tokens=True, 
  clean_up_tokenization_spaces=False)[0]
```

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -45,6 +45,8 @@ from optimum.utils.save_utils import maybe_load_preprocessors
 from .utils import _MAX_UNCOMPRESSED_SIZE, MULTI_MODAL_TEXT_GENERATION_MODELS, clear_class_registry
 
 
+FORCE_ATTN_MODEL_CLASSES = {"phi3-v": "eager"}
+
 if TYPE_CHECKING:
     from optimum.intel.openvino.configuration import OVConfig
 
@@ -264,6 +266,13 @@ def main_export(
 
         if is_transformers_version(">=", "4.36") and model_type in SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED:
             loading_kwargs["attn_implementation"] = "eager"
+        
+        # some models force flash_attn attention by default thta is not available for cpu
+        logger.warn(model_type)
+        if is_transformers_version(">=", "4.36") and model_type in FORCE_ATTN_MODEL_CLASSES:
+            loading_kwargs["_attn_implementation"] = FORCE_ATTN_MODEL_CLASSES[model_type]
+        
+        logger.warn(loading_kwargs)
         # there are some difference between remote and in library representation of past key values for some models,
         # for avoiding confusion we disable remote code for them
         if (
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -266,13 +266,10 @@ def main_export(
 
         if is_transformers_version(">=", "4.36") and model_type in SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED:
             loading_kwargs["attn_implementation"] = "eager"
-        
+
         # some models force flash_attn attention by default thta is not available for cpu
-        logger.warn(model_type)
         if is_transformers_version(">=", "4.36") and model_type in FORCE_ATTN_MODEL_CLASSES:
             loading_kwargs["_attn_implementation"] = FORCE_ATTN_MODEL_CLASSES[model_type]
-        
-        logger.warn(loading_kwargs)
         # there are some difference between remote and in library representation of past key values for some models,
         # for avoiding confusion we disable remote code for them
         if (
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -267,7 +267,7 @@ def main_export(
         if is_transformers_version(">=", "4.36") and model_type in SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED:
             loading_kwargs["attn_implementation"] = "eager"
 
-        # some models force flash_attn attention by default thta is not available for cpu
+        # some models force flash_attn attention by default that does not support load model on cpu
         if is_transformers_version(">=", "4.36") and model_type in FORCE_ATTN_MODEL_CLASSES:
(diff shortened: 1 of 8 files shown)


tests/openvino/test_modeling.py
```diff
@@ -1880,9 +1880,10 @@ class OVModelForVisualCausalLMIntegrationTest(unittest.TestCase):
     if is_transformers_version(">=", "4.40.0"):
         SUPPORTED_ARCHITECTURES += ["llava_next", "nanollava"]
     if is_transformers_version(">=", "4.45.0"):
-        SUPPORTED_ARCHITECTURES += ["minicpmv", "internvl2"]
+        SUPPORTED_ARCHITECTURES += ["minicpmv", "internvl2", "phi3_v"]
+    REMOTE_CODE_MODELS = ["minicpmv", "nanollava", "phi3_v"]
```
eaidova, line 1884:
```suggestion
```

AlexKoff88 (approved):

optimum/exporters/openvino/__main__.py
```diff
@@ -264,6 +266,10 @@ def main_export(
 
         if is_transformers_version(">=", "4.36") and model_type in SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED:
             loading_kwargs["attn_implementation"] = "eager"
+
+        # some models force flash_attn attention by default thta is not available for cpu
```
eaidova, line 270:
```suggestion
        # some models force flash_attn attention by default that does not support load model on cpu
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -2196,3 +2197,210 @@ def patch_model_for_export(
...
+        """
+        Creates a config for different behaviour.
+        Args:
+            behavior ([`ConfigBehavior`]):
+                The behavior to use for the new instance.
+        """
+        if isinstance(behavior, str) and not isinstance(behavior, Phi3VisionConfigBehavior):
+            behavior = Phi3VisionConfigBehavior(behavior)
+
+        if behavior == Phi3VisionConfigBehavior.TEXT_EMBEDDINGS:
```
nikita-savelyevv, line 2301: The logic for `Phi3VisionConfigBehavior.TEXT_EMBEDDINGS` and `Phi3VisionConfigBehavior.LANGUAGE` is mostly the same. Can we avoid code duplication in this case?

optimum/exporters/openvino/model_configs.py
```diff
@@ -2196,3 +2197,210 @@ def patch_model_for_export(
...
+            return self.__class__(
+                self._orig_config,
+                task=self.task,
+                int_dtype=self.int_dtype,
+                float_dtype=self.float_dtype,
+                behavior=behavior,
+                preprocessors=self._preprocessors,
+            )
+
+    def get_model_for_behavior(self, model, behavior: Union[str, Phi3VisionConfigBehavior]):
```
nikita-savelyevv, line 2378: Can be static

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -230,7 +230,14 @@ def forward(self, image_feature, pos_embed, key_padding_mask):
         return result
 
 
-MODEL_PARTS_CLS_MAPPING = {"resampler": OVResampler}
+class OVVisionProjection(OVModelPart):
+    _model_name = "vision_projection"
+
+    def forward(self, img_features):
+        return self.request(img_features)[0]
```
nikita-savelyevv, line 237:
```suggestion
    def forward(self, img_features):
        self._compile()
        return self.request(img_features)[0]
```

nikita-savelyevv (commented):
👍

optimum/exporters/openvino/convert.py
```diff
@@ -712,7 +712,18 @@ def export_from_model(
...
+        if preprocessors is not None:
+            # phi3-vision processor does not have chat_template attribute that breaks Processor saving on disk
+            if is_transformers_version(">=", "4.45") and model_type == "phi3-v" and len(preprocessors) > 1:
+                if not hasattr(preprocessors[1], "chat_template"):
+                    preprocessors[1].chat_template = getattr(preprocessors[0], "chat_template", None)
+            for processor in preprocessors:
+                try:
+                    processor.save_pretrained(output)
+                except Exception as ex:
+                    logger.error(f"Saving {type(processor)} failed with {ex}")
```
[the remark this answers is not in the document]
eaidova: generally, I would like to avoid model conversion failed due to issues with preprocessing savings. in case, if model exported using from_pretrained with export=True, preprocessor saved to tmp dir and not accessible for public usage and there is still possibility to use original model_id for initializing them. So even if no possibility to load or save preprocessor, it is still possible to use model itself. We have similar logic for generation_conifg (in some cases models on hf hub contains invalid outdated configs that can not be saved, but model still valid and can be used)

optimum/exporters/openvino/model_configs.py

optimum/exporters/openvino/model_configs.py
eaidova: reused code where it is possible

PR #984, opened 2024-10-31
fix order of hidden states in text encoder

eaidova:
# What does this PR do?

using sorting breaks topological order of outputs, it keeps output names in alphabetical order, that is different from topological if model has > 10 layers (e.g. hiden_states.0, hidden_states.1, ..., hidden_states.9, hidden_states.10 we got hidden_states.0, hidden_tates.1, hidden_states.10, hidden_states.2, ...., hidden_states.9) that significantly impact model accuracy for sdxl.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -995,9 +995,9 @@ class OVPipelinePart(ConfigMixin):
 class OVModelTextEncoder(OVPipelinePart):
     def __init__(self, model: openvino.runtime.Model, parent_pipeline: OVDiffusionPipeline, model_name: str = ""):
         super().__init__(model, parent_pipeline, model_name)
-        self.hidden_states_output_names = sorted(
-            {name for out in self.model.outputs for name in out.names if name.startswith("hidden_states")}
-        )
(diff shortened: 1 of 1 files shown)


AlexKoff88 (approved):

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -995,9 +995,9 @@ def modules(self):
 class OVModelTextEncoder(OVPipelinePart):
     def __init__(self, model: openvino.runtime.Model, parent_pipeline: OVDiffusionPipeline, model_name: str = ""):
         super().__init__(model, parent_pipeline, model_name)
-        self.hidden_states_output_names = sorted(
-            {name for out in self.model.outputs for name in out.names if name.startswith("hidden_states")}
-        )
+        self.hidden_states_output_names = {
+            name for out in self.model.outputs for name in out.names if name.startswith("hidden_states")
+        }
```
eaidova, line 1000:
```suggestion
        self.hidden_states_output_names = [
            name for out in self.model.outputs for name in out.names if name.startswith("hidden_states")
        ]
```

PR #985, opened 2024-11-04
Add a full OV testing GHA workflow for nightly

nikita-savelyevv:
# What does this PR do?

**Changes:**
1. Introduce a new GHA workflow for full OV testing on 4 possible combinations of OV and NNCF stable/nightly versions.
2. Remove OV and NNCF stable/nightly options from slow test.

**Reason for changes:**
During slow testing almost none quantization tests are run which is critical for assessing NNCF version impact. That is why all OV tests need to be run on a nightly basis. 

Instead of expanding the current slow test workflow to do full testing, a new workflow is added for that. This is to conserve computational resources, as different OS and transformers version is not strictly required to assess OV nightly / NNCF develop impact.

## Before submitting


eaidova (approved):
accuracy issues for runs with openvino nightly not related to this PR. Will be addressed in separated PR

.github/workflows/test_openvino_full.yml
```diff
@@ -0,0 +1,88 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install lowest compatible transformers version
+        if: ${{ matrix.transformers-version != 'latest' }}
+        run: pip install transformers==${{ matrix.transformers-version }}
```
IlyasMoutawwakil, line 56: installing transformers before `pip install .[tests]` will result in the later overriding the intended version

nikita-savelyevv: Thanks! Fixed

.github/workflows/test_openvino_full.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: Install openvino-nightly
+        if: ${{ matrix.openvino == 'ov-nightly' }}
+        run: pip install --pre -U openvino openvino-tokenizers --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
+
+      - name: Install openvino release
+        if: ${{ matrix.openvino == 'ov-stable' }}
+        run: pip install .[openvino]
```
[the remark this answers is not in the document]
nikita-savelyevv: Right👍

IlyasMoutawwakil (approved):

.github/workflows/test_openvino_slow.yml
```diff
@@ -65,7 +58,7 @@ jobs:
 
       - name: Install dependencies (slow)
         run: |
-          pip install ${{ matrix.nncf }}
+          pip install .[nncf]
```
IlyasMoutawwakil, line 61: not necessary for now since nncf doesn't constrain transformers but just in case it did
```suggestion
          pip install .[nncf]

      - if: ${{ matrix.transformers-version != 'latest' }}
        name: Downgrade Transformers and Accelerate
        run: pip install transformers==${{ matrix.transformers-version }} accelerate==0.*
```

PR #986, opened 2024-11-04
fix loading safety checker in pipeline

eaidova:
# What does this PR do?

fixing broken tests for textual inversion due to incorrect safety checker imports

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -409,9 +409,9 @@ class OVDiffusionPipeline(OVBaseModel, DiffusionPipeline):
             "tokenizer_2": None,
             "tokenizer_3": None,
             "feature_extractor": None,
-            "image_encoder": None,
-            "safety_checker": None,
         }
+
+        additional_submodels = ["image_encoder", "safety_checker"]
         for name in submodels.keys():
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -409,9 +409,9 @@ def _from_pretrained(
             "tokenizer_2": None,
             "tokenizer_3": None,
             "feature_extractor": None,
-            "image_encoder": None,
-            "safety_checker": None,
         }
+
+        additional_submodels = ["image_encoder", "safety_checker"]
```
IlyasMoutawwakil, line 414: This doesn't solves the issue as it will just ignore and not load the `safety_checker` for example.
I provide a fix based on how diffusers handle the case of `library_name` being a module in pipelines https://github.com/huggingface/optimum-intel/pull/975#discussion_r1827642118

eaidova: pk, thanks, than I can close my pr?

IlyasMoutawwakil: yes, let's try merging the other PR today, it should accelerate per-commit openvino CI back to ~20mins

PR #988, opened 2024-11-05
Introduce support for NF4 data type for OV weight compression

l-bat:
# What does this PR do?

* Added a new possible value for `--weight-format` CLI argument: `nf4`.
* Added a couple of tests for compression with the  `nf4` data type. 

## Usage examples:
Optimum CLI:
```bash
optimum-cli export openvino -m facebook/opt-125m --task text-generation-with-past --weight-format nf4 ./tmp
```
Optimum Python API:
```python
    model_id="gpt2",
    export=True,
    quantization_config=OVWeightQuantizationConfig(bits=4, weight_format="nf4"),
)
```
## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -347,7 +347,7 @@ class OVWeightQuantizationConfig(OVQuantizationConfigBase):
             Indicates whether to apply a scale estimation algorithm that minimizes the L2 error between the original and
             compressed layers. Providing a dataset is required to run scale estimation.
         weight_format (`str`, defaults to 'int'):
-            Data format weights are compressed to. Possible values: ['int4', 'int8', 'mxfp4'].
+            Data format weights are compressed to. Possible values: ['int4', 'int8', 'mxfp4', 'nf4].
         qptq (`bool`, *optional*):
             Whether to apply GPTQ algorithm. GPTQ optimizes compressed weights in a layer-wise fashion to minimize the
             difference between activations of a compressed and original layer. Dataset is required to run GPTQ.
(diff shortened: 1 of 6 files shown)


l-bat:

tests/openvino/test_exporters_cli.py
```diff
@@ -107,6 +107,8 @@ class OVCLIExportTestCase(unittest.TestCase):
         ("text-generation-with-past", "opt125m", "int4 --sym --group-size 128", {"int8": 4, "int4": 72}),
         ("text-generation-with-past", "opt125m", "int4 --group-size 64", {"int8": 4, "int4": 144}),
         ("text-generation-with-past", "opt125m", "mxfp4", {"int8": 4, "f4e2m1": 72, "f8e8m0": 72}),
+        ("text-generation-with-past", "opt125m", "nf4", {"int8": 4, "nf4": 72}),
+        ("text-generation-with-past", "opt125m", "nf4  --awq --dataset wikitext2", {"int8": 4, "nf4": 72}),
```
AlexKoff88, line 111: The tests are pretty slow already. Can we just keep data-free test for "nf4"? It should be enough at this point.

optimum/intel/openvino/configuration.py
```diff
@@ -462,6 +462,11 @@ def post_init(self):
             if self.gptq:
                 raise ValueError("The GPTQ algorithm is not supported for 'mxfp4' weight format")
 
+        if self.weight_format in "nf4" and self.bits != 4:
+            raise ValueError(
+                f"When applying weight compression with 'nf4' weight format the `bits` parameters must be set to 4, but found {self.bits}"
+            )
```
nikita-savelyevv, line 468: I believe AWQ and SE are not supported for NF4, so corresponding checks should be added. GPTQ is supported though. Possibly it could be combined with conditions for mxfp4 above to avoid code duplication.

optimum/intel/openvino/configuration.py
```diff
@@ -462,6 +462,11 @@ def post_init(self):
             if self.gptq:
                 raise ValueError("The GPTQ algorithm is not supported for 'mxfp4' weight format")
 
+        if self.weight_format in "nf4" and self.bits != 4:
```
nikita-savelyevv, line 465:
```suggestion
        if self.weight_format == "nf4" and self.bits != 4:
```

nikita-savelyevv (approved):
Thanks!

AlexKoff88 (approved):

AlexKoff88:
@l-bat, can you fix the style, please?

PR #989, opened 2024-11-06
add patching for update_causal_mask to falcon for >= 4.45

eaidova:
* reflect to alignment for attention mask preparation in transformers https://github.com/huggingface/transformers/commit/a30c865f991dfec9452cc64bd9a97bfbb96be036

current approach brings accuracy issues on platforms that natively supports bf16/fp16 (e.g. iGPU, ARM CPU) if model loaded and converted as fp32 (due to numeric overflow representing torch_dtype.min for fp32 as bf16 or fp16 constant), reused the same patching like in llama and gemma for other models
* enable back codegen2 model with remote code (we can not update remote code for test model as it will be incompatible with reference model  https://huggingface.co/Salesforce/codegen2-1B_P. Added opportunity to use legacy patching for such case
## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -1914,8 +1914,7 @@ def _codegen_wrapped_scaled_dot_product_legacy(
     else:
         query_length, key_length = query.size(-2), key.size(-2)
 
-        # causal_mask is always [True, ..., True] otherwise, so executing this
-        # is unnecessary
+        # causal_mask is always [True, ..., True] otherwise, so executing this is unnecessary
         if query_length > 1:
             causal_mask = self.causal_mask[:, :, key_length - query_length : key_length, :key_length].to(torch.bool)
 
@@ -1944,8 +1943,8 @@ class CodeGenModelPatcher(DecoderModelPatcher):
 
         attn_fn = codegen_wrapped_scaled_dot_product
         if is_torch_version(">=", "2.1.0") and is_transformers_version(">=", "4.45"):
-            # in transformers 4.45 removed causal_mask const buffer from model
-            # if it is still exists, it means legacy remote code loaded
+            # in transformers 4.45 causal_mask const buffer was removed from the model
+            # if it still exists, it means legacy remote code was loaded
             if hasattr(self._model.transformer.h[0].attn, "causal_mask"):
                 attn_fn = _codegen_wrapped_scaled_dot_product_legacy
 
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -109,11 +109,19 @@ def patch_model_with_bettertransformer(model):
     return model
 
 
-def patch_update_causal_mask(model, transformers_version):
+def patch_update_causal_mask(model, transformers_version, inner_model_name="model"):
     if is_transformers_version(">=", transformers_version):
-        inner_model = getattr(model, "model", getattr(model, "transformer", None))
+        inner_model = getattr(model, inner_model_name, None)
         if inner_model is not None:
+            if hasattr(inner_model, "_update_causal_mask"):
+                inner_model._orig_update_causal_mask = inner_model._update_causal_mask
             inner_model._update_causal_mask = types.MethodType(_llama_gemma_update_causal_mask, inner_model)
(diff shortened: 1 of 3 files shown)


AlexKoff88 (approved):

AlexKoff88:
@eaidova, please take a look at the failed tests.

eaidova:
> @eaidova, please take a look at the failed tests.

@AlexKoff88  it does not seems to be related to my changes:
```
FAILED tests/openvino/test_modeling.py::OVModelIntegrationTest::test_load_model_from_hub_private_with_token - huggingface_hub.errors.RepositoryNotFoundError: 401 Client Error. (Request ID: Root=1-672b4212-1f28ca04615798037e8d4cf0;d3d5573b-1008-445b-ac83-e1c2e85cb525)
```
looks like ci used account does not have access to download repo using private token in this test, I sow this issue previously time to time

AlexKoff88:
@slyalin, @nikita-savelyevv, please help with the review.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1865,6 +1872,68 @@ def __exit__(self, exc_type, exc_value, traceback):
                 layer.self_attn.forward = layer.self_attn._orig_forward
 
 
+# copied from  https://github.com/huggingface/optimum/blob/v1.22.0/optimum/bettertransformer/models/attention.py#L169
```
nikita-savelyevv, line 1875: Can you please add a fixed commit in the link so that it will be valid later?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1865,6 +1872,68 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            # https://github.com/huggingface/transformers/blob/ae54e3c3b18bac0832ad62ea9b896dfd52a09850/src/transformers/models/gpt2/modeling_gpt2.py#L195
+            # is [True, ..., True] so actually not causal
+            sdpa_result = torch.nn.functional.scaled_dot_product_attention(
+                query, key, value, attn_mask=None, dropout_p=dropout_p, is_causal=False
+            )
+    else:
+        query_length, key_length = query.size(-2), key.size(-2)
+
+        # causal_mask is always [True, ..., True] otherwise, so executing this
+        # is unnecessary
```
nikita-savelyevv, line 1918:
```suggestion
        # causal_mask is always [True, ..., True] otherwise, so executing this is unnecessary
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1873,14 +1942,23 @@ def __enter__(self):
         # For avoiding breaking model on tracing stage, we reduce area of bettertransformer patch only for _attn.
         from optimum.bettertransformer.models.attention import codegen_wrapped_scaled_dot_product
 
+        attn_fn = codegen_wrapped_scaled_dot_product
+        if is_torch_version(">=", "2.1.0") and is_transformers_version(">=", "4.45"):
+            # in transformers 4.45 removed causal_mask const buffer from model
```
nikita-savelyevv, line 1947:
```suggestion
            # in transformers 4.45 causal_mask const buffer was removed from the model
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1873,14 +1942,23 @@ def __enter__(self):
...
+            # if it is still exists, it means legacy remote code loaded
```
nikita-savelyevv, line 1948:
```suggestion
            # if it still exists, it means legacy remote code was loaded
```

optimum/exporters/openvino/model_patcher.py
eaidova:
```suggestion
# copied from  https://github.com/huggingface/optimum/blob/2112e99122d7f23a1da1a9d263fef64301050ea7/optimum/bettertransformer/models/attention.py#L168
```

nikita-savelyevv: > v1.22.0 it is tag in the repo https://github.com/huggingface/optimum/releases/tag/v1.22.0 it is already pointed to specific commit. Do you mean what happens if somebody remove this tag?

Ah, I see. I think it's fine then.

PR #994, opened 2024-11-11
Added a fix for FP16 overflow issue on GPU/NPU

AlexKoff88:


optimum/exporters/openvino/convert.py
```diff
@@ -91,9 +91,31 @@
...
+    path: str,
+    ov_config: Optional["OVConfig"] = None,
+    library_name: Optional[str] = None,
+    config: OnnxConfig = None,
+):
     compress_to_fp16 = ov_config is not None and ov_config.dtype == "fp16"
     model = _add_version_info_to_model(model, library_name)
+
+    if hasattr(config, "runtime_options"):
+        model = _add_runtime_options_to_rt_info(model, config.runtime_options)
```
eaidova, line 118: maybe better to provide runtime options as argument instead of config if it is needed only for providing these options here?

eaidova (approved):

eaidova (approved):

PR #1009, opened 2024-11-22
unify xpu and cpu backend and use paged attention

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting


yao-matrix:
@IlyasMoutawwakil @echarlaix , pls help review, we can also have a meeting to review it if needed. Thx.

jiqing-feng:
Hi @IlyasMoutawwakil , please also merge this PR #1024. Thanks!

.github/workflows/test_ipex.yml
```diff
@@ -18,8 +18,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        torch-version: ["2.2.0", "2.3.*", "2.4.*"]
-        transformers-version: ["4.39.0", "4.44.*"]
```
IlyasMoutawwakil, line 22: what's the minimum supported version with the new modeling ?

jiqing-feng: To avoid too many tests and extra efforts for previous issues, we only keep the latest released version torch and the last 1 released version torch.

IlyasMoutawwakil: and for transformers it's one version (4.45) ?

jiqing-feng: As, transformers models refactoring so fast, we only keep 1 version for now, it should be 4.46 now, please merge #1024 . Thanks

IlyasMoutawwakil: merged, let's pin versions in setup that reflect this limited support

tests/neural_compressor/test_ipex.py
```diff
@@ -17,6 +17,7 @@
 
 import os
 import tempfile
+import unittest
```
IlyasMoutawwakil, line 20:
```suggestion
```

tests/neural_compressor/test_ipex.py
```diff
@@ -79,5 +80,5 @@ def test_ipex_static_quantization_with_smoothquant(self, task, model_arch, expec
                 is_static=True,
                 num_samples=num_samples,
                 load_inc_model=False,
-                load_ipex_model=True,
+                load_ipex_model=False,
```
IlyasMoutawwakil, line 83: why not ? my understanding is this test is specifically written to do "INC quant -> IPEX load"

jiqing-feng: The function (IPEX load inc model) is deprecated because IPEXModel is no longer support jit mode, and the model saved by INC is jit mode. And yes, the test name is quiet confused, we will change it.

optimum/intel/ipex/modeling_base.py
```diff
@@ -676,118 +514,53 @@ def _prepare_generation_config(
...
-        if self._is_ipex_exported and kwargs.get("assistant_model", None):
+        # Patch functions to support paged cache
+        if self._add_patch:
+            transformers.generation.utils.NEED_SETUP_CACHE_CLASSES_MAPPING["paged"] = IPEXPagedCache
+            self.generation_config.cache_implementation = "paged"
+            if is_transformers_version(">=", "4.45.0"):
+                if "paged" not in transformers.generation.configuration_utils.ALL_CACHE_IMPLEMENTATIONS:
+                    transformers.generation.configuration_utils.ALL_CACHE_IMPLEMENTATIONS.append("paged")
+            if new_kwargs.get("generation_config", None):
+                new_kwargs["generation_config"].cache_implementation = "paged"
```
IlyasMoutawwakil, line 537: "ipex_paged" makes more sense, to avoid conflicts with more generic paged attention in https://github.com/huggingface/transformers/pull/27597

setup.py
```diff
@@ -64,7 +64,7 @@
     "nncf": ["nncf>=2.11.0"],
     "openvino": ["nncf>=2.11.0", "openvino==2024.5.0", "openvino-tokenizers==2024.5.0"],
     "neural-compressor": ["neural-compressor[pt]>3.0", "accelerate", "transformers<4.46"],
-    "ipex": ["intel-extension-for-pytorch", "transformers>=4.39,<4.45"],
+    "ipex": ["intel-extension-for-pytorch", "transformers>=4.39,<4.47"],
```
IlyasMoutawwakil, line 67: if my understanding is correct
```suggestion
    "ipex": ["intel-extension-for-pytorch>=2.4", "transformers==4.46"],
```

optimum/exporters/ipex/modeling_utils.py
```diff
@@ -156,103 +253,335 @@ def _llama_model_forward(
...
         self.config = config
-        self.ipex_scale_dot_product = IndirectAccessKVCacheAttention(text_max_length=config.max_position_embeddings)
-        if hasattr(config, "rope_theta"):
-            self.ipex_rope = RotaryEmbedding(
-                config.max_position_embeddings,
-                config.hidden_size // config.num_attention_heads,
-                config.rope_theta,
-                config.architectures[0],
-            )
+        self.module_device = next(module.parameters()).device.type
```
IlyasMoutawwakil, line 510: do we wanna save only the device type here ?
`torch.xpu` device supports device ordinal like `torch.cuda`, wouldn't that create a problem with `self.kv_head_mapping` initialized on `xpu:0` by default ?

.github/workflows/test_ipex.yml
```diff
@@ -18,8 +18,8 @@ jobs:
...
+        transformers-version: ["4.46.*"]
+        torch-version: ["2.4.0", "2.5.0"]
```
IlyasMoutawwakil, line 22:
```suggestion
        torch-version: ["2.4.0", "2.5.*"]
```

jiqing-feng: Unfortunately, ipex sometimes only supports big versions like 2.4.0 and 2.5.0. IPEX has conflict in little versions torch like 2.5.1

optimum/exporters/ipex/modeling_utils.py
jiqing-feng: This parameter is only for choosing the type of fused linear, so we only need to know the device type. Optimum-intel should not support cuda.

jiqing-feng: It's just for selecting cpu/xpu linear to patch, we don't have any ops like `.to(self.module_device)`

optimum/intel/ipex/modeling_base.py

setup.py
jiqing-feng: Fixed, we have 2 versions of transformers in tests: [4.46.0, 4.46.3]

tests/neural_compressor/test_ipex.py
jiqing-feng: fixed

jiqing-feng:
Hi @IlyasMoutawwakil . I have replied and fixed your comments, please take the 2nd  round review. Thanks~

optimum/exporters/ipex/modeling_utils.py
IlyasMoutawwakil: there's a tensor constructor call right after this line that uses `device=self.module_device`

optimum/intel/ipex/modeling_base.py
```diff
@@ -676,118 +324,53 @@ def _prepare_generation_config(
...
 
+    def _reorder_cache(self, *args, **kwargs):
+        return self.model._reorder_cache(*args, **kwargs)
+
+    def prepare_inputs_for_generation(self, *args, **kwargs):
+        return self.model.prepare_inputs_for_generation(*args, **kwargs)
+
     def generate(self, *args, **kwargs):
-        if is_ipex_version("<", "2.4.0") and self._is_ipex_exported and kwargs.get("assistant_model", None):
+        new_kwargs = copy.deepcopy(kwargs)
```
IlyasMoutawwakil, line 334: original `kwargs` is never used after this point, I think it's better to use it than to make a deepcopy (most would use `model.generate(**inputs)` which means tensor inputs will be in kwargs as well.

optimum/intel/ipex/modeling_base.py
```diff
@@ -168,47 +84,25 @@ def __init__(
...
 
-        self.model.to(self._device)
-        self._dtype = self.config.torch_dtype if self.config.torch_dtype is not None else torch.float32
+        self._dtype = self.model.dtype if self.model.dtype is not None else torch.float32
+        self.use_cache = kwargs.get("use_cache", False)
         self.model_save_dir = model_save_dir
-        self._is_ipex_exported = _is_patched_with_ipex(model, self.export_feature)
+        self._add_patch = _is_patched_with_ipex(model, self.export_feature, self.use_cache)
 
-        if isinstance(model, torch.jit.RecursiveScriptModule):
```
echarlaix, line 198: TorchScript models will not be compatible anymore which is an important breaking change, we need to catch this to inform users

optimum/intel/ipex/modeling_base.py
```diff
@@ -287,82 +177,32 @@ def _from_pretrained(
...
-        else:
-            model_cache_path = hf_hub_download(repo_id=model_id, filename=file_name, **model_kwargs)
-            model_save_dir = Path(model_cache_path).parent
-
-        model = torch.jit.load(model_cache_path)
-        torch.jit.freeze(model.eval())
-
-        return cls(model, config=config, model_save_dir=model_save_dir, **kwargs)
+        task = cls.export_feature
+        model = TasksManager.get_model_from_task(
```
echarlaix, line 181: why not use `cls.auto_model_class` ?

optimum/intel/ipex/modeling_base.py
kaixuanliu: The reason we use a `deepcopy`  is to avoid scenarios like [here](https://github.com/huggingface/optimum-intel/pull/1009/files#diff-e59fd816d97650a343b56217d6530d47b18a08440fcc5a86a136b1908a6d69f8R345-R346), as we changed `kwargs["generation_config"].cache_implementation` to `ipex_paged` , if we pass this `kwargs` to a `transformer` model, it will crash

IlyasMoutawwakil: I see what it fixes but I'm not sure it's the optimal way. why not `deepcopy` only the `kwargs["generation_config"]` in kwargs itself ? that way you have an independent copy of the config without duplicating inputs.

kaixuanliu: Yeah, it's a good advice, will follow it.

tests/ipex/test_modeling.py
```diff
@@ -335,46 +340,29 @@ def test_ipex_patching_beam_search(self, test_name, model_arch, use_cache):
...
-            return_tensors="pt",
-            return_token_type_ids=False if model_arch in ("llama", "llama2") else None,
-        )
-        inputs = ipex_model.prepare_inputs_for_generation(**tokens)
-        ipex_outputs = ipex_model(**inputs)
-        exported_outputs = exported_model(**inputs)
-        self.assertTrue(torch.allclose(ipex_outputs.logits, exported_outputs.logits, atol=1e-7))
-
     def test_compare_with_and_without_past_key_values(self):
-        model_id = "echarlaix/tiny-random-gpt2-torchscript"
```
echarlaix, line 364: Would be great to add a test to make sure an IPEX model that has been saved and then pushed on the hub is still compatible with the latest the optimum-intel version and that the model can still be correctly loaded and used for inference. Also could make sense to check the resulting output (comparing it with a transformers model) wdyt ?

optimum/intel/ipex/modeling_base.py
kaixuanliu: @IlyasMoutawwakil , pls review https://github.com/huggingface/optimum-intel/pull/1035

optimum/exporters/ipex/modeling_utils.py
kaixuanliu: Fixed here: https://github.com/huggingface/optimum-intel/pull/1035

tests/ipex/test_modeling.py
jiqing-feng: I am afraid we cannot support the previous exported language model anymore because these jit models use IAVK cache which has a different logic with paged attention. It will make the code extremely massive and hard to maintain, and it will confuse users too. Besides, these jit models are also out of date compared to the current transformers.

jiqing-feng: The only way is to fallback IPEXModelForCausalLM to TSModelForCausalLM when loading a jit model, but it requires `config.torchscript == True` so we can know it's a torch script model. So you might need to update the "echarlaix/tiny-random-gpt2-torchscript" config parameter.

jiqing-feng:
Hi @echarlaix @IlyasMoutawwakil , please review the new changes. Thanks

optimum/intel/ipex/modeling_base.py
```diff
@@ -240,129 +145,30 @@ def _from_pretrained(
...
-                library_name="transformers",
-                trust_remote_code=trust_remote_code,
-                torch_dtype=torch_dtype,
-                _commit_hash=commit_hash,
-                **model_kwargs,
+        if getattr(config, "torchscript", False):
+            logger.warning(
+                "IPEXModel will not support torch script model in the future, fallback to TSModelForCausalLM"
             )
+            return TSModelForCausalLM.from_pretrained(model_id, **kwargs)
```
echarlaix, line 153: An instance of `TSModelForCausalLM` will be created for all `IPEXModel` (even for encoder models) which doesn't really make sense to me. Also it's not tested anywhere from what I see, I prefer to raise an error here instead of keeping support that we're not sure works / is compatible with the previous integration

tests/ipex/test_modeling.py
echarlaix: what I'm suggesting here is to create a new model with this implementation, push it on the hub and have a test to make sure this model is still compatible / can be correctly loaded and inference works as expected (pushing an other model than "echarlaix/tiny-random-gpt2-torchscript")

optimum/intel/ipex/modeling_base.py
```diff
@@ -240,129 +145,30 @@ def _from_pretrained(
...
-        else:
-            model_cache_path = hf_hub_download(repo_id=model_id, filename=file_name, **model_kwargs)
-            model_save_dir = Path(model_cache_path).parent
-
-        model = torch.jit.load(model_cache_path)
-        torch.jit.freeze(model.eval())
-
-        return cls(model, config=config, model_save_dir=model_save_dir, **kwargs)
+        model = cls.auto_model_class.from_pretrained(model_id, **kwargs)
+        return cls(model, config=model.config, export=True, **kwargs)
```
echarlaix, line 156: why would export be needed ?  

```suggestion
        return cls(model, config=model.config, **kwargs)
```

optimum/intel/ipex/modeling_base.py
echarlaix: also we need to update the documentation https://github.com/huggingface/optimum-intel/blob/ad8a4cb468b418c82d4b34dfce6ba6f1de323b18/docs/source/ipex/inference.mdx?plain=1#L18

jiqing-feng: done.

tests/ipex/test_modeling.py
jiqing-feng: Yes, I understand it now! The test model `Intel/tiny_random_llama2_ipex_model` is uploaded from `IPEXModelForCausalLM`.

optimum/intel/ipex/modeling_base.py
jiqing-feng: done.

optimum/intel/ipex/modeling_base.py
jiqing-feng: done.

jiqing-feng:
Hi @echarlaix  @IlyasMoutawwakil . Please review the new changes.

docs/source/ipex/inference.mdx
```diff
@@ -14,8 +14,8 @@ Optimum Intel can be used to load models from the [Hub](https://huggingface.co/m
 
 ## Loading
 
-You can load your model and apply IPEX optimizations (including weight prepacking and graph mode). For supported architectures like LLaMA, BERT and ViT, further optimizations will be applied by patching the model to use custom operators.
-For now, support is only enabled for CPUs and the original model will be exported via TorchScript. In the future `torch.compile` will be used and model exported via TorchScript will get deprecated.
+You can load your model and apply IPEX optimizations (apply torch.compile for non-generation tasks). For supported architectures like LLaMA, BERT and ViT, further optimizations will be applied by patching the model to use custom operators.
+For now, support is enabled for Intel CPU/GPU. The TorchScript is deprecated.
```
echarlaix, line 18:
```suggestion
For now, support is enabled for Intel CPU/GPU. Previous models converted to TorchScript will be deprecated in v1.22.
```

tests/ipex/test_modeling.py
echarlaix: thanks @jiqing-feng !

tests/ipex/test_pipelines.py
```diff
@@ -125,23 +125,24 @@ def test_fill_mask_pipeline_inference(self, model_arch):
...
             self.assertAlmostEqual(transformers_output[i]["score"], ipex_output[i]["score"], delta=1e-4)
 
     @parameterized.expand(TEXT_GENERATION_SUPPORTED_ARCHITECTURES)
     def test_text_generation_pipeline_inference(self, model_arch):
         model_id = MODEL_NAMES[model_arch]
-        transformers_generator = transformers_pipeline("text-generation", model_id)
-        ipex_generator = ipex_pipeline("text-generation", model_id, accelerator="ipex")
+        dtype = torch.float32
+        if IS_XPU:
+            dtype = torch.float16
```
echarlaix, line 137:
```suggestion
        dtype = torch.float16 if is_xpu_available else torch.float32
```

tests/ipex/utils_tests.py
```diff
@@ -11,8 +11,11 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+from transformers import is_torch_xpu_available
 
 
+IS_XPU = is_torch_xpu_available(check_device=True)
```
echarlaix, line 17:
```suggestion
is_xpu_available = is_torch_xpu_available(check_device=True)
```

tests/ipex/test_pipelines.py
```diff
@@ -56,7 +59,6 @@ class PipelinesIntegrationTest(unittest.TestCase):
         "gpt2",
         "gpt_neo",
         "gpt_neox",
-        "llama",
```
echarlaix, line 59: why not keep it ?

tests/ipex/test_modeling.py
```diff
@@ -473,38 +462,22 @@ def test_compare_to_transformers(self, model_arch):
...
+        model = self.IPEX_MODEL_CLASS.from_pretrained(model_id)
         preprocessor = AutoFeatureExtractor.from_pretrained(model_id)
         pipe = pipeline("image-classification", model=model, feature_extractor=preprocessor)
         outputs = pipe("http://images.cocodataset.org/val2017/000000039769.jpg")
         self.assertEqual(pipe.device, model.device)
         self.assertGreaterEqual(outputs[0]["score"], 0.0)
         self.assertTrue(isinstance(outputs[0]["label"], str))
-
-    @unittest.skipIf(is_ipex_version("<", "2.3.0"), reason="Only ipex version > 2.3.0 supports ipex model patching")
-    def test_patched_model(self):
```
echarlaix, line 498: can we keep this test ? (using a new model)

jiqing-feng:
Hi @echarlaix. I have fixed all your comments, please take a review. For the change of using generate instead of forward in causal lm tests, we need to pass a cache_class (IPEXPagedCache) in forward just like StaticCache. Otherwise, the past_key_values.update will raise an error because past_key_values is None. The only way to support forward with use_cache=True is to create a cache class in forward, but it's not reasonable because generate already created the cache class.  I would like to hear your opinion. Thanks!!

jiqing-feng:
Hi @echarlaix . We plan to merge this PR this year which means we need the review done before Xmas. Appreciate it if you could prioritize this PR. Thanks!

sywangyi:
> > For the change of using generate instead of forward in causal lm tests, we need to pass a cache_class (IPEXPagedCache) in forward just like StaticCache. Otherwise, the past_key_values.update will raise an error because past_key_values is None.
> 
> I'm pretty sure all calls to `past_key_values.update` in the `forward` are guarded with an `if past_key_values is not None`.

because if past_key_values is None and use_cache = True. DynamicCache will be created in modeling. see 
https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L863, as an example. we do not maintain dynamicCache logic in optimum-intel ipex modeling now. ipex modeling only support pagedCache.

jiqing-feng:
Hi @sywangyi . The `use_cache` is from `generation_config`, which means it's just a parameter for generation and should not block the model forward without `past_key_values`. We can only guarantee the inputs have a cache in generation if `use_cache=True`, but not for just calling forward. 

Hi @IlyasMoutawwakil . I have fixed it by your comment, please check if there are any changes required before merging. Thanks!

IlyasMoutawwakil (approved):

Issue #965, opened 2024-10-24
Follow-up request after #916 due to issues on the OpenVINO side


andrei-kochin:
T5 encoder is not working on a tiny model which prevents the model to be validated on a regular basis

https://github.com/huggingface/optimum-intel/pull/916#discussion_r1811266332

andrei-kochin:
@IlyasMoutawwakil  please assign this on @eaidova

eaidova:
blocked by https://github.com/openvinotoolkit/openvino/pull/27215

eaidova:
fixed as part of https://github.com/huggingface/optimum-intel/pull/1016

PR #1016, opened 2024-11-22
enable t5 in SD3 pipe testing and flux img2img and inpaint

eaidova:
# What does this PR do?
* enable text_encoder_3 in SD3 testing for resolving https://github.com/huggingface/optimum-intel/issues/965
* enable flux img2img pipeline (was introduced in diffusers 0.31.0, previous version was based on 0.30.0)
* added test that save_pretrained returns correct model_index.json

## Before submitting


AlexKoff88 (approved):

tests/openvino/test_diffusion.py
```diff
@@ -398,6 +408,7 @@ class OVPipelineForImage2ImageTest(unittest.TestCase):
     SUPPORTED_ARCHITECTURES = ["stable-diffusion", "stable-diffusion-xl", "latent-consistency"]
     if is_transformers_version(">=", "4.40.0"):
         SUPPORTED_ARCHITECTURES.append("stable-diffusion-3")
+        SUPPORTED_ARCHITECTURES.append("flux")
```
nikita-savelyevv, line 411: Why is flux addition conditioned on transformers version and not on `is_diffusers_version(">=", "0.31.0")` as above?

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -95,6 +95,13 @@
     FluxPipeline = StableDiffusionPipeline
 
 
+if is_diffusers_version(">=", "0.31.0"):
+    from diffusers import FluxImg2ImgPipeline, FluxInpaintPipeline
+else:
+    FluxImg2ImgPipeline = StableDiffusionImg2ImgPipeline
+    FluxInpaintPipeline = StableDiffusionInpaintPipeline
```
nikita-savelyevv, line 102: With this, will ov flux pipeline still work on `diffusers<0.31`?

nikita-savelyevv (commented):
After I run
```
optimum-cli export openvino -m katuni4ka/tiny-random-flux ./tmp
```

I don't see a model for `text_encoder_3`. Based on PR description I would expect that it should be there. Am I getting something wrong?

tests/openvino/test_diffusion.py
eaidova: problem in optimum-intel gone not from diffusers version, in github actions there is division by transformers version, diffusers always used the latest. probmel with models imports comes from saving text encoders that are part of transformers with upper version of tokenizers than supported in transformers < 4.40 as the result test will fail in precommit

optimum/intel/openvino/modeling_diffusion.py
nikita-savelyevv: I see. Can this be done then? 
```
    FluxImg2ImgPipeline = object
    FluxInpaintPipeline = object
```

If so, I think this is a better option because it is more clear that in such case we don't expect it to work/be instantiated.

nikita-savelyevv: Or even
```
    FluxImg2ImgPipeline = None
    FluxInpaintPipeline = None
```

eaidova: applied

PR #1020, opened 2024-11-22
fix imports in openvino cli command

eaidova:
# What does this PR do?

when user has optimum-intel installed and not installed openvino any optimum-cli command 
(e.g. `optimum-cli export onnx  --model akreal/tiny-random-bert test-bert`
failed with
```
  File "/home/ea/work/py311/lib/python3.11/site-packages/optimum/exporters/openvino/__main__.py", line 28, in <module>
    from openvino.runtime import Core, Type, save_model
ModuleNotFoundError: No module named 'openvino
```

The root cause is wrong place of framework-specific import in module with CLI command.
Also moved "guilty" function to correct place where there is no openvino imports for convenience of its reusage

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -21,7 +21,6 @@ from typing import TYPE_CHECKING, Optional
 from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
 
 from ...exporters import TasksManager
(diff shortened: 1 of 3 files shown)


AlexKoff88:
@hshen14, we fixed the issue that we discussed but there are other failed tests. Can your team take a look at them?

AlexKoff88 (approved):

PR #1021, opened 2024-11-22
Add new features from NNCF 2.14

nikita-savelyevv:
# What does this PR do?

Changes:
- Added API for running weights compression with LoRA Correction algorithm. Usage: `optimum-cli export openvino ... --lora-correction`, or `OVWeightQuantizationConfig(..., lora_correction=True)`
- Added API for selecting weights compression backup precision. Usage: `optimum-cli export openvino ... --backup-precision <none|int8_sym|int8_asym>` or `OVWeightQuantizationConfig(..., backup_precision=<"none"|"int8_sym"|"int8_asym">)`.
- Added an "auto" dataset option for compressing `OVModelForCausalLM`. In this case `num_samples` strings are generated according to https://github.com/openvinotoolkit/nncf/blob/release_v2140/nncf/data/generators.py#L23 . Usage: `optimum-cli export openvino ... --dataset auto` or 
`OVWeightQuantizationConfig(..., dataset="auto")`.

## Before submitting

### DIFF
--- a/docs/source/openvino/export.mdx
+++ b/docs/source/openvino/export.mdx
@@ -29,7 +29,7 @@ optimum-cli export openvino --model local_llama --task text-generation-with-past
 
 Check out the help for more options:
 
-```
+```text
 usage: optimum-cli export openvino [-h] -m MODEL [--task TASK] [--framework {pt,tf}] [--trust-remote-code]
                                    [--weight-format {fp32,fp16,int8,int4,mxfp4,nf4}]
                                    [--library {transformers,diffusers,timm,sentence_transformers,open_clip}]
--- a/docs/source/openvino/export.mdx
+++ b/docs/source/openvino/export.mdx
@@ -29,7 +29,7 @@ optimum-cli export openvino --model local_llama --task text-generation-with-past
 
 Check out the help for more options:
 
-```bash
+```
 usage: optimum-cli export openvino [-h] -m MODEL [--task TASK] [--framework {pt,tf}] [--trust-remote-code]
                                    [--weight-format {fp32,fp16,int8,int4,mxfp4,nf4}]
                                    [--library {transformers,diffusers,timm,sentence_transformers,open_clip}]
@@ -50,14 +50,14 @@ Required arguments:
 
 Optional arguments:
   --task TASK           The task to export the model for. If not specified, the task will be auto-inferred based on
-                        the model. Available tasks depend on the model, but are among: ['image-to-image', 'image-
-                        segmentation', 'inpainting', 'sentence-similarity', 'text-to-audio', 'image-to-text',
-                        'automatic-speech-recognition', 'token-classification', 'text-to-image', 'audio-
-                        classification', 'feature-extraction', 'semantic-segmentation', 'masked-im', 'audio-xvector',
+                        the model. Available tasks depend on the model, but are among: ['image-to-image',
+                        'image-segmentation', 'inpainting', 'sentence-similarity', 'text-to-audio', 'image-to-text',
+                        'automatic-speech-recognition', 'token-classification', 'text-to-image', 'audio-classification',
+                        'feature-extraction', 'semantic-segmentation', 'masked-im', 'audio-xvector',
                         'audio-frame-classification', 'text2text-generation', 'multiple-choice', 'depth-estimation',
                         'image-classification', 'fill-mask', 'zero-shot-object-detection', 'object-detection',
                         'question-answering', 'zero-shot-image-classification', 'mask-generation', 'text-generation',
-                        'text-classification']. For decoder models, use `xxx-with-past` to export the model using past
+                        'text-classification']. For decoder models, use 'xxx-with-past' to export the model using past
                         key values in the decoder.
   --framework {pt,tf}   The framework to use for the export. If not provided, will attempt to use the local
                         checkpoint's original framework or what is available in the environment.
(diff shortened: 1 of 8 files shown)


nikita-savelyevv:
@ljaljushkin  @andreyanufr @l-bat  please take a look

optimum/commands/export/openvino.py
```diff
@@ -167,12 +183,22 @@ def parse_args_openvino(parser: "ArgumentParser"):
             "applying GPTQ takes additional memory and time."
         ),
     )
+    optional_group.add_argument(
+        "--lora",
```
ljaljushkin, line 187: I'm afraid this name might be confusing, people can think about efficient fine-tuning, but there's no gradient update and no tuning for a specific task. I'd explicitly name it `lora-correction` and it would be aligned with `nncf` name as well.

optimum/intel/openvino/configuration.py
```diff
@@ -464,14 +496,17 @@ def post_init(self):
...
-            if self.scale_estimation:
-                raise ValueError(
-                    f"The Scale Estimation algorithm is not supported for '{self.weight_format}' weight format"
-                )
-            if self.weight_format == "mxfp4" and self.gptq:
-                raise ValueError("The GPTQ algorithm is not supported for 'mxfp4' weight format")
+            if self.weight_format == "mxfp4":
+                if self.quant_method == OVQuantizationMethod.AWQ:
+                    raise ValueError("The AWQ algorithm is not supported for 'mxpf4' weight format")
+                if self.scale_estimation:
```
ljaljushkin, line 502: Why not extend `OVQuantizationMethod`? Can also substitute `self.quant_method.name` to a single error message.

ljaljushkin (changes requested):

optimum/intel/openvino/configuration.py
nikita-savelyevv: > Why not extend OVQuantizationMethod?

Then we won't be able to apply AWQ + LoRA correction at the same time. I'm afraid, we shouldn't have relied on `OVQuantizationMethod` for enabling AWQ in the first place. But still stick to it now for backward-compatability.

optimum/commands/export/openvino.py
nikita-savelyevv: Thanks for the suggestion! I agree. Changed.

optimum/commands/export/openvino.py
```diff
@@ -117,14 +117,30 @@ def parse_args_openvino(parser: "ArgumentParser"):
...
         "--dataset",
         type=str,
         default=None,
         help=(
             "The dataset used for data-aware compression or quantization with NNCF. "
-            "You can use the one from the list ['wikitext2','c4','c4-new'] for language models "
-            "or ['conceptual_captions','laion/220k-GPT4Vision-captions-from-LIVIS','laion/filtered-wit'] for diffusion models."
+            "For language models you can use the one from the list ['auto','wikitext2','c4','c4-new']. With 'auto' the "
+            "dataset will be collected from model's generations. "
+            "For diffusion models -- on of ['conceptual_captions','laion/220k-GPT4Vision-captions-from-LIVIS',"
```
andreyanufr, line 141: Not clear what means ' -- on of'

nikita-savelyevv: Updated

docs/source/openvino/export.mdx
```diff
@@ -82,10 +83,18 @@ Optional arguments:
...
-                        from the list ['wikitext2','c4','c4-new'] for language models or
-                        ['conceptual_captions','laion/220k-GPT4Vision-captions-from-LIVIS','laion/filtered-wit'] for
-                        diffusion models.
+  --backup-precision {none,int8_sym,int8_asym}
+                        Defines a backup precision for mixed-precision weight compression. Only valid for int4 weight
+                        format. If not provided, backup precision is int8_asym. 'none' stands for original floating-
+                        point precision of the model weights, in this case weights are retained in their original
+                        precision without any quantization. 'int8_sym' stands for 8-bit integer symmetric quantization
+                        without zero point. 'int8_asym' stands for 8-bit integer asymmetric quantization with a
+                        typical non-fixed zero point.
```
AlexKoff88, line 92: It is not clear what `non-fixed` zero point means. I would rephrase it to `with zero point per each quantization group`.

docs/source/openvino/export.mdx
```diff
@@ -98,6 +107,9 @@ Optional arguments:
   --gptq                Indicates whether to apply GPTQ algorithm that optimizes compressed weights in a layer-wise
                         fashion to minimize the difference between activations of a compressed and original layer.
                         Please note, that applying GPTQ takes additional memory and time.
+  --lora-correction     Indicates whether to apply LoRA Correction algorithm. When enabled, this algorithm mitigates
```
AlexKoff88, line 110: I would add additional clarification what it costs and why it is helpful, e.g. `When enabled, this algorithm introduces low-rank adaptation layers in the model that can recover accuracy after weight compression at some cost of inference latency. Please note, ...`

docs/source/openvino/export.mdx
nikita-savelyevv: Updated

docs/source/openvino/export.mdx
nikita-savelyevv: Updated

PR #1026, opened 2024-11-25
Relaxed requirements for OpenVINO version

AlexKoff88:
@helena-intel, I think this is something we agreed on.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ QUALITY_REQUIRE = ["black~=23.1", "ruff==0.4.4"]
 
 EXTRAS_REQUIRE = {
     "nncf": ["nncf>=2.11.0"],
-    "openvino": ["nncf>=2.11.0", "openvino==2024.5.0", "openvino-tokenizers==2024.5.0"],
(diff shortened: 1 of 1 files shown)


eaidova (approved):

setup.py
```diff
@@ -62,7 +62,7 @@
 
 EXTRAS_REQUIRE = {
     "nncf": ["nncf>=2.11.0"],
-    "openvino": ["nncf>=2.11.0", "openvino==2024.5.0", "openvino-tokenizers==2024.5.0"],
+    "openvino": ["nncf>=2.11.0", "openvino>=2024.4.0", "openvino-tokenizers>=2024.4.0"],
```
eaidova, line 65: 2024.4.0 or 2024.5.0 is lowerbound should be? I see in Helena's PR 2024.5 https://github.com/huggingface/optimum-intel/pull/1005

but both variants works for me

eaidova: 2024.4.1 was not released and is not planned to be as I know, only prelrease package exists after that infra switched for preparing 2024.5

AlexKoff88: I am oik with 2024.5. Closing this PR

PR #1028, opened 2024-11-25
fix model removal from tmp dir windows

eaidova:
# What does this PR do?

Additional fixes for issue with removing model from tmp dir on Windows:
1. For python3.11+ cleanup stage called twice with ignore_cleanup_erros=True (in our tmp dir class) and ignore_cleanup_erros=False (comes from base tmp dir class) that may fail if we use model with api
2. For models that consist from several submodels where not all of them big enough for compression by default based on model size (e.g. vlm with small vision encoder and big llm part) added removal submodel after reading if compression will be not applied.

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -445,11 +445,13 @@ def main_export(
         else:
             quantization_config = ov_config.quantization_config
         if quantization_config is None:
+            del submodel
+            gc.collect()
             continue
 
         if not is_nncf_available():
             raise ImportError("Quantization of the weights requires nncf, please install it with `pip install nncf`")
-
+ 
         from optimum.intel.openvino.quantization import _weight_only_quantization
 
         _weight_only_quantization(submodel, quantization_config)
@@ -457,6 +459,7 @@ def main_export(
         compressed_submodel_path = submodel_path.parent / f"{submodel_path.stem}_compressed.xml"
         save_model(submodel, compressed_submodel_path, compress_to_fp16=False)
         del submodel
+        gc.collect()
 
         submodel_path.unlink()
         submodel_path.with_suffix(".bin").unlink()
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -432,6 +432,7 @@ def main_export(
             for op in submodel.get_ops():
                 if op.get_type_name() == "Constant" and op.get_element_type() in [Type.f16, Type.f32, Type.bf16]:
                     num_parameters += reduce(operator.mul, op.shape, 1)
+                del op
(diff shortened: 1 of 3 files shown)


optimum/exporters/openvino/__main__.py
```diff
@@ -445,18 +445,21 @@ class StoreAttr(object):
...
             quantization_config = ov_config.quantization_config
         if quantization_config is None:
+            del submodel
+            gc.collect()
             continue
 
         if not is_nncf_available():
             raise ImportError("Quantization of the weights requires nncf, please install it with `pip install nncf`")
-
+
```
eaidova, line 454:
```suggestion

```

optimum/exporters/openvino/__main__.py
```diff
@@ -445,18 +446,22 @@ class StoreAttr(object):
...
 
         from optimum.intel.openvino.quantization import _weight_only_quantization
 
-        _weight_only_quantization(submodel, quantization_config)
+        compressed_submodel = _weight_only_quantization(submodel, quantization_config)
 
         compressed_submodel_path = submodel_path.parent / f"{submodel_path.stem}_compressed.xml"
-        save_model(submodel, compressed_submodel_path, compress_to_fp16=False)
+        save_model(compressed_submodel, compressed_submodel_path, compress_to_fp16=False)
+        del compressed_submodel
```
eaidova, line 462:
```suggestion
        save_model(submodel, compressed_submodel_path, compress_to_fp16=False)
```

optimum/intel/openvino/utils.py
```diff
@@ -485,13 +486,13 @@ def __init__(self, suffix=None, prefix=None, dir=None, ignore_cleanup_errors=Tru
         )
 
     @classmethod
-    def _cleanup(cls, name, warn_message, ignore_errors=False, delete=True):
+    def _cleanup(cls, name, warn_message, ignore_errors=True, delete=True):
```
nikita-savelyevv, line 489: Why `ignore_errors=True` is needed? Is it safe to do so?

nikita-savelyevv (approved):
Tested locally on Python 3.12.6. No issues with tmp directory arise.

optimum/intel/openvino/utils.py
eaidova: this is needed because on windows time of running destructor for ov.Model happens after tmp dir cleanup by system if in python script there is no sys.exit(). Another possible solution is create tmp dir handle without automatic cleanup and remove it after we make sure that ov model is destroyed, in OVModelForXxx class destructor, but previously on my end such approach failed on some complex models (looks like something removed several times) with Segmentation Fault. Unfortunatly, I'm not replicated such case without optimum-intel yet to get advice from python API engineering team. It will be better to switch to this approach, but it requires more time for investigation, i think.

AlexKoff88 (approved):

PR #1029, opened 2024-11-25
support minicpm3

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -192,6 +192,55 @@ class MiniCPMOpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
 
 
+class OVMiniCPM3DummyPastKeyValuesGenerator(MistralDummyPastKeyValuesGenerator):
+    def __init__(
+        self,
+        task: str,
+        normalized_config: NormalizedTextConfig,
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        sequence_length: int = DEFAULT_DUMMY_SHAPES["sequence_length"],
+        random_batch_size_range: Optional[Tuple[int, int]] = None,
+        random_sequence_length_range: Optional[Tuple[int, int]] = None,
+        **kwargs,
+    ):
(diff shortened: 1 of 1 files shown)


tests/openvino/test_modeling.py
```diff
@@ -1182,8 +1184,8 @@ def test_default_filling_attention_mask_and_position_ids(self):
         gc.collect()
 
     @parameterized.expand(SUPPORTED_ARCHITECTURES)
-    @pytest.mark.run_slow
-    @slow
+    # @pytest.mark.run_slow
+    # @slow
```
eaidova, line 1188:
```suggestion
    @pytest.mark.run_slow
    @slow
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3237,3 +3237,148 @@ def __init__(
...
+                k have the shape [batch_size, heads, seq_len, head_dim], then setting unsqueeze_dim=1 makes
+                cos[position_ids] and sin[position_ids] broadcastable to the shapes of q and k. Similarly, if q and k have
+                the shape [batch_size, seq_len, heads, head_dim], then set unsqueeze_dim=2.
+        Returns:
+            `tuple(torch.Tensor)` comprising of the query and key tensors rotated using the Rotary Position Embedding.
+        """
+        # cos = cos[position_ids].unsqueeze(unsqueeze_dim)
+        # sin = sin[position_ids].unsqueeze(unsqueeze_dim)
+        # q_embed = (q * cos) + (rotate_half(q) * sin)
+        # k_embed = (k * cos) + (rotate_half(k) * sin)
```
nikita-savelyevv, line 3280: I see quite a few commented lines here and below. Are they required?

nikita-savelyevv (approved):

optimum/exporters/openvino/model_patcher.py
eaidova: it is for demonstration how original code looks

eaidova, line 3280:
```suggestion
```

nikita-savelyevv: I see. I'm ok with keeping them then

PR #1030, opened 2024-11-25
use small model for testing textual inversion

eaidova:
# What does this PR do?

Current textual_inversion tests use runwayml/stable-diffusion-v1-5, that is more than 5GB in size and due to slowness and required amount of resources tests are executed only in slow tests scope. This PR propose to replace it with light-weight tiny model and adapter trained for it to speedup test and reduce required amount of memory for downloading test models

## Before submitting

### DIFF
--- a/tests/openvino/test_diffusion.py
+++ b/tests/openvino/test_diffusion.py
@@ -386,8 +386,8 @@ class OVPipelineForText2ImageTest(unittest.TestCase):
         # for now we only test for stable-diffusion
         # this is very slow and costly to run right now
 
-        model_id = "runwayml/stable-diffusion-v1-5"
-        ti_id = "sd-concepts-library/cat-toy"
+        model_id = "hf-internal-testing/tiny-stable-diffusion-torch"
+        ti_id = "katuni4ka/textual_inversion_cat"
 
         inputs = self.generate_inputs()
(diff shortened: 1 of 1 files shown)


tests/openvino/test_diffusion.py
```diff
@@ -379,15 +377,13 @@ def test_height_width_properties(self, model_arch: str):
...
         self.assertEqual(ov_pipeline.width, width)
 
-    @pytest.mark.run_slow
-    @slow
     @require_diffusers
     def test_textual_inversion(self):
         # for now we only test for stable-diffusion
         # this is very slow and costly to run right now
 
-        model_id = "runwayml/stable-diffusion-v1-5"
```
nikita-savelyevv, line 389: I guess the comment "this is very slow and costly to run right now" can be removed now

nikita-savelyevv (approved):
Thanks! `runwayml/stable-diffusion-v1-5` is more than 5GB in size so this allows to save us some downloading time and model cache size

tests/openvino/test_diffusion.py
eaidova: removed

IlyasMoutawwakil (approved):
LGTM !

PR #1039, opened 2024-11-27
Expand activation scaling to other submodels of diffusion pipelines

AlexKoff88:


tests/openvino/test_export.py
```diff
@@ -140,6 +140,18 @@ def _openvino_export(
                     self.assertTrue(
                         ov_model.vae_decoder.model.has_rt_info(["runtime_options", "ACTIVATIONS_SCALE_FACTOR"])
                     )
+                    if hasattr(ov_model, "text_encoder") and ov_model.text_encoder:
```
eaidova, line 143: should we also check text_encoder_2 and text_encoder_3 models too?

eaidova (approved):

PR #1040, opened 2024-11-27
[OV] Quantization of Whisper pipeline

nikita-savelyevv:
# What does this PR do?

- Added full quantization of Whisper pipeline trough python API. All three models (encoder, decoder, decoder with past) are fully quantized, including activations.
- Added calibration dataset collection for speech-to-text models from a string. Currently only `librispeech` is supported.

Some demo generations are below.

| Model                          | Quantized | Inference Time (sec.) | Transcription                                                                       |
|--------------------------------|-----------|----------------|-------------------------------------------------------------------------------------|
| openai/whisper-tiny            | False     | 0.338          | Joel Keaton disapproved of films and buster also had reservations about the medium. |
| openai/whisper-tiny            | True      | 0.271 (0.80x)  | Jo Kitson disapproved of films and buster also had reservations about the medium.   |
| openai/whisper-small           | False     | 1.069          | Joe Keaton disapproved of films and Buster also had reservations about the medium.  |
| openai/whisper-small           | True      | 0.834 (0.78x)  | Jow Keeton disapproved of films and Buster also had reservations about the medium.  |
| distil-whisper/distil-large-v3 | False     | 1.865          | Joe Keaton disapproved of films and Buster also had reservations about the medium.  |
| distil-whisper/distil-large-v3 | True      | 1.180 (0.63x)  | J. Keeton disapproved of films and Buster also had reservations about the median.   |
| openai/whisper-large-v3-turbo  | False     | 3.475          | Joel Keaton disapproved of films and Buster also had reservations about the medium. |
| openai/whisper-large-v3-turbo  | True      | 2.254 (0.65x)  | Joe Keaton disapproved of films and Buster also had reservations about the medium.  |
| openai/whisper-large-v2        | False     | 5.300          | Joel Keaton disapproved of films and Buster also had reservations about the medium. |
| openai/whisper-large-v2        | True      | 3.157 (0.60x)  | Joe Keaton disapproved of films and Buster also had reservations about the medium.  |

The code:

```python
import time
from pathlib import Path

import numpy as np
from datasets import load_dataset

from optimum.intel.openvino import OVConfig, OVModelForSpeechSeq2Seq, OVQuantizationConfig, OVQuantizer


model_id = "openai/whisper-tiny"
# model_id = "openai/whisper-base"
# model_id = "openai/whisper-small"
# model_id = "distil-whisper/distil-large-v3"
# model_id = "openai/whisper-large-v3-turbo"
# model_id = "openai/whisper-large-v2"

save_dir = Path("./whisper") / model_id.split("/")[-1]


def extract_input_features(sample):
    audio = sample["audio"]["array"]
    sampling_rate = sample["audio"]["sampling_rate"]
    if sampling_rate != 16000:
        duration = audio.shape[0] / sampling_rate
        resampled_data = np.zeros(shape=(int(duration * 16000)), dtype=np.float32)
        x_old = np.linspace(0, duration, audio.shape[0], dtype=np.float32)
        x_new = np.linspace(0, duration, resampled_data.shape[0], dtype=np.float32)
        audio = np.interp(x_new, x_old, audio)

    input_features = processor(audio, sampling_rate=16000, return_tensors="pt").input_features

    return input_features

# Example from common voice
dataset = load_dataset("mozilla-foundation/common_voice_13_0", "en", split="test", streaming=True)
sample = next(iter(dataset))

# Example from librispeech (subset of librispeech is also used during calibration data collection)
# dataset = load_dataset("hf-internal-testing/librispeech_asr_dummy", "clean", split="validation")
# sample = dataset[0]

input_features = extract_input_features(sample)

# Load & run original model
original_model_path = save_dir / "original"
if not original_model_path.exists():
    ov_model = OVModelForSpeechSeq2Seq.from_pretrained(
        model_id,
        compile=False,
        load_in_8bit=False,
    )
    ov_model.save_pretrained(original_model_path)
ov_model = OVModelForSpeechSeq2Seq.from_pretrained(original_model_path, compile=True)

start_time = time.perf_counter()
transcription = processor.batch_decode(ov_model.generate(input_features), skip_special_tokens=True)[0]
print(f'\nOriginal model transcription: "{transcription.strip()}"')
print(f"Generation time: {time.perf_counter() - start_time:.3f}\n")

# Obtain & run the quantized model
quantized_model_path = save_dir / "quantized"
if not quantized_model_path.exists():
    OVQuantizer(ov_model).quantize(
        ov_config=OVConfig(
            quantization_config=OVQuantizationConfig(
                num_samples=10,
                dataset="librispeech",
                processor=model_id,
                smooth_quant_alpha=0.9,
            )
        ),
        save_directory=quantized_model_path,
    )
ov_model = OVModelForSpeechSeq2Seq.from_pretrained(quantized_model_path, compile=True)

start_time = time.perf_counter()
transcription = processor.batch_decode(ov_model.generate(input_features), skip_special_tokens=True)[0]
print(f'\nQuantized model transcription: "{transcription.strip()}"')
print(f"Generation time: {time.perf_counter() - start_time:.3f}\n")
```

## Before submitting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ EXTRAS_REQUIRE = {
     "diffusers": ["diffusers"],
     "quality": QUALITY_REQUIRE,
     "tests": TESTS_REQUIRE,
-    "tests-openvino": ["datasets[audio]>=1.4.0"]
+    "tests-openvino": ["datasets[audio]>=1.4.0"],
 }
 
 setup(


docs/source/openvino/optimization.mdx
```diff
@@ -166,6 +166,26 @@ calibration_dataset = quantizer.get_calibration_dataset(
 The `quantize()` method applies post-training static quantization and export the resulting quantized model to the OpenVINO Intermediate Representation (IR). The resulting graph is represented with two files: an XML file describing the network topology and a binary file describing the weights. The resulting model can be run on any target Intel device.
 
 
+#### Whisper Quantization
+
+
```
AlexKoff88, line 171: Extra new line

optimum/intel/openvino/configuration.py
```diff
@@ -535,6 +552,12 @@ def __init__(
         model_type: str = "transformer",
         fast_bias_correction: bool = True,
         overflow_fix: str = "disable",
+        dataset: Optional[str] = None,
+        tokenizer: Optional[str] = None,
+        processor: Optional[str] = None,
+        trust_remote_code: bool = False,
+        matmul_sq_alpha: Optional[float] = None,
```
AlexKoff88, line 559: Can you please rename `matmul_sq_alpha` to `smoothquant_alpha`

docs/source/openvino/optimization.mdx
```diff
@@ -166,6 +166,26 @@ calibration_dataset = quantizer.get_calibration_dataset(
 The `quantize()` method applies post-training static quantization and export the resulting quantized model to the OpenVINO Intermediate Representation (IR). The resulting graph is represented with two files: an XML file describing the network topology and a binary file describing the weights. The resulting model can be run on any target Intel device.
 
 
+#### Whisper Quantization
```
AlexKoff88, line 169: Should not it be `Speech-to-text Models Quantization`

optimum/intel/openvino/configuration.py
```diff
@@ -557,11 +580,44 @@ def __init__(
...
+                A transformers processor used to process inputs for multi-modal models. You can pass either:
+                    - A string, the *model id* of a predefined processor hosted inside a model repo on huggingface.co.
+                    - A path to a *directory* containing files required by the processor, for instance saved
+                        using the [`~AutoProcessor.save_pretrained`] method, e.g., `./my_model_directory/`.
+            trust_remote_code (`bool`, defaults to `False`):
+                Allows to use custom code for the modeling hosted in the model repository. This option should only be set
+                for repositories you trust and in which you have read the code, as it will execute on your local machine
+                arbitrary code present in the model repository.
+            matmul_sq_alpha (`float`, *optional*):
+                SmoothQuant alpha parameter for MatMul layers.
```
AlexKoff88, line 602:
```suggestion
                SmoothQuant alpha parameter that improves the distribution of activations before MatMul layers and reduces quantization error.
```

optimum/intel/openvino/quantization.py
```diff
@@ -809,8 +808,75 @@ def _prepare_visual_causal_lm_dataset(self, config: OVWeightQuantizationConfig):
         calibration_dataset = nncf.Dataset(calibration_dataset)
         return calibration_dataset
 
+    def _prepare_whisper_dataset(self, config: OVQuantizationConfigBase):
```
AlexKoff88, line 811:
```suggestion
    def _prepare_whisper_calibration_data(self, config: OVQuantizationConfigBase):
```

AlexKoff88:
@nikita-savelyevv, overall it looks fine. Thank you.

The one thing that is missing here is quantization via optimum-cli. We need to make sure that it works.

nikita-savelyevv:
> @nikita-savelyevv, overall it looks fine. Thank you.
> 
> The one thing that is missing here is quantization via optimum-cli. We need to make sure that it works.

Thanks! I would like to add it in a separate PR. https://github.com/huggingface/optimum-intel/pull/1040#issuecomment-2506546235

optimum/intel/openvino/configuration.py
nikita-savelyevv: Won't there be a confusion with `convolution_sq_alpha`?

AlexKoff88: agreed offline to keep only one parameter.

AlexKoff88:
> > @nikita-savelyevv, overall it looks fine. Thank you.
> > The one thing that is missing here is quantization via optimum-cli. We need to make sure that it works.
> 
> Thanks! I would like to add it in a separate PR. [#1040 (comment)](https://github.com/huggingface/optimum-intel/pull/1040#issuecomment-2506546235)

I am fine with this.

docs/source/openvino/optimization.mdx
nikita-savelyevv: Removed

docs/source/openvino/optimization.mdx
nikita-savelyevv: Renamed. The reason I named it Whisper is because as I understand it is the only speech-to-text model currently supported.
cc @eaidova

optimum/intel/openvino/quantization.py
nikita-savelyevv: Renamed. Also renamed other similar methods.

AlexKoff88 (approved):

AlexKoff88:
@echarlaix, @IlyasMoutawwakil, here we enabled quantization of the Whisper model that has worked for quite a while but it is not exposed still. Please take a look.

eaidova (approved):

AlexKoff88:
@IlyasMoutawwakil, @echarlaix, PR is ready for review from your side.

setup.py
```diff
@@ -68,6 +68,7 @@
     "diffusers": ["diffusers"],
     "quality": QUALITY_REQUIRE,
     "tests": TESTS_REQUIRE,
+    "tests-openvino": ["datasets[audio]>=1.4.0"],
```
nikita-savelyevv, line 71: `datasets[audio]` is required for collecting calibration data for Whisper model. 

I had to add a separate test extra for this because when installing `datasets[audio]` in a general tests extra, INC tests fail. `datasets[audio]` require `numba` package which leads to INC tests failure.

IlyasMoutawwakil: I would prefer if it's only installed in the openvino CI rather than have it being an extra

IlyasMoutawwakil: also is all of `datasets[audio]` necessary ?

nikita-savelyevv: > I would prefer if it's only installed in the openvino CI rather than have it being an extra

But this way if someone wants to run tests locally, it will be less clear why did they fail, because with this it is clearly stated that these requirements are needed for openvino testing and otherwise one would have to look into workflow files. Do you think it's worth it?

> also is all of datasets[audio] necessary ?

A regular `datasets` package is installed together with optimum-intel https://github.com/huggingface/optimum-intel/blob/main/setup.py#L33 . `datasets[audio]` additionally installs these these three packages https://github.com/huggingface/datasets/blob/main/setup.py#L144-L146 . The first two of them (`soundfile` and `librosa`) are necessary for the added test. The third one (`soxr`) is not. But I decided to install it as `datasets[audio]` as a safer option, for example in case some additional version requirements on `soundfile` or `librosa` will appear at `datasets[audio]`, or something like that.

IlyasMoutawwakil: I prefer adding `datasets[audio]` to test require in general, and `tbb` as well, to fix numba issue https://github.com/intel/neural-compressor/issues/2031#issuecomment-2419904575

nikita-savelyevv: I believe INC bit packing works differently whether `numba` package is installed or not https://github.com/intel/neural-compressor/blob/master/neural_compressor/torch/algorithms/weight_only/modules.py#L456-L457 . If we add a `numba` requirement to general tests requirement, then it will have some effect on INC tests, i.e. different methods will be called. This seems like a change that should be addressed in a separate PR.

nikita-savelyevv: @IlyasMoutawwakil What do you think? 

I can open a separate PR after this one where we add `datasets[audio]` to general tests requirements and invite someone from INC side for a review.

IlyasMoutawwakil (approved):
LGTM ! agree to merge but would like to get rid of the `tests-openvino` extra as it complicates and fragments the developer experience

AlexKoff88:
Merging as the test failures unrelated to the changes in the PR.

PR #1048, opened 2024-12-04
import OVQuantizer only if it is required in VLM

eaidova:
# What does this PR do?

in some cases, import OVQuantizer may lead to issues with onnx (import check_dummy_inputs trigger whole initialization of onnx part in optimum), for avoiding this issue in analogy with OVModelForCausalLM I suggest to import it conditionally only in place where it is required as short-term solution. As long-term, we need to extract common part of helper functions from optimum.exporters.onnx to general optimum.exporters.utils (imports will be switched after merge https://github.com/huggingface/optimum/pull/2114 in separated PR)

## Before submitting


AlexKoff88 (approved):

AlexKoff88:
@echarlaix, @IlyasMoutawwakil, this is a hotfix.

IlyasMoutawwakil (approved):

echarlaix (approved):

PR #1049, opened 2024-12-04
Support AWQ models

mvafin:
# What does this PR do?

Add support for AWQ models after this support was added by OpenVINO in 2024.6 (https://github.com/openvinotoolkit/openvino/pull/27859)
Patching for 16bit can be also used for GPTQ and AWQ models to support FP16/BF16 regions in the model.

Fixes # (issue)

## Before submitting


eaidova (approved):

AlexKoff88:
@mvafin, can we have a test for this, e.g. with "hf-internal-testing/Mixtral-tiny-AWQ" or any other dummy model?

IlyasMoutawwakil:
> @mvafin, can we have a test for this, e.g. with "hf-internal-testing/Mixtral-tiny-AWQ" or any other dummy model?

yes a test would be great

mvafin:
@AlexKoff88 @IlyasMoutawwakil Test added

AlexKoff88:
@mvafin, some of the tests failed:
```
=========================== short test summary info ============================
FAILED tests/openvino/test_modeling.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_18_mixtral_awq - ValueError: `.float()` is not supported for quantized model. Please use the model as it is, since the model has already been casted to the correct `dtype`.
FAILED tests/openvino/test_modeling.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_23_qwen - AssertionError: Torch not compiled with CUDA enabled
FAILED tests/openvino/test_modeling.py::OVModelForSeq2SeqLMIntegrationTest::test_compare_with_and_without_past_key_values - AssertionError: False is not true : With pkv latency: 227.547 ms, without pkv latency: 240.161 ms, speedup: 1.055
FAILED tests/openvino/test_modeling.py::OVModelForSTFeatureExtractionIntegrationTest::test_compare_to_transformers_0_st_bert - AssertionError: Torch not compiled with CUDA enabled
FAILED tests/openvino/test_modeling.py::OVModelForSTFeatureExtractionIntegrationTest::test_compare_to_transformers_1_st_mpnet - AssertionError: Torch not compiled with CUDA enabled
```

Can you please check on your side?

mvafin:
> @mvafin, some of the tests failed:
> 
> ```
> =========================== short test summary info ============================
> FAILED tests/openvino/test_modeling.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_18_mixtral_awq - ValueError: `.float()` is not supported for quantized model. Please use the model as it is, since the model has already been casted to the correct `dtype`.
> FAILED tests/openvino/test_modeling.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_23_qwen - AssertionError: Torch not compiled with CUDA enabled
> FAILED tests/openvino/test_modeling.py::OVModelForSeq2SeqLMIntegrationTest::test_compare_with_and_without_past_key_values - AssertionError: False is not true : With pkv latency: 227.547 ms, without pkv latency: 240.161 ms, speedup: 1.055
> FAILED tests/openvino/test_modeling.py::OVModelForSTFeatureExtractionIntegrationTest::test_compare_to_transformers_0_st_bert - AssertionError: Torch not compiled with CUDA enabled
> FAILED tests/openvino/test_modeling.py::OVModelForSTFeatureExtractionIntegrationTest::test_compare_to_transformers_1_st_mpnet - AssertionError: Torch not compiled with CUDA enabled
> ```
> 
> Can you please check on your side?

Fixed

IlyasMoutawwakil:
still getting awq cuda errors

mvafin:
> still getting awq cuda errors

That might be because optimum is not tested with 2024.6 version of openvino

AlexKoff88 (approved):

.github/workflows/test_openvino.yml
```diff
@@ -50,6 +50,11 @@ jobs:
         name: Install specific dependencies and versions required for older transformers
         run: |
           pip install transformers==${{ matrix.transformers-version }} accelerate==0.* peft==0.13.* diffusers==0.30.* transformers_stream_generator
+      
+      - if: ${{ matrix.transformers-version == 'latest' && matrix.test-pattern == '*modeling*'}}
+        name: Install auto-gptq, autoawq
+        run: |
+          pip install auto-gptq autoawq --extra-index-url https://download.pytorch.org/whl/cpu
```
IlyasMoutawwakil, line 57: these are not valid extra urls for auto-gptq and awq

IlyasMoutawwakil:
The PR in its current form seems to degrade the user experience greatly. My understanding is that it suggests we pass the responsibility of patching autoawq and autogptq to the user ? like writing their own `mock_torch_cuda_is_available` and `patch_awq_for_inference` to use autogptq/autoawq. Please correct me if I'm wrong.

.github/workflows/test_openvino.yml
eaidova: this is neede for preventing reinstalling torch with cuda during installing third-party, packages themselves should be installed from regular source, torch-dependent libs (the difference from --index-url and --extra-index-url that first redefine source index completely, the second one parameter used for usage index URL as additional source if library present in that source) will be tried to install from torch cpu url

eaidova:
> The PR in its current form seems to degrade the user experience greatly. My understanding is that it suggests we pass the responsibility of patching autoawq and autogptq to the user ? like writing their own `mock_torch_cuda_is_available` and `patch_awq_for_inference` to use autogptq/autoawq. Please correct me if I'm wrong.

these manual patches requires only for running original torch model without cuda in test environment, general flow for model conversion does not requires these patches, so no any impact on optimum-intel user experience as running original model in torch is out of our responsibility

IlyasMoutawwakil:
> these manual patches requires only for running original torch model without cuda in test environment

Thanks for the explanation ! I thought it's also required for the quantization process on cpu but I see that patch is still in its place.

optimum/exporters/openvino/__main__.py
```diff
@@ -247,7 +247,10 @@ def main_export(
             trust_remote_code=trust_remote_code,
         )
         quantization_config = getattr(config, "quantization_config", None)
-        do_gptq_patching = quantization_config and quantization_config["quant_method"] == "gptq"
+        supported_quant_methods = ["gptq"]
+        if is_openvino_version(">=", "2024.6.0"):
+            supported_quant_methods.append("awq")
+        do_gptq_patching = quantization_config and quantization_config["quant_method"] in supported_quant_methods
```
IlyasMoutawwakil, line 253: why patch `auto-gptq` lib when method is `awq` ?

tests/openvino/test_modeling.py
```diff
@@ -978,20 +978,30 @@ def test_compare_to_transformers(self, model_arch):
...
+            with patch_awq_for_inference("awq" in model_arch):
+                transformers_outputs = transformers_model(**tokens)
 
         # Compare tensor outputs
         atol = 1e-3 if model_arch == "minicpm" else 1e-4
+        # quantized models have higher tolerance
+        if "awq" in model_arch:
+            atol = 1e-2
+        elif "gptq" in model_arch:
+            atol = 0.6
```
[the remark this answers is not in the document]
eaidova: slow test with beam search testing and generated sequences in the same test are equal. logits can be represented in denormalized format and it is hard to say which threshold is meaningful as we do not know range of values. If I apply softmax on output I see 1-e5 difference between torch and ov result, argsort of both tensors shows small permutations after 1000 tokens in logits probs (it is just misplacing 1027 and 1028 tokens), so I believe it is enough accurate to avoid impact on generated strings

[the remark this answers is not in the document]
eaidova: added skipping logits comparision for quantized models

IlyasMoutawwakil (approved):
LGTM ! Merging approved as long as tests are passing ! Thanks @eaidova for following up on this.

eaidova:
@IlyasMoutawwakil  looks like all tests passing now, let me fix code style after rebasing and I think everything is done for this PR

AlexKoff88:
Merging as we have few approvals.

PR #1053, opened 2024-12-08
ipex Page attn xpu support bug fix

kaixuanliu:


kaixuanliu:
@echarlaix @IlyasMoutawwakil , pls help review

tests/ipex/test_modeling.py
```diff
@@ -50,6 +50,7 @@
 
 SEED = 42
 torch.use_deterministic_algorithms(True)
+device_map = "xpu:0" if IS_XPU_AVAILABLE else None
```
IlyasMoutawwakil, line 53: it would make sense to define this as "DEVICE", and then use it both for `device_map=DEVICE` and in the `.to(DEVICE)`
which should robustify the tests when xpu is available.

kaixuanliu: Have updated the code per your suggestion.

tests/ipex/test_modeling.py
```diff
@@ -50,6 +50,7 @@
 
 SEED = 42
 torch.use_deterministic_algorithms(True)
+DEVICE = "xpu:0" if IS_XPU_AVAILABLE else None
```
IlyasMoutawwakil, line 53:
```suggestion
DEVICE = "xpu:0" if IS_XPU_AVAILABLE else "cpu"
```

IlyasMoutawwakil (approved):

PR #1054, opened 2024-12-09
Enable Text2text task on ipex

jiqing-feng:
Add IPEXModelForSeq2SeqLM support, optimized by torch.compile. It forced `attn_implementation="static"` to be compatible with torch.compile.

### DIFF
--- a/docs/source/ipex/inference.mdx
+++ b/docs/source/ipex/inference.mdx
@@ -14,7 +14,7 @@ Optimum Intel can be used to load models from the [Hub](https://huggingface.co/m
 
 ## Loading
 
-You can load your model and apply IPEX optimizations (apply torch.compile for non-generation tasks). For supported architectures like LLaMA, BERT and ViT, further optimizations will be applied by patching the model to use custom operators.
+You can load your model and apply IPEX optimizations (apply torch.compile except text-generation tasks). For supported architectures like LLaMA, BERT and ViT, further optimizations will be applied by patching the model to use custom operators.
 For now, support is enabled for Intel CPU/GPU. Previous models converted to TorchScript will be deprecated in v1.22.
 
 ```diff
--- a/docs/source/ipex/inference.mdx
+++ b/docs/source/ipex/inference.mdx
@@ -43,3 +43,4 @@ As shown in the table below, each task is associated with a class enabling to au
 | `IPEXModelForMaskedLM`               | `fill-mask`                          |
 | `IPEXModelForAudioClassification`    | `audio-classification`               |
 | `IPEXModelForCausalLM`               | `text-generation`                    |
+| `IPEXModelForSeq2SeqLM`              | `text2text-generation`               |
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -186,7 +186,7 @@ class IPEXModel(OptimizedModel):
 
     def maybe_apply_torch_compile(self):
         if (
-            not self.model.device.type != "cpu"
+            self.model.device.type != "cpu"
             or self.config.model_type in _COMPILE_NOT_READY_MODEL_TYPES
             or is_ipex_version("<", _IPEX_MINIMUM_VERSION_FOR_COMPILE)
         ):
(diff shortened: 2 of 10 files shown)


jiqing-feng:
Hi @echarlaix  @IlyasMoutawwakil . Please review this PR, thanks! BTW, the failed tests are not related to my changes.

docs/source/ipex/inference.mdx
```diff
@@ -14,7 +14,7 @@ Optimum Intel can be used to load models from the [Hub](https://huggingface.co/m
 
 ## Loading
 
-You can load your model and apply IPEX optimizations (apply torch.compile for non-generation tasks). For supported architectures like LLaMA, BERT and ViT, further optimizations will be applied by patching the model to use custom operators.
+You can load your model and apply IPEX optimizations (apply torch.compile except text-generation tasks). For supported architectures like LLaMA, BERT and ViT, further optimizations will be applied by patching the model to use custom operators.
```
IlyasMoutawwakil, line 17: I don't understand what you mean here, I see torch.compile being applied to `text-generation` task, why does it say "except text-generation tasks" here ?

jiqing-feng: Actually, we didn't apply torch.compile in text-generation task which means IPEXModelForCausal LM doesn't have torch.compile in __init__. And generation tasks are also excluded in IPEXModel.__init__ when calling torch.compile

optimum/intel/ipex/modeling_base.py
```diff
@@ -106,9 +108,9 @@ def __init__(
 
         # Non-generation tasks can use torch.compile to get acceleration.
         if (
-            model.device.type == "cpu"
-            and self.export_feature not in _IPEX_EXPORTED_GENERATION_TASKS
-            and config.model_type not in _COMPILE_NOT_READY_MODEL_TYPES
+            self.model.device.type == "cpu"
+            and self.export_feature not in _IPEX_SUPPORTED_GENERATION_TASKS
```
jiqing-feng, line 112: Hi @IlyasMoutawwakil . You can see we excluded the text-generation tasks when calling torch.compile here.

IlyasMoutawwakil: hmm I see, can you give some explanation as to why it's not supported ? I believe many CausalLM models in transformers are now `torch.compile` compatible when static cache is in use. The use `_IPEX_SUPPORTED_GENERATION_TASKS` is confusing here, as it states "supported", maybe we can have a variable for `_IPEX_NOT_SUPPORTED_TORCH_COMPILE_TASKS` for example.

Also I see that for text2text, torch.compile is applied in the `__init__` and then skipped here, which is very confusing readability wise, it makes more sense to have an `apply_torch_compile` method that is overridden in IPEXSeq2Seq (for example to warmup with generate) removing `text2text` from "`_IPEX_NOT_SUPPORTED_TORCH_COMPILE_TASKS`"

jiqing-feng: Hi @IlyasMoutawwakil . Sorry for not making it clear when I said text-generation is not compatible with torch.compile, it's because we patched a lot of ipex ops in text-generation models (such as llama; but compile works well for non-patching models as mistral). These patching ops will cause performance regression and we are fixing it, but need some time. I will indicate it and enable compile for non-patching models in text-generation task.

jiqing-feng:
Hi @IlyasMoutawwakil . I have fixed your comments, please take the 2nd round review. Thanks!

yao-matrix (approved):

IlyasMoutawwakil (approved):

optimum/intel/ipex/modeling_base.py
```diff
@@ -335,6 +352,83 @@ def generate(self, *args, **kwargs):
 
         return result
 
+    def _init_warmup(self):
+        inputs = prepare_jit_inputs(self.model, self.export_feature, False)
+        self.generate(input_ids=inputs["input_ids"], attention_mask=inputs["attention_mask"], max_new_tokens=4)
+        self.generate(input_ids=inputs["input_ids"], attention_mask=inputs["attention_mask"], max_new_tokens=4)
+        logger.info("Warm up end")
```
IlyasMoutawwakil, line 359: shouldn't this be in `IPEXModelForCausalLM` as well ?

jiqing-feng:
Hi @IlyasMoutawwakil . It should be ready to merge, thanks!

PR #1059, opened 2024-12-10
Remove tests-openvino extra requirement

nikita-savelyevv:
# What does this PR do?

**Changes**

- Remove `tests-openvino` introduced in #1040
- Add `datasets[audio]>=1.4.0` and `tbb` to general tests requirements

## Before submitting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ TESTS_REQUIRE = [
     "sentence-transformers",
     "open_clip_torch>=2.26.1",
     "peft",
+    "datasets[audio]>=1.4.0",
 ]
 
 QUALITY_REQUIRE = ["black~=23.1", "ruff==0.4.4"]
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -56,6 +56,8 @@
     "sentence-transformers",
     "open_clip_torch>=2.26.1",
     "peft",
+    "datasets[audio]>=1.4.0",
+    "tbb",
```
nikita-savelyevv, line 60: Installing `datasets[audio]` leads to `numba` package also being installed. Because of this, `tbb` needs to be installed as well. This is because with `numbda` installed, but `tbb` not installed, the following error is raised inside INC: 
```
ValueError: No threading layer could be loaded.
HINT:
Intel TBB is required, try:
$ conda/pip install tbb
```

After both `numba` and `tbb` are installed, INC starts to pack bits relying on these packages. For example: https://github.com/intel/neural-compressor/blob/master/neural_compressor/torch/algorithms/weight_only/modules.py#L456-L457 

@IlyasMoutawwakil As we've agreed before, could you please invite somebody from INC side for a review?

IlyasMoutawwakil:
INC seems to work ! thanks for following up on this.

IlyasMoutawwakil (approved):

PR #1061, opened 2024-12-10
[OV] Introduce `--quant-mode` cli argument enabling full quantization via optimum-cli

nikita-savelyevv:
# What does this PR do?

- Introduce `--quant-mode` export parameter. Currently the only supported value is `int8` for int8 weights and activations. In the future it can be extended to other combinations, for example `int8/fp8`.
- Introduce `weight_format` and `activation_format` fields to `OVQuantizationConfig`. Currently they are not used as `int8/int8` is the only supported option.
- Enable quantization of speech-to-text pipelines via optimum-cli. Example command is below.

```
optimum-cli export openvino -m openai/whisper-tiny.en --quant-mode int8 --dataset librispeech --num-samples 32 --smooth-quant-alpha 0.9 ./whisper-tiny-en
```

## Before submitting


optimum/commands/export/openvino.py
```diff
@@ -75,6 +75,17 @@ def parse_args_openvino(parser: "ArgumentParser"):
         default=None,
         help="The weight format of the exported model.",
     )
+    optional_group.add_argument(
+        "--quant-mode",
+        type=str,
+        choices=["int8/int8"],
```
AlexKoff88, line 81:
```suggestion
        choices=["int8"],
```
My take is that it can be simplified to just one value if both weights and activations are quantized to the same data type.

AlexKoff88: Please fix the description above accordingly

docs/source/openvino/export.mdx
```diff
@@ -155,6 +163,12 @@ Models larger than 1 billion parameters are exported to the OpenVINO format with
 </Tip>
 
 
+Besides weight-only quantization, you can also apply full model quantization including activations by setting `--quant-mode` to `int8/int8`. This will quantize both weights and activations of Linear, Convolutional and some other layers to int8. Currently this is only supported for speech-to-text models. Please see example below.
```
AlexKoff88, line 166: @echarlaix, @IlyasMoutawwakil, the idea is to expose the capabilities of OVQuantizer in the optimum-cli.

optimum/commands/export/openvino.py

AlexKoff88:
@echarlaix, @IlyasMoutawwakil, the PR is ready for your review

AlexKoff88 (approved):

nikita-savelyevv:
@echarlaix @IlyasMoutawwakil could you please review this PR some time this week? I'm on a vacation starting from the next week. Thanks!

AlexKoff88:
@eaidova, can you please take a look as well.

echarlaix (approved):
Looks good thanks @nikita-savelyevv, I added a minor comment about an argument name (no worries for us to take care of this part in case you're away at this time, won't block merging)

docs/source/openvino/export.mdx
```diff
@@ -66,6 +67,10 @@ Optional arguments:
                         on your local machine arbitrary code present in the model repository.
   --weight-format {fp32,fp16,int8,int4,mxfp4,nf4}
                         The weight format of the exported model.
+  --quant-mode {int8}
```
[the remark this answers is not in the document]
AlexKoff88: No problem, @echarlaix. The issue with activation-format is that we will let user so much freedom to mix different types of weights and activations data types that we don't support, e.g. W-FP8/A-INT8, W-INT4/A-FP8, etc. We target `--weight-format` for weight-only quantization and it corresponds `nncf.compress_weights()` API while `--quant-mode` controls the type of static full model quantization and corresponds `nncf.quantize(..., mode=)` parameter, e.g. `INT8`, `FP8`, etc.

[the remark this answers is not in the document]
AlexKoff88: BTW, we don't assume that `--weight-format` and `--quant-mode` are used all together.

AlexKoff88:
Merging as we have few approvals and no strong objections.

PR #1062, opened 2024-12-11
fix autotp linear check

jiqing-feng:
If the original model is a deepspeed auto TP model, the linear will be [LinearLayer](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/module_inject/layers.py#L124) which don't have `out_features`, so we can use `weight.shape[0]` which has the same value as `out_features`.


jiqing-feng:
Hi @sywangyi @IlyasMoutawwakil , please review this change. Thanks

sywangyi (approved):

IlyasMoutawwakil (approved):

PR #1063, opened 2024-12-11
[OV]: Removed checks for transformers==4.36

AlexKoff88:
Agreed with @helena-intel to remove validation for transformers==4.36.

### DIFF
--- a/.github/workflows/test_openvino.yml
+++ b/.github/workflows/test_openvino.yml
@@ -26,7 +26,7 @@ jobs:
             "*training*",
             "*export*",
           ]
-        transformers-version: ["4.36.0", "latest"]
+        transformers-version: ["latest"]
 
     runs-on: ubuntu-22.04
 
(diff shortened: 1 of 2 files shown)


AlexKoff88:
@IlyasMoutawwakil, @echarlaix, please take a look and merge if you are ok with this.

.github/workflows/test_openvino.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             "*training*",
             "*export*",
           ]
-        transformers-version: ["4.36.0", "latest"]
```
echarlaix, line 29: I'm not at all against dropping v4.36 support, but would keep tests for the lowest "official" transformers version supported (we can an additional check for this), would that work ?

helena-intel: Thanks Ella! Agreed, we should definitely keep testing the lowest supported version. I hope 4.37 works.

helena-intel: @AlexKoff88 Yes, testing now, I will create a new PR when everything works (because some minor changes are needed). Looks like 4.37 will work.

PR #1033, opened 2024-11-26
Load diffusers in native FP16/BF16 precision to reduce the memory usage

mvafin:
# What does this PR do?

Load diffusers model in native `float16` or `bfloat16` precision and apply 16bit patching for that case.
| Model | Before | After |
| -------- | ------- | ------- |
| FLUX-schnell | ~120Gb image | ~60Gb image |
| stable-diffusion-3.5-medium | ~38Gb | ~18Gb |

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1064, opened 2024-12-11
Do not check trace for diffusers, saving memory and time for FLUX

mvafin:
# What does this PR do?

This is further optimization of memory consumption of diffusers conversion, continuation of this PR: #1033 
When `check_trace=True` the TorchScript graph is generated second time and is compared with the graph generated first time. It is useful to catch incorrect traced graph sometimes, but in optimum we control which models are supported and such issues shouldn't happen.
Currently only introduce this for `diffusers`, but can be done for all the models.
The most impact on memory is demonstrated for FLUX, for other diffusers it significantly reduces conversion time.

| Model | Before | After |
| ------- | ------- | ------ |
| FLUX | ~58Gb/~240s image | ~33Gb/~130s image |
| SD-3.5-medium | ~18Gb/~250s image | ~18Gb/~100s image 
 |

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -16,7 +16,7 @@ import gc
 import logging
 import operator
 import warnings
-from functools import reduce
+from functools import reduce, partial
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Union
 
@@ -234,6 +234,7 @@ def main_export(
     do_gptq_patching = False
     custom_architecture = False
     patch_16bit = False
+    orig_trace = None
     loading_kwargs = model_loading_kwargs or {}
     if library_name == "transformers":
         config = AutoConfig.from_pretrained(
@@ -350,6 +351,8 @@ def main_export(
         if dtype in [torch.float16, torch.bfloat16]:
             loading_kwargs["torch_dtype"] = dtype
             patch_16bit = True
+        orig_trace = torch.jit.trace
+        torch.jit.trace = partial(torch.jit.trace, check_trace=False)
 
     if library_name == "open_clip":
         model = _OpenClipForZeroShotImageClassification.from_pretrained(model_name_or_path, cache_dir=cache_dir)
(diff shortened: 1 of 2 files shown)


AlexKoff88:
Awesome! Thanks @mvafin.

Please check and fix the quality: 
```
black --check .
ruff check .
```

AlexKoff88 (approved):

mvafin:
> ruff check .

Done

AlexKoff88:
+ @alexsu52 . Shall we have an option that disables this behavior during model export? Such an option can be enabled by default.

IlyasMoutawwakil (approved):
Not a fan of patches, long term solution would be decouple the export process by doing `torch.jit.trace` with the arguments we want then `ov.convert_model` with the resulting model.

AlexKoff88:
> Not a fan of patches, long term solution would be decouple the export process by doing `torch.jit.trace` with the arguments we want then `ov.convert_model` with the resulting model.

I agree. I see that torch.jit.trace in one place in the Optimum-Intel code. @mvafin, why did you just not pass the argument there?

optimum/exporters/openvino/__main__.py
```diff
@@ -46,6 +46,7 @@
     MULTI_MODAL_TEXT_GENERATION_MODELS,
     clear_class_registry,
     deduce_diffusers_dtype,
+    patch_not_check_trace,
```
IlyasMoutawwakil, line 49:
```suggestion
    disable_check_trace,
```

optimum/exporters/openvino/utils.py
```diff
@@ -288,3 +290,15 @@ def save_preprocessors(
                 logger.error(f"Saving {type(processor)} failed with {ex}")
     else:
         maybe_save_preprocessors(model_name_or_path, output, trust_remote_code=trust_remote_code)
+
+
+@contextmanager
+def patch_not_check_trace(to_patch):
```
IlyasMoutawwakil, line 296:
```suggestion
def disable_check_trace(to_patch):
```

IlyasMoutawwakil:
okay, we can go with a patch but let's at least make it a clean one, the naming in the PR is a bit confusing and I don't think the patching should be in `__main__.py` but rather in `convert` where lower level logic is defined. What about this:
- add an argument `check_trace=None` to the export function(s) so that users can control this behavior.
- if None, set this value to False in the case of diffusers.
- do the patching around the `convert_model` function with `patch_torch_jit_trace(check_trace=check_trace)`.

PR #1067, opened 2024-12-12
fix when attention_mask=None

jiqing-feng:
Fix IPEX patched model forward when attention_mask is None.


IlyasMoutawwakil (approved):

jiqing-feng:
Hi @IlyasMoutawwakil , it should be ready to merge.

PR #1070, opened 2024-12-12
Reduce img size in dummy_inputs for FLUX

mvafin:
# What does this PR do?

In addition to #1064 this change further reduce memory usage for FLUX conversion

| Before (with #1064 change) | After |
| ------- | ------ |
| ~33Gb image | ~13Gb image |

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -1804,12 +1804,12 @@ class DummyFluxTransformerInputGenerator(DummyVisionInputGenerator):
         normalized_config: NormalizedVisionConfig,
         batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
         num_channels: int = DEFAULT_DUMMY_SHAPES["num_channels"],
-        width: int = DEFAULT_DUMMY_SHAPES["width"],
-        height: int = DEFAULT_DUMMY_SHAPES["height"],
+        width: int = DEFAULT_DUMMY_SHAPES["width"] // 4,
+        height: int = DEFAULT_DUMMY_SHAPES["height"] // 4,
         **kwargs,
     ):
         # Reduce img shape by 4 for FLUX to reduce memory usage on conversion
-        super().__init__(task, normalized_config, batch_size, num_channels, width // 4, height // 4, **kwargs)
+        super().__init__(task, normalized_config, batch_size, num_channels, width, height, **kwargs)
         if getattr(normalized_config, "in_channels", None):
             self.num_channels = normalized_config.in_channels // 4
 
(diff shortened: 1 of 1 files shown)


optimum/exporters/openvino/model_configs.py
```diff
@@ -1808,7 +1808,8 @@ def __init__(
         height: int = DEFAULT_DUMMY_SHAPES["height"],
         **kwargs,
     ):
-        super().__init__(task, normalized_config, batch_size, num_channels, width, height, **kwargs)
+        # Reduce img shape by 4 for FLUX to reduce memory usage on conversion
+        super().__init__(task, normalized_config, batch_size, num_channels, width // 4, height // 4, **kwargs)
```
IlyasMoutawwakil, line 1812: Cool ! This will also override the user given values (from cli or python api), I would prefer changing these values in the function signature defaults.

optimum/exporters/openvino/model_configs.py
```diff
@@ -1804,10 +1804,11 @@ def __init__(
         normalized_config: NormalizedVisionConfig,
         batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
         num_channels: int = DEFAULT_DUMMY_SHAPES["num_channels"],
-        width: int = DEFAULT_DUMMY_SHAPES["width"],
-        height: int = DEFAULT_DUMMY_SHAPES["height"],
+        width: int = DEFAULT_DUMMY_SHAPES["width"] // 4,
+        height: int = DEFAULT_DUMMY_SHAPES["height"] // 4,
         **kwargs,
     ):
+        # Reduce img shape by 4 for FLUX to reduce memory usage on conversion
```
IlyasMoutawwakil, line 1811:
```suggestion
        width: int = DEFAULT_DUMMY_SHAPES["width"] // 4,
        height: int = DEFAULT_DUMMY_SHAPES["height"] // 4,
        # Reduce img shape by 4 for FLUX to reduce memory usage on conversion
        **kwargs,
    ):
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -1804,8 +1804,9 @@ def __init__(
         normalized_config: NormalizedVisionConfig,
         batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
         num_channels: int = DEFAULT_DUMMY_SHAPES["num_channels"],
-        width: int = DEFAULT_DUMMY_SHAPES["width"],
-        height: int = DEFAULT_DUMMY_SHAPES["height"],
+        width: int = DEFAULT_DUMMY_SHAPES["width"] // 4,
```
mvafin, line 1807: This unfortunately doesn't work as defaults are set in the optimum code and this will not change the behaviour

IlyasMoutawwakil: @echarlaix do you know where exactly we can change these values for a specific architecture like Flux ?
I thought dummy inputs generators will use the defaults in their signature by default unless a user passes input shapes.

PR #1075, opened 2024-12-16
Add a note about data-aware mixed precision assignment

nikita-savelyevv:
# What does this PR do?

Recently there appeared a confusion regarding when the `dataset` argument actually affects the resulting model. I've added some notes to help clarifying this.

Namely, if users provide compression `ratio<1` and a dataset at the same time, according to NNCF data-aware mixed precision using `MAX_ACTIVATION_VARIANCE` criterion will be applied. https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/weight_compression/algorithm.py#L96

Otherwise, if no data-aware compression algorithm is selected, `dataset` argument is not used.

## Before submitting

### DIFF
--- a/docs/source/openvino/export.mdx
+++ b/docs/source/openvino/export.mdx
@@ -78,7 +78,8 @@ Optional arguments:
   --ratio RATIO         A parameter used when applying 4-bit quantization to control the ratio between 4-bit and 8-bit
                         quantization. If set to 0.8, 80% of the layers will be quantized to int4 while 20% will be
                         quantized to int8. This helps to achieve better accuracy at the sacrifice of the model size
-                        and inference latency. Default value is 1.0.
+                        and inference latency. Default value is 1.0. Note: If dataset is provided, and the ration is
+                        less than 1.0, then data-aware mixed precision assignment will be applied.
   --sym                 Whether to apply symmetric quantization
   --group-size GROUP_SIZE
                         The group size to use for quantization. Recommended value is 128 and -1 uses per-column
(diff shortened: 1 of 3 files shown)


helena-intel (approved):
Thanks @nikita-savelyevv !

docs/source/openvino/export.mdx
```diff
@@ -78,7 +78,8 @@ Optional arguments:
   --ratio RATIO         A parameter used when applying 4-bit quantization to control the ratio between 4-bit and 8-bit
                         quantization. If set to 0.8, 80% of the layers will be quantized to int4 while 20% will be
                         quantized to int8. This helps to achieve better accuracy at the sacrifice of the model size
-                        and inference latency. Default value is 1.0.
+                        and inference latency. Default value is 1.0. Note: If dataset is provided, and the ratio is
+                        less than 1.0, then data-aware mixed precision assignment will be applied.
   --sym                 Whether to apply symmetric quantization
```
echarlaix, line 83: What do you think about adding a warning when a dataset is provided while the quantization methodology is default and `ratio=1` to specify that the dataset will be ignored in this case?

echarlaix (approved):
Thanks for the addition @nikita-savelyevv !

docs/source/openvino/export.mdx
nikita-savelyevv: Thanks for the idea. Added.

echarlaix: thanks a lot @nikita-savelyevv

PR #1077, opened 2024-12-17
fix import bug when ipex is available but sentence_transformers is no…

kaixuanliu:
fix import bug when ipex is available but sentence_transformers is not; 
add `accelerate` dependency to run with XPU


kaixuanliu:
@IlyasMoutawwakil @echarlaix , pls help review

IlyasMoutawwakil (approved):

PR #1078, opened 2024-12-18
merge decoder and decoder with past to stateful for seq2seq

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -32,6 +32,10 @@ from optimum.exporters.utils import (
 )
 from optimum.exporters.utils import (
     get_diffusion_models_for_export,
+    DECODER_NAME,
+    DECODER_WITH_PAST_NAME,
+    ENCODER_NAME,
+    _get_submodels_for_export_encoder_decoder,
 )
 from optimum.intel.utils.import_utils import (
     _diffusers_version,
@@ -43,6 +47,7 @@ from optimum.intel.utils.import_utils import (
     _torch_version,
     _transformers_version,
     compare_versions,
+    is_openvino_version,
     is_openvino_tokenizers_version,
     is_tokenizers_version,
     is_transformers_version,
@@ -624,10 +629,14 @@ def export_from_model(
 
         logger.info(f"Automatic task detection to: {task}.")
 
+    is_encoder_decoder = getattr(getattr(model, "config", {}), "is_encoder_decoder", False)
+    model_type = getattr(getattr(model, "config", {}), "model_type", "")
     stateful = stateful and (
-        ensure_export_task_support_stateful(task)
-        or ensure_model_type_support_stateful(getattr(getattr(model, "config", {}), "model_type", ""))
+        ensure_export_task_support_stateful(task, is_encoder_decoder) or ensure_model_type_support_stateful(model_type)
     )
+
+    if stateful and is_encoder_decoder and not getattr(model, "_supports_cache_class", False):
+        stateful = False
     # TODO: support onnx_config.py in the model repo
     if custom_architecture and custom_export_configs is None:
         raise ValueError(
@@ -666,6 +675,11 @@ def export_from_model(
     if library_name == "diffusers":
         export_config, models_and_export_configs = get_diffusion_models_for_export_ext(model, exporter="openvino")
         stateful_submodels = False
+    elif stateful and is_encoder_decoder and not custom_architecture:
+        export_config, models_and_export_configs = _get_encoder_decoder_stateful_models_for_export(
+            model=model, task=task, preprocessors=preprocessors, library_name=library_name, _variant="default"
+        )
+        stateful_submodels = [False, True]
     else:
         logging.disable(logging.INFO)
         export_config, models_and_export_configs, stateful_submodels = _get_submodels_and_export_configs(
@@ -1188,3 +1202,43 @@ def get_flux_models_for_export(pipeline, exporter, int_dtype, float_dtype):
         models_for_export["text_encoder_2"] = (text_encoder_2, export_config)
 
     return models_for_export
+
+
+def _get_encoder_decoder_stateful_models_for_export(
+    model: Union["PreTrainedModel", "TFPreTrainedModel"],
+    task: str,
+    _variant: str,
+    library_name: str,
+    int_dtype: str = "int64",
+    float_dtype: str = "fp32",
+    preprocessors: Optional[List[Any]] = None,
+):
+    export_config_constructor = TasksManager.get_exporter_config_constructor(
+        model=model, exporter="openvino", task=task, library_name=library_name
+    )
+    export_config = export_config_constructor(
+        model.config,
+        int_dtype=int_dtype,
+        float_dtype=float_dtype,
+        preprocessors=preprocessors,
+        legacy=False,
+    )
+
+    export_config.variant = _variant
+    all_variants = "\n".join([f"    - {name}: {description}" for name, description in export_config.VARIANTS.items()])
+    logger.info(f"Using the export variant {export_config.variant}. Available variants are:\n{all_variants}")
+
+    models_for_export = _get_submodels_for_export_encoder_decoder(model, use_past=True)
+
+    encoder_export_config = export_config.with_behavior("encoder")
+    models_for_export[ENCODER_NAME] = (models_for_export[ENCODER_NAME], encoder_export_config)
+
+    decoder_export_config_with_past = export_config.with_behavior("decoder", use_past=True, use_past_in_inputs=True)
+
+    decoder_export_config_with_past.stateful = True
+    decoder_with_past_model = models_for_export.pop(DECODER_WITH_PAST_NAME)
+    models_for_export[DECODER_NAME] = (
+        decoder_with_past_model,
+        decoder_export_config_with_past,
+    )
+    return None, models_for_export
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -20,6 +20,7 @@ from packaging import version
 from transformers import AutoConfig, PretrainedConfig, PreTrainedModel, TFPreTrainedModel
 from transformers.utils import is_tf_available
 
+from optimum.exporters.onnx.base import ConfigBehavior
 from optimum.exporters.onnx.config import OnnxConfig, TextDecoderOnnxConfig, TextDecoderWithPositionIdsOnnxConfig
 from optimum.exporters.onnx.model_configs import (
     CLIPOnnxConfig,
@@ -36,8 +37,10 @@ from optimum.exporters.onnx.model_configs import (
     MistralOnnxConfig,
     MPTOnnxConfig,
     PhiOnnxConfig,
+    T5OnnxConfig,
     UNetOnnxConfig,
     VisionOnnxConfig,
+    WhisperOnnxConfig,
 )
 from optimum.exporters.onnx.model_patcher import ModelPatcher
 from optimum.exporters.tasks import TasksManager
@@ -90,6 +93,7 @@ from .model_patcher import (
     Phi3VisionImageEmbeddingsPatcher,
     QwenModelPatcher,
     RotaryEmbPatcher,
+    StatefulSeq2SeqDecoderPatcher,
     UpdateCausalMaskModelPatcher,
     XverseModelPatcher,
 )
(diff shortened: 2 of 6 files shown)


tests/openvino/test_quantization.py
```diff
@@ -106,7 +106,7 @@ class OVQuantizerTest(unittest.TestCase):
                 weight_only=False,
                 smooth_quant_alpha=0.95,
             ),
-            (14, 22, 21) if is_transformers_version("<=", "4.42.4") else (14, 22, 25),
+            (14, 22, 21) if is_transformers_version("<=", "4.42.4") else (14, 26, 25),
```
eaidova, line 109:
```suggestion
            (14, 22, 21) if is_transformers_version("<=", "4.42.4") else (14, 22, 25),
```

optimum/exporters/openvino/stateful.py
```diff
@@ -190,18 +190,93 @@ def ensure_stateful_is_available(warn=True):
...
 
 
-def ensure_export_task_support_stateful(task: str):
+_ENCODER_DECODER_TASKS_WITH_PAST = (
+    "automatic-speech-recognition",
+    "text2text-generation",
+)
+
+
+def ensure_export_task_support_stateful(task: str, is_encoder_decoder: bool = False):
```
echarlaix, line 199: could there be any case where the task is "automatic-speech-recognition" or "text2text-generation" and the model is a decoder only model ? I'd imagine that the `task` argument should be enough here and that we can drop `is_encoder_decoder`

optimum/exporters/openvino/convert.py
```diff
@@ -1188,3 +1201,43 @@ def get_flux_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+        float_dtype=float_dtype,
+        preprocessors=preprocessors,
+        legacy=False,
+    )
+
+    export_config.variant = _variant
+    all_variants = "\n".join([f"    - {name}: {description}" for name, description in export_config.VARIANTS.items()])
+    logger.info(f"Using the export variant {export_config.variant}. Available variants are:\n{all_variants}")
+
+    models_for_export = _get_submodels_for_export_encoder_decoder(model, use_past=True)
```
echarlaix, line 1230: looks like we could also set `use_past=False` and use  `models_for_export[DECODER_NAME]` directly no ? 
https://github.com/huggingface/optimum/blob/d21256c2964945fc3fe4623f7befb21082b69a25/optimum/exporters/utils.py#L203

tests/openvino/test_quantization.py
```diff
@@ -208,8 +208,12 @@ def test_ov_model_static_quantization_with_auto_dataset(
             ov_model.save_pretrained(tmp_dir)
```
echarlaix, line 209: Not directly related to this PR but the conversion step is getting more complex so adding it here : I'd be in favor of removing the onnx conversion step fallback a least for every model which are composed of multiple subcomponents https://github.com/huggingface/optimum-intel/blob/753f84db6e0966580eb9eaa74a808213be730631/optimum/exporters/openvino/convert.py#L465

echarlaix (commented):
Looks great @eaidova :fire:

optimum/exporters/openvino/convert.py
eaidova: not sure that I understand your suggestion, we still need past as outputs and inputs for making transformation (their filling just will be hided inside model and performed automatically, but for applying our transformations we need them anyway), that is why decoder_with_past is more suitable for such case than just decoder.

I think I can figure out how to achieve the same effect with model without past (changing decoder config during export)

tests/openvino/test_quantization.py
eaidova: Yes, I think we can do that in separate PR after some experiments to check that we do not have modes that require fallback

optimum/exporters/openvino/convert.py
echarlaix: When calling `_get_submodels_for_export_encoder_decoder(model, use_past=True)` it looks like the models `models_for_export[DECODER_NAME]` and `models_for_export[DECODER_WITH_PAST_NAME]` will be equivalent 

https://github.com/huggingface/optimum/blob/d21256c2964945fc3fe4623f7befb21082b69a25/optimum/exporters/utils.py#L213-L215

so  I was suggesting to set `use_past=False` and use  `models_for_export[DECODER_NAME]` directly, would that make sense ?

optimum/exporters/openvino/stateful.py
eaidova: it was copied from my old implementation that requires additional guidance for correct stateful seq2seq. there is also Vision2Seq models that also usually have 2 decoders, but not all - vision-encoder-decoder architecture uses gpt2 based model decoder that does not have cross attention, so it has only 1 decoder, but I do not see why model with single decoder can not be stateful if it works in the same way like in text-generation model task only getting image as input instead of text

optimum/exporters/openvino/convert.py
eaidova: yes, thanks,I applied your suggestion and it looks better now

optimum/exporters/openvino/stateful.py
eaidova: removed

optimum/exporters/openvino/model_configs.py
```diff
@@ -2611,3 +2615,69 @@ def patch_model_for_export(
...
+    ) -> ModelPatcher:
+        if getattr(self, "stateful", False) and self._behavior == ConfigBehavior.DECODER:
+            return StatefulSeq2SeqDecoderPatcher(self, model, model_kwargs)
+        return super().patch_model_for_export(model, model_kwargs)
+
+    @property
+    def inputs(self):
+        common_inputs = super().inputs
+        if getattr(self, "stateful", False) and self._behavior == ConfigBehavior.DECODER:
+            common_inputs["decoder_input_ids"] = {0: "batch_size", 1: "seq_length"}
```
IlyasMoutawwakil, line 2643:
```suggestion
            common_inputs["decoder_input_ids"] = {0: "batch_size", 1: "decoder_sequence_length"}
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -2611,3 +2615,69 @@ def patch_model_for_export(
...
```
IlyasMoutawwakil, line 2664:
```suggestion
            common_inputs["decoder_input_ids"] = {0: "batch_size", 1: "decoder_sequence_length"}
```

optimum/intel/openvino/modeling_base_seq2seq.py
```diff
@@ -257,7 +257,21 @@ def _from_pretrained(
             if not compile_only:
                 encoder = cls.load_model(file_names["encoder"], quantization_config)
                 decoder = cls.load_model(file_names["decoder"], quantization_config)
-                if use_cache:
+                if use_cache and not model_has_state(decoder):
+                    model_file_names["decoder_with_past"] = decoder_with_past_file_name
+                    model_file_names["decoder_with_past_bin"] = decoder_with_past_file_name.replace(".xml", ".bin")
```
IlyasMoutawwakil, line 262: this should only be added and downloaded `if not from_onnx` right ?

optimum/intel/openvino/modeling_base_seq2seq.py
```diff
@@ -75,7 +76,8 @@ def __init__(
         if self.is_dynamic and not self._compile_only:
             encoder = self._reshape(encoder, -1, -1, is_decoder=False)
             decoder = self._reshape(decoder, -1, -1)
-            decoder_with_past = self._reshape(decoder_with_past, -1, -1) if self.use_cache else None
+            if decoder_with_past is not None:
+                decoder_with_past = self._reshape(decoder_with_past, -1, -1) if self.use_cache else None
```
IlyasMoutawwakil, line 80:
```suggestion
            if self.use_cache and decoder_with_past is not None:
                decoder_with_past = self._reshape(decoder_with_past, -1, -1)
```

optimum/intel/openvino/modeling_base_seq2seq.py
eaidova: yes, onnx format does not support stateful models, so there is no relevant changes for from_onnx path

IlyasMoutawwakil: my problem is that now decoder_with_past_bin will be (attempted to be) downloaded even when `from_onnx` is True.
because these binaries used to be added to `model_file_names` under the condition `if not from_onnx:` `for key in list(model_file_names.keys()):`, now that list only has encoder and decoder at that stage, to which we add `decoder_with_past` (ok) and `decoder_with_past_bin` (not necessarily exist) unconditionally.

IlyasMoutawwakil:
```suggestion
                    if not from_onnx:
                        model_file_names["decoder_with_past_bin"] = decoder_with_past_file_name.replace(".xml", ".bin")
```

IlyasMoutawwakil: but the next loop will have to be adapted as well

optimum/intel/openvino/modeling_base_seq2seq.py
```diff
@@ -232,8 +234,6 @@ def _from_pretrained(
         # Load model from hub
         else:
             model_file_names = {"encoder": encoder_file_name, "decoder": decoder_file_name}
-            if use_cache:
```
IlyasMoutawwakil, line 235: one suggestion would be to follow the same pattern you used to check for statefulness when loading from local files based on existence/absence of `decoder_with_past_file_name`, we can check here as well using
`use_cache and decoder_with_past_file_name in repo_files ` with `repo_files = huggingface_hub.list_repo_files(model_name_or_path, revision=revision, token=token, ...)`
avoiding changes to the onnx path and the subsequent changes.

eaidova: file existance not always means that it should be loaded, sometimes people save model multiple time in the same directory (so if previous run uses stateless and current run stateful) it will be incorrect

IlyasMoutawwakil: Totally agree with that, I'm only suggesting it because you already used this "file existence/absence" in the local files loading path (right above this one).

eaidova: improved from_onnx usage and conditions for loading decoder_with_past

IlyasMoutawwakil (approved):
LGTM !

optimum/exporters/openvino/convert.py
```diff
@@ -635,10 +641,14 @@ def export_from_model(
 
         logger.info(f"Automatic task detection to: {task}.")
 
+    is_encoder_decoder = getattr(getattr(model, "config", {}), "is_encoder_decoder", False)
+    model_type = getattr(getattr(model, "config", {}), "model_type", "")
     stateful = stateful and (
-        ensure_export_task_support_stateful(task)
-        or ensure_model_type_support_stateful(getattr(getattr(model, "config", {}), "model_type", ""))
+        ensure_export_task_support_stateful(task) or ensure_model_type_support_stateful(model_type)
```
echarlaix, line 647: did a test with t5 and looks like it wasn't exported as a stateful model coming from https://github.com/huggingface/optimum-intel/blob/fe55db5da2e20f8df704d27c795390e3c501b606/optimum/exporters/openvino/stateful.py#L200, I think we need to include t5, mt5 and whisper to `ensure_model_type_support_stateful`

tests/openvino/test_modeling.py
```diff
@@ -1718,7 +1720,7 @@ def test_generate_utils(self, model_arch):
         gc.collect()
```
echarlaix, line 1721: Could we also add a test to make sure that the architectures that supports it are indeed exported in stateful mode ? (so for now for t5, mt5 and whisper)

echarlaix (approved):
Great PR thanks @eaidova ! Left two comments to make sure stateful compatible models are exported as expected, good to merge once resolved

optimum/exporters/openvino/model_configs.py
echarlaix: keeps compatibility with optimum so works for me!

eaidova:
failed flux tests are not related, they are caused by update models on hf hub, I prepared PR for fixing this issue
https://github.com/huggingface/optimum-intel/pull/1108

tests/openvino/test_modeling.py
echarlaix: thanks!

optimum/exporters/openvino/convert.py
echarlaix: Actually I tested again and everything works fine : I was using a tiny random model ([hf-internal-testing/tiny-random-T5Model](https://huggingface.co/hf-internal-testing/tiny-random-T5Model)) without specifying the task which was automatially detected to '`feature-extraction-with-past'`, sorry for the false alert!

PR #1080, opened 2024-12-19
disable kv cache compression for fp vlm

eaidova:
# What does this PR do?

Fixes issue with failed minicpmv with ov nightly. starting from 2024.6 openvino will use kv cache compression by default enabled, that may impact model accuracy, but identify when it should be disabled can not be predicted on runtime level, so we proposed addition of specific hint for such models (by our agreement it should be done for noncompressed models only) - extended this approach to handle language models as part visual language models

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -100,9 +100,15 @@ def _set_runtime_options(
     for model_name in models_and_export_configs.keys():
         _, sub_export_config = models_and_export_configs[model_name]
         sub_export_config.runtime_options = {}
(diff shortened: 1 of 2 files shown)


AlexKoff88 (approved):

nikita-savelyevv (approved):

eaidova:
@echarlaix  @IlyasMoutawwakil could you please take a look? ov 2024/6 release couple of hours ago, so this minicpmv test failure should be visible on main branch

echarlaix (approved):
Thanks for the rapid fix @eaidova

PR #1072, opened 2024-12-13
Fix test for transformers 4.36

helena-intel:
Instead of https://github.com/huggingface/optimum-intel/pull/1063

When fixing the test for 4.37 I tried with 4.36 and tests passed with that too, so we can support this a little longer.

Also added workflow_dispatch trigger to OpenVINO test. This makes it easier to test these kinds of changes in a branch in a fork. workflow_dispatch only works if it is also in main, so without having this in main you always have to add it to the main branch of your fork first. If there's a reason not to have it, I'll remove it.

@AlexKoff88

PR #1082, opened 2024-12-19
skip private mdoel loading test for external contributors

echarlaix:
https://github.com/huggingface/optimum-intel/pull/1072#issuecomment-2545175142


echarlaix:
cc @helena-intel fixing the issue from https://github.com/huggingface/optimum-intel/pull/1072#issuecomment-2545175142

PR #1087, opened 2024-12-20
support any input resolution in stable diffusion models

eaidova:
# What does this PR do?

due to UNet architecture limitations, original stable diffusion pipeline implementation requires that input was always divisible on 64 (https://github.com/CompVis/stable-diffusion/issues/60#issuecomment-1240294667). In diffusers-based pipelines this issue was fixed and model can generate if input resolution nondivisible to 64 too (adjusting sizes between unet blocks), but during tracing we still note that openvino model preserve behaviour with requirement divisibility to 64 (because path with adjusting data resolution is not activated during tracing). This PR fixes this limitation

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -1785,21 +1785,19 @@ class DummyTransformerTimestpsInputGenerator(DummyTimestepInputGenerator):
 
 class DummyUnetVisionInputGenerator(DummyVisionInputGenerator):
     def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
-        print("HERE")
-        
-        if not input_name in ["sample", "latent_sample"]:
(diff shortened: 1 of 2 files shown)


optimum/exporters/openvino/model_configs.py
```diff
@@ -1783,6 +1783,25 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
         return super().generate(input_name, framework, int_dtype, float_dtype)
 
 
+class DummyUnetVisionInputGenerator(DummyVisionInputGenerator):
+    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+        print("HERE")
```
eaidova, line 1788:
```suggestion
```

AlexKoff88 (approved):

IlyasMoutawwakil (approved):

PR #1088, opened 2024-12-20
Support transformers 4.47

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -421,9 +421,9 @@ def _llama_gemma_update_causal_mask_legacy(self, attention_mask, input_tensor, c
                 offset = 0
             mask_shape = attention_mask.shape
             mask_slice = (attention_mask.eq(0.0)).to(dtype=dtype) * min_dtype
-            causal_mask[
-                : mask_shape[0], : mask_shape[1], offset : mask_shape[2] + offset, : mask_shape[3]
-            ] = mask_slice
+            causal_mask[: mask_shape[0], : mask_shape[1], offset : mask_shape[2] + offset, : mask_shape[3]] = (
+                mask_slice
+            )
 
     if (
         self.config._attn_implementation == "sdpa"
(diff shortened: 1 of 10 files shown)


optimum/exporters/openvino/model_patcher.py
```diff
@@ -2712,7 +2712,14 @@ def patched_forward(*args, **kwargs):
 
     def __enter__(self):
         super().__enter__()
-        if is_transformers_version(">=", "4.45.0"):
+
+        if is_transformers_version(">=", "4.47.0"):
+            from transformers.models.gemma2.modeling_gemma2 import GEMMA2_ATTENTION_FUNCTION
+
+            GEMMA2_ATTENTION_FUNCTION["original_eager"] = GEMMA2_ATTENTION_FUNCTION["eager"]
+            GEMMA2_ATTENTION_FUNCTION["eager"] = GEMMA2_ATTENTION_FUNCTION["sdpa"]
```
echarlaix, line 2720: why is this needed ? also looks like GEMMA2_ATTENTION_FUNCTION will be removed in v4.48 
https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma2/modeling_gemma2.py

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2712,7 +2712,14 @@ def patched_forward(*args, **kwargs):
```
echarlaix, line 2712: Looks like the mismatch for qwen2_vl is coming from https://github.com/huggingface/transformers/pull/33487/files (mismatch between the `position_ids` when not given at inference)

echarlaix (commented):
thanks @IlyasMoutawwakil !

optimum/exporters/openvino/model_patcher.py
eaidova: openvino attention optimizations and paged attention support va vllm requires to use sdpa, while default implementation for gemma2 is eager in transformers due to some accuracy issues with running model on cude in bf16. I think it is better to move such patching on model loading level:
https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/__main__.py#L52

for avoid necessarily to align this code refactoring sensitive patch

IlyasMoutawwakil: > why is this needed ? also looks like GEMMA2_ATTENTION_FUNCTION will be removed in v4.48

It's the new equivalent to the lines that come afterwards, transformers seem to be migrating from a class based attention dispatching to a functional one.

>  I think it is better to move such patching on model loading level

Thanks ! will use that instead then.

IlyasMoutawwakil:
only one test failure that is unrelated

echarlaix:
LGTM, thanks @IlyasMoutawwakil !

PR #1091, opened 2024-12-23
restore input format for stable diffusion and export configs mapping

eaidova:
# What does this PR do?

changes on optimum side for support sd3 make export configs mapping incompatible with our registered configuration (we need own clip config for preserving SDPA in model) and changed sd timestep input format. That is not expected on our side and may causes errors in users's custom pipelines

## Before submitting


AlexKoff88 (approved):

AlexKoff88:
Merging as a hotfix.

PR #1093, opened 2024-12-23
update min dtype in falcon for prevent bf16 execution issue

eaidova:
# What does this PR do?

in 4.45 when _prepare_4d_causal_attention_mask_with_cache_position was function, it had argument for passing min_dtype,  in 4.46 it was excluded, this PR updates patching for resolving bf16 execution issues

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -2501,6 +2501,40 @@ class RotaryEmbPatcher(DecoderModelPatcher):
                 _reinitialize_cos_sin_cached_fp32(layer.self_attn.rotary_emb)
 
 
+# Adopted from https://github.com/huggingface/transformers/blob/31f9a289a6207be6cae746e009d8e0db523be203/src/transformers/models/falcon/modeling_falcon.py#L1138
+def _falcon_prepare_4d_causal_attention_mask_with_cache_position(
+    attention_mask: torch.Tensor,
+    sequence_length: int,
+    target_length: int,
+    dtype: torch.dtype,
+    device: torch.device,
+    cache_position: torch.Tensor,
+    batch_size: int,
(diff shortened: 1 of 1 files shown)


optimum/exporters/openvino/model_patcher.py
```diff
@@ -2501,6 +2501,40 @@ def __enter__(self):
...
+    device: torch.device,
+    cache_position: torch.Tensor,
+    batch_size: int,
+    **kwargs,
+):
+    if attention_mask is not None and attention_mask.dim() == 4:
+        # In this case we assume that the mask comes already in inverted form and requires no inversion or slicing.
+        causal_mask = attention_mask
+    else:
+        # different from original allow to provide min_dtype as parameter
```
eaidova, line 2519:
```suggestion
        # different from original: allow to provide min_dtype as parameter
```

eaidova:
nightly tests failure is not related, it is caused by openvino python API unexpected changes

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2501,6 +2501,40 @@ def __enter__(self):
                 _reinitialize_cos_sin_cached_fp32(layer.self_attn.rotary_emb)
 
 
+# Adopted from https://github.com/huggingface/transformers/blob/31f9a289a6207be6cae746e009d8e0db523be203/src/transformers/models/falcon/modeling_falcon.py#L1138
```
echarlaix, line 2504:
```suggestion
# Adapted from https://github.com/huggingface/transformers/blob/31f9a289a6207be6cae746e009d8e0db523be203/src/transformers/models/falcon/modeling_falcon.py#L1138
```

echarlaix (approved):
Thanks for the fix @eaidova !

PR #1103, opened 2025-01-08
Disable graph check while tracing

mvafin:
# What does this PR do?
This is alternative to #1064. It will only work with `openvino>=2025.0` (https://github.com/openvinotoolkit/openvino/pull/28328)
By disabling checking graph after trace we reduce memory usage in diffusers.

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -429,9 +429,9 @@ def export_pytorch(
 
             patcher.patched_forward = ts_patched_forward
 
-            decoder_kwargs = {}
+            ts_decoder_kwargs = {}
             if library_name == "diffusers" and is_openvino_version(">=", "2025.0"):
-                decoder_kwargs["trace_kwargs"] = {"check_trace": False}
+                ts_decoder_kwargs["trace_kwargs"] = {"check_trace": False}
 
             with patcher:
                 if patch_16bit_model:
@@ -440,9 +440,9 @@ def export_pytorch(
                     __make_16bit_traceable(model)
                 check_dummy_inputs_are_allowed(model, dummy_inputs)
                 input_info = _get_input_info(model, config, dummy_inputs)
-                decoder = TorchScriptPythonDecoder(model, example_input=dummy_inputs, **decoder_kwargs)
+                ts_decoder = TorchScriptPythonDecoder(model, example_input=dummy_inputs, **ts_decoder_kwargs)
                 ov_model = convert_model(
-                    decoder,
+                    ts_decoder,
                     example_input=dummy_inputs,
                     input=[(item.shape, item.type) for item in input_info],
                 )
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -45,9 +45,9 @@ from optimum.intel.utils.import_utils import (
     _transformers_version,
     compare_versions,
     is_openvino_tokenizers_version,
+    is_openvino_version,
     is_tokenizers_version,
     is_transformers_version,
-    is_openvino_version,
 )
 from optimum.utils import DEFAULT_DUMMY_SHAPES, is_diffusers_available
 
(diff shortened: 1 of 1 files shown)


AlexKoff88 (approved):

optimum/exporters/openvino/convert.py
```diff
@@ -23,6 +23,7 @@
 from transformers.generation import GenerationMixin
 from transformers.utils import is_tf_available, is_torch_available
 
+from openvino.frontend.pytorch.ts_decoder import TorchScriptPythonDecoder
```
IlyasMoutawwakil, line 26: I would prefer if we used something like `TorchScriptPythonWrapper` or `TorchScriptPythonModel`, the decoder part doesn't make sense especially since we are using this wrapper on top of diffusers text_encoder/unet/transformer etc.
Will this wrapper ever change and do decoder-specific stuff ? if npt why does it have `Decoder` in its name.

mvafin: This "decoder" in a sense that it is used to decode the pytorch model. It is used in openvino as an abstraction to get the graph from the model. It is an internal class, but it can be used to provide such custom parameters for tracing.

eaidova: this naming does not have relation to decoder models, Decoder is internal component of openvino responsible for decoding of original framework model during conversion. All that time we use it internally inside convert_model, now it is created explicitly for resolving advanced conversion options

eaidova (approved):

optimum/exporters/openvino/convert.py
IlyasMoutawwakil: oh okay, confusing name but okay if it's part of the openvino naming conventions.

AlexKoff88: Can we merge this then?

optimum/exporters/openvino/convert.py
```diff
@@ -427,15 +429,20 @@ def ts_patched_forward(*args, **kwargs):
 
             patcher.patched_forward = ts_patched_forward
 
+            decoder_kwargs = {}
+            if library_name == "diffusers" and is_openvino_version(">=", "2025.0"):
+                decoder_kwargs["trace_kwargs"] = {"check_trace": False}
```
IlyasMoutawwakil, line 434:
```suggestion
            ts_decoder_kwargs = {}
            if library_name == "diffusers" and is_openvino_version(">=", "2025.0"):
                ts_decoder_kwargs["trace_kwargs"] = {"check_trace": False}
```

optimum/exporters/openvino/convert.py
```diff
@@ -427,15 +429,20 @@ def ts_patched_forward(*args, **kwargs):
...
+
             with patcher:
                 if patch_16bit_model:
                     from openvino.frontend.pytorch.patch_model import __make_16bit_traceable
 
                     __make_16bit_traceable(model)
                 check_dummy_inputs_are_allowed(model, dummy_inputs)
                 input_info = _get_input_info(model, config, dummy_inputs)
+                decoder = TorchScriptPythonDecoder(model, example_input=dummy_inputs, **decoder_kwargs)
```
IlyasMoutawwakil, line 443:
```suggestion
                ts_decoder = TorchScriptPythonDecoder(model, example_input=dummy_inputs, **decoder_kwargs)
```

optimum/exporters/openvino/convert.py
```diff
@@ -427,15 +429,20 @@ def ts_patched_forward(*args, **kwargs):
...
                 ov_model = convert_model(
-                    model,
+                    decoder,
```
IlyasMoutawwakil, line 445:
```suggestion
                    ts_decoder,
```

mvafin:
Per @eaidova ask I moved decoder input inside function, can be useful in case openvino was custom built without pytorch support

IlyasMoutawwakil (approved):
LGTM, can merge once tests pass.

PR #1104, opened 2025-01-08
Remove height and width from default shapes pre-configuration

mvafin:
# What does this PR do?

Fixes #1070 as it was merged with changes that will not work because `height` and `width` are pre-configured. After this change dummy input generators will use height and width from defaults of `__init__`.

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -664,6 +664,9 @@ def export_from_model(
     # Get the shapes to be used to generate dummy inputs
     input_shapes = {}
     for input_name in DEFAULT_DUMMY_SHAPES.keys():
+        if input_name in ["height", "width"]:
+            # use H and W from generator defaults
+            continue
         input_shapes[input_name] = (
             kwargs_shapes[input_name] if input_name in kwargs_shapes else DEFAULT_DUMMY_SHAPES[input_name]
         )


optimum/exporters/openvino/convert.py
```diff
@@ -664,6 +664,9 @@ def export_from_model(
     # Get the shapes to be used to generate dummy inputs
     input_shapes = {}
     for input_name in DEFAULT_DUMMY_SHAPES.keys():
+        if input_name in ["height", "width"]:
+            # use H and W from generator defaults
+            continue
         input_shapes[input_name] = (
             kwargs_shapes[input_name] if input_name in kwargs_shapes else DEFAULT_DUMMY_SHAPES[input_name]
         )
```
IlyasMoutawwakil, line 672: I think this entire part is unnecessary and could/should be removed, as we already have the default shapes in each input generator signature. This part enforces the values in DEFAULT_DUMMY_SHAPES on all model inputs generators, restricting model-specific export optimizations (like reducing input dimensions for a dynamic axis to lower export memory requirements). It also populates input shapes with irrelevant shapes.
@echarlaix wdyt ?

mvafin: I agree, but I feel that it may be dangerous to remove it now as it was there for a long time and some models may rely on it. As for `height` and `weight` the impact is much less, because not many models receive image in the first place.

IlyasMoutawwakil: I will merge for now and follow up with a PR investigating the possibility of its removal.

IlyasMoutawwakil: Thanks @mvafin

Issue #1073, opened 2024-12-13
issue with converting vlm model "InternVL2_5-1B"


endomorphosis:
devel@workstation:/tmp$ optimum-cli export openvino --model OpenGVLab/InternVL2_5-1B InternVL2_5-1B --trust-remote-code  >> /tmp/openvino.txt

```
Flash Attention 2.0 only supports torch.float16 and torch.bfloat16 dtypes, but the current dype in Qwen2ForCausalLM is torch.float32. You should run training or inference using Automatic Mixed-Precision via the `with torch.autocast(device_type='torch_device'):` decorator, or load the model with the `torch_dtype` argument. Example: `model = AutoModel.from_pretrained("openai/whisper-tiny", attn_implementation="flash_attention_2", torch_dtype=torch.float16)`
The class `optimum.bettertransformers.transformation.BetterTransformer` is deprecated and will be removed in a future release.
WARNING:root:Cannot apply model.to_bettertransformer because of the exception:
The model type qwen2 is not yet supported to be used with BetterTransformer. Feel free to open an issue at https://github.com/huggingface/optimum/issues if you would like this model type to be supported. Currently supported models are: dict_keys(['albert', 'bark', 'bart', 'bert', 'bert-generation', 'blenderbot', 'bloom', 'camembert', 'blip-2', 'clip', 'codegen', 'data2vec-text', 'deit', 'distilbert', 'electra', 'ernie', 'fsmt', 'gpt2', 'gptj', 'gpt_neo', 'gpt_neox', 'hubert', 'layoutlm', 'm2m_100', 'marian', 'markuplm', 'mbart', 'opt', 'pegasus', 'rembert', 'prophetnet', 'roberta', 'roc_bert', 'roformer', 'splinter', 'tapas', 't5', 'vilt', 'vit', 'vit_mae', 'vit_msn', 'wav2vec2', 'xlm-roberta', 'yolos']).. Usage model with stateful=True may be non-effective if model does not contain torch.functional.scaled_dot_product_attention
`loss_type=None` was set in the config but it is unrecognised.Using the default loss: `ForCausalLMLoss`.
We detected that you are passing `past_key_values` as a tuple of tuples. This is deprecated and will be removed in v4.47. Please convert your cache or use an appropriate `Cache` class (https://huggingface.co/docs/transformers/kv_cache#legacy-cache-format)
[...]

PR #1105, opened 2025-01-10
disable flash_attn during export internvl2

eaidova:
# What does this PR do?
InternVL2 models forces flash attention if it is available in user environment that may breaks model export. In the same time, if flash attn disabled, it selects eager attention for some cases, that gives suboptimal perf on openvino side. Updated model patching for sdpa attention realization as main path

Fixes https://github.com/huggingface/optimum-intel/issues/1073

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -21,6 +21,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 import torch
 import torch.nn.functional as F
+from transformers import PreTrainedModel, TFPreTrainedModel
 from transformers.modeling_outputs import BaseModelOutputWithPast, BaseModelOutputWithPooling
 from transformers.utils import is_tf_available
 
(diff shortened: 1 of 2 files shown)


AlexKoff88 (approved):

eaidova:
@echarlaix  @IlyasMoutawwakil could you please take a look?

andrei-kochin:
@echarlaix @IlyasMoutawwakil any news here?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2992,11 +2993,91 @@ def __init__(
...
+        internal_patcher_cls = patcher_for_model_type.get(model_type)
+        if internal_patcher_cls is not None:
+            self._internal_patcher = internal_patcher_cls(config, model, model_kwargs)
+            self._patched_forward = self._internal_patcher.patched_forward
+        super().__init__(config, model, model_kwargs)
+
+    @property
+    def patched_forward(self):
+        if self._internal_patcher is not None:
+            return self._internal_patcher.patched_forward
```
echarlaix, line 3033: isn't it equivalent to returning `self._patched_forward` ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2992,11 +2993,91 @@ def __init__(
...
+class InternVL2ChatLangModelPatcher(DecoderModelPatcher):
+    def __init__(
+        self, config: "OnnxConfig", model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Dict[str, Any]
+    ):
+        model_type = model.config.model_type
+        patcher_for_model_type = {
+            "llama": LlamaModelPatcher,
+            "qwen2": UpdateCausalMaskModelPatcher,
+            "phi3": Phi3ModelPatcher,
+            "internlm2": InternLM2Patcher,
```
echarlaix, line 3020: could make sense to add some additional tests for these cases to make sure this doesn't break in the future

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2992,11 +2993,91 @@ def __init__(
...
+        if self._internal_patcher is not None:
+            self._internal_patcher.patched_forward = fn
+
+    def __enter__(self):
+        if is_torch_version(">=", "2.1.0"):
+            if self._model.config.model_type == "qwen2" and self._model.config._attn_implementation != "sdpa":
+                from transformers.models.qwen2.modeling_qwen2 import QWEN2_ATTENTION_CLASSES
+
+                sdpa_attn = QWEN2_ATTENTION_CLASSES["sdpa"]
+                self._model.config._orig_attn_implementation = self._model.config._attn_implementation
```
echarlaix, line 3048: small note : `QWEN2_ATTENTION_CLASSES` is deprecated in transformers v4.48 (don't have to be included as we currently don't officially support 4.48 but just adding it to know that this needs to be changed soon)

echarlaix (approved):
LGTMn thanks @eaidova

optimum/exporters/openvino/model_patcher.py
eaidova: no, it may contains some extra patching

optimum/exporters/openvino/model_patcher.py
eaidova: do you mean more models with different llms or what? I believe we will be not able test flash attention selection in cpu only env as it required to have cuda to install and import lib

optimum/exporters/openvino/model_patcher.py
eaidova: thanks for note

PR #1106, opened 2025-01-13
sana support

eaidova:
# What does this PR do?

* provided opportunity to select model checkpoint variant for loading
* supported Sana image generation model family

Example of usage:
```
import torch

from optimum.intel.openvino import OVDiffusionPipeline


pipe = OVDiffusionPipeline.from_pretrained("Efficient-Large-Model/Sana_600M_512px_diffusers", variant="fp16")
prompt = 'a cyberpunk cat in neon style'
image = pipe(
    prompt=prompt,
    generator=torch.Generator(device="cpu").manual_seed(142),
)[0]

image[0].save("sana_fp16.png")
```

Generated image:
sana_fp16

## Before submitting


optimum/exporters/openvino/__main__.py
```diff
@@ -364,6 +369,7 @@ class StoreAttr(object):
         if library_name == "open_clip":
             model = _OpenClipForZeroShotImageClassification.from_pretrained(model_name_or_path, cache_dir=cache_dir)
         else:
+            logger.warn(loading_kwargs)
```
eaidova, line 372:
```suggestion
```

AlexKoff88 (approved):

andrei-kochin:
@echarlaix   @IlyasMoutawwakil  could you please take a look?

tests/openvino/test_diffusion.py
```diff
@@ -130,30 +139,38 @@ def test_num_images_per_prompt(self, model_arch: str):
...
         ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
-        diffusers_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
-
-        for output_type in ["latent", "np", "pt"]:
-            inputs["output_type"] = output_type
+        diffusers_pipeline = auto_cls.from_pretrained(MODEL_NAMES[model_arch])
 
-            ov_output = ov_pipeline(**inputs, generator=get_generator("pt", SEED)).images
-            diffusers_output = diffusers_pipeline(**inputs, generator=get_generator("pt", SEED)).images
+        with torch.no_grad():
```
IlyasMoutawwakil, line 148: I think this not necessary ? all diffusers call methods have no_grad decorators

tests/openvino/test_diffusion.py
```diff
@@ -104,12 +111,14 @@ def test_load_vanilla_model_which_is_not_supported(self):
...
-        auto_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
+        auto_cls = self.get_auto_cls(model_arch)
+        auto_pipeline = DiffusionPipeline if model_arch != "sana" else auto_cls
+        auto_pipeline = auto_cls.from_pretrained(MODEL_NAMES[model_arch])
         ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
 
         self.assertEqual(ov_pipeline.auto_model_class, auto_pipeline.__class__)
 
-        auto_pipeline = DiffusionPipeline.from_pretrained(MODEL_NAMES[model_arch])
+        auto_pipeline = auto_pipeline.from_pretrained(MODEL_NAMES[model_arch])
```
IlyasMoutawwakil, line 121: why ? `DiffusionPipeline` should be able to return the right SanaPipeline

tests/openvino/test_diffusion.py
```diff
@@ -94,6 +94,13 @@ def generate_inputs(self, height=128, width=128, batch_size=1):
 
         return inputs
 
+    def get_auto_cls(self, model_arch):
+        if model_arch == "sana":
+            from diffusers import SanaPipeline
+
+            return SanaPipeline
+        return self.AUTOMODEL_CLASS
```
IlyasMoutawwakil, line 102: is Sana pipeline not registered as a text to image pipeline ?

tests/openvino/test_diffusion.py
```diff
@@ -77,8 +77,8 @@ class OVPipelineForText2ImageTest(unittest.TestCase):
     SUPPORTED_ARCHITECTURES = ["stable-diffusion", "stable-diffusion-xl", "latent-consistency"]
     NEGATIVE_PROMPT_SUPPORT_ARCHITECTURES = ["stable-diffusion", "stable-diffusion-xl", "latent-consistency"]
     if is_transformers_version(">=", "4.40.0"):
-        SUPPORTED_ARCHITECTURES.extend(["stable-diffusion-3", "flux"])
-        NEGATIVE_PROMPT_SUPPORT_ARCHITECTURES.append("stable-diffusion-3")
+        SUPPORTED_ARCHITECTURES.extend(["stable-diffusion-3", "flux", "sana"])
+        NEGATIVE_PROMPT_SUPPORT_ARCHITECTURES.append(["stable-diffusion-3"])
```
IlyasMoutawwakil, line 81: extend

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1052,6 +1067,11 @@ def forward(
         self._compile()
         model_inputs = {"input_ids": input_ids}
 
+        if "attention_mask" in self.input_names:
+            model_inputs["attention_mask"] = (
+                attention_mask if attention_mask is not None else torch.ones(input_ids.shape, dtype=torch.long)
```
IlyasMoutawwakil, line 1072: are you sure we wanna create `torch.ones` here ? it doesn't work for batched prompts (batch of samples with different lengths)

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -838,17 +846,23 @@ def reshape(
 
         if self.text_encoder is not None:
             self.text_encoder.model = self._reshape_text_encoder(
-                self.text_encoder.model, batch_size, self.tokenizer.model_max_length
+                self.text_encoder.model,
+                batch_size,
+                self.tokenizer.model_max_length if "Gemma" not in self.tokenizer.__class__.__name__ else -1,
```
IlyasMoutawwakil, line 851:
```suggestion
                getattr(self.tokenizer_3, "model_max_length", -1),
```

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -838,17 +846,23 @@ def reshape(
...
             )
 
         if self.text_encoder_2 is not None:
             self.text_encoder_2.model = self._reshape_text_encoder(
-                self.text_encoder_2.model, batch_size, self.tokenizer_2.model_max_length
+                self.text_encoder_2.model,
+                batch_size,
+                self.tokenizer_2.model_max_length if "Gemma" not in self.tokenizer.__class__.__name__ else -1,
```
IlyasMoutawwakil, line 858: the check should be `if "Gemma" not in self.tokenizer_2.__class__.__name__` ?
```suggestion
                getattr(self.tokenizer_2, "model_max_length", -1),
```

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -838,17 +846,23 @@ def reshape(
...
             )
 
         if self.text_encoder_3 is not None:
             self.text_encoder_3.model = self._reshape_text_encoder(
-                self.text_encoder_3.model, batch_size, self.tokenizer_3.model_max_length
+                self.text_encoder_3.model,
+                batch_size,
+                self.tokenizer_3.model_max_length if "Gemma" not in self.tokenizer.__class__.__name__ else -1,
```
IlyasMoutawwakil, line 865: the check should be `if "Gemma" not in self.tokenizer_3.__class__.__name__` ?
```suggestion
                getattr(self.tokenizer_3, "model_max_length", -1),
```

optimum/intel/openvino/modeling_diffusion.py
eaidova: it does not solve issue, because actually this field exists but equal to max int

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -588,6 +590,7 @@ def _from_transformers(
             force_download=force_download,
             ov_config=ov_config,
             library_name=cls._library_name,
+            model_variant=variant,
```
IlyasMoutawwakil, line 593: switching between `model_variant` and `variant` is confusing, we should probably stick to one, `model_variant`  or `weight_variant` should be fine as this more of a weights file specification

tests/openvino/test_diffusion.py
eaidova: unfortunatly, only on main branch https://github.com/huggingface/diffusers/pull/10538
diffusers pypi package (0.32.2) does not have sana in autopipeline registry

tests/openvino/test_diffusion.py
eaidova: in the latest released diffuers not yet, see comment above

tests/openvino/test_diffusion.py
IlyasMoutawwakil: hmm yeah for autopipeline I see but `DiffusionPipeline` is not part of the autopipeline registery, it's the main entrypoint of all pipelines (it just reads the `_class_name` in https://huggingface.co/Efficient-Large-Model/Sana_1600M_1024px_BF16_diffusers/blob/main/model_index.json#L2)

optimum/intel/openvino/modeling_diffusion.py
eaidova: yes, I'm sure, this common practice to create tensor of ones in case if attention mask required but not provided from user. Otherwise we need to have access to tokenizer and used pad token, that is not always possible, but in case where input with padding provided - tokenizer returns both input_ids and attention_mask, so it should not be problem to use tokenizer prepared attention_mask

optimum/intel/openvino/modeling_diffusion.py
eaidova: sana will file on reshaping stage as model_max_length too large and can not be used for reshaping

IlyasMoutawwakil: what about `if "Gemma" not in self.tokenizer.__class__.__name__` should be the same in all three checks ?

optimum/intel/openvino/modeling_diffusion.py
IlyasMoutawwakil: yeah let's avoid `variant` and stick to `weights_variant` in this case

optimum/intel/openvino/modeling_diffusion.py
eaidova: not enough data to verify yet as we have only 1 model where gemma used and it has only 1 text_encoder, so I'll return back fro text_encoder_2 and 3

tests/openvino/test_diffusion.py
IlyasMoutawwakil: https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/sana/pipeline_sana.py#L673

optimum/intel/openvino/modeling_diffusion.py
eaidova: but in this case, it will be unaligned with original from_pretrained argument
https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained.variant

is it ok or better to keep as variant?

tests/openvino/test_diffusion.py
eaidova: yes, you are right, I added it during debugging accuracy issue

optimum/intel/openvino/modeling_diffusion.py
eaidova: removed

optimum/intel/openvino/modeling_diffusion.py
eaidova: updated

optimum/intel/openvino/modeling_diffusion.py
eaidova: I'll use weights_variant everywhere except from_pretrained now, could you please take a look?

IlyasMoutawwakil: > but in this case, it will be unaligned with original from_pretrained argument
https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained.variant
is it ok or better to keep as variant in from_pretrainedt?

`variant` makes more sense now since it's the argument the core libs are using especially because it's used in the model cards https://huggingface.co/Efficient-Large-Model/Sana_1600M_512px_MultiLing_diffusers#1-how-to-use-sanapipeline-with-%F0%9F%A7%A8diffusers. @echarlaix wdyt

eaidova: renamed

tests/openvino/test_diffusion.py
eaidova: fixed

eaidova:
@IlyasMoutawwakil all comments applied, could you please take a look one more time?
Thanks!

tests/openvino/test_diffusion.py
```diff
@@ -104,12 +104,12 @@ def test_load_vanilla_model_which_is_not_supported(self):
     @parameterized.expand(SUPPORTED_ARCHITECTURES)
     @require_diffusers
     def test_ov_pipeline_class_dispatch(self, model_arch: str):
-        auto_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
+        auto_pipeline = DiffusionPipeline.from_pretrained(MODEL_NAMES[model_arch])
```
IlyasMoutawwakil, line 107: let's simply use an if else here on the sana case, because this test is supposed to check whether the `self.OVMODEL_CLASS` returns the equivalent of what `self.AUTOMODEL_CLASS` returns, but now it's comparing `self.OVMODEL_CLASS` with `DiffusionPipeline`

tests/openvino/test_diffusion.py
```diff
@@ -130,30 +130,36 @@ def test_num_images_per_prompt(self, model_arch: str):
...
 
         # test on inputs nondivisible on 64
         height, width, batch_size = 96, 96, 1
 
         for output_type in ["latent", "np", "pt"]:
             inputs["output_type"] = output_type
+            if model_arch == "sana":
+                # resolution binning will lead to resize output to standard resolution and back that can interpolate floating-point deviations
+                inputs["use_resolution_binning"] = False
+            atol = 6e-3
```
IlyasMoutawwakil, line 157:
```suggestion
```

tests/openvino/test_diffusion.py
```diff
@@ -130,30 +130,36 @@ def test_num_images_per_prompt(self, model_arch: str):
...
 
             ov_output = ov_pipeline(**inputs, generator=get_generator("pt", SEED)).images
             diffusers_output = diffusers_pipeline(**inputs, generator=get_generator("pt", SEED)).images
 
-            np.testing.assert_allclose(ov_output, diffusers_output, atol=6e-3, rtol=1e-2)
+            np.testing.assert_allclose(ov_output, diffusers_output, atol=atol, rtol=1e-2)
```
IlyasMoutawwakil, line 162:
```suggestion
            np.testing.assert_allclose(ov_output, diffusers_output, atol=6e-3, rtol=1e-2)
```

tests/openvino/test_diffusion.py
```diff
@@ -130,25 +130,30 @@ def test_num_images_per_prompt(self, model_arch: str):
...
         ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
-        diffusers_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
+        diffusers_pipeline = DiffusionPipeline.from_pretrained(MODEL_NAMES[model_arch])
 
         for output_type in ["latent", "np", "pt"]:
             inputs["output_type"] = output_type
+            if model_arch == "sana":
+                # resolution binning will lead to resize output to standard resolution and back that can interpolate floating-point deviations
+                inputs["use_resolution_binning"] = False
+            atol = 1e-4
```
IlyasMoutawwakil, line 143:
```suggestion
```

tests/openvino/test_diffusion.py
```diff
@@ -130,25 +130,30 @@ def test_num_images_per_prompt(self, model_arch: str):
...
 
             ov_output = ov_pipeline(**inputs, generator=get_generator("pt", SEED)).images
             diffusers_output = diffusers_pipeline(**inputs, generator=get_generator("pt", SEED)).images
-
-            np.testing.assert_allclose(ov_output, diffusers_output, atol=6e-3, rtol=1e-2)
+            np.testing.assert_allclose(ov_output, diffusers_output, atol=atol, rtol=1e-2)
```
IlyasMoutawwakil, line 147:
```suggestion
            np.testing.assert_allclose(ov_output, diffusers_output, atol=6e-3, rtol=1e-2)
```

tests/openvino/test_diffusion.py
eaidova: done

tests/openvino/test_diffusion.py
```diff
@@ -104,12 +104,13 @@ def test_load_vanilla_model_which_is_not_supported(self):
...
     def test_ov_pipeline_class_dispatch(self, model_arch: str):
-        auto_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
+        auto_cls = self.AUTOMODEL_CLASS if model_arch != "sana" else DiffusionPipeline
+        auto_pipeline = auto_cls.from_pretrained(MODEL_NAMES[model_arch])
         ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
 
         self.assertEqual(ov_pipeline.auto_model_class, auto_pipeline.__class__)
 
-        auto_pipeline = DiffusionPipeline.from_pretrained(MODEL_NAMES[model_arch])
+        auto_pipeline = auto_pipeline.from_pretrained(MODEL_NAMES[model_arch])
```
IlyasMoutawwakil, line 113:
```suggestion
        auto_pipeline = DiffusionPipeline.from_pretrained(MODEL_NAMES[model_arch])
```

tests/openvino/test_diffusion.py
```diff
@@ -130,25 +131,29 @@ def test_num_images_per_prompt(self, model_arch: str):
     @parameterized.expand(SUPPORTED_ARCHITECTURES)
     @require_diffusers
     def test_compare_to_diffusers_pipeline(self, model_arch: str):
-        height, width, batch_size = 128, 128, 1
+        height, width, batch_size = 64, 64, 1
         inputs = self.generate_inputs(height=height, width=width, batch_size=batch_size)
-
         ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
-        diffusers_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
+        diffusers_pipeline = DiffusionPipeline.from_pretrained(MODEL_NAMES[model_arch])
```
IlyasMoutawwakil, line 137:
```suggestion
        auto_cls = self.AUTOMODEL_CLASS if model_arch != "sana" else DiffusionPipeline
        diffusers_pipeline = auto_cls.from_pretrained(MODEL_NAMES[model_arch])
```

IlyasMoutawwakil (approved):

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -325,6 +327,7 @@ def _from_transformers(
             stateful=stateful,
             model_loading_kwargs=model_loading_kwargs,
             library_name=cls._library_name,
+            weughts_variant=variant,
```
echarlaix, line 330: typo :
```suggestion
            variant=variant,
```

tests/openvino/test_diffusion.py
```diff
@@ -77,8 +77,8 @@ class OVPipelineForText2ImageTest(unittest.TestCase):
     SUPPORTED_ARCHITECTURES = ["stable-diffusion", "stable-diffusion-xl", "latent-consistency"]
     NEGATIVE_PROMPT_SUPPORT_ARCHITECTURES = ["stable-diffusion", "stable-diffusion-xl", "latent-consistency"]
```
echarlaix, line 78: now that `variant` was added, it could make sense to add a test to make sure this works as expected / doesn't break in the future, can be added in a following PR

echarlaix (approved):
Thanks for the addition @eaidova

tests/openvino/test_diffusion.py
echarlaix: added in https://github.com/huggingface/optimum-intel/pull/1106/commits/b27e6f79743e273578eead971126d88c2cf6d297 thanks a lot @eaidova !

optimum/intel/__init__.py
```diff
@@ -150,6 +151,7 @@
             "OVFluxImg2ImgPipeline",
             "OVFluxInpaintPipeline",
             "OVFluxFillPipeline",
+            "OVSanaPipeline",
```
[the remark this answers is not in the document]
eaidova: added

tests/openvino/test_quantization.py
```diff
@@ -543,6 +544,7 @@ class OVWeightCompressionTest(unittest.TestCase):
             [
                 (OVStableDiffusion3Pipeline, "stable-diffusion-3", 9, 65),
                 (OVFluxPipeline, "flux", 7, 56),
+                (OVSanaPipeline, "sana", 19, 53),
```
openvino-dev-samples, line 547: @nikita-savelyevv any idea to get these predefined num of quantized node ?

nikita-savelyevv: @openvino-dev-samples For example, you can run the newly added test locally and see how many reference nodes there are based on an assertion error.

PR #1115, opened 2025-01-16
Give library name when inferring task during export

echarlaix:


helena-intel (approved):
Thank you! Confirmed that with this you get an expected error message when trying to export a non-transformers model with and passing --library

echarlaix:
Thanks a lot for taking the time to check this @helena-intel !

PR #1122, opened 2025-01-20
Raise an error when OVQuantizer is invoked on an compressed model

l-bat:
# What does this PR do?

For models exceeding 1 billion parameters, optimum-intel applies weight compression by default during the from_pretrained method call. This behavior impacts the functionality of subsequent quantization algorithms because patterns cannot be matched properly, leading to incorrect results (ref: 160308).

This PR introduces a mechanism to detect if the model was compressed by checking runtime information. If a user attempts to quantize a compressed model, an error is raised to prevent incorrect behavior.

Fixes # (issue)
ticket: 160760

## Before submitting

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -315,31 +315,6 @@ class OVQuantizer(OptimumQuantizer):
         else:
             raise TypeError(f"Unsupported model type: {type(self.model)}")
 
-    def _check_model_state(self, sub_model_names: List[str] = None):
-        message_template = (
-            "Couldn't apply optimization to the model because it was already compressed with config: {}. "
-            "To avoid this issue, set load_in_8bit=False in the from_pretrained method when using the optimum-intel API, "
-            "or explicitly specify the desired weight format using --weight_format fp16/fp32 for CLI."
-        )
-
-        def check_rt_info(ov_model):
-            rt_info = ov_model.get_rt_info()
-            if "nncf" in rt_info:
-                model_weight_compression_config = rt_info["nncf"].get("weight_compression", None)
-                model_quantization_config = rt_info["nncf"].get("quantization", None)
-                if model_weight_compression_config is not None:
-                    raise RuntimeError(message_template.format(model_weight_compression_config))
-                elif model_quantization_config is not None:
-                    raise RuntimeError(message_template.format(model_quantization_config))
-
-        if sub_model_names is None:
-            check_rt_info(self.model.model)
-        else:
-            for name in sub_model_names:
-                if hasattr(self.model, name):
-                    ov_model = getattr(self.model, name).model
-                    check_rt_info(ov_model)
-
     def _quantize_ovbasemodel(
         self,
         ov_config: OVConfig,
@@ -350,7 +325,7 @@ class OVQuantizer(OptimumQuantizer):
         remove_unused_columns: bool = True,
         **kwargs,
     ):
-        from optimum.intel.openvino.modeling_seq2seq import _OVModelForWhisper, OVModelForSeq2SeqLM
+        from optimum.intel.openvino.modeling_seq2seq import _OVModelForWhisper
         from optimum.intel.openvino.modeling_visual_language import OVModelForVisualCausalLM
 
         if is_diffusers_available():
@@ -429,7 +404,6 @@ class OVQuantizer(OptimumQuantizer):
                         "text_encoder_2",
                         "text_encoder_3",
                     ]
-                    self._check_model_state(sub_model_names)
                     sub_models = filter(lambda x: x, (getattr(self.model, name) for name in sub_model_names))
                     for sub_model in sub_models:
                         _weight_only_quantization(sub_model.model, quantization_config_copy, **kwargs)
@@ -447,7 +421,6 @@ class OVQuantizer(OptimumQuantizer):
                     self.model.clear_requests()
                 else:
                     # The model may be for example OVModelForImageClassification, OVModelForAudioClassification, etc.
-                    self._check_model_state()
                     self.model.model = _hybrid_quantization(
                         self.model.model, quantization_config, calibration_dataset, **kwargs
                     )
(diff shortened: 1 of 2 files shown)


l-bat:
@AlexKoff88, @nikita-savelyevv please take a look

tests/openvino/test_quantization.py
```diff
@@ -698,6 +698,23 @@ def test_ovmodel_load_with_compressed_weights(self, model_cls, model_type, trust
...
+            trust_remote_code=trust_remote_code,
+        )
+        quantization_config = OVWeightQuantizationConfig(bits=4, sym=True)
+        quantizer = OVQuantizer(model)
+        if isinstance(model, OVModelOpenCLIPForZeroShotImageClassification):
+            with pytest.raises(TypeError):
+                quantizer.quantize(ov_config=OVConfig(quantization_config=quantization_config))
+        else:
+            with pytest.raises(RuntimeError):
+                quantizer.quantize(ov_config=OVConfig(quantization_config=quantization_config))
```
nikita-savelyevv, line 716: Could you please also specify a message with which the error should be raised?

optimum/intel/openvino/quantization.py
```diff
@@ -315,6 +315,31 @@ def quantize(
         else:
             raise TypeError(f"Unsupported model type: {type(self.model)}")
 
+    def _check_model_state(self, sub_model_names: List[str] = None):
+        message_template = (
+            "Couldn't apply optimization to the model because it was already compressed with config: {}. "
```
AlexKoff88, line 320:
```suggestion
            "Cannot apply optimization to the model because it was already optimized with the following config: {}. "
```

optimum/intel/openvino/quantization.py
```diff
@@ -315,6 +315,31 @@ def quantize(
...
+            "To avoid this issue, set load_in_8bit=False in the from_pretrained method when using the optimum-intel API, "
```
AlexKoff88, line 321:
```suggestion
            "To avoid this issue, check that you set load_in_8bit=False or not using quantization_config at export in the .from_pretrained(),"
```

tests/openvino/test_quantization.py
```diff
@@ -698,6 +698,23 @@ def test_ovmodel_load_with_compressed_weights(self, model_cls, model_type, trust
             _, num_weight_nodes = get_num_quantized_nodes(model)
             self.assertEqual(expected_ov_int8[i], num_weight_nodes["int8"])
 
+    @parameterized.expand(SUPPORTED_ARCHITECTURES_WITH_AUTO_COMPRESSION)
```
AlexKoff88, line 701: We need to check that the exception is thrown if we compress quantized or compressed model and if we quantize quantized or compressed model.

optimum/intel/openvino/quantization.py
```diff
@@ -436,19 +463,31 @@ def _quantize_ovbasemodel(
...
                     self.model.clear_requests()
+                elif isinstance(self.model, OVModelForSeq2SeqLM):
+                    sub_model_names = ["encoder", "decoder"]
+                    if self.model.decoder_with_past is not None:
+                        sub_model_names.append("decoder_with_past")
+                    self._check_model_state(sub_model_names)
+                    sub_models = [getattr(self.model, name) for name in sub_model_names]
+                    for sub_model in sub_models:
+                        _weight_only_quantization(sub_model, quantization_config, **kwargs)
+                    self.model.clear_requests()
```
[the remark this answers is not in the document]
l-bat: ticket: 160958

tests/openvino/test_quantization.py
```diff
@@ -698,6 +698,23 @@ def test_ovmodel_load_with_compressed_weights(self, model_cls, model_type, trust
...
+    def test_raise_error_WC_over_WC(self, model_cls, model_type, trust_remote_code):
```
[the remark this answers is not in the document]
l-bat: The test time for all cases is 0:03:15 on my machine. I can incorporate this into `test_ovmodel_load_with_compressed_weights` to minimize the impact on testing time.

tests/openvino/test_quantization.py
```diff
@@ -698,6 +698,23 @@ def test_ovmodel_load_with_compressed_weights(self, model_cls, model_type, trust
...
+            MODEL_NAMES[model_type],
+            export=True,
+            load_in_8bit=True,
+            trust_remote_code=trust_remote_code,
+        )
+        quantization_config = OVWeightQuantizationConfig(bits=4, sym=True)
+        quantizer = OVQuantizer(model)
+        if isinstance(model, OVModelOpenCLIPForZeroShotImageClassification):
+            with pytest.raises(TypeError):
+                quantizer.quantize(ov_config=OVConfig(quantization_config=quantization_config))
```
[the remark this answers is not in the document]
l-bat: Quantization is not supported for `OVModelOpenCLIPForZeroShotImageClassification`. I'll skip it in test.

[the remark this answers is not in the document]
l-bat: ticket: 161043

AlexKoff88 (approved):

optimum/intel/openvino/quantization.py
```diff
@@ -1010,6 +1010,7 @@ def _weight_only_quantization(
     calibration_dataset: Optional[Union[nncf.Dataset, Iterable]] = None,
     **kwargs,
 ) -> openvino.runtime.Model:
+    _verify_not_optimized(model)
```
nikita-savelyevv, line 1013: 👍

tests/openvino/test_quantization.py
```diff
@@ -214,6 +214,9 @@ def preprocess_function(examples, tokenizer):
             # Verify that the configuration is correctly saved and loaded
             loaded_config = OVConfig.from_pretrained(tmp_dir)
             self.assertEqual(ov_config.quantization_config.to_dict(), loaded_config.quantization_config.to_dict())
+            check_optimization_not_applicable_to_optimized_model(
```
nikita-savelyevv, line 217: I see that you've added checks for quantized models when compression is applied to them. Could you please also add this check to some tests which apply compression? For example `test_ovmodel_load_with_compressed_weights` and `test_ovmodel_4bit_auto_compression_with_config`.

nikita-savelyevv (approved):

tests/openvino/test_quantization.py
nikita-savelyevv: I now see that check for `test_ovmodel_load_with_compressed_weights` is already added

echarlaix (approved):
LGTM, thanks for the addition @l-bat !

PR #1127, opened 2025-01-23
NF4+F8E4M3 support

KodiaqQ:
# What does this PR do?

Implements 153357

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -717,7 +717,10 @@ class OVConfig(BaseConfig):
         self.compression = kwargs.get(
             "compression", None
         )  # A field for backward-compatability of training-time compression parameters
-        self.dtype = dtype
+        if self.quantization_config is not None:
+            self.dtype = self.quantization_config.weight_format
+        else:
+            self.dtype = dtype
 
     def add_input_info(self, model_inputs: Dict, force_batch_one: bool = False):
         self.input_info = [
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -1026,8 +1026,8 @@ class OVQuantizeOptions:
                     overflow_fix=self.advanced_parameters["overflow_fix"],
                 )
                 if "smooth_quant_alphas" in self.advanced_parameters:
-                    advanced_parameters.smooth_quant_alphas = (
-                        nncf.AdvancedSmoothQuantParameters(**self.advanced_parameters["smooth_quant_alphas"]),
+                    advanced_parameters.smooth_quant_alphas = nncf.AdvancedSmoothQuantParameters(
+                        **self.advanced_parameters["smooth_quant_alphas"]
                     )
 
             self._nncf_dict = {
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -717,10 +717,7 @@ class OVConfig(BaseConfig):
         self.compression = kwargs.get(
             "compression", None
         )  # A field for backward-compatability of training-time compression parameters
-        if self.quantization_config is not None:
-            self.dtype = self.quantization_config.weight_format
-        else:
-            self.dtype = dtype
+        self.dtype = dtype
 
     def add_input_info(self, model_inputs: Dict, force_batch_one: bool = False):
         self.input_info = [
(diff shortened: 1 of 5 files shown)


KodiaqQ:
@nikita-savelyevv, @AlexKoff88, could you please review it? Thanks.

optimum/intel/openvino/configuration.py
```diff
@@ -775,3 +772,348 @@ def to_dict(self) -> Dict[str, Any]:
...
+    def to_dict(self) -> Dict:
+        return copy.deepcopy(self.__dict__)
+
+
+class OVGeneralQuantizationConfig(QuantizationConfigMixin):
+    def __init__(
+        self,
+        ignored_scope: Optional[Dict] = None,
+        num_samples: Optional[int] = None,
+        compress_weights_options: Optional[OVCompressWeightsOptions] = None,
```
AlexKoff88, line 1054: Why did not you use the existing `OVWeightQuantizationConfig` and `OVQuantizationConfig`?

optimum/intel/openvino/configuration.py
```diff
@@ -775,3 +772,348 @@ def to_dict(self) -> Dict[str, Any]:
...
+            }
+            return self._nncf_dict
+
+        raise ImportError("NNCF is required to execute this method. Please install nncf first.")
+
+    def to_dict(self) -> Dict:
+        return copy.deepcopy(self.__dict__)
+
+
+class OVGeneralQuantizationConfig(QuantizationConfigMixin):
```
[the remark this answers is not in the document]
KodiaqQ: If weight compression or quantize-specific options are absent, the method's main purpose would be only to fully quantize or compress the models' weights. Does the name `Mixed` reflect this behaviour? I don't think so.

optimum/intel/openvino/configuration.py
KodiaqQ: Because `OVWeightQuantizationConfig` and `OVQuantizationConfig` contain general parameters like `tokenizer` or `dataset` that are not method-specific options. For parameters like `sym` or `bits` it is even not clear how they should be used in the new approach.

KodiaqQ:
The issue with:
`NotImplementedError: NNCF is not yet supported OpenVINO data type: nf4.`
Already was fixed on the NNCF's side: https://github.com/openvinotoolkit/nncf/pull/3209.

optimum/intel/openvino/configuration.py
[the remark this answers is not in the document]
AlexKoff88: ok, please name it `OVGenericQuantizationConfig` then.

optimum/intel/openvino/configuration.py
```diff
@@ -775,3 +775,348 @@ def to_dict(self) -> Dict[str, Any]:
...
+            }
+            return self._nncf_dict
+
+        raise ImportError("NNCF is required to execute this method. Please install nncf first.")
+
+    def to_dict(self) -> Dict[str, Any]:
+        return copy.deepcopy(self.__dict__)
+
+
+class OVQuantizeOptions:
```
AlexKoff88, line 925:
```suggestion
class OVQuantizationOptions:
```

optimum/intel/openvino/configuration.py
```diff
@@ -775,3 +775,348 @@ def to_dict(self) -> Dict[str, Any]:
 
     def to_diff_dict(self) -> Dict[str, Any]:
         return self._to_dict_safe(to_diff_dict=True)
+
+
+class OVCompressWeightsOptions:
```
AlexKoff88, line 780:
```suggestion
class OVWeightsCompressionOptions:
```

AlexKoff88: Let's use this `OVXXXOptions` classes as base entities for corresponding `OVXXXConfig`s.

optimum/intel/openvino/configuration.py
```diff
@@ -775,3 +775,348 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+        raise ImportError("NNCF is required to execute this method. Please install nncf first.")
+
+    def to_dict(self) -> Dict:
+        return copy.deepcopy(self.__dict__)
+
+
+class OVGeneralQuantizationConfig(QuantizationConfigMixin):
+    def __init__(
+        self,
```
AlexKoff88, line 1054: Constructor of the class should be extended with `dataset`, `tokenizer` and the rest of the options.

optimum/intel/openvino/configuration.py
```diff
@@ -775,3 +775,348 @@ def to_dict(self) -> Dict[str, Any]:
...
+            }
+            return self._nncf_dict
+
+        raise ImportError("NNCF is required to execute this method. Please install nncf first.")
+
+    def to_dict(self) -> Dict:
+        return copy.deepcopy(self.__dict__)
+
+
+class OVGeneralQuantizationConfig(QuantizationConfigMixin):
```
nikita-savelyevv, line 1052: I believe it should inherit from `OVQuantizationConfigBase` as other configs do because `ov_config.quantization_config` currently supports such instances.

nikita-savelyevv (commented):
I've left several comments, some are minor, but there are a couple of major ones.

The major ones are related to (1) interaction of `OVGeneralQuantizationConfig` with `OVConfig`, e.g. `dict`-conversion support, (2) complexity of config creation with `OVGeneralQuantizationConfig`.

I've sketched out a proposal where I've implemented a way that satisfies some of the issues I've highlighted. But possibly, this is not the only way. It supports dict-to-config-class conversion and inherits `OVQuantizationConfigBase`. I've also refactored the base class itself to contain dataset-related logic, assuming that compression and quantization will rely on the same dataset. Unfortunately I don't have much time for this so possibly there are some rough points still.

@KodiaqQ @AlexKoff88, please take a look: https://github.com/huggingface/optimum-intel/compare/main...nikita-savelyevv:optimum-intel:ns/nf4_f8e4m3_proposal

optimum/intel/openvino/configuration.py
```diff
@@ -775,3 +775,348 @@ def to_dict(self) -> Dict[str, Any]:
...
+        target_device: str = "any",
+        fast_bias_correction: bool = True,
+        model_type: Optional[str] = None,
+        advanced_parameters: Optional[Dict] = None,
+    ):
+        """
+        Class containing specific nncf.quantize method's options.
+        Args:
+            mode (`str`, *optional*):
+                Defines special quantization modes. Possible values: ['fp8_e4m3', 'fp8_e5m2'].
```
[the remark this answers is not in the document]
KodiaqQ: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/parameters.py#L144-L154

optimum/intel/openvino/configuration.py
```diff
@@ -775,3 +775,348 @@ def to_dict(self) -> Dict[str, Any]:
...
+        if smooth_quant_alpha:
+            advanced_parameters["smooth_quant_alphas"] = {"matmul": smooth_quant_alpha}
+
+        mode = None
+        if activation_format:
+            mode_map = {
+                "f8e4m3": "fp8_e4m3",
+                "f8e5m2": "fp8_e5m2",
+            }
+            mode = mode_map[activation_format]
```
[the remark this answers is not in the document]
KodiaqQ: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/parameters.py#L144-L154

optimum/intel/openvino/configuration.py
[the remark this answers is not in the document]
KodiaqQ: Do you have a better proposal?

KodiaqQ: `OVQuantizationConfigBase` was introduced for weight compression purposes. I do not understand, why we should use such a specific "base" everywhere. Are there any reasons?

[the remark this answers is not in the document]
KodiaqQ: `_quantization_config_from_dict` contains hard-coded configuration detection and instantiation. Do you think is it a scalable approach to extend?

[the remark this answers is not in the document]
nikita-savelyevv: We currently support providing quantization config as dict for weight and activation quantization. I believe we should also support this for mixed precision quantization.

This should also become useful when implementing ticket 157523. The plan there was to allow specifying quantization config as json file from cli in which case dict to config class conversion should be supported.

`_quantization_config_from_dict` implementation may not be ideal, but it does the job. Feel free to suggest improvements if you see how it can be done better.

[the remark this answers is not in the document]
nikita-savelyevv: Yes. I propose to use `OVGeneralQuantizationConfig` only for mixed precision quantization. It can then be named `OVMixedQuantizationConfig`. I've linked an example of how it could look [above](https://github.com/huggingface/optimum-intel/compare/main...nikita-savelyevv:optimum-intel:ns/nf4_f8e4m3_proposal).

nikita-savelyevv: We currently rely on the quantization config class to belong to `OVQuantizationConfigBase` in some parts of the code. For example:
- https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_seq2seq.py#L1027
- https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/quantization.py#L728

This is useful for clarity, when accessing some shared properties of quantization configs.

PR #1130, opened 2025-01-28
align rt_info int8 models compressed by default and via config

eaidova:
# What does this PR do?

currently --weight-format int8 and models compressed to int8 by default based on number of parameters has slightly different behaviour in respect of kv cache compression. 

see example (left default, right with --weight-format int8):

image

## Before submitting


AlexKoff88:
Merging as a hotfix

PR #1133, opened 2025-01-29
update tests for compatibility with ov 2025.0

eaidova:
# What does this PR do?

* openvino tokenizers in 2025.0 supports export detokenizer for bert and distilbert tokenizers
* mt5 slightly dropped in accuracy due to internal sdpa fusion (in encoder part)

## Before submitting

### DIFF
--- a/tests/openvino/test_quantization.py
+++ b/tests/openvino/test_quantization.py
@@ -1085,7 +1085,7 @@ class OVQuantizerQATest(unittest.TestCase):
 
 
 class OVTrainerTest(unittest.TestCase):
-    SUPPORTED_ARCHITECTURES_WITH_EXPECTED_QUANTIZED_MATMULS = (("albert", 63, 39),)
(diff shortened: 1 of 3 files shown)


tests/openvino/test_quantization.py
```diff
@@ -1085,7 +1085,7 @@ def preprocess_function(examples, tokenizer):
 
 
 class OVTrainerTest(unittest.TestCase):
-    SUPPORTED_ARCHITECTURES_WITH_EXPECTED_QUANTIZED_MATMULS = (("albert", 63, 39),)
```
nikita-savelyevv, line 1088: This change is due to the fact that this test is somewhat non-deterministic, please see ticket 142945

nikita-savelyevv (approved):
Thanks!

IlyasMoutawwakil (approved):

PR #1140, opened 2025-02-04
support janus model

eaidova:
# What does this PR do?
conversion required fix on optimum side: https://github.com/huggingface/optimum/pull/2179

model loading required to install janus package before conversion (and for preprocessing usage)

```
pip install "git+https://github.com/deepseek-ai/Janus.git"
```

```
from io import BytesIO
from pathlib import Path

import requests
from janus.models import VLChatProcessor
from PIL import Image
from transformers import TextStreamer

model_id = "deepseek-ai/Janus-Pro-1B"

processor = VLChatProcessor.from_pretrained(model_id)
```

## Multimodal understanding
```
input_prompt = "Describe image in details"
image_path = Path("cat_in_box.png")

if not image_path.exists():
    response = requests.get(
        "https://github.com/openvinotoolkit/openvino_notebooks/assets/29454499/d5fbbd1a-d484-415c-88cb-9986625b7b11"
    )
    image = Image.open(BytesIO(response.content)).convert("RGB")
    image.save(image_path)

image = Image.open(image_path)

inputs = model.preprocess_inputs(input_prompt, image, processor)
streamer = TextStreamer(processor.tokenizer, skip_prompt=True, skip_special_tokens=True)

model.generate(**inputs, streamer=streamer, max_new_tokens=100, do_sample=False)
```

Answer:
> The image shows a gray tabby cat lying inside an open cardboard box on a light-colored carpet. The cat is lying on its back with its belly exposed, legs up in the air, and its tail curled around its body. The background includes a beige couch and a bright, airy room with natural light streaming in, creating a cozy and relaxed atmosphere.

## Text to Image generation

```
image_gen_prompt = "A cute and adorable baby fox with big brown eyes, autumn leaves in the background enchanting,immortal,fluffy, shiny mane,Petals,fairyism,unreal engine 5 and Octane Render,highly detailed, photorealistic, cinematic, natural colors."

images = model.generate_image(processor, image_gen_prompt, parallel_size=1)

images[0].save("fox.png")
```

Generated Image
fox

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -234,7 +234,6 @@ def main_export(
         library_name=library_name,
     )
 
-
     do_gptq_patching = False
     do_quant_patching = False
     custom_architecture = False
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -213,8 +213,7 @@ def main_export(
             subfolder=subfolder,
             revision=revision,
             cache_dir=cache_dir,
-            token=token,
-            library_name=library_name,
+            token=token
         )
         if library_name == "sentence_transformers":
             logger.warning(
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -13,7 +13,6 @@
 #  limitations under the License.
 
 import gc
-import importlib
 import logging
 import operator
 import warnings
@@ -40,13 +39,13 @@ from optimum.intel.utils.modeling_utils import (
     _infer_library_from_model_name_or_path,
     _OpenClipForZeroShotImageClassification,
 )
-from optimum.utils.save_utils import maybe_load_preprocessors
 
 from .utils import (
     _MAX_UNCOMPRESSED_SIZE,
     MULTI_MODAL_TEXT_GENERATION_MODELS,
     clear_class_registry,
     deduce_diffusers_dtype,
+    load_preprocessors,
 )
 
 
@@ -193,6 +192,7 @@ def main_export(
     ```
     """
     from optimum.exporters.openvino.convert import export_from_model
+
     if use_auth_token is not None:
         warnings.warn(
             "The `use_auth_token` argument is deprecated and will be removed soon. Please use the `token` argument instead.",
@@ -214,7 +214,7 @@ def main_export(
             revision=revision,
             cache_dir=cache_dir,
             token=token,
-            library_name=library_name
+            library_name=library_name,
         )
         if library_name == "sentence_transformers":
             logger.warning(
@@ -445,7 +445,7 @@ def main_export(
                 possible_synonyms = ""
             logger.info(f"Automatic task detection to {task}{possible_synonyms}.")
 
-        preprocessors = maybe_load_preprocessors(
+        preprocessors = load_preprocessors(
             model_name_or_path, subfolder=subfolder, trust_remote_code=trust_remote_code
         )
 
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -442,7 +442,7 @@ def main_export(
             logger.info(f"Automatic task detection to {task}{possible_synonyms}.")
 
         preprocessors = load_preprocessors(
-            model_name_or_path, subfolder=subfolder, trust_remote_code=trust_remote_code
+            model_name_or_path, subfolder=subfolder, trust_remote_code=trust_remote_code, model_type=model_type
         )
 
         submodel_paths = export_from_model(
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -213,7 +213,7 @@ def main_export(
             subfolder=subfolder,
             revision=revision,
             cache_dir=cache_dir,
-            token=token
+            token=token,
         )
         if library_name == "sentence_transformers":
             logger.warning(
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 
 import gc
+import importlib
 import logging
 import operator
 import warnings
@@ -192,7 +193,6 @@ def main_export(
     ```
     """
     from optimum.exporters.openvino.convert import export_from_model
-
     if use_auth_token is not None:
         warnings.warn(
             "The `use_auth_token` argument is deprecated and will be removed soon. Please use the `token` argument instead.",
@@ -214,6 +214,7 @@ def main_export(
             revision=revision,
             cache_dir=cache_dir,
             token=token,
+            library_name=library_name
         )
         if library_name == "sentence_transformers":
             logger.warning(
@@ -232,6 +233,9 @@ def main_export(
         token=token,
         library_name=library_name,
     )
+
+    logger.warn(task)
+    logger.warn(library_name)
 
     do_gptq_patching = False
     do_quant_patching = False
@@ -458,6 +462,7 @@ def main_export(
             device=device,
             trust_remote_code=trust_remote_code,
             patch_16bit_model=patch_16bit,
+            library_name=library_name,
             **kwargs_shapes,
         )
 
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -234,8 +234,6 @@ def main_export(
         library_name=library_name,
     )
 
-    logger.warn(task)
-    logger.warn(library_name)
 
     do_gptq_patching = False
     do_quant_patching = False
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -615,6 +615,7 @@ def export_from_model(
     device: str = "cpu",
     trust_remote_code: bool = False,
     patch_16bit_model: bool = False,
+    library_name: Optional[str] = None,
     **kwargs_shapes,
 ):
     model_kwargs = model_kwargs or {}
@@ -624,9 +625,9 @@ def export_from_model(
             f"Compression of the weights to {ov_config.quantization_config} requires nncf, please install it with `pip install nncf`"
         )
 
-    library_name = _infer_library_from_model_or_model_class(model)
+    library_name = _infer_library_from_model_or_model_class(model, library_name=library_name)
     if library_name != "open_clip":
-        TasksManager.standardize_model_attributes(model)
+        TasksManager.standardize_model_attributes(model, library_name=library_name)
 
     if hasattr(model.config, "export_model_type"):
         model_type = model.config.export_model_type.replace("_", "-")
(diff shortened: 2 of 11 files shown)


optimum/exporters/openvino/__main__.py
```diff
@@ -233,6 +234,9 @@ def main_export(
         library_name=library_name,
     )
 
+    logger.warn(task)
+    logger.warn(library_name)
```
eaidova, line 238:
```suggestion
```

optimum/exporters/openvino/convert.py
```diff
@@ -756,7 +757,7 @@ def export_from_model(
                 logger.warning(
                     f"The generation config will not be saved, saving failed with following error:\n{exception}"
                 )
-
+        logger.warn(preprocessors)
```
eaidova, line 760:
```suggestion
```

tests/openvino/utils_tests.py
```diff
@@ -137,6 +137,7 @@
     "stable-diffusion-xl": "echarlaix/tiny-random-stable-diffusion-xl",
     "stable-diffusion-xl-refiner": "echarlaix/tiny-random-stable-diffusion-xl-refiner",
     "stable-diffusion-3": "yujiepan/stable-diffusion-3-tiny-random",
+    "stable-diffusion-variant": "katuni4ka/tiny-stable-diffusion-torch-custom-variant",
```
eaidova, line 140:
```suggestion
```

optimum/exporters/openvino/convert.py
```diff
@@ -615,9 +616,10 @@ def export_from_model(
             f"Compression of the weights to {ov_config.quantization_config} requires nncf, please install it with `pip install nncf`"
         )
 
-    library_name = _infer_library_from_model_or_model_class(model)
+   if library_name is None:
+        library_name = _infer_library_from_model_or_model_class(model)
```
eaidova, line 620:
```suggestion
    if library_name is None:
        library_name = _infer_library_from_model_or_model_class(model)
```

AlexKoff88:
I wonder why we need to keep `VLChatProcessor` instance outside the model class and if we can move it inside?

AlexKoff88 (approved):

eaidova:
> I wonder why we need to keep `VLChatProcessor` instance outside the model class and if we can move it inside?

not sure that I understand your question. This is standard preprocessing/postprocessing part for transformers-based models (like any other stuff - tokenizers, feature_extractors, image processors, e.t.c), usually it is an independent object (except diffusers case). It may be helpful for VLM models to move it closer as it becomes more complicated and bounded. So possibly we can consider keeping processors for other models as well (it may be helpful for alignment result of save_pretrained and optimum-cli, which also save processors and tokenizers if they are available)

AlexKoff88:
> model = OVModelForVisualCausalLM.from_pretrained(model_id, trust_remote_code=True)
> 
> processor = VLChatProcessor.from_pretrained(model_id)

To clarity, I am just looking at the code in the PR description and wondering why it could not look like this:
```python
...
inputs = model.preprocess_inputs(input_prompt, image)
streamer = TextStreamer(model.tokenizer, skip_prompt=True, skip_special_tokens=True)
model.generate(**inputs, streamer=streamer, max_new_tokens=100, do_sample=False)
...
images = model.generate_image(image_gen_prompt, parallel_size=1)
```

So, `processor` is loaded inside the model and hidden from the user but it can be acquired like `model.processor`.

But from what I understood your implementation is aligned with diffusers, right?

eaidova:
@IlyasMoutawwakil  @echarlaix  could you please take a look?

optimum/exporters/openvino/convert.py
```diff
@@ -615,9 +616,9 @@ def export_from_model(
             f"Compression of the weights to {ov_config.quantization_config} requires nncf, please install it with `pip install nncf`"
         )
 
-    library_name = _infer_library_from_model_or_model_class(model)
+    library_name = _infer_library_from_model_or_model_class(model, library_name=library_name)
     if library_name != "open_clip":
-        TasksManager.standardize_model_attributes(model)
+        TasksManager.standardize_model_attributes(model, library_name=library_name)
```
echarlaix, line 621: why is this needed here ? (should only be an issue for open_clip no ?)

optimum/exporters/openvino/model_configs.py
```diff
@@ -134,6 +139,17 @@ def init_model_configs():
...
+    TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS[
+        "any-to-any"
+    ] = TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS["text-generation"]
+
+    # for model registration in auto transformers classses
+    if importlib.util.find_spec("janus") is not None:
+        try:
+            from janus.models import MultiModalityCausalLM  # noqa: F401
+        except ImportError:
+            pass
```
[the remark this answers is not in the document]
eaidova: janus code stored in separated repo and required this package installing. init module contains logic for registering model into AutoModel class in transformers, after such import we can load model with

AutoModel.from_pretrained

optimum/exporters/openvino/model_configs.py
```diff
@@ -134,6 +139,17 @@ def init_model_configs():
         "image-text-to-text"
     ] = TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS["text-generation"]
 
+    TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS[
+        "any-to-any"
```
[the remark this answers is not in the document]
eaidova: such task is automatically detected for [janus model](https://huggingface.co/deepseek-ai/Janus-Pro-7B), as model doing multiple things simulteniusly (image-text-to-text and text-to-image). it also used for some complex pipelines like minicpmv2.6-o (asr+vlm + tts), moshi (asr+llm+tts) that can be also supported in future

as there is no associated dedicated autoclass for loading, we need guide task manager for loading pipelines for such task

optimum/exporters/openvino/utils.py
```diff
@@ -310,3 +312,20 @@ def save_preprocessors(
...
+        src_name_or_path, subfolder=subfolder, trust_remote_code=trust_remote_code
+    )
+    if importlib.util.find_spec("janus") is not None:
+        from janus.models import VLChatProcessor
+
+        try:
+            processor = VLChatProcessor.from_pretrained(
+                src_name_or_path, subfolder=subfolder, trust_remote_code=trust_remote_code
+            )
+            preprocessors.append(processor)
```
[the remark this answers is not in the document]
eaidova: I can provide model_type and load processor only if janus selected. The problem that without explicit call it is not registered in AutoProcessor, that is why I added it explicitly. If we will not save it during conversion, it may bring inconvenience with usage optimum-cli that usually includes saving preprocessing steps.

optimum/exporters/openvino/model_configs.py
```diff
@@ -2773,3 +2805,274 @@ class MT5OpenVINOConfig(T5OpenVINOConfig):
...
+                [self.batch_size, 576],
+                max_value=self.normalized_config.config.params.image_token_size,
+                framework=framework,
+                dtype=int_dtype,
+            )
+        if input_name == "image_shape":
+            import torch
+
+            return torch.tensor(
+                [self.batch_size, self.normalized_config.config.params.n_embed, 24, 24], dtype=torch.int64
```
echarlaix, line 2866: could you add a comment to explain where 24 is coming from ?

optimum/exporters/openvino/model_configs.py
```diff
@@ -2773,3 +2805,274 @@ class MT5OpenVINOConfig(T5OpenVINOConfig):
...
+        if input_name == "image_ids":
+            return self.random_int_tensor(
+                [self.sequence_length],
+                max_value=self.normalized_config.config.params.image_token_size,
+                framework=framework,
+                dtype=int_dtype,
+            )
+        if input_name == "code_b":
+            return self.random_int_tensor(
+                [self.batch_size, 576],
```
echarlaix, line 2857: same could you add a comment to explain where 576 is coming from ?

echarlaix (commented):
Thanks for the addition @eaidova !

optimum/exporters/openvino/convert.py
eaidova: I believe calling this explicitly is legacy code and can be removed now. standartize_model_attributes now is part of get_model_for_task
https://github.com/huggingface/optimum/blob/6335599da3bad65609620d1f64512e8241fd5aaa/optimum/exporters/tasks.py#L2303

Should I remove it?

optimum/exporters/openvino/utils.py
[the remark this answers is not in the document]
eaidova: added model_type checking for loading preprocessing

optimum/exporters/openvino/convert.py
eaidova: with passing library_name, such method should work with open_clip as well.
initial reason why I added this. library_name getting from model instance uses submodules and requires to have 'transformers`, while for janus some of model classes starts with own namespace and this check failed, however file based approach treats it as transformers based model.

optimum/exporters/openvino/model_configs.py
eaidova: I used such values during debugging, replaced with calculation used config

optimum/exporters/openvino/model_configs.py
eaidova: I used such values during debugging, replaced with calculation used config

echarlaix:
Apologies for the delay @eaidova!!  We could need to make sure we are compatible with https://github.com/huggingface/transformers/pull/36053 before merging so that we don't have any issues when transformers v4.50 is out

eaidova:
> Apologies for the delay @eaidova!! We could need to make sure we are compatible with [huggingface/transformers#36053](https://github.com/huggingface/transformers/pull/36053) before merging so that we don't have any issues when transformers v4.50 is out

I carefully examinated provided PR, it seems to be that required reconvert model from original janus modeling to transformers realization. My PR is working with original Janus provided models, so there should not be issues because `model_type` in configs are different with what we use. This transformers-compatible models enabling will require additional efforts to add conversion configs and model class for inference, but it is non blocker for this PR and can be supported when transforemrs PR will be merged.

So my suggestion is to resolve conflicts and merge this PR if there is no any additional issues

eaidova:
@IlyasMoutawwakil could you please take a look?

eaidova:
@echarlaix , I understand your position regarding remote code and agree with that, but also would like to note that besides long time waiting for reviewing, it also brings a lot of inconvenience for uses:

1. code in transformers is compatible only with community models versions converted from original checkpoints 
https://huggingface.co/deepseek-community while original models still exists and working with remote code https://huggingface.co/collections/deepseek-ai/janus-6711d145e2b73d369adfd3cc and required reconvert model
2. it is not possible to support transformers released version without usage features that currently available only in transformers main branch (no any released package support it and I speak about changes related generate method parameters for images, so it will be hard to support just copying from transformers overriding methods) - extra time required waiting next transformers package to do that... non speaking about how long we already review this PR

From my point of view, it will be better to integrate remote code model for now and deprecate it in next release (when transformers that support janus officially released) in flavor usage of official transformers code (as anyway models live by own independent life on huggingdace hub)

echarlaix:
@eaidova I agree with your point, waiting for the transformers integration + release can result in a delayed release, also it will be only compatible with the subset of models which are compatible with this integration. In my opinion we should stop adding any official new export to optimum-intel for specific modeling (needing `trust_remote_code=True`) that requires a modeling patch, as for these export many issues could arise on which we don't have any control on. Now for cases where we need it, what do you think about adding it in an example export script instead specifying what types of models this export is targeted for ? An other option could be to push it on a dedicated branch, what do you think of this option ? Also concerning janus support, it's already available through the [v4.51.3-Janus-preview]( https://github.com/huggingface/transformers/releases/tag/v4.51.3-Janus-preview) tag that we can use for testing, and we can add official support once v4.52 is released (~2 weeks). This PR has been opened for a while, apologies for this. Let me know if you'd like me to take care of the janus export using the v4.51.3-Janus-preview tag!

PR #1145, opened 2025-02-05
maira2 support

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


optimum/exporters/openvino/model_configs.py
```diff
@@ -1488,6 +1488,7 @@ def __init__(
         float_dtype: str = "fp32",
         behavior: LlavaConfigBehavior = LlavaConfigBehavior.VISION_EMBEDDINGS,
         preprocessors: Optional[List[Any]] = None,
+        **kwargs,
```
echarlaix, line 1491: why is it needed for MAIRA-2 models ?

echarlaix (commented):
LGTM, thanks @eaidova! Could you add a test before we can merge ?

echarlaix (approved):

optimum/exporters/openvino/model_configs.py
eaidova: maira processor has own arguments different from standard, this is needed to support inference for this case

eaidova: automatic task detection in some cases select text-generation-with-past as task this is required to pass some extra arguments to config

eaidova:
@echarlaix I added test and model into supported list. Could you please one more time?

eaidova:
@echarlaix could you please help to rerun precommit. I updated model remote code for compatibility with python3.9

echarlaix:
> @echarlaix could you please help to rerun precommit. I updated model remote code for compatibility with python3.9

Done, looks like one last thing needs to be updated https://huggingface.co/katuni4ka/tiny-random-maira2/blob/main/processing_maira2.py#L347 to fix :
```
SyntaxError: non-default argument follows default argument
```

eaidova:
> > @echarlaix could you please help to rerun precommit. I updated model remote code for compatibility with python3.9
> 
> Done, looks like one last thing needs to be updated https://huggingface.co/katuni4ka/tiny-random-maira2/blob/main/processing_maira2.py#L347 to fix :
> 
> ```
> SyntaxError: non-default argument follows default argument
> ```

fixed

PR #1148, opened 2025-02-06
[OV] Add support for `nf4_f8e4m3` quantization mode

nikita-savelyevv:
# What does this PR do?

### Changes

- Added `OVMixedQuantizationConfig` for mixed precision quantization scenario. It is initialized with an instance of `OVWeightQuantizationConfig` and an instance of `OVQuantizationConfig`.
- Added `nf4_f8e4m3`, `int4_f8e4m3`, `nf4_f8e5m2`, `int4_f8e5m2` as possible values of `--quant-mode` CLI argument. This performs mixed precision quantization, compressing weights to `nf4`/`int4` precision and activations to `f8e4m3`/`f8e5m2`.
- Quantization configs refactoring. `OVQuantizationConfigBase` now contains only model-related parameters. Added `to_nncf_dict()` method to quantization configs for convenience.
- Renamed `OVWeightQuantizationConfig.weight_format` to `OVWeightQuantizationConfig.dtype` and `OVQuantizationConfig.activation_format` to `OVQuantizationConfig.dtype`. The latter is done because when OVQuantizationConfig is used, not only activations but also weights are quantized, so `activation_format` does not correctly represent what actually happens. `OVWeightQuantizationConfig.weight_format` is renamed for consistency.
- `OVBaseModel._prepare_quantization_config()` can now create instances of configs other than `OVWeightQuantizationConfig`.

### Examples
CLI
```shell
optimum-cli export openvino -m meta-llama/Llama-3.1-8B --quant-mode nf4_f8e4m3 --dataset wikitext2 ./llama-3.1-8b_nf4_f8e4m3
```

Python API:
```python
    model_id="meta-llama/Llama-3.1-8B",
    quantization_config=OVMixedQuantizationConfig(
        weight_quantization_config=OVWeightQuantizationConfig(bits=4, dtype="nf4"),
        full_quantization_config=OVQuantizationConfig(dtype="f8e4m3"),
        dataset="wikitext2",
    )
)
```

Some of these changes were implemented thanks to @nikita-malininn .

## Before submitting


optimum/commands/export/openvino.py
```diff
@@ -375,16 +359,22 @@ def run(self):
...
-                "bits": 8,
-                "sym": self.args.sym or False,
-                "dataset": self.args.dataset,
-                "num_samples": self.args.num_samples,
-                "smooth_quant_alpha": self.args.smooth_quant_alpha,
-                "trust_remote_code": self.args.trust_remote_code,
-            }
+            if self.args.quant_mode == "nf4_f8e4m3":
+                wc_config = prepare_wc_config(self.args, _DEFAULT_4BIT_CONFIG)
+                wc_config["weight_format"] = "nf4"
```
AlexKoff88, line 364: We need to set FP8 as a backup precision for the layers that are not in `nf4` by default.

tests/openvino/test_quantization.py
```diff
@@ -119,16 +119,26 @@ class OVQuantizerTest(unittest.TestCase):
...
             ),
             (13,),
             (16,),
         ),
+        (
+            OVModelForCausalLM,
+            "llama",
+            dict(
+                weight_quantization_config=dict(bits=4, weight_format="nf4", group_size=16),
+                full_quantization_config=dict(activation_format="f8e4m3"),
```
AlexKoff88, line 135: What is the weight format in this case when you define just `activation_format`?

optimum/commands/export/openvino.py
nikita-savelyevv: I will need to prepare this feature in NNCF first

tests/openvino/test_quantization.py
nikita-savelyevv: Discussed offline: weights will also be quantized to `activation_format` data type

optimum/commands/export/openvino.py
nikita-savelyevv: Implemented a workaround by providing a "none" backup precision. This keep sensitive weights in original precision after weight compression step. They will later be quantized to FP8 by quantization step.

AlexKoff88:
@l-bat, can you please review this PR as well?

optimum/intel/openvino/quantization.py
```diff
@@ -1008,6 +1014,7 @@ def _weight_only_quantization(
     model: openvino.runtime.Model,
     quantization_config: Union[OVWeightQuantizationConfig, Dict],
     calibration_dataset: Optional[Union[nncf.Dataset, Iterable]] = None,
+    remove_kv_cache_precision_flag: Optional[bool] = True,
```
l-bat, line 1017: Could you please explain to me why we need this `remove_kv_cache_precision_flag`? Are there any plans to remove it in the future when the accuracy issue with KV cache compression is fixed?

optimum/intel/openvino/quantization.py
```diff
@@ -1172,31 +1135,80 @@ def _hybrid_quantization(
...
+    wc_config = quantization_config.clone()
+    wc_config.ignored_scope = {}
+    if any(op.get_type_name() == "Convolution" for op in model.get_ops()):
+        wc_config.ignored_scope["types"] = ["Convolution"]
+
+    q_config_ignored_scope = {"names": _collect_ops_with_weights(model)}
+    q_config = OVQuantizationConfig(
+        ignored_scope=q_config_ignored_scope,
+        num_samples=quantization_config.num_samples or 200,
+        smooth_quant_alpha=-1,
```
l-bat, line 1148:
```suggestion
        # SQ algo should be disabled for MatMul nodes because their weights are already compressed
        advanced_parameters=nncf.AdvancedQuantizationParameters(
            smooth_quant_alphas=AdvancedSmoothQuantParameters(matmul=-1)
        ),
```

l-bat: What about other quantization parameters, such as `model_type=nncf.ModelType.TRANSFORMER`?

optimum/intel/openvino/configuration.py
```diff
@@ -542,29 +535,63 @@ def post_init(self):
...
+        signed_bitness = {4: "int4", 8: "int8"}
+        mode = self.dtype if self.dtype else signed_bitness[self.bits]
+        if mode in signed_bitness.values():
+            mode += "_sym" if self.sym else "_asym"
+        if mode == "mxfp4":
+            mode = "e2m1"
+        mode = nncf.CompressWeightsMode(mode)
+
+        awq = None
+        if self.quant_method == "awq" or self.quant_method == OVQuantizationMethod.AWQ:
```
l-bat, line 575: I believe that `self.quant_method` is an instance of `OVQuantizationMethod`, according to https://github.com/huggingface/optimum-intel/blob/f61b7e86264c2719610338a56ac1b971c6047d2a/optimum/intel/openvino/configuration.py#L442

optimum/intel/openvino/configuration.py
```diff
@@ -782,3 +848,92 @@ def to_dict(self) -> Dict[str, Any]:
...
+            weight_quantization_config = weight_quantization_config.clone()
+        self.weight_quantization_config = weight_quantization_config
+        wqc = self.weight_quantization_config
+
+        if isinstance(full_quantization_config, dict):
+            full_quantization_config = OVQuantizationConfig.from_dict(full_quantization_config)
+        else:
+            full_quantization_config = full_quantization_config.clone()
+        self.full_quantization_config = full_quantization_config
+        fqc = self.full_quantization_config
```
l-bat, line 911: What do you think about using the `_initialize_quantization_config` method to handle the initialization of `weight_quantization_config` and `full_quantization_config` to avoid code duplication?

optimum/intel/openvino/configuration.py
```diff
@@ -782,3 +848,92 @@ def to_dict(self) -> Dict[str, Any]:
...
+                A transformers processor used to process the dataset inputs.
+            trust_remote_code (`bool`, defaults to `False`):
+                Allows to use custom code for the modeling hosted in the model repository. This option should only be
+                set for repositories you trust and in which you have read the code, as it will execute on your local
+                machine arbitrary code present in the model repository.
+            **kwargs:
+        """
+        if isinstance(weight_quantization_config, dict):
+            weight_quantization_config = OVWeightQuantizationConfig.from_dict(weight_quantization_config)
+        else:
```
AlexKoff88, line 901: You probably need to check the type in the `else` case:
```
elif isinstance(weight_quantization_config, OVWeightQuantizationConfig):
```

The same comment for `full_quantization_config`

optimum/intel/openvino/quantization.py
AlexKoff88: I also did get why we need it. The logic around setting `KV_CACHE_PRECISION` to `fp16` is simple. We enable it only for FP16/FP32 models. For the rest of the cases, we don't specify these options and the runtime enabled `u8` cache by default. Thus, we should always remove `KV_CACHE_PRECISION` when we quantize the model.

nikita-savelyevv: My motivation was not to remove this flag for mixed quantization the same way it is not removed for full quantization. 

If we agree to remove it in case of any type of compression/quantization, I should also remove it for full quantization. But I would like to check whisper performance/quality first before adding this change.

optimum/intel/openvino/quantization.py
nikita-savelyevv: There is possibly some misunderstanding here. `q_config` is an instance `OVQuantizationConfig` which accepts SQ alpha as `smooth_quant_alpha` argument, because this is our wrapper over nncf quantization parameters. Also, model_type is set to "transformer" by default there.

optimum/intel/openvino/quantization.py
```diff
@@ -467,6 +464,15 @@ def _quantize_ovbasemodel(
                 )
                 self.model.model = quantized_model
                 self.model.request = None
+        elif isinstance(quantization_config, OVMixedQuantizationConfig):
+            if calibration_dataset is None:
+                raise ValueError("Calibration dataset is required to run quantization.")
```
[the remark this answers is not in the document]
nikita-savelyevv: Discussed offline: this change is not worth it

optimum/intel/openvino/configuration.py
nikita-savelyevv: Removed

optimum/intel/openvino/quantization.py
nikita-savelyevv: I've done some checks for a couple of whisper architectures and do not observe any differences in generation speed or quality depending on kv cache precision. So I add removal of f16 kv cache precision flag to full_quantization method too.

optimum/intel/openvino/configuration.py
nikita-savelyevv: Added

AlexKoff88 (approved):

l-bat (approved):

optimum/intel/openvino/modeling_base.py
```diff
@@ -250,6 +250,13 @@ def fix_op_names_duplicates(model: openvino.runtime.Model):
 
             from optimum.intel.openvino.quantization import _weight_only_quantization
 
+            if not isinstance(quantization_config, dict) and not isinstance(
+                quantization_config, OVWeightQuantizationConfig
+            ):
```
echarlaix, line 255: can be simplified :
```suggestion
            if not isinstance(quantization_config, (dict, OVWeightQuantizationConfig)):
```

optimum/intel/openvino/modeling_base.py
```diff
@@ -250,6 +250,13 @@ def fix_op_names_duplicates(model: openvino.runtime.Model):
...
+                raise RuntimeError(
+                    "Expected quantization_config to be a dictionary or OVWeightQuantizationConfig object."
+                )
```
echarlaix, line 258: (can be updated but not very important)
```suggestion
                raise TypeError(
                    f"Expected `quantization_config` to be either a dictionary or OVWeightQuantizationConfig object, got {type(quantization_config)}."
                )
```

optimum/intel/openvino/configuration.py
```diff
@@ -493,10 +483,13 @@ def post_init(self):
                 "quantization algorithm is selected and compression ratio is 1.0."
             )
 
+        if self.dtype in ["int4", "int8"]:
+            self.bits = 4 if self.dtype == "int4" else 8
+
```
echarlaix, line 488: should we add a warning if `dtype` and `bits` were originally set to be incompatible to warn users ?

optimum/intel/openvino/configuration.py
```diff
@@ -719,13 +772,22 @@ def __init__(
...
+            quantization_config = self.quantization_config_from_dict(quantization_config)
         self.quantization_config = quantization_config
         self.compression = kwargs.get(
             "compression", None
         )  # A field for backward-compatability of training-time compression parameters
         if self.quantization_config is not None:
-            self.dtype = self.quantization_config.weight_format
+            if isinstance(self.quantization_config, OVWeightQuantizationConfig) or isinstance(
+                self.quantization_config, OVQuantizationConfig
+            ):
```
echarlaix, line 783: can be simplified 
```suggestion
            if isinstance(self.quantization_config, (OVWeightQuantizationConfig, OVQuantizationConfig)):
```

optimum/intel/openvino/configuration.py
```diff
@@ -389,8 +379,8 @@ class OVWeightQuantizationConfig(OVQuantizationConfigBase):
         scale_estimation (`bool`, *optional*):
             Indicates whether to apply a scale estimation algorithm that minimizes the L2 error between the original and
             compressed layers. Providing a dataset is required to run scale estimation.
-        weight_format (`str`, *optional*):
-            Data format weights are compressed to. Possible values: ['int4', 'int8', 'mxfp4', 'nf4'].
+        dtype (`str`, *optional*):
```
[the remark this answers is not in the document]
echarlaix: yes could make sense to add a warning (+ potentially keep compatibility for one or two releases by setting `dtype` in case `weight_format` is provided)

echarlaix (approved):
LGTM, thanks for the addition @nikita-savelyevv

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: Thanks! Changed

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: Done

optimum/intel/openvino/configuration.py
nikita-savelyevv: Done

echarlaix:
Failing tests unrelated so merging, thanks @nikita-savelyevv

PR #1152, opened 2025-02-08
Ignore deprecation warnings from modeling_decoder.py

wgzintel:
There are too many deprecationWarning prints from modeling_decoder.py when running openvino.genai/tools/llm_bench with the latest optimum-intel main branch and openvino master branch.

The prints are as below:
python3.10/site-packages/optimum/intel/openvino/modeling_decoder.py:501: DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments.

The prints are caused by openvino commit  2b442b43689efea4faff581c55a4b020a0af537c
https://github.com/openvinotoolkit/openvino/commit/2b442b43689efea4faff581c55a4b020a0af537c

Test cmd:
python openvino.genai/tools/llm_bench/benchmark.py -m model_path --optimum


eaidova:
@wgzintel I do not think that it is related and should be fixed in such way.

warning and commit that you point are not directly related. commit adds warning to openvino, while you supporters warning produced by numpy

wgzintel:
> @wgzintel I do not think that it is related and should be fixed in such way.
> 
> warning and commit that you point are not directly related. commit adds warning to openvino, while you supporters warning produced by numpy

@eaidova  When I installed the previous commit id of OV(2b442b4),  no deprecationWarning was printed from modeling_decoder.py.  Do you have any other suggestions for ignoring or removing these prints? Because there are so many of them.  Can np.array() be replaced by np.asarray()?

eaidova:
@wgzintel , I believe, it should be fixed in openvino python API side (see my comment for provided commit). Generally it is not good practice to force warnings visibility globally (I mean not only ignore like in your PR, but also filtering in python API commit). It is usually recommended to use filter once instead of always for avoid such issues.

Also I do not think that asarray has difference in behaviour relative to __array__ instance. I believe correct result should be acienved using tensor.numpy() instead see https://github.com/huggingface/optimum-intel/pull/1153

PR #1153, opened 2025-02-10
use Tensor.numpy() instead np.array(Tensor)

eaidova:
# What does this PR do?

* numpy 2.0 produces deprecation warning when np.array(torch.tensor()) used  https://github.com/pytorch/pytorch/issues/136264
* in some cases usage model with pipelines assign input tensor on different device (e.g. on M1 MacOS, default device for pipeline in mps:0) that may lead to unexpected error inside with attempt to cast to array non-cpu offloaded tensor

## Before submitting


eaidova:

echarlaix (approved):
LGTM!

PR #1159, opened 2025-02-12
fix misplaced configs saving

eaidova:
# What does this PR do?

during diffusion pipeline loading, `_name_or_path` field overridden only for main pipeline config, while its components still contains paths for original pipeline. As te result, ioriginal config loaded using diffusers pytorch pipeline and saved on disk for ov diffusers pipeline are different (e.g. for too old models where config does not contains some important fields) it may lead to incorrect loading.

## Before submitting


eaidova:
@l-bat could you please review

AlexKoff88 (approved):

AlexKoff88:
It looks like that the logic in the tests should be changed as well

l-bat (approved):
Thanks!

eaidova:
@echarlaix  @IlyasMoutawwakil  could you please take a look?

echarlaix (approved):
Thanks @eaidova

PR #770, opened 2024-06-17
Fixed issues in the Hybrid quantization notebook

AlexKoff88:

PR #1162, opened 2025-02-14
Remove unnecessary SD reload from saved dir

l-bat:
# What does this PR do?

I only deleted one line where `int8_pipe` was reloaded from the saved directory. I believe we can compare the latency of the initial int8_pipe optimized via `OVQuantizer`.
image

## Before submitting


l-bat:
@AlexKoff88, please take a look. It was added in https://github.com/huggingface/optimum-intel/pull/770

AlexKoff88 (approved):

PR #1164, opened 2025-02-17
avoid extra reshaping to max_model_lenght for unet

eaidova:
# What does this PR do?

unet in stable diffusion pipelines always is saved with dynamic shapes during export and uses reshape to set max_model_lenght from tokenizer before starting inference. Suggestion is remove reshape every model loading and set max_model_len as default value for encoder_hidden_state during export if it is known

## Before submitting


echarlaix (approved):
LGTM!

PR #1167, opened 2025-02-18
Deprecate OVTrainer

nikita-savelyevv:
# What does this PR do?

Deprecate `OVTrainer` and related logic as stated in #890.

## Before submitting


echarlaix (approved):

echarlaix:
Works for me, @AlexKoff88 feel free to merge

AlexKoff88:
@echarlaix, we decided to deprecate and remove it due to low interest. However, we are working on QAT for LLMs which could be contributed as an example by inheriting and implementing the HF Trainer interface instead of bringing it to the core functionality of Optimum-Intel.

AlexKoff88:
@ljaljushkin, please take a look as well.

AlexKoff88:
> > @echarlaix, we decided to deprecate and remove it due to low interest. However, we are working on QAT for LLMs which could be contributed as an example by inheriting and implementing the HF Trainer interface instead of bringing it to the core functionality of Optimum-Intel.
> 
> Why deprecating the `OVTrainer` in this context, what would be the benefit of having in the examples and not in optimum-intel directly as currently the case ?

The main reason is to get flexibility in this flow:
- In the example, we will show how to implement QAT Trainer for LLMs with NNCF so that it is PyTorch model in and out
- User can modify the Trainer from the example according to the model training pipeline it has. NNCF part remains the same.
- User decides how to use optimized model: go to OV IR, use torch.compile, executorch, export to ONNX for ORT OV EP, etc.

AlexKoff88 (approved):

PR #1173, opened 2025-02-24
deprecate TSModelForCausalLM

echarlaix:


jiqing-feng:
Need to rewrite the `prepare_jit_inputs` in ipex model

PR #1175, opened 2025-02-25
enable phi

jiqing-feng:
Enable phi model.

Tests will be ready after #1176 merged


jiqing-feng:
The failed ipex test is quiet strange, this change didn't have any impack on llama. Besides, I can pass all ipex tests locally with same env.
`FAILED tests/ipex/test_modeling.py::IPEXModelForCausalLMTest::test_patched_model_0_llama2 - AssertionError: False is not true`
```
tests/ipex/test_modeling.py:445: in test_patched_model
    self.assertTrue(torch.allclose(ipex_outputs.logits[0], exported_outputs.logits[0], atol=1e-7))
E   AssertionError: False is not true
```

I guess that's because we added a new testing model `phi` which changed the previous tests parallelism strategy (All tests run together and compete the computation resources). We can change the `atol` to pass the test.

PR #1176, opened 2025-02-26
Update ipex Ci to torch 2.6

jiqing-feng:
Update ipex tests to torch 2.6

As ipex tests usually takes too much time, we'd like to only keep 1 test in CI.

Hi @IlyasMoutawwakil , please review these changes. Thanks!


optimum/intel/ipex/modeling_base.py
```diff
@@ -61,8 +61,8 @@
 _IPEX_SUPPORT_MODEL_TYPES = ("llama", "bert", "vit", "falcon", "gpt2", "qwen2")
 _IPEX_EXPORTED_GENERATION_METHODS = ("sample", "greedy_search", "beam_sample", "beam_search", "assisted_generation")
 _IPEX_MINIMUM_VERSION_FOR_COMPILE = "2.5.0"
-# TODO: Some models are already fixed in torch 2.6, will enable them when torch upgrading to 2.6
-_COMPILE_NOT_READY_MODEL_TYPES = ("electra", "roformer", "gpt_neox", "beit", "llama", "falcon", "gpt2", "qwen2")
+# Page attention model cannot use torch.compile for now.
+_COMPILE_NOT_READY_MODEL_TYPES = ("llama", "falcon", "gpt2", "qwen2")
```
echarlaix, line 65: should we check the torch version and adapt `_COMPILE_NOT_READY_MODEL_TYPES` depending on it ?

.github/workflows/test_ipex.yml
```diff
@@ -18,8 +18,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        transformers-version: ["4.47.0", "4.47.1"]
-        torch-version: ["2.4.0", "2.5.*"]
+        transformers-version: ["4.47.*"]
+        torch-version: ["2.6.0"]
```
echarlaix, line 22: we can keep a test with a lower torch version, both option works for me

echarlaix (approved):

optimum/intel/ipex/modeling_base.py
jiqing-feng: Not really, the ipex op is not compatible with torch.compile for now. We will remove it once ipex fixed this issue.

.github/workflows/test_ipex.yml
jiqing-feng: I think there is no need to keep the previous version of pytorch tests, as the ipex tests take too much time which might be hanged if too many tasks run together.

echarlaix: ok works for me

optimum/intel/ipex/modeling_base.py
echarlaix: my question is more : shouldn't `_COMPILE_NOT_READY_MODEL_TYPES` depend on the torch version ?

```python
if is_torch_version("<", "2.6"):
    _COMPILE_NOT_READY_MODEL_TYPES = ("electra", "roformer", "gpt_neox", "beit", "llama", "falcon", "gpt2", "qwen2")
else:
    _COMPILE_NOT_READY_MODEL_TYPES = ("llama", "falcon", "gpt2", "qwen2")
```

echarlaix: thanks for adding in https://github.com/huggingface/optimum-intel/pull/1176/commits/af1536ffc9afc7f45439cf7eea1c2122fa2262be @jiqing-feng, merging now

PR #1179, opened 2025-02-27
Propagate OV*QuantizationConfig kwargs to nncf calls

nikita-savelyevv:
# What does this PR do?

Introduce ability to provide additional arguments via kwargs for `OVWeightQuantizationConfig` and `OVQuantizationConfig`. These additional kwargs will be propagated to `nncf.compress_weights()` or `nncf.quantize()` calls correspondingly.

For example this allows to perform the following:
```python
    <model_id>,
    quantization_config=OVWeightQuantizationConfig(
        bits=8,
        advanced_parameters=nncf.AdvancedCompressionParameters(statistics_path="asd"),    # will be passed to nncf.compress_weights() call
    )
)
```

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -294,7 +294,8 @@ class OVQuantizationConfigBase(QuantizationConfigMixin):
         dataset: Optional[Union[str, List[str]]] = None,
         tokenizer: Optional[str] = None,
         processor: Optional[str] = None,
-        trust_remote_code: bool = False,
+        trust_remote_code: Optional[bool] = False,
+        init_kwargs: Optional[dict] = None,
         **kwargs,
     ):
         """
@@ -314,6 +315,8 @@ class OVQuantizationConfigBase(QuantizationConfigMixin):
                 Allows to use custom code for the modeling hosted in the model repository. This option should only be
                 set for repositories you trust and in which you have read the code, as it will execute on your local
                 machine arbitrary code present in the model repository.
+            init_kwargs ('dict', *optional*):
+                Additional parameters for NNCF calls. This explicit argument is needed for deserialization from dict.
         """
         self.num_samples = num_samples
         self.dataset = dataset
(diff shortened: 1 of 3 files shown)


nikita-savelyevv:
@ljaljushkin, please take a look

docs/source/openvino/optimization.mdx
```diff
@@ -87,6 +87,8 @@ quantization_config = OVWeightQuantizationConfig(
 )
 ```
 
+Note: `OVWeightQuantizationConfig` also accepts key word arguments not listed in its constructor. In this case such arguments will be passed directly to `nncf.compress_weights()` call. This is useful for passing additional parameters to the quantization algorithm.
```
AlexKoff88, line 90:
```suggestion
Note: `OVWeightQuantizationConfig` also accepts keyword arguments that are not listed in its constructor. In this case such arguments will be passed directly to `nncf.compress_weights()` call. This is useful for passing additional parameters to the quantization algorithm.
```

AlexKoff88 (approved):

optimum/intel/openvino/configuration.py
```diff
@@ -624,6 +634,7 @@ def to_nncf_dict(self) -> Dict[str, Any]:
             "gptq": self.gptq,
             "lora_correction": self.lora_correction,
             "backup_mode": backup_mode,
+            **self.init_kwargs,
```
ljaljushkin, line 637: Why do you need 2 different names `init_kwargs` and `kwargs`?

nikita-savelyevv: My reasoning was not to create a property named `kwargs` at `OVQuantizationConfigBase` to avoid confusion as name "kwargs" is commonly used to refer to additional keyword arguments not listed in a method signature. That's why I chose `init_kwargs` to convey the meaning that these are the keyword arguments which were provided at config initialization.

ljaljushkin (approved):

optimum/intel/openvino/configuration.py
nikita-savelyevv: Renamed

echarlaix (approved):

PR #1187, opened 2025-03-05
align loading dtype logic for diffusers with other models

eaidova:
# What does this PR do?

Currently diffusers loading behaviour is different from logic that we have for transformers, because users preferences are ignored and model loaded with auto detected dtype. Potentially it may lead to some accuracy issues (e.g. pipeline dtype determined by unet/transformer dtype, but text encoders may have unaligned types, when fp32 enforced, we will load bf16 if main model is bf16)and unexpected by users behaviour e.g. if user requests fp16 model using --weight-format argument in optimum-cli and original model is bf16, model will be loaded in bf16 and some weights remains in bf16 after export

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -240,6 +240,9 @@ def main_export(
     loading_kwargs = model_loading_kwargs or {}
     if variant is not None:
         loading_kwargs["variant"] = variant
+    dtype = loading_kwargs.get("torch_dtype", None)
+    if isinstance(dtype, str):
+        dtype = getattr(torch, dtype) if dtype != "auto" else dtype
     if library_name == "transformers":
         config = AutoConfig.from_pretrained(
             model_name_or_path,
@@ -302,9 +305,8 @@ def main_export(
                 "Please provide custom export config if you want load model with remote code."
             )
             trust_remote_code = False
-        dtype = loading_kwargs.get("torch_dtype")
-        if isinstance(dtype, str):
-            dtype = getattr(config, "torch_dtype") if dtype == "auto" else getattr(torch, dtype)
+        if dtype == "auto":
+            dtype = getattr(config, "torch_dtype")
 
         if (
             dtype is None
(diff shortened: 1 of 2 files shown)


optimum/exporters/openvino/__main__.py
```diff
@@ -351,16 +351,22 @@ class StoreAttr(object):
                 GPTQQuantizer.post_init_model = post_init_model
     elif library_name == "diffusers" and is_openvino_version(">=", "2024.6"):
```
echarlaix, line 352: shouldn't your modifications also be applied for lower openvino version ? also is the version check for `deduce_diffusers_dtype` in which case would you mind adding a comment ?

optimum/exporters/openvino/__main__.py
```diff
@@ -351,16 +351,22 @@ class StoreAttr(object):
...
-            local_files_only=local_files_only,
-            force_download=force_download,
-            trust_remote_code=trust_remote_code,
-            **_loading_kwargs,
-        )
+        dtype = loading_kwargs.pop("torch_dtype", None)
+        if isinstance(dtype, str):
+            dtype = None if dtype == "auto" else getattr(torch, dtype)
+        if ov_config is not None and ov_config.dtype in {"fp16", "fp32"}:
+            dtype = torch.float16 if ov_config.dtype == "fp16" else torch.float32
```
echarlaix, line 358: what do you think about moving this before the library check as this code should be the same for transformers and diffusers models (expect for `deduce_diffusers_dtype` when dtype not inferred)
also currently it seems that for transformers models `torch_dtype` is prioritized when defined and if not then we check for the `ov_config` (only modified if  `torch_dtype` is not `None`), here it looks like the other way around (`dtype` will be overwritten)
https://github.com/huggingface/optimum-intel/blob/6cceb30aaf1a0126d054d726842ed4945c0227ba/optimum/exporters/openvino/__main__.py#L306-L319

echarlaix (commented):
LGTM!

optimum/exporters/openvino/__main__.py
eaidova: I though about that but the problem that auto treated differently for diffusers and transformers (in transformers, if auto specified, specified in config dtype will be used, in diffusers there is no similar behaviour, only explicit dtype can be used and there is no config file that contains this info, that kis why we use deduce_diffusers_dtype to resolve auto dtype)

optimum/exporters/openvino/__main__.py
eaidova: for lower versions we have no functional support for bf16/fp16 models loading for conversion in openvino. Default case, if dtype is not specified deduce will be used

optimum/exporters/openvino/__main__.py
eaidova: I moved common part (before resolving auto before library-specific code and also add ov_config precision check only for case when precision is not specified. Please check one more time

echarlaix: nice, that works for me!

echarlaix (approved):
will merge once the tests pass, thanks @eaidova !

PR #1188, opened 2025-03-05
fix logits_to_keep

jiqing-feng:
Update parameter name as transformers deprecated `num_logits_to_keep`

### DIFF
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -489,7 +489,7 @@ class IPEXModelForSeq2SeqLM(IPEXModel, GenerationMixin):
     def _supports_num_logits_to_keep(self) -> bool:
         """
         Will be deprecated after we no longer support transformers < 4.49
-    
+
         Return True if the current model supports the keyword argument `num_logits_to_keep` in forward()
         to save memory. Checking it in this way allows to avoid using a new model attribute.
         """
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -358,8 +358,14 @@ class IPEXModelForCausalLM(IPEXModel, GenerationMixin):
         Return True if the current model supports the keyword argument `logits_to_keep` in forward()
         to save memory. Checking it in this way allows to avoid using a new model attribute.
         """
-        logits_to_keep_name = "logits_to_keep" if is_transformers_version(">", "4.49") else "num_logits_to_keep"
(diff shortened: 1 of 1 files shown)


jiqing-feng:
Hi @IlyasMoutawwakil . The transformers have deprecated the `num_logits_to_keep` [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/phi3/modeling_phi3.py#L860). Please review this PR. Thanks!

optimum/intel/ipex/modeling_base.py
```diff
@@ -353,12 +353,12 @@ def _reorder_cache(self, *args, **kwargs):
...
 
-    def _supports_num_logits_to_keep(self) -> bool:
+    def _supports_logits_to_keep(self) -> bool:
         """
-        Return True if the current model supports the keyword argument `num_logits_to_keep` in forward()
+        Return True if the current model supports the keyword argument `logits_to_keep` in forward()
         to save memory. Checking it in this way allows to avoid using a new model attribute.
         """
-        return "num_logits_to_keep" in set(inspect.signature(self.model.forward).parameters.keys())
+        return "logits_to_keep" in set(inspect.signature(self.model.forward).parameters.keys())
```
echarlaix, line 361: what about something to keep compatibility with transformers version lower than 4.49 
```suggestion
        logits_to_keep_name = "logits_to_keep" if is_transformers_version(">", "4.49") else "num_logits_to_keep"
        return logits_to_keep_name in set(inspect.signature(self.model.forward).parameters.keys())
```

optimum/intel/ipex/modeling_base.py
```diff
@@ -353,12 +353,12 @@ def _reorder_cache(self, *args, **kwargs):
     def prepare_inputs_for_generation(self, *args, **kwargs):
         return self.model.prepare_inputs_for_generation(*args, **kwargs)
 
-    def _supports_num_logits_to_keep(self) -> bool:
+    def _supports_logits_to_keep(self) -> bool:
```
echarlaix, line 356: also we can keep `_supports_num_logits_to_keep` until versions lower than 4.49 are deprecated (just calling `_supports_logits_to_keep` until then + add a comment to remove once only > v4.49)

echarlaix (approved):

optimum/intel/ipex/modeling_base.py
jiqing-feng: transformers 4.49 has already deprecated `_supports_num_logits_to_keep`, the only way is keep both...

PR #1196, opened 2025-03-12
upgrade transformers to 4.49 for patching models

jiqing-feng:
Hi @echarlaix @IlyasMoutawwakil . I updated the transformers version to 4.49.0 in ipex backend. Please review it, thanks!

### DIFF
--- a/optimum/exporters/ipex/cache_utils.py
+++ b/optimum/exporters/ipex/cache_utils.py
@@ -1,6 +1,7 @@
 import os
 from typing import List, Optional, Tuple
 
+import intel_extension_for_pytorch as ipex
 import torch
 from intel_extension_for_pytorch.llm.modules import PagedAttention
 from transformers import Cache, PretrainedConfig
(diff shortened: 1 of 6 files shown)


jiqing-feng:
Hi @echarlaix @IlyasMoutawwakil . Please review this PR, thanks!

sywangyi (approved):

yao-matrix (approved):

optimum/exporters/ipex/cache_utils.py
```diff
@@ -38,13 +39,14 @@ def __init__(
...
         max_cache_len: int,
-        device,
+        device=None,
         dtype=None,
-        layer_device_map=None,
         **kwargs,
     ) -> None:
         super().__init__()
         self.max_batch_size = max_batch_size
+        default_device = torch.device("xpu") if ipex._C._has_xpu() else torch.device("cpu")
```
IlyasMoutawwakil, line 48: this seems more like a global variable, also if I understand correctly, if xpu device is available and ipex has xpu, it's okay, but what if xpu device is available and current ipex doesn't have xpu support, maybe warning the user and then proceding with cpu device makes sense.

setup.py
```diff
@@ -67,7 +67,7 @@
     "nncf": ["nncf>=2.14.0"],
     "openvino": ["nncf>=2.14.0", "openvino>=2024.5.0", "openvino-tokenizers>=2024.5.0"],
     "neural-compressor": ["neural-compressor[pt]>3.0", "accelerate", "transformers<4.46"],
-    "ipex": ["intel-extension-for-pytorch>=2.4", "transformers>4.46,<4.48", "accelerate"],
+    "ipex": ["intel-extension-for-pytorch>=2.4", "transformers>4.48,<4.50", "accelerate"],
```
IlyasMoutawwakil, line 70: we are only testing ipex 2.6.0, shouldn't we test the minimum version as well ? if it still works ofc, otherwise we should update the minimum version.

IlyasMoutawwakil (approved):
overall LGTM, just a couple nits.

jiqing-feng:
Hi @IlyasMoutawwakil . I have fixed your comments!

IlyasMoutawwakil (approved):

PR #1199, opened 2025-03-14
fix generation for statically reshaped diffusion pipeline

eaidova:
# What does this PR do?

in case if pipeline was reshaped to generate images of given size and this size is different from user requested or used in pipeline by default generation raised shape mismatch error

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -962,7 +962,7 @@ class OVDiffusionPipeline(OVBaseModel, DiffusionPipeline):
         components = {k: v for k, v in components.items() if v is not None}
         return components
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self, *args, height=None, width=None, **kwargs):
         # we do this to keep numpy random states support for now
         # TODO: deprecate and add warnings when a random state is passed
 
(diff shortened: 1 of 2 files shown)


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -889,9 +889,7 @@ def reshape(
             )
 
         if self.text_encoder_3 is not None:
-            self.text_encoder_3.model = self._reshape_text_encoder(
-                self.text_encoder_3.model, batch_size, getattr(self.tokenizer_3, "model_max_length", -1)
-            )
+            self.text_encoder_3.model = self._reshape_text_encoder(self.text_encoder_3.model, batch_size, -1)
```
IlyasMoutawwakil, line 892: why not `getattr(self.tokenizer_3, "model_max_length", -1)`

eaidova: in our test model model_max_length is 256, while default argument in pipeline_call__ is 512

tests/openvino/test_diffusion.py
```diff
@@ -438,6 +438,24 @@ def test_load_custom_weight_variant(self):
...
+        height, width, batch_size = 64, 64, 1
+        inputs = self.generate_inputs(height=height, width=width, batch_size=batch_size)
+        inputs["output_type"] = "pil"
+        image = pipeline(**inputs).images[0]
+        self.assertTupleEqual(image.size, (32, 64))
+        # generation without height / width provided
+        inputs.pop("height")
+        inputs.pop("width")
+        image = pipeline(**inputs).images[0]
+        self.assertTupleEqual(image.size, (32, 64))
```
IlyasMoutawwakil, line 457: please add an assertion for warning/not warning when shapes are mismatched/matched.

AlexKoff88 (approved):

tests/openvino/test_diffusion.py
IlyasMoutawwakil: Here are some examples from unittest docs https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns
I believe you should use https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs if the warning is done by the logger. This can be used to make sure the warning is there when we want it to be there and avoid false positives.

IlyasMoutawwakil (approved):

PR #1200, opened 2025-03-17
Patch

jiqing-feng:
Hi @echarlaix @IlyasMoutawwakil  @sywangyi .

This is our 1st step to enable patched model + torch.compile. We plan to disable all ipex fuse linear patch because torch.compile also has fuse opimization. We cannot defaultly use torch.compile for patched models because there are still incompatible issues for flash attention.

We plan to enable the patched model + torch.compile in ipex 2.7 or 2.8 

This PR works for non-generation models like bert as patched bert model has no flash attention which will block the torch.compile .


optimum/intel/ipex/modeling_base.py
```diff
@@ -231,24 +233,27 @@ def to(self, device: Union[torch.device, str]):
     def can_generate(self):
         return isinstance(self, GenerationMixin)
 
-    def maybe_apply_torch_compile(self):
+    def can_compile(self):
```
IlyasMoutawwakil, line 236: Please make this method private, let's not add public interface methods that are removed in a couple PRs.

optimum/intel/ipex/modeling_base.py
```diff
@@ -146,7 +146,8 @@ def __init__(
         self.use_cache = kwargs.get("use_cache", False)
         self.model_save_dir = model_save_dir
         self._add_patch = _is_patched_with_ipex(model, self.export_feature, self.use_cache)
-        self.compiled = False
+        self.compile = self.can_compile()
+        model.config.compile = self.compile
```
IlyasMoutawwakil, line 150: same thing, "self.compile" don't need to be public.

IlyasMoutawwakil (approved):

optimum/intel/ipex/modeling_base.py

optimum/intel/ipex/modeling_base.py

jiqing-feng:
Hi @IlyasMoutawwakil . I have fixed your comments, please check the new changes. Thanks!

PR #1202, opened 2025-03-17
add support got-ocr2

eaidova:
# What does this PR do?

```
from transformers import AutoProcessor, TextStreamer

model_id = "stepfun-ai/GOT-OCR-2.0-hf"

model = OVModelForVisualCausalLM.from_pretrained(model_id)

image = "https://huggingface.co/datasets/hf-internal-testing/fixtures_got_ocr/resolve/main/image_ocr.jpg"

inputs = processor(image, return_tensors="pt")

generate_ids = model.generate(
    **inputs,
    do_sample=False,
    tokenizer=processor.tokenizer,
    stop_strings="<|im_end|>",
    max_new_tokens=4096,
    streamer=TextStreamer(processor.tokenizer, skip_prompt=True, skip_special_tokens=True),
)
```

Fixes # https://github.com/huggingface/optimum-intel/issues/918

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -79,6 +79,7 @@ from .model_patcher import (
     FalconModelPatcher,
     FluxTransfromerModelPatcher,
     Gemma2ModelPatcher,
+    GotOCR2ImageEmbeddingsModelPatcher,
     GptBigCodeModelPatcher,
     GptJModelPatcher,
     GptNeoModelPatcher,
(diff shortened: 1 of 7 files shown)


optimum/exporters/openvino/model_configs.py
```diff
@@ -3001,3 +3002,16 @@ def patch_model_for_export(
...
+
+@register_in_tasks_manager("got-ocr2", *["image-to-text", "image-text-to-text"], library_name="transformers")
+class GotOCR2OpenVINOConfig(LlavaOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.49.0"
+
+    def patch_model_for_export(
+        self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ):
+        model_kwargs = model_kwargs or {}
+        if self._behavior != LlavaConfigBehavior.VISION_EMBEDDINGS:
```
IlyasMoutawwakil, line 3015: why is `VISION_EMBEDDINGS` not added directly to `ConfigBehavior` ? using `LlavaConfigBehavior` looks wrong here

IlyasMoutawwakil (approved):
LGTM, just one question/suggestion to address.

optimum/exporters/openvino/model_configs.py
eaidova: not sure that I understand your comment, there is no separated config behaviour for this model class. it inherits it from parent class - LllavaOpenVINOConfig. vision embeddings is part of it and its LlavaConfigBehavior. This overriding required to use got-ocr specific patching instead of llava (because they are different)

eaidova: every multimodal model may have own partition. it is difficult to maintain everything using only one enum class and will required additional logic for maintenance and validation of model specific structure. 

it maybe refactored later if you have better solution

PR #1205, opened 2025-03-18
fix tests running with nightly

eaidova:
# What does this PR do?

- for tokenizers failed bugs, it is expected and it failed due to new feature support, updated condition for expected number of toeknizers
- for loading sd model from hub, some additional investigation from openvino side needed (I crated internal issue). Seems that model is too old and has some incompatible with new ov expected by cpu plugin values (I was not able to find info how it was created unfortunately, but it was uploaded more that 2 years ago...), as workaround suggest to update model

## Before submitting

### DIFF
--- a/tests/openvino/test_modeling.py
+++ b/tests/openvino/test_modeling.py
@@ -149,7 +149,7 @@ class OVModelIntegrationTest(unittest.TestCase):
         self.OV_MODEL_ID = "echarlaix/distilbert-base-uncased-finetuned-sst-2-english-openvino"
         self.OV_DECODER_MODEL_ID = "helenai/gpt2-ov"
         self.OV_SEQ2SEQ_MODEL_ID = "echarlaix/t5-small-openvino"
-        self.OV_SD_DIFFUSION_MODEL_ID = "hf-internal-testing/tiny-stable-diffusion-openvino"
+        self.OV_SD_DIFFUSION_MODEL_ID = "katuni4ka/tiny-stable-diffusion-openvino"
         self.OV_FLUX_DIFFUSION_MODEL_ID = "katuni4ka/tiny-random-flux-ov"
         self.OV_VLM_MODEL_ID = "katuni4ka/tiny-random-llava-ov"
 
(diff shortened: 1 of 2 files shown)


nikita-savelyevv (commented):
> Seems that model is too old and has some incompatible with new ov expected by cpu plugin values.

Is it possible that some old but still popular models are affected too?

eaidova:
> > Seems that model is too old and has some incompatible with new ov expected by cpu plugin values.
> 
> Is it possible that some old but still popular models are affected too?

it affect only very old IRs stored on hub, original models are not affected.
I do not think that there are some popular among them

nikita-savelyevv (approved):

IlyasMoutawwakil (approved):
LGTM, okay with updating the tiny diffusion model.

tests/openvino/test_modeling.py
```diff
@@ -149,7 +149,7 @@ def __init__(self, *args, **kwargs):
         self.OV_MODEL_ID = "echarlaix/distilbert-base-uncased-finetuned-sst-2-english-openvino"
         self.OV_DECODER_MODEL_ID = "helenai/gpt2-ov"
         self.OV_SEQ2SEQ_MODEL_ID = "echarlaix/t5-small-openvino"
-        self.OV_SD_DIFFUSION_MODEL_ID = "hf-internal-testing/tiny-stable-diffusion-openvino"
+        self.OV_SD_DIFFUSION_MODEL_ID = "katuni4ka/tiny-stable-diffusion-openvino"
```
IlyasMoutawwakil, line 152: I would however suggest we add a warning that notifies the user if a very old IR is being used and prompt them to re export for optimal perf.

eaidova: it is good idea if there is robust way to estimate what is old IR and what is not, but it is not so now... I think I need to think dipper about that and implement separately

IlyasMoutawwakil: please open an issue to track this item, I believe an estimator is the one you just used (2022/2023 -> old).

PR #1208, opened 2025-03-19
fix checking available files if from_onnx=True

eaidova:
# What does this PR do?
 model classes _from_pretrained has from_onnx flag for opportunity to load onnx model instead of openvino ir. This functionality becomes broken after enabling auto detection  for export, because checking available files does not take into account from_onnx flag.

This PR restores opportunity to load onnx model files (in some cases, it still useful)

Also fixed logic for explicit export=False flag loading if model is not found (for me, if there is no model and user specified to not use export, error message that openvino model is not found looks more clear than when it is failed that can not find original model files during attempt of export)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -445,7 +445,7 @@ class OVBaseModel(OptimizedModel):
             logger.info("Offline mode: forcing local_files_only=True")
             local_files_only = True
 
-        _export = export or False
+        _export = export
         try:
             if local_files_only:
                 object_id = model_id.replace("/", "--")


nikita-savelyevv (approved):

optimum/intel/openvino/modeling_base.py
```diff
@@ -459,7 +459,7 @@ def from_pretrained(
 
             ov_files = _find_files_matching_pattern(
                 model_dir,
-                pattern=r"(.*)?openvino(.*)?\_model(.*)?.xml$",
+                pattern=r"(.*)?openvino(.*)?\_model(.*)?.xml$" if kwargs.get("from_onnx", False) else "*.onnx",
```
eaidova, line 462:
```suggestion
                pattern=r"(.*)?openvino(.*)?\_model(.*)?.xml$" if not kwargs.get("from_onnx", False) else "*.onnx",
```

AlexKoff88 (approved):

optimum/intel/openvino/modeling_base.py
```diff
@@ -473,10 +473,13 @@ def from_pretrained(
...
                 else:
-                    logger.warning(
-                        f"No OpenVINO files were found for {model_id}, setting `export=True` to convert the model to the OpenVINO IR. "
-                        "Don't forget to save the resulting model with `.save_pretrained()`"
-                    )
+                    if export is None:
+                        logger.warning(
+                            f"No OpenVINO files were found for {model_id}, setting `export=True` to convert the model to the OpenVINO IR. "
+                            "Don't forget to save the resulting model with `.save_pretrained()`"
+                        )
```
IlyasMoutawwakil, line 480: it says "setting `export=True` to convert the model to the OpenVINO IR" but nothing is set ?

IlyasMoutawwakil (changes requested):
I'm not sure I understand what's the goal here. fixing the onnx model loading looks good, but the export behavior changes are unreadable, I would suggest either clearly explaining the flow you're targetting or doing it in another PR to not block this one.

optimum/intel/openvino/modeling_base.py
eaidova: at this moment _export variable already equal to True in code above
https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_base.py#L467

we have 3 scenarios when _export  ^ export
1. export = True _export = False (user explicitly specifies to convert model, but model already exists, force reexport
2. export = None, _export = True (auto detection of is there is need export or not) say that there is no ov model and model will be converted
3. export = False _export = True (user explicitly force to not use export, but something went wrong and there is no openvino model) - stay forced export False and let model failed on reading stage - that is what changed comparing with current behavior (now it will try to export model even if user provided export=False and that is what I want to change)

eaidova: ok, lets rework it separately

optimum/intel/openvino/modeling_base.py
```diff
@@ -445,7 +445,7 @@ def from_pretrained(
             logger.info("Offline mode: forcing local_files_only=True")
             local_files_only = True
 
-        _export = export
+        _export = export or False
```
eaidova, line 448:
```suggestion
        _export = export
```

PR #1210, opened 2025-03-20
add support SmolVLM and Idefics3 models

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -88,6 +88,7 @@ from .model_patcher import (
     GptNeoxModelPatcher,
     GraniteMoEModelPatcher,
     IBertModelPatcher,
+    Idefics3ImageEmbeddingsModelPatcher,
     InputEmbeddingPatcher,
     InternLM2Patcher,
     InternLMModelPatcher,
@@ -147,6 +148,14 @@ def init_model_configs():
     TasksManager._CUSTOM_CLASSES[("pt", "gemma3", "image-text-to-text")] = (
         "transformers",
         "Gemma3ForConditionalGeneration",
+    )
+    TasksManager._CUSTOM_CLASSES[("pt", "idefics3", "image-text-to-text")] = (
+        "transformers",
+        "AutoModelForImageTextToText",
+    )
+    TasksManager._CUSTOM_CLASSES[("pt", "smolvlm", "image-text-to-text")] = (
+        "transformers",
+        "AutoModelForImageTextToText",
     )
 
     TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS[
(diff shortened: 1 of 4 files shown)


AlexKoff88 (approved):

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3241,6 +3244,144 @@ def preprocess_inputs(
...
+        processor: Optional[AutoImageProcessor] = None,
+        tokenizer: Optional[PreTrainedTokenizer] = None,
+        config: Optional[PretrainedConfig] = None,
+        video: Optional["VideoInput"] = None,
+    ):
+        if processor is None:
+            raise ValueError("Processor is required.")
+
+        if video is not None:
+            raise ValueError("video input is not supported")
```
nikita-savelyevv, line 3336: In the PR description I see that you provide a video input. However, here it says that video input is not supported. Why is that?

eaidova: it supports only loading video as part of chat template (or at least providing metadata) and does not work if video loaded outside of it unfortunately.

It requires slightly redesign API for input preprocessing in our case, that is why I would like to postpone this feature

nikita-savelyevv: Have you considered doing it like this?
```python
    def preprocess_inputs(
        text: str,
        image: Optional["Image"] = None,
        processor: Optional[AutoImageProcessor] = None,
        tokenizer: Optional[PreTrainedTokenizer] = None,
        config: Optional[PretrainedConfig] = None,
        video: Optional[str] = None,
    ):
        if processor is None:
            raise ValueError("Processor is required.")

        conversation = [
            {
                "role": "user",
                "content": [
                    {"type": "text", "text": text},
                ],
            }
        ]
        if video is not None:
            conversation[0]["content"].append({"type": "video", "url": video})
        if image is not None:
            conversation[0]["content"].insert(0, {"type": "image"})

        text_prompt = processor.apply_chat_template(conversation, add_generation_prompt=True)

        inputs = processor(images=image, text=text_prompt, return_tensors="pt")
        return inputs
```

This will be inconsistent with other models though..

eaidova: I want to extend video parameter for 2 cases in all models that support it:
1. str/Path to process it as part of chat template (probably for old versions, it will also require to provide reading capabilities)
2. accepts tuple  video data + video metadata like load_video returns.

As it is quit huge reworking that affects other models, I would like to do that in separated PR

eaidova:
@IlyasMoutawwakil  could you please take a look?

optimum/exporters/openvino/model_configs.py
```diff
@@ -1507,26 +1516,24 @@ def get_vlm_text_generation_config(
     return export_config
 
 
-class LlavaConfigBehavior(str, enum.Enum):
+class VLMConfigBehavior(str, enum.Enum):
```
IlyasMoutawwakil, line 1519: AWSOME !

optimum/exporters/openvino/model_configs.py
```diff
@@ -1610,12 +1613,49 @@ def patch_model_for_export(
         self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
     ):
         model_kwargs = model_kwargs or {}
-        if self._behavior != LlavaConfigBehavior.VISION_EMBEDDINGS:
+        if self._behavior != VLMConfigBehavior.VISION_EMBEDDINGS:
+            return super().patch_model_for_export(model, model_kwargs)
+        return CommonImageEmbeddingsModelPatcher(self, model, model_kwargs)
+
+
+class VLMOpenVINOConfig(BaseVLMOpenVINOConfig):
```
IlyasMoutawwakil, line 1621: having both `VLMOpenVINOConfig` and `BaseVLMOpenVINOConfig` is confusing, what's the difference ?

eaidova: remains only one

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4519,3 +4519,200 @@ def forward(self, attention_mask, position_ids, past_key_values, token_type_ids,
...
+                is_causal=is_causal,
+                scale=self.scale,
+            )
+
+            attn_output = attn_output.transpose(1, 2).contiguous()
+            attn_output = attn_output.view(batch_size, q_len, self.embed_dim)
+
+            attn_output = self.out_proj(attn_output)
+
+            return attn_output, None
```
IlyasMoutawwakil, line 4699: probably not in this PR but we definitely need to move these methods to separate files per model_type.

IlyasMoutawwakil (approved):
LGTM thansk @eaidova  !
I will do another review in the afternoon and merge if everything goes right.

PR #1216, opened 2025-03-25
Update default int4 config for llama-2-7b-chat-hf

nikita-savelyevv:
# What does this PR do?

WWB similarity  before: 78.10%, after: 92.65%.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -71,7 +71,7 @@ _DEFAULT_4BIT_CONFIGS = {
         "bits": 4,
         "sym": True,
         "group_size": 128,
-        "ratio": 0.8,
+        "ratio": 1.0,
         "dataset": "wikitext2",
         "quant_method": OVQuantizationMethod.AWQ,
         "scale_estimation": True,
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/configuration.py
```diff
@@ -67,7 +67,15 @@ class OVQuantizationMethod(str, Enum):
         "quant_method": OVQuantizationMethod.AWQ,
     },
     "meta-llama/Llama-2-7b-hf": {"bits": 4, "sym": True, "group_size": 128, "ratio": 0.6},
-    "meta-llama/Llama-2-7b-chat-hf": {"bits": 4, "sym": True, "group_size": 128, "ratio": 0.8},
+    "meta-llama/Llama-2-7b-chat-hf": {
+        "bits": 4,
+        "sym": True,
+        "group_size": 128,
+        "ratio": 0.8,
```
MaximProshin, line 74: Did you check the Similarity with ratio=1 with all other remaining parameters? If it's in our threshold, I would use the default ratio, i.e. 1

nikita-savelyevv: Updated to 1.0. Surprisingly, similarity increased by another 2%.

MaximProshin (approved):

AlexKoff88 (approved):

PR #1224, opened 2025-04-04
Add HF_TOKEN environment variable to other OV tests

nikita-savelyevv:
# What does this PR do?

Similar to #1222, add `HF_TOKEN` environment variable to the rest of openvino tests.

## Before submitting


eaidova (approved):

nikita-savelyevv:
@IlyasMoutawwakil Also, seems like some ov tests for which HF_TOKEN environment variable was added in #1222 started to fail again :thinking:

PR #1177, opened 2025-02-26
Add `ov_submodels` property to `OVBaseModel`

nikita-savelyevv:
# What does this PR do?

Add `ov_submodels` property to `OVBaseModel` containing all instances of `openvino.runtime.Model` that the model pipeline consists of. This simplifies the logic of handling such models.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1229, opened 2025-04-08
add OVSamModel for inference

eaidova:
# What does this PR do?

added inference class for SAM model (export part reused from optimum)

Example of usage
```
import requests
from PIL import Image
from transformers import SamProcessor

from optimum.intel.openvino import OVSamModel


model = OVSamModel.from_pretrained("facebook/sam-vit-base")
processor = SamProcessor.from_pretrained("facebook/sam-vit-base")

img_url = "https://huggingface.co/ybelkada/segment-anything/resolve/main/assets/car.png"
raw_image = Image.open(requests.get(img_url, stream=True).raw).convert("RGB")
input_points = [[[450, 600]]]  # 2D localization of a window

inputs = processor(raw_image, input_points=input_points, return_tensors="pt")
outputs = model(**inputs)
masks = processor.image_processor.post_process_masks(
    outputs.pred_masks.cpu(), inputs["original_sizes"].cpu(), inputs["reshaped_input_sizes"].cpu()
)
scores = outputs.iou_scores
print(scores)

import matplotlib.pyplot as plt
import numpy as np
from PIL import Image


def show_mask(mask, ax, random_color=False):
    if random_color:
        color = np.concatenate([np.random.random(3), np.array([0.6])], axis=0)
    else:
        color = np.array([30 / 255, 144 / 255, 255 / 255, 0.6])
    h, w = mask.shape[-2:]
    print(mask.shape)
    mask_image = mask[0][0].reshape(h, w, 1) * color.reshape(1, 1, -1)
    ax.imshow(mask_image)


plt.imshow(np.array(raw_image))
ax = plt.gca()
for mask in masks:
    show_mask(mask, ax=ax, random_color=True)
plt.axis("off")
plt.show()
````
sam

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_sam.py
+++ b/optimum/intel/openvino/modeling_sam.py
@@ -21,6 +21,7 @@ from .utils import (
     OV_PROMPT_ENCODER_MASK_DECODER_MODEL_NAME,
     OV_VISION_ENCODER_MODEL_NAME,
     TemporaryDirectory,
+    model_has_dynamic_inputs,
 )
 
 
@@ -88,7 +89,6 @@ class OVSamModel(OVBaseModel):
         self.config = config
         self._model_save_dir = model_save_dir
         self._device = device.upper()
-        self.is_dynamic = dynamic_shapes
         self.ov_config = {} if ov_config is None else {**ov_config}
         self.preprocessors = kwargs.get("preprocessors", [])
         self.vision_encoder_model = vision_encoder_model
@@ -97,6 +97,9 @@ class OVSamModel(OVBaseModel):
         enable_compilation = kwargs.get("compile", True)
         self.vision_encoder = OVSamVisionEncoder(self.vision_encoder_model, self)
         self.prompt_encoder_mask_decoder = OVSamPromptEncoder(self.prompt_encoder_mask_decoder_model, self)
+
+        if dynamic_shapes and not self.is_dynamic and not self._compile_only:
+            self.reshape()
 
         if enable_compilation and not self._compile_only:
             self.compile()
@@ -243,8 +246,8 @@ class OVSamModel(OVBaseModel):
             file_names = {}
             for name, file_name in model_file_names.items():
                 model_cache_path = cls._cached_file(
-                    repo_id=model_id,
-                    filename=file_name,
+                    model_path=model_id,
+                    file_name=file_name,
                     token=token,
                     revision=revision,
                     cache_dir=cache_dir,
@@ -285,7 +288,37 @@ class OVSamModel(OVBaseModel):
         model_names = ["vision_encoder_model", "prompt_encoder_mask_decoder_model"]
         return model_names
 
-    def reshape(self, batch_size: int, sequence_length: int):
+    def reshape(self, batch_size: int = -1, point_batch_size: int = -1, num_points_per_image: int = -1):
+        """
+        Propagates the given input shapes on the model's layers, fixing the inputs shapes of the model.
+
+        Arguments:
+            batch_size (`int`):
+                The batch size.
+            point_batch_size (`int`):
+                The point batch size.
+            num_points_per_image (`int`, *optional*):
+                The number of points per image
+        """
+        if self._compile_only:
+            raise ValueError(
+                "`reshape()` is not supported with `compile_only` mode, please intialize model without this option"
+            )
+        vision_encoder_shapes = {}
+        for inputs in self.vision_encoder_model.inputs:
+            vision_encoder_shapes[inputs] = inputs.get_partial_shape()
+            vision_encoder_shapes[inputs][0] = batch_size
+        self.vision_encoder_model.reshape(vision_encoder_shapes)
+        self.vision_encoder.request = None
+        mask_decoder_shapes = {}
+        for inputs in self.prompt_encoder_mask_decoder_model.inputs:
+            mask_decoder_shapes[inputs] = inputs.get_partial_shape()
+            mask_decoder_shapes[inputs][0] = batch_size
+            if inputs.get_any_name() in ["input_points", "input_labels"]:
+                mask_decoder_shapes[inputs][1] = point_batch_size
+                mask_decoder_shapes[inputs][2] = num_points_per_image
+        self.prompt_encoder_mask_decoder_model.reshape(mask_decoder_shapes)
+        self.prompt_encoder_mask_decoder.request = None
         return self
 
     def half(self):
@@ -363,3 +396,9 @@ class OVSamModel(OVBaseModel):
 
     def get_image_features(self, pixel_values, *args, **kwargs):
         return torch.from_numpy(self.vision_encoder(pixel_values).image_embeddings)
+
+    @property
+    def is_dynamic(self):
+        return model_has_dynamic_inputs(self.vision_encoder_model) or model_has_dynamic_inputs(
+            self.prompt_encoder_mask_decoder_model
+        )
--- a/optimum/intel/openvino/modeling_sam.py
+++ b/optimum/intel/openvino/modeling_sam.py
@@ -110,7 +110,7 @@ class OVSamModel(OVBaseModel):
     def clear_requests(self):
         if self._compile_only:
             raise ValueError(
-                "`clear_requests()` is not supported with `compile_only` mode, please intialize model without this option"
+                "`clear_requests()` is not supported with `compile_only` mode, please initialize model without this option"
             )
 
         for _, component in self.components.items():
--- a/optimum/intel/openvino/modeling_sam.py
+++ b/optimum/intel/openvino/modeling_sam.py
@@ -1,0 +1,409 @@
+import logging
+import os
+import warnings
+from pathlib import Path
+from typing import Dict, Optional, Union
+
+import numpy as np
+import openvino as ov
+import torch
+from huggingface_hub import hf_hub_download
+from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
+from openvino._offline_transformations import apply_moc_transformations, compress_model_transformation
+from transformers import AutoConfig, PretrainedConfig, SamModel
+from transformers.modeling_outputs import ModelOutput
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: tests/openvino/test_modeling.py [hunk 1]
```diff
@@ -2070,12 +2139,12 @@ class OVModelForPix2StructIntegrationTest(unittest.TestCase):
...
-            "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/transformers/tasks/ai2d-demo.jpg",
-            stream=True,
-        ).raw
-    )
+    # IMAGE = Image.open(
+    #     requests.get(
+    #         "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/transformers/tasks/ai2d-demo.jpg",
+    #         stream=True,
+    #     ).raw
+    # )
```
eaidova, line 2147:
```suggestion
    
      IMAGE = Image.open(
       requests.get(
             "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/transformers/tasks/ai2d-demo.jpg",
             stream=True,
         ).raw
    )
```

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,409 @@
...
+            self.auto_model_class.register(AutoConfig, self.__class__)
+        except AttributeError:
+            pass
+        self.shared_image_embedding = SamPositionalEmbedding(config.vision_config)
+        self.positional_embeddings = self.get_image_wide_positional_embeddings()
+
+    def clear_requests(self):
+        if self._compile_only:
+            raise ValueError(
+                "`clear_requests()` is not supported with `compile_only` mode, please intialize model without this option"
```
nikita-savelyevv, line 113:
```suggestion
                "`clear_requests()` is not supported with `compile_only` mode, please initialize model without this option"
```

nikita-savelyevv (approved):

nikita-savelyevv:
Also, the list of supported models in the documentation should be updated.

eaidova:
> Also, the list of supported models in the documentation should be updated.

thanks, added

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,409 @@
...
+                "`compile_only` mode will be disabled because it does not support model export."
+                "Please provide openvino model obtained using optimum-cli or saved on disk using `save_pretrained`"
+            )
+            compile_only = False
+        save_dir = TemporaryDirectory()
+        save_dir_path = Path(save_dir.name)
+
+        # This attribute is needed to keep one reference on the temporary directory, since garbage collecting
+        # would end-up removing the directory containing the underlying OpenVINO model
+        cls._model_save_dir_tempdirectory_instance = save_dir
```
IlyasMoutawwakil, line 289: the tmpdir should be affected to the instance not the class, if I create two OVSamModels, the first will be lost and garbage collected

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,409 @@
...
+                standard cache should not be used.
+            encoder_file_name(`str`, *optional*):
+                The encoder model file name. Overwrites the default file name openvino_encoder_model.xml and allows one to
+                load the encoder model with a different name.
+            decoder_file_name(`str`, *optional*):
+                The decoder model file name. Overwrites the default file name openvino_decoder_model.xml and allows one to
+                load the decoder model with a different name.
+            decoder_with_past_file_name(`str`, *optional*):
+                The decoder with past key values model file name overwriting the default file name
+                openvino_decoder_with_past_model.xml, allowing to load the decoder model with a different name.
```
IlyasMoutawwakil, line 194: ?

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,409 @@
...
+        )
+
+    @property
+    def _component_names(self):
+        base_components = ["vision_encoder", "prompt_encoder_mask_decoder"]
+        return base_components
+
+    @property
+    def components(self):
+        return {component_name: getattr(self, component_name) for component_name in self._component_names}
```
IlyasMoutawwakil, line 325: sam is a transformers model, we only added the components property is for diffusion pipelines no ?

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,409 @@
...
+
+        if enable_compilation and not self._compile_only:
+            self.compile()
+
+        # Avoid warnings when creating a transformers pipeline
+        AutoConfig.register(self.base_model_prefix, AutoConfig)
+        try:
+            self.auto_model_class.register(AutoConfig, self.__class__)
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
eaidova: I think it is better to ask @echarlaix because I added that for alignment with other models that using the same logic for registering for pipleine

optimum/intel/openvino/modeling_sam.py
eaidova: how is that connected? no, components property used for correct getting models in multi-model pipelines in weights compression case, it does not depends from which library used for export, it is needed to make compression process model class independent

IlyasMoutawwakil: 😅I remember adding it first time in the diffusion refactoring, it seems you added it later to vlms at some point, we can make it a standard in the lib, but for now it's far from it, also it's not used for compression 🙂 https://github.com/search?q=repo%3Ahuggingface%2Foptimum-intel+%22components%22&type=code

eaidova: compression used ov_submodels property starting from this commit https://github.com/huggingface/optimum-intel/pull/1177, I thought that components takes part for its preparation
ok if you insist I can remove components

IlyasMoutawwakil: no worries, I'm not insisting, I think it's neither bad not good, but I don't see the point of adding it as a public property if it doesn't serve an actual purpose.

optimum/intel/openvino/modeling_sam.py
eaidova: moved

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,427 @@
...
+        prompt_encoder_mask_decoder_model: ov.Model,
+        config: PretrainedConfig = None,
+        device: str = "CPU",
+        dynamic_shapes: bool = True,
+        ov_config: Optional[Dict[str, str]] = None,
+        model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
+        **kwargs,
+    ):
+        self.config = config
+        self.use_cache = kwargs.get("use_cache", True)
```
IlyasMoutawwakil, line 91: it doesn't use cache anywhere

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,427 @@
...
+
+        compile_only = kwargs.get("compile_only", False)
+        if os.path.isdir(model_id):
+            # Load model from a local directory
+            model_save_dir = Path(model_id)
+            file_names = {k: os.path.join(model_id, model_file_names[k]) for k in model_file_names}
+        else:
+            file_names = {}
+            for name, file_name in model_file_names.items():
+                model_cache_path = hf_hub_download(
```
IlyasMoutawwakil, line 248: please use the `_cached_file` method whenever possible, it abstracts the from_onnx logic and It is more robust against CI rate limiting

IlyasMoutawwakil: https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_base.py#L522

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,427 @@
...
+            revision=revision,
+            cache_dir=cache_dir,
+            token=token,
+            local_files_only=local_files_only,
+            force_download=force_download,
+            trust_remote_code=trust_remote_code,
+            stateful=False,
+            variant=variant,
+        )
+        config = AutoConfig.from_pretrained(save_dir_path, trust_remote_code=trust_remote_code)
```
IlyasMoutawwakil, line 332: config is literally passed above, why reload another one ?

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,427 @@
...
+        config: PretrainedConfig,
+        use_auth_token: Optional[Union[bool, str]] = None,
+        token: Optional[Union[bool, str]] = None,
+        revision: Optional[str] = None,
+        force_download: bool = False,
+        cache_dir: str = HUGGINGFACE_HUB_CACHE,
+        subfolder: str = "",
+        local_files_only: bool = False,
+        task: Optional[str] = None,
+        use_cache: bool = True,
```
IlyasMoutawwakil, line 299: where does this method differ from the original ?

optimum/intel/openvino/modeling_sam.py
eaidova: reused, PR was opened before your changes

optimum/intel/openvino/modeling_sam.py
eaidova: removed

optimum/intel/openvino/modeling_sam.py
```diff
@@ -0,0 +1,365 @@
...
+        )
+
+        return model
+
+    @property
+    def _ov_submodel_names(self):
+        model_names = ["vision_encoder_model", "prompt_encoder_mask_decoder_model"]
+        return model_names
+
+    def reshape(self, batch_size: int, sequence_length: int):
```
IlyasMoutawwakil, line 288: let's log a warning that it's not yet supported

eaidova: I provided reshape support

IlyasMoutawwakil (approved):
LGTM ! thanks @eaidova 
I suggested one last nit.

PR #1230, opened 2025-04-09
Support SpeechT5 text-to-speech pipeline by OpenVINO

rkazants:
# What does this PR do?

This PR introduces support of SpeechT5 text-to-speech pipeline using OpenVINO. Here is a demo code:

```py
from optimum.intel import OVModelForSpeechSeq2Seq, OVModelForTextToSpeechSeq2Seq
from transformers import SpeechT5Processor, SpeechT5ForTextToSpeech, SpeechT5HifiGan
import torch
import soundfile as sf
from datasets import load_dataset
from optimum.exporters import tasks

model_id = "microsoft/speecht5_tts"
vocoder_id = "microsoft/speecht5_hifigan"

ov_model = OVModelForTextToSpeechSeq2Seq.from_pretrained(model_id, export=True, vocoder=vocoder_id)
ov_model.save_pretrained("speecht5_tts")
ov_model = OVModelForTextToSpeechSeq2Seq.from_pretrained("speecht5_tts")

processor = SpeechT5Processor.from_pretrained(model_id)

inputs = processor(text="Hello, my dog is cute.", return_tensors="pt")

# load xvector containing speaker's voice characteristics from a dataset
embeddings_dataset = load_dataset("Matthijs/cmu-arctic-xvectors", split="validation")
speaker_embeddings = torch.tensor(embeddings_dataset[7306]["xvector"]).unsqueeze(0)

speech = ov_model.generate(input_ids=inputs["input_ids"],
                           speaker_embeddings=speaker_embeddings)

sf.write("speech.wav", speech.numpy()[0], samplerate=16000)
```

## Before submitting


eaidova:
@rkazants could you please provide tests?

optimum/exporters/openvino/model_configs.py
```diff
@@ -3236,3 +3239,168 @@ def get_model_for_behavior(self, model, behavior: Union[str, VLMConfigBehavior])
...
+            if self.variant == "with-past" and self.use_past_in_inputs:
+                self.add_past_key_values(common_inputs, direction="inputs")
+        elif self.is_postnet:
+            common_inputs["raw_spectrogram"] = {
+                0: "n_spectrums",
+                1: "batch_size",
+                2: "reduction_factor",
+                3: "num_mel_bins",
+            }
+        elif self.is_vocoder:
```
eaidova, line 3376: maybe better use config behaviour is better than boolean flags?

optimum/exporters/openvino/model_configs.py
```diff
@@ -3236,3 +3239,168 @@ def get_model_for_behavior(self, model, behavior: Union[str, VLMConfigBehavior])
...
+            common_inputs["spectrogram"] = {0: "batch_size", 1: "n_spectrums", 2: "num_mel_bins"}
+        elif self.is_postnet_and_vocoder:
```
eaidova, line 3378: I do not see this parameter initialization in config, is it expected?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4782,3 +4783,366 @@ def __exit__(self, exc_type, exc_value, traceback):
         self._model.vision_model.embeddings.forward = self._model.vision_model.embeddings._orig_forward
         for layer in self._model.vision_model.encoder.layers:
             layer.self_attn.forward = layer.self_attn._orig_forward
+
+
+def speecht5_decoder_prenet_forward(
+    self,
```
eaidova, line 4789: could you please provide comment about on which code it is based

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -0,0 +1,634 @@
...
+        if "tensor_for_check" in result:
+            tensor_for_check = torch.from_numpy(result["tensor_for_check"])
+
+        # numpy_past_key_values = []
+        # for var_idx in range(6):
+        #    key_tensor_name = "present.{}.decoder.key".format(var_idx)
+        #    value_tensor_name = "present.{}.decoder.value".format(var_idx)
+        #    numpy_past_key_values.append((result[key_tensor_name], result[value_tensor_name]))
+
+        # TODO: (rkazants) recover code below
```
eaidova, line 133: is it still actual?

eaidova:
@rkazants please update also import structure for added model classes:
https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/__init__.py#L163
https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/utils/dummy_openvino_objects.py
https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/__init__.py

optimum/exporters/openvino/model_configs.py

optimum/exporters/openvino/model_configs.py
rkazants: Removed

optimum/intel/openvino/modeling_text2speech.py
rkazants: Removed

optimum/exporters/openvino/model_patcher.py
rkazants: Done.

rkazants:
> @rkazants could you please provide tests?

Done

rkazants:
@eaidova, @IlyasMoutawwakil, @echarlaix, could you please review PR?

Thanks,
Roman

rkazants:
CI failures does not relate my changes in PR. For example, I see issue with whisper model. Please correct me if I am wrong.

optimum/intel/openvino/modeling_base_seq2seq.py
```diff
@@ -431,6 +431,7 @@ def _from_transformers(
             ov_config=ov_config,
             stateful=stateful,
             variant=variant,
+            model_kwargs=kwargs,
```
rkazants, line 434:
```suggestion
```

eaidova:
@IlyasMoutawwakil  could you please rerun ci, thanks

IlyasMoutawwakil:
@eaidova done, was this issue fixed ?
```
FAILED tests/openvino/test_modeling.py::OVModelForTextToSpeechSeq2SeqIntegrationTest::test_compare_to_transformers_0_speecht5 - RuntimeError: The size of tensor a (512) must match the size of tensor b (32) at non-singleton dimension 1
```

eaidova:
> @eaidova done, was this issue fixed ?
> 
> ```
> FAILED tests/openvino/test_modeling.py::OVModelForTextToSpeechSeq2SeqIntegrationTest::test_compare_to_transformers_0_speecht5 - RuntimeError: The size of tensor a (512) must match the size of tensor b (32) at non-singleton dimension 1
> ```

@IlyasMoutawwakil  thanks, yes @rkazants  is working on the fix

eaidova:
@rkazants could you also show how to convert model using cli. I'm wondering how is it possible to provide vocoder id using optimum-cli command (it may require to update some arguments I think)

optimum/exporters/openvino/model_configs.py
```diff
@@ -3266,3 +3269,211 @@ def get_model_for_behavior(self, model, behavior: Union[str, VLMConfigBehavior])
...
+
+        if direction == "inputs":
+            decoder_sequence_name = "past_decoder_sequence_length"
+            name = "past_key_values"
+        else:
+            decoder_sequence_name = "past_decoder_sequence_length + 1"
+            name = "present"
+
+        for i in range(self._normalized_config.decoder_num_layers):
+            inputs_or_outputs[f"{name}.{i}.decoder.key"] = {0: "total_batch_size", 2: decoder_sequence_name}
```
eaidova, line 3384: what is the difference between total_batch_size used for past_key_values and batch_size used in other inputs?

optimum/exporters/openvino/model_configs.py
```diff
@@ -3266,3 +3269,211 @@ def get_model_for_behavior(self, model, behavior: Union[str, VLMConfigBehavior])
...
+            common_inputs["encoder_hidden_states"] = {0: "batch_size", 1: "encoder_sequence_length"}
+            common_inputs["encoder_attention_mask"] = {0: "batch_size", 1: "encoder_sequence_length"}
+            if self.variant == "with-past" and self.use_past_in_inputs:
+                self.add_past_key_values(common_inputs, direction="inputs")
+        elif self._behavior is SpeechT5ConfigBehavior.POSTNET:
+            common_inputs["raw_spectrogram"] = {
+                0: "n_spectrums",
+                1: "batch_size",
+                2: "reduction_factor",
+                3: "num_mel_bins",
```
eaidova, line 3406: all of these inputs should be dynamic or some of them static?

tests/openvino/test_modeling.py
```diff
@@ -2969,3 +2970,75 @@ def test_huggingface_pipeline_streaming(self, model_arch):
...
+            raise Exception("{} unknown processor for text-to-speech".format(model_arch))
+
+    def _get_model(self, model_id, model_arch):
+        if model_arch == "speecht5":
+            from transformers import SpeechT5ForTextToSpeech
+
+            model = SpeechT5ForTextToSpeech.from_pretrained(model_id)
+            return model
+        else:
+            raise Exception("{} unknown model for text-to-speech".format(model_arch))
```
eaidova, line 3002: not sure if these check for unsupported models are needed as test design pattern uses supported models list for parametrization, and it guarantees that nothing unsupported by tests will be passed

rkazants: Don't think that there is smth bad to have this check. Imagine, someone adds new arch into that list but forgot to update this test to load the model. Then this check will mention it.

optimum/exporters/openvino/model_configs.py
rkazants: What is a concern here, please elaborate what to do?

optimum/exporters/openvino/model_configs.py
rkazants: I think no difference. We can use `batch_size`.

optimum/exporters/openvino/model_configs.py
eaidova: These values in inputs signify that dimensions are dynamic and can be different from run to run. It may reduce area of optimizations on runtime side if we mark dynamic dimensions that never change their values. Also it may lead to user errors if they will attempt to manually fill such inputs with invalid by shape data (but which satisfy dynamic shape signature)

rkazants: Understood, I will clean up this dict. Indeed, `num_mel_bins` and some others are static.

optimum/exporters/openvino/model_configs.py
rkazants: Fixed.

rkazants:
> @rkazants could you also show how to convert model using cli. I'm wondering how is it possible to provide vocoder id using optimum-cli command (it may require to update some arguments I think)

Added model-kwargs option to pass vocoder.

rkazants:
@echarlaix, @IlyasMoutawwakil, please review.

Thanks,
Roman

optimum/exporters/openvino/convert.py
```diff
@@ -1346,3 +1351,55 @@ def _get_encoder_decoder_stateful_models_for_export(
...
+        model.config,
+        int_dtype=int_dtype,
+        float_dtype=float_dtype,
+        preprocessors=preprocessors,
+        legacy=False,
+    )
+
+    export_config.variant = "default"
+    all_variants = "\n".join([f"    - {name}: {description}" for name, description in export_config.VARIANTS.items()])
+    logger.info(f"Using the export variant {export_config.variant}. Available variants are:\n{all_variants}")
```
echarlaix, line 1386: by setting the variant to default we will only have the case where it's set to "with-past" so no other potential variants no ?  
```suggestion
```

optimum/exporters/openvino/convert.py
```diff
@@ -1346,3 +1351,55 @@ def _get_encoder_decoder_stateful_models_for_export(
...
+
+    config = export_config
```
echarlaix, line 1388: why is it needed ?

echarlaix (commented):
Thanks for the addition @rkazants !

optimum/exporters/openvino/convert.py
```diff
@@ -1346,3 +1351,55 @@ def _get_encoder_decoder_stateful_models_for_export(
...
+
+def _get_speecht5_tss_model_for_export(
+    model: Union["PreTrainedModel", "TFPreTrainedModel"],
+    task: str,
+    library_name: str,
+    int_dtype: str,
+    float_dtype: str,
+    preprocessors: Optional[List[Any]] = None,
+    model_kwargs: Optional[Dict] = None,
+    stateful: bool = True,
```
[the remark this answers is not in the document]
rkazants: Removed

optimum/exporters/openvino/convert.py
rkazants: Removed

optimum/exporters/openvino/convert.py
```diff
@@ -1346,3 +1351,55 @@ def _get_encoder_decoder_stateful_models_for_export(
...
+):
+    if model_kwargs is None or "vocoder" not in model_kwargs:
+        # use the default vocoder if it is not specified
+        vocoder_id = "microsoft/speecht5_hifigan"
```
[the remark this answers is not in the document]
rkazants: Made `vocoder` manadatory to pass for exporting SpeechT5 TTS

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -0,0 +1,678 @@
...
+            revision=revision,
+            cache_dir=cache_dir,
+            token=token,
+            local_files_only=local_files_only,
+            force_download=force_download,
+            trust_remote_code=trust_remote_code,
+            ov_config=ov_config,
+            stateful=stateful,
+            variant=variant,
+            model_kwargs=kwargs,
```
echarlaix, line 542: why not add this modification here directly ;
https://github.com/huggingface/optimum-intel/blob/0abe991c2d89ba306cd6078f22f01ebab603cd2a/optimum/intel/openvino/modeling_base_seq2seq.py#L356
(could be useful for more models, if we really want to make sure this is only for text2speech model for now we can still limit it by checking `export_feature`, also avoid code duplication)

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -0,0 +1,678 @@
...
+            model_id=save_dir_path,
+            config=config,
+            use_cache=use_cache,
+            load_in_8bit=load_in_8bit,
+            quantization_config=quantization_config,
+            compile_only=compile_only,
+            **kwargs,
+        )
+
+    def generate(
```
echarlaix, line 555: could you add a link to the original code ? https://github.com/huggingface/transformers/blob/af6d2756d9ba7d83388f617b612009242ca46681/src/transformers/models/speecht5/modeling_speecht5.py#L2427

also isn't it specific to SpeechT5 in which case we might want to move this part from OVModelForTextToSpeechSeq2Seq to something more specific to speecht5

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -0,0 +1,678 @@
...
+
+
+@add_start_docstrings(
+    """
+    text-to-audio models
+    """,
+    INPUTS_DOCSTRING,
+)
+class OVModelForTextToSpeechSeq2Seq(OVModelForSeq2SeqLM):
+    auto_model_class = SpeechT5ForTextToSpeech
```
echarlaix, line 156: this is specific to speecht5 so should be replaced, why not use `AutoModelForTextToSpectrogram` instead ?

echarlaix (approved):

optimum/intel/openvino/modeling_text2speech.py
eaidova: Maybe it will be better make similar like for SpeechToText (where general class serves as factory for creation model specific class for whisper
https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_seq2seq.py#L999

it will simplify enabling new models that required different behavior in future (as I understand, different text2speech may have different subcomponents and generation flow)

optimum/exporters/openvino/model_configs.py
```diff
@@ -3281,3 +3284,209 @@ def get_model_for_behavior(self, model, behavior: Union[str, VLMConfigBehavior])
...
+    @property
+    def outputs(self) -> Dict[str, Dict[int, str]]:
+        common_outputs = {}
+        if self._behavior == SpeechT5ConfigBehavior.ENCODER:
+            common_outputs = {
+                "last_hidden_state": {1: "encoder_sequence_length"},
+                "encoder_attention_mask": {1: "encoder_sequence_length"},
+            }
+        elif self._behavior is SpeechT5ConfigBehavior.DECODER:
+            common_outputs = {}
```
IlyasMoutawwakil, line 3439:
```suggestion
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4874,3 +4875,376 @@ def __exit__(self, exc_type, exc_value, traceback):
         self._model.vision_model.embeddings.forward = self._model.vision_model.embeddings._orig_forward
         for layer in self._model.vision_model.encoder.layers:
             layer.self_attn.forward = layer.self_attn._orig_forward
+
+
+# Adopted from https://github.com/huggingface/transformers/blob/v4.51.3/src/transformers/models/speecht5/modeling_speecht5.py#L698
+# this is a patch to avoid PyTorch FE issue
+# with the same tensor names on input and intermediate tensor for speaker_embeddings
```
IlyasMoutawwakil, line 4882: you mean torch can't export something if it has `x = op(x)` ? 😳

rkazants: This is not Torch issue. This is an issue of OpenVINO functionality that converts Pytorch models to OV format

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -0,0 +1,678 @@
...
+        model_names = ["encoder_model", "decoder_model", "postnet_model", "vocoder_model"]
+        return model_names
+
+    @property
+    def submodels(self):
+        return {submodel_name: getattr(self, submodel_name) for submodel_name in self._submodel_names}
+
+    @property
+    def ov_submodels(self) -> Dict[str, ov.runtime.Model]:
+        return {submodel_name: getattr(self, submodel_name).model for submodel_name in self._submodel_names}
```
IlyasMoutawwakil, line 253: we should only define properties we need, components and submodels are the same thing here, we only defined components for diffusion pipelines because `DiffusionPipeline` class needs that property.

optimum/intel/openvino/modeling_text2speech.py
IlyasMoutawwakil: Also can't we use `SpeechT5ForTextToSpeech.generate` directly, the same way we do for Whisper for example, we don't need to rewrite the whole logic if our submodels are compliant and behave similarly to the transformers submodels. We also don't wanna port every fix or change from transformers to optimum-intel manually.

IlyasMoutawwakil (approved):
LGTM, great addition !
I think there's still some redundancy / room to make the implementation leaner, for example with [Whisper](https://github.com/huggingface/optimum/blob/fa87c66967595b8af4de529500868840a3443611/optimum/onnxruntime/modeling_seq2seq.py#L1277C1-L1300C16) which has a custom generate method, we only make sure our class is compliant with its behavior and use the method directly from transformers.

optimum/intel/openvino/modeling_text2speech.py
rkazants: Done, thanks

optimum/intel/openvino/modeling_text2speech.py
rkazants: Removed unneeded submodule properties and leave only components that are use in `_save_pretrained` method

optimum/intel/openvino/modeling_text2speech.py
rkazants: @echarlaix, added a link to the original `generate` method.
@IlyasMoutawwakil, unfortunately, it is not possible to re-use original `generate` method because we combined several decoder parts (prenet, wrapper_decoder, and feat_out you can see in the original code) used in `generate` method into the single piece decoder. We did this to have 4 IRs (encoder, decoder, postnet and vocoder) instead of 4 + 2 IRs. It will be more efficient in terms of performance, etc. I would rather lean to refactor SpeechT5 code to align.

optimum/intel/openvino/modeling_text2speech.py
rkazants: Agreed with @eaidova. Created common interface for text-to-audio models using OpenVINO.
@echarlaix, changed to `AutoModelForTextToWaveform` becuase spectrogram looks raw (or unprocessed) format of audio

rkazants:
> LGTM, great addition ! I think there's still some redundancy / room to make the implementation leaner, for example with [Whisper](https://github.com/huggingface/optimum/blob/fa87c66967595b8af4de529500868840a3443611/optimum/onnxruntime/modeling_seq2seq.py#L1277C1-L1300C16) which has a custom generate method, we only make sure our class is compliant with its behavior and use the method directly from transformers.

Responded here https://github.com/huggingface/optimum-intel/pull/1230#discussion_r2061242034
Thanks

echarlaix (approved):
left couple of minor comments, good to merge once resolved

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -0,0 +1,535 @@
...
+        enable_compilation = kwargs.get("compile", True)
+        self.generation_config = kwargs.get("generation_config", GenerationConfig.from_model_config(config))
+        self._openvino_config = None
+        if quantization_config:
+            self._openvino_config = OVConfig(quantization_config=quantization_config)
+        self._set_ov_config_parameters()
+        self.encoder_model = OVTextToSpeechEncoder(self.encoder_model, self)
+        self.decoder_model = OVTextToSpeechDecoder(self.decoder_model, self)
+        self.postnet_model = OVTextToSpeechPostNet(self.postnet_model, self)
+        self.vocoder_model = OVTextToSpeechVocoder(self.vocoder_model, self)
```
rkazants, line 228:
```suggestion
        self.encoder_model = OVTextToSpeechEncoder(encoder_model, self)
        self.decoder_model = OVTextToSpeechDecoder(decoder_model, self)
        self.postnet_model = OVTextToSpeechPostNet(postnet_model, self)
        self.vocoder_model = OVTextToSpeechVocoder(vocoder_model, self)
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -5238,3 +5239,377 @@ def __exit__(self, exc_type, exc_value, traceback):
             from transformers.models.marian.modeling_marian import MarianAttention
 
             modulewise_unpatch(self._model, MarianAttention)
+
+
```
rkazants, line 5243:
```suggestion
```

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -0,0 +1,535 @@
...
+    ):
+        if "SpeechT5ForTextToSpeech" in config.architectures:
+            return _OVModelForSpeechT5ForTextToSpeech._from_pretrained(model_id, config, **kwargs)
+        else:
+            raise ValueError(f"{config.architectures} are not supported text-to-audio model using OpenVINO")
+
+            return super()._from_pretrained(model_id, config, **kwargs)
+
+
+class _OVModelForSpeechT5ForTextToSpeech(OVModelForTextToSpeechSeq2Seq):
```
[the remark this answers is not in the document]
rkazants: I am sure that four submodules (encoder, decoder, postnet and vocoder) are only specific to SpeechT5 arch. Other text-to-speech models may not have vocoder and postnet at all. So let us keep it now for `_OVModelForSpeechT5ForTextToSpeech` and we will see what next arch will require. Just moving step-by-step:)

optimum/exporters/openvino/model_patcher.py
```diff
@@ -5238,3 +5239,375 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        if self.real_config._behavior == "encoder":
+            self.patched_forward = patched_encoder_forward
+        elif self.real_config._behavior == "decoder":
+            self.patched_forward = patched_decoder_forward
+        elif self.real_config._behavior == "postnet":
+            self.patched_forward = patched_postnet_forward
+        elif self.real_config._behavior == "vocoder":
+            self.patched_forward = patched_vocoder_forward
+        else:
+            raise ValueError("Unknown ")
```
rkazants, line 5613:
```suggestion
            raise ValueError("Unknown ")
        self.orig_forward = self.patched_forward
```

eaidova (approved):

echarlaix (approved):

PR #1232, opened 2025-04-09
Introduce OVCalibrationDatasetBuilder (part 1/2)

nikita-savelyevv:
# What does this PR do?

This is the first part of the changes introduced in #1191.

This is a preparation PR for the introduction of `OVPipelineQuantizationConfig` which purpose is to contain inside of it multiple quantization configs per pipeline model.

### Changes:
- Introduce a `OVCalibrationDataset` class representing _Dict[str, nncf.Dataset]_. The reasoning behind is that most recent model pipelines contain many ov.Model components and not a single _model.model_ as it was before. To quantize such pipelines in a data-aware fashion we need a dataset per ov submodel. This is what `OVCalibrationDataset` represents.
- Moved calibration dataset collection logic into a newly added `OVCalibrationDatasetBuilder` class. It is not exposed to users and is instantiated only inside _OVQuantizer_. The class supports the following methods:
  - `load_dataset(...)` -- basically does the same as the current _OVQuantizer.get_calibation_dataset()_ method.
  - `build_from_dataset_name(dataset, ...) -> OVCalibrationDataset` -- builds nncf calibration dataset from an instance of _datasets.Dataset_.
  - `build_from_dataset_name(dataset_name, ...) -> OVCalibrationDataset` -- builds calibration dataset by first calling _load_dataset()_ method and then _build_from_dataset()_ method. 
  - `build_from_quantization_config(q_config) -> OVCalibrationDataset` -- builds calibration dataset from just a quantization config, using _q_config.dataset_ field and others. Relies on _build_from_dataset_name()_.

### Deprecations:
All changes are backward compatible. However some deprecation warnings (until v1.25) are added to avoid some rare corner case API scenarios which I believe should be avoided:
- Providing _calibration_dataset_ to `OVQuantizer.quantize()` as a list. Currently this is only supported for a list of prompts for hybrid quantization of diffusion models. The list is better to be provided via `quantization_config.dataset` field as it is done for causal models quantization.
- Deprecate `remove_unused_columns` argument for `OVQuantizer.quantize()` because it relies on the fact that `model.forward()` is used for inference. However we can't always guarantee that. For example there is `model.generate()` method. In general I believe the user should remove the unnecessary columns themselves if they provide a `datasets.Dataset` instance which is already a quite advanced usage scenario.
- Currently all columns except `caption` column are removed from a `datasets.Dataset` instance in case of a hybrid quantization. I haven't found usages like this, I believe it is not currently used, but added a deprecation warning anyway.

### Future plans:
In one of the next PRs an `OVPipelineQuantizationConfig` will be introduced. The changes implemented here will be helpful when implementing support for this type of config.

For the further future (past v1.25) I would consider to change `OVQuantizer.get_calibration_dataset` signature. I believe it would be better for it to return _OVCalibrationDataset_ instance from either _datasets.Dataset_ instance or a dataset name as input. This way _OVQuantizer.quantize()_ will accept fully ready _OVCalibrationDataset_ instance and _batch_size_ and _data_collator_ arguments can be removed. For example in the future it could look like this:
```
calibration_dataset: OVCalibrationDataset = ov_quantizer.get_calibration_dataset(
  ov_config,
  dataset_name,
  ...,
  batch_size,
  data_collator,
)
ov_quantizer.quantize(calibration_dataset, ov_config)
```

## Before submitting

### DIFF
--- a/notebooks/openvino/stable_diffusion_hybrid_quantization.ipynb
+++ b/notebooks/openvino/stable_diffusion_hybrid_quantization.ipynb
@@ -168,8 +168,7 @@
     "    return {\"prompt\": example[\"caption\"]}\n",
     "\n",
     "NUM_SAMPLES = 200\n",
-    "dataset = dataset.take(NUM_SAMPLES)\n",
-    "calibration_dataset = dataset.map(lambda x: preprocess_fn(x), remove_columns=dataset.column_names)"
+    "prompts = [example[\"caption\"] for example in dataset.take(NUM_SAMPLES)]"
    ]
   },
   {
@@ -1038,8 +1037,7 @@
     "quantization_config = OVWeightQuantizationConfig(bits=8, num_samples=NUM_SAMPLES, quant_method=OVQuantizationMethod.HYBRID)\n",
     "quantizer = OVQuantizer(int8_pipe)\n",
     "quantizer.quantize(\n",
-    "    ov_config=OVConfig(quantization_config=quantization_config),\n",
-    "    calibration_dataset=calibration_dataset,\n",
+    "    ov_config=OVConfig(quantization_config=quantization_config, dataset=prompts),\n",
     "    save_directory=int8_model_path\n",
     ")"
    ]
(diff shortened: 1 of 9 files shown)


eaidova (approved):

IlyasMoutawwakil (approved):
Looks good ! let's make sure future PRs with proper documentation of this API are included.

nikita-savelyevv:
The failed tests are not related to the changes

cc @IlyasMoutawwakil

PR #1239, opened 2025-04-14
fix bettertransformers import in codegen legacy

eaidova:
# What does this PR do?

one more fix codegen2 tests

## Before submitting


optimum/exporters/openvino/model_patcher.py
```diff
@@ -2024,9 +2024,8 @@ def _codegen_wrapped_scaled_dot_product_legacy(
     attention_mask: Optional[torch.Tensor] = None,
     head_mask: Optional[torch.Tensor] = None,
 ):
-    from optimum.bettertransformer.models.attention import raise_on_head_mask
-
-    raise_on_head_mask(head_mask)
+    if head_mask is not None:
+        raise ValueError("`head_mask` input argu,emt is not supported")
```
eaidova, line 2028:
```suggestion
        raise ValueError("`head_mask` input argument is not supported")
```

PR #1240, opened 2025-04-14
move sentencepiece to test requirements for unblocking python3.13

eaidova:
# What does this PR do?

there is known issue with installing sentencepiece https://github.com/google/sentencepiece/issues/1104
as this package used only specific models and no longer strong requirement for transformers (only own extra)
 https://github.com/huggingface/transformers/blob/main/setup.py#L329
I suggest to move it to test dependencies to unblock optimum-intel usage with python 3.13

## Before submitting


IlyasMoutawwakil:
wouldn't this cause BC problems to users ? they will need to add a `pip install sentencepiece` everywhere 🤔

eaidova:
> wouldn't this cause BC problems to users ? they will need to add a `pip install sentencepiece` everywhere 🤔

from my understanding, before running model that require sentencepiece with optimum, user usually run this model with transformers and probably already have message to install it before and have it installed. Majority of modern models do not required to have it installed at all, so I think it should not be big issue

PR #1228, opened 2025-04-08
Fix more ci

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1243, opened 2025-04-15
Install diffusers requirement during OV Full and Slow tests

nikita-savelyevv:
# What does this PR do?

After #1228 `diffusers` package is removed from general optimum-intel requirements list. Because of that OV Full and Slow tests started to fail:
- https://github.com/huggingface/optimum-intel/actions/runs/14460853601/job/40553085854
- https://github.com/huggingface/optimum-intel/actions/runs/14459777951/job/40550103640

## Before submitting


nikita-savelyevv:
Seems like the disk space is not enough again, will investigate

IlyasMoutawwakil:
can you please fix or skip the two failing models

PR #1254, opened 2025-04-18
restore cache_position input in whisper

eaidova:
# What does this PR do?
cache_position input was removed in optimum for whisper after this commit https://github.com/huggingface/optimum/commit/414afab57943eb6629181ff9a351a73c4eeb879e

 our internal optimizations and transformations rely on having this input in model, so restore it for openvino config

## Before submitting


IlyasMoutawwakil:
> our internal optimizations and transformations rely on having this input in model, so restore it for openvino config

can you please elaborate ?

eaidova:
> > our internal optimizations and transformations rely on having this input in model, so restore it for openvino config
> 
> can you please elaborate ?

we need it for running model with static shapes on NPU device

PR #1256, opened 2025-04-21
fix tests for working with transformers 4.51

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/loaders.py
+++ b/optimum/intel/openvino/loaders.py
@@ -65,6 +65,8 @@ class InsertTextEmbedding(MatcherPass):
 
                 # Use new operation for additional matching
                 self.register_new_node(add_ti)
+                print("FOUND")
+                return True
 
             # Root node wasn't replaced or changed
             return False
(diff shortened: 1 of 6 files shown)


optimum/intel/openvino/loaders.py
```diff
@@ -48,23 +48,25 @@ def __init__(self, tokens_ids, embeddings):
...
-                        name=f"{TEXTUAL_INVERSION_EMBEDDING_KEY}.textual_inversion_{token_id}",
+                        name=f"{friendly_name}.textual_inversion_{token_id}",
                     )
 
                 for consumer in consumers:
                     consumer.replace_source_output(add_ti.output(0))
 
                 # Use new operation for additional matching
                 self.register_new_node(add_ti)
+                return True
```
IlyasMoutawwakil, line 69: why does the callback need to return True ?

IlyasMoutawwakil:
Thanks @eaidova 🙏

PR #1237, opened 2025-04-14
Upgrade transformers to latest version for IPEX part

kaixuanliu:

PR #1260, opened 2025-04-23
Fix ipex CI

jiqing-feng:
Fix ipex test:
1. uninstall triton which is introduced by autoawq to avoid conflict with ipex
2. update tests for the bert model as the testing model updated.


jiqing-feng:
Hi @IlyasMoutawwakil . I found the transformers have bug in squeezebert, it will be fixed in the transformers 4.51.3. So I will disable squeezebert int this PR and it will be re enabled in #1237

IlyasMoutawwakil (approved):

IlyasMoutawwakil:
@jiqing-feng we should definitely distribute the ipex tests on multiple runners, they take a whole hour 🫠

jiqing-feng:
> @jiqing-feng we should definitely distribute the ipex tests on multiple runners, they take a whole hour 🫠

True.... Do you have any ideas? I think we can disable non-patch models tests because they were just enabled torch.compile for transformers models. WDYT?

IlyasMoutawwakil:
@jiqing-feng here's an example of distributing tests by test_file

jiqing-feng:
> @jiqing-feng here's an example of distributing tests by test_file

Do you have any links on it?

IlyasMoutawwakil:
my bad, I thought I added a link https://github.com/huggingface/optimum/pull/2234/files (top file)

jiqing-feng:
> my bad, I thought I added a link https://github.com/huggingface/optimum/pull/2234/files (top file)

I see, thanks for the instruction. I will delete some unnecessary tests first and then check if we need separate tests.

PR #1264, opened 2025-04-24
support LTX-video

eaidova:
# What does this PR do?

```
from diffusers.utils import export_to_video

from optimum.intel.openvino import OVDiffusionPipeline


pipe = OVDiffusionPipeline.from_pretrained("Lightricks/LTX-Video")

prompt = "The camera pans across a cityscape of tall buildings...The camera pans across a cityscape of tall buildings with a circular building in the center. The camera moves from left to right, showing the tops of the buildings and the circular building in the center. The buildings are various shades of gray and white, and the circular building has a green roof. The camera angle is high, looking down at the city. The lighting is bright, with the sun shining from the upper left, casting shadows from the buildings. The scene is computer-generated imagery."
negative_prompt = "worst quality, inconsistent motion, blurry, jittery, distorted"

video = pipe(
    prompt=prompt,
    negative_prompt=negative_prompt,
    width=704,
    height=480,
    num_frames=48,
    num_inference_steps=50,
).frames[0]
export_to_video(video, "output.mp4", fps=24)
```
ltx

## Before submitting


optimum/intel/__init__.py
```diff
@@ -124,6 +126,8 @@
         "OVPipelineForImage2Image",
         "OVPipelineForText2Image",
         "OVPipelineForInpainting",
+        "OVPipelineForText2Video",
+        "OVPipelineForImage2Video",
```
eaidova, line 130:
```suggestion
```

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1156,6 +1204,9 @@ def forward(
         if "attention_mask" in self.input_names:
             model_inputs["attention_mask"] = attention_mask
 
+        print(self.request)
+        print(model_inputs)
```
eaidova, line 1208:
```suggestion
```

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1266,7 +1322,19 @@ def forward(
...
+            model_inputs["height"] = height
+        if width is not None:
+            model_inputs["width"] = width
+        if rope_interpolation_scale is not None:
+            if not isinstance(rope_interpolation_scale, torch.Tensor):
+                rope_interpolation_scale = torch.tensor(rope_interpolation_scale)
+            model_inputs["rope_interpolation_scale"] = rope_interpolation_scale
+
+        print(model_inputs)
+        print(self.request.inputs)
```
eaidova, line 1337:
```suggestion
```

optimum/exporters/openvino/__main__.py
```diff
@@ -410,10 +410,10 @@ class StoreAttr(object):
                     )
                 model.config.pad_token_id = pad_token_id
 
-        if hasattr(model.config, "export_model_type"):
+        if hasattr(model.config, "export_model_type") and model.config.export_model_type is not None:
```
IlyasMoutawwakil, line 413: why would it be there and None ?

optimum/exporters/openvino/__main__.py
```diff
@@ -410,10 +410,10 @@ class StoreAttr(object):
...
             model_type = model.config.export_model_type.replace("_", "-")
         else:
-            model_type = model.config.model_type.replace("_", "-")
+            model_type = (getattr(model.config, "model_type", "") or "").replace("_", "-")
```
IlyasMoutawwakil, line 416: same as above

optimum/exporters/openvino/convert.py
```diff
@@ -1052,11 +1052,90 @@ def get_diffusion_models_for_export_ext(
...
+    vae_decoder = copy.deepcopy(pipeline.vae)
+    vae_decoder.register_to_config(
+        **{
+            "latents_mean_data": vae_decoder.latents_mean.tolist(),
+            "latents_std_data": vae_decoder.latents_std.tolist(),
+        }
+    )
+
+    def _vae_decode_forward(self, latent_sample, timestep=None):
+        return self.decode(z=latent_sample, temb=timestep)
```
IlyasMoutawwakil, line 1119: why not why not a lambda function here as well ?

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1378,6 +1445,18 @@ class OVModelVae:
...
+            patch_size = self.decoder.config.patch_size
+            patch_size_t = self.decoder.config.patch_size_t
+            spatio_temporal_scaling = self.decoder.config.spatio_temporal_scaling
+            self.spatial_compression_ratio = patch_size * 2 ** sum(spatio_temporal_scaling)
+            self.temporal_compression_ratio = patch_size_t * 2 ** sum(spatio_temporal_scaling)
+        self.latents_mean, self.latents_std = None, None
+        if hasattr(self.decoder.config, "latents_mean_data"):
+            self.latents_mean = torch.tensor(self.decoder.config.latents_mean_data)
+        if hasattr(self.decoder.config, "latents_std_data"):
+            self.latents_std = torch.tensor(self.decoder.config.latents_std_data)
```
IlyasMoutawwakil, line 1459: This should rather be done in the encoder/decoder using register_to_config (see https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/autoencoders/autoencoder_kl_ltx.py#L1144), the wrapper class is just interfacing the two components.

optimum/exporters/openvino/__main__.py
eaidova: FrozenDict behavior returns None if field does not exists in config and does not allow to set any additional fields

optimum/intel/openvino/modeling_diffusion.py
eaidova: not sure that I understand you. In VAE, `latents_std` and `latents_mean` are trainable parameters, we can not load them using the same way like pytorch because in general case, we have not access to pyotrch model weights.  To handle that I register data of these tensors in config and adds to vae manually

optimum/exporters/openvino/convert.py
eaidova: the second input is optional depends from config parameters, it will be more complicated if we will use lambda

optimum/exporters/openvino/__main__.py
eaidova: removed

optimum/exporters/openvino/__main__.py
eaidova: removed

optimum/exporters/openvino/convert.py
eaidova: used lambda

optimum/intel/openvino/modeling_diffusion.py
eaidova: @IlyasMoutawwakil  any ideas or comments?

IlyasMoutawwakil: makes sense, thanks for the explanation !

tests/openvino/test_diffusion.py
```diff
@@ -1003,3 +1004,159 @@ def test_textual_inversion(self):
         ov_output = ov_pipeline(**inputs, generator=get_generator("pt", SEED)).images
 
         np.testing.assert_allclose(ov_output, diffusers_output, atol=1e-4, rtol=1e-2)
+
+
+@unittest.skipIf(is_transformers_version("<", "4.45"), "Required transformers >= 4.45")
+class OVPipelineForText2VideoTest(unittest.TestCase):
+    SUPPORTED_ARCHITECTURES = []
+    if is_transformers_version(">=", "4.45.0"):
```
IlyasMoutawwakil, line 1012:
```suggestion
    if is_diffusers_version(">=", "0.28.2") and is_transformers_version(">=", "4.45.0"):
```

PR #1270, opened 2025-04-25
Clean ipex tests

jiqing-feng:
This PR will make the ipex tests faster than before:
1. Delete test pipeline in test_modeling.py as we already have test_pipelines.py
2. Delete some models which is out-of-date, duplicated, or not common.
3. Separate test model and test pipeline.


IlyasMoutawwakil:
Nice ! and very much cleaner !

jiqing-feng:
Hi  @IlyasMoutawwakil . All checks passed.

PR #1271, opened 2025-04-28
fix gptj export for transformers>4.49

eaidova:
# What does this PR do?

BetterTransformers is not applicable for transformers>=4.49, however not all of supported by it models have sdpa attention by default in transformers library now. In our case, usage eager attention instead of sdpa for this case, may lead to significant perf degradations and inability to apply PagedAttention later.

For avoid this:
* restored sdpa for models where it was covered by BetterTransformers (bloom, blenderbot, blenderbot-small, gptj, pegasus, marian)
* added checking number of SDPA operations after model export in tests

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -2920,7 +2920,7 @@ class GptJModelPatcher(DecoderModelPatcher):
     def __enter__(self):
         super().__enter__()
         patch_update_causal_mask(self._model, "4.45.0", "transformer")
-        if is_transformers_version(">", "4.49"):
+        if is_transformers_version(">=", "4.49"):
             self._model.config._orig_attn_implementation = self._model.config._attn_implementation
             self._model.config._attn_implementation = "sdpa"
             for block in self._model.transformer.h:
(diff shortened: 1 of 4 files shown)


optimum/exporters/openvino/model_patcher.py
```diff
@@ -2854,6 +2953,98 @@ def __exit__(self, exc_type, exc_value, traceback):
         unpatch_update_causal_mask(self._model, "gpt_neox_japanese")
 
 
+def _bloom_attn_forward(
```
echarlaix, line 2956: missing like to the original code

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4874,3 +5065,177 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        # key/value_states (first "if" case)
+        # if uni-directional self-attention (decoder) save Tuple(torch.Tensor, torch.Tensor) of
+        # all previous decoder key/value_states. Further calls to uni-directional self-attention
+        # can concat previous decoder key/value_states to current projected key/value_states (third "elif" case)
+        # if encoder bi-directional self-attention `past_key_value` is always `None`
+        past_key_value = (key_states, value_states)
+
+    query_states = self._shape(query_states, tgt_len, bsz)
+    key_states = key_states
+    value_states = value_states
```
echarlaix, line 5131:
```suggestion
```

echarlaix (approved):
LGTM!

optimum/exporters/openvino/model_patcher.py
eaidova: done

optimum/exporters/openvino/model_patcher.py
eaidova: done

PR #1273, opened 2025-04-29
support Zero-shot-Image-Classification

eaidova:
# What does this PR do?
Added inference class for zero-shot-classification models. Export part is standard and supported by optimum
```
from PIL import Image
import requests

from transformers import CLIPProcessor, CLIPModel
from optimum.intel.openvino import OVModelForZeroShotImageClassification

model = OVModelForZeroShotImageClassification.from_pretrained("openai/clip-vit-base-patch16")
processor = CLIPProcessor.from_pretrained("openai/clip-vit-base-patch16")

url = "http://images.cocodataset.org/val2017/000000039769.jpg"

inputs = processor(text=["a photo of a cat", "a photo of a dog"], images=image, return_tensors="pt", padding=True)

outputs = model(**inputs)
logits_per_image = outputs.logits_per_image # this is the image-text similarity score
probs = logits_per_image.softmax(dim=1) # we can take the softmax to get the label probabilities

print(probs)
```

## Before submitting

### DIFF
--- a/optimum/intel/__init__.py
+++ b/optimum/intel/__init__.py
@@ -191,6 +191,7 @@ else:
             "OVModelOpenCLIPVisual",
             "OVModelOpenCLIPText",
             "OVModelOpenCLIPForZeroShotImageClassification",
+            "OVModelForZeroShotImageClassification",
             "OVSamModel",
         ]
     )
--- a/optimum/intel/__init__.py
+++ b/optimum/intel/__init__.py
@@ -357,6 +357,7 @@ if TYPE_CHECKING:
             OVModelForTokenClassification,
             OVModelForVision2Seq,
             OVModelForVisualCausalLM,
+            OVModelForZeroShotImageClassification,
             OVModelOpenCLIPForZeroShotImageClassification,
             OVModelOpenCLIPText,
             OVModelOpenCLIPVisual,
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -35,6 +35,7 @@ from transformers import (
     AutoModelForQuestionAnswering,
     AutoModelForSequenceClassification,
     AutoModelForTokenClassification,
+    AutoModelForZeroShotImageClassification,
     PretrainedConfig,
 )
 from transformers.file_utils import add_start_docstrings, add_start_docstrings_to_model_forward
@@ -49,6 +50,7 @@ from transformers.modeling_outputs import (
     TokenClassifierOutput,
     XVectorOutput,
 )
+from transformers.models.clip.modeling_clip import CLIPOutput
 
 from ..utils.import_utils import is_timm_available, is_timm_version
 from .modeling_base import OVBaseModel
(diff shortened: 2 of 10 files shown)


optimum/intel/__init__.py
```diff
@@ -185,6 +185,7 @@
             "OVModelOpenCLIPVisual",
             "OVModelOpenCLIPText",
             "OVModelOpenCLIPForZeroShotImageClassification",
+            "OVModelForZeroShotImageClassification",
             "OVSamModel",
         ]
     )
```
echarlaix, line 191: could you also add in https://github.com/huggingface/optimum-intel/blob/f53b878b98f7680902e83f869767dc62d69038ac/optimum/intel/__init__.py#L339

optimum/intel/openvino/modeling.py
```diff
@@ -952,3 +954,45 @@ def forward(self, **kwargs):
...
+
+        np_inputs = isinstance(input_ids, np.ndarray)
+        if not np_inputs:
+            input_ids = input_ids.cpu().numpy()
+            pixel_values = pixel_values.cpu().numpy()
+            attention_mask = attention_mask.cpu().numpy() if attention_mask is not None else attention_mask
+        inputs = {"input_ids": input_ids, "pixel_values": pixel_values}
+        # Add the attention_mask when needed
+        if "attention_mask" in self.input_names:
+            inputs["attention_mask"] = attention_mask
```
echarlaix, line 974: here shouldn't we check that `attention_mask` is given before inference when needed (and raise an error if None or create a default one) ? not specific to this PR as it's following the same behavior as for other OVModels so no need to include anything just wanted to raise the point

echarlaix (approved):
LGTM!Thanks a lot for the addition @eaidova

nikita-savelyevv (approved):

optimum/intel/__init__.py
eaidova: added

optimum/intel/openvino/modeling.py
eaidova: added

optimum/intel/__init__.py
```diff
@@ -191,6 +191,7 @@
             "OVModelOpenCLIPVisual",
             "OVModelOpenCLIPText",
             "OVModelOpenCLIPForZeroShotImageClassification",
+            "OVModelForZeroShotImageClassification",
```
rkazants, line 194: why do we need `OVModelOpenCLIPForZeroShotImageClassification` then?

eaidova: for open_clip library support and its models, new one added for transformers. open clip has different API (splitting models on 2 parts, different configuration approach and pre- and post- processing steps

rkazants: well, I can't find `AutoModelOpenCLIPForZeroShotImageClassification`. How it is loaded? I thought the concept of moving to OpenVINO backend is always replacing `AutoModel` to `OvModel`.

rkazants, line 194: well, I can't find AutoModelOpenCLIPForZeroShotImageClassification. How it is loaded? I thought the concept of moving to OpenVINO backend is always replacing AutoModel to OvModel.

eaidova: AutoModel is transformers library integration way, this class for different library https://github.com/mlfoundations/open_clip it has different API. Models in such format are also available on huggingface https://huggingface.co/docs/hub/open_clip

I do not think that this PR is good place for such discussion and I do not think that removing open clip integration is good idea, not all models can be loaded by transformers as well that not all transformers supported clip models are compatible with open_clip

echarlaix:
failing test unrelated, merging

PR #1276, opened 2025-04-30
fix bug when bs > 1 and do not provide `position_ids` for  Mistral model

kaixuanliu:
1. fix bug when bs > 1 and do not provide `position_ids` for  Mistral model
2. update in-correct comments
3. update related test cases

### DIFF
--- a/tests/ipex/test_modeling_causal_lm.py
+++ b/tests/ipex/test_modeling_causal_lm.py
@@ -62,15 +62,8 @@ class IPEXModelForCausalLMTest(unittest.TestCase):
     GENERATION_LENGTH = 100
     SPEEDUP_CACHE = 1.0
 
-    @parameterized.expand(
-        grid_parameters(
-            {
-                "model_arch": SUPPORTED_ARCHITECTURES,
-                "prepare_inputs": [True, False],
-            }
-        )
-    )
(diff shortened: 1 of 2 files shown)


tests/ipex/test_modeling_causal_lm.py
```diff
@@ -62,43 +62,61 @@ class IPEXModelForCausalLMTest(unittest.TestCase):
...
+            tokens = tokenizer(
+                text,
+                return_tensors="pt",
+                return_token_type_ids=False if model_arch in ("llama2",) else None,
+            ).to(DEVICE)
+            if prepare_inputs:
+                inputs = ipex_model.prepare_inputs_for_generation(**tokens)
+                outputs = ipex_model(**inputs)
+            else:
+                outputs = ipex_model(**tokens)
```
echarlaix, line 91: both case can be included at the same time so that we don't have to re-export the model in both case with something like 
```suggestion
            outputs = ipex_model(**tokens)
            inputs = ipex_model.prepare_inputs_for_generation(**tokens)
            outputs_2 = ipex_model(**inputs)
            self.assertTrue(torch.allclose(outputs.logits, outputs_2.logits, atol=1e-3))
```

echarlaix (approved):
LGTM thanks @kaixuanliu

PR #1277, opened 2025-04-30
upgrade torch to 2.7

jiqing-feng:


jiqing-feng:
Hi @IlyasMoutawwakil . I hhave upgraded the ipex and torch version to 2.7 and disable 2.6 as there is no reason to keep 2 versions check. Please review it. Thanks!

PR #1278, opened 2025-04-30
Remove OVBaseModelForSeq2SeqLM 

echarlaix:
Remove `OVBaseModelForSeq2SeqLM` which has no specific reason to exist, and move the model loading /saving logic to OVModelForSeq2SeqLM directly


optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -187,10 +186,10 @@ class _OVModelForSpeechT5ForTextToSpeech(OVModelForTextToSpeechSeq2Seq):
...
     def __init__(
         self,
-        encoder_model: ov.Model,
-        decoder_model: ov.Model,
-        postnet_model: ov.Model,
-        vocoder_model: ov.Model,
+        encoder: ov.Model,
+        decoder: ov.Model,
+        postnet: ov.Model,
+        vocoder: ov.Model,
```
echarlaix, line 192: to stay consistent with other seq2seq models  https://github.com/huggingface/optimum-intel/blob/bdeea95b61a80b9b9217fb730b3609657b55d208/optimum/intel/openvino/modeling_seq2seq.py#L320

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -200,7 +199,7 @@ def __init__(
         **kwargs,
     ):
         self.config = config
-        self.use_cache = kwargs.get("use_cache", True)
+        self.use_cache = model_has_state(decoder)
```
echarlaix, line 202: to stay consistent with other seq2seq models  https://github.com/huggingface/optimum-intel/blob/bdeea95b61a80b9b9217fb730b3609657b55d208/optimum/intel/openvino/modeling_seq2seq.py#L332-L333

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -296,7 +295,6 @@ def _from_pretrained(
         cls,
         model_id: Union[str, Path],
         config: "PretrainedConfig",
-        use_auth_token: Optional[Union[bool, str]] = None,
```
echarlaix, line 299: not used so removing

tests/openvino/test_modeling.py
```diff
@@ -547,6 +549,45 @@ def test_load_from_hub_and_save_sam_model(self):
         del model
         gc.collect()
 
+    def test_load_from_hub_and_save_text_speech_model(self):
```
echarlaix, line 552: adding missing test to check we keep compatibility with openvino model pushed on the hub

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -455,16 +779,12 @@ def clear_requests(self):
             raise ValueError(
                 "`clear_requests()` is not supported with `compile_only` mode, please initialize model without this option"
             )
-        self.encoder.request = None
-        self.decoder.request = None
-        if self.decoder_with_past is not None:
-            self.decoder_with_past.request = None
+        for submodel_name in self._ov_submodel_names:
+            getattr(self, submodel_name).request = None
```
IlyasMoutawwakil, line 783:
```suggestion
        for submodel in self.ov_submodels.values():
            submodel.request = None
```
```suggestion
        for submodel in self.ov_submodels.values():
            submodel.request = None
```

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -310,7 +318,7 @@
     """,
     INPUTS_DOCSTRING,
 )
-class OVModelForSeq2SeqLM(OVBaseModelForSeq2SeqLM, GenerationMixin):
+class OVModelForSeq2SeqLM(OVBaseModel, GenerationMixin):
```
rkazants, line 321: can we give some description to this class: architecture we expect, format of inputs/outputs and their number?
It will clarify to developers for what cases to utilize this class.

optimum/intel/openvino/modeling_seq2seq.py
nikita-savelyevv: I don't think this will work out because `submodel` is an instance of `openvino.Model` but should be an instance of `OVEncoder` or `OVDecoder`.

optimum/intel/openvino/quantization.py
```diff
@@ -715,16 +715,14 @@ def _prepare_speech_to_text_calibration_data(
...
         collected_inputs: Dict[str, List[Dict[str, Any]]] = {}
-        for submodel_name in self.model.ov_submodels:
-            ov_component_name = "_".join(submodel_name.split("_")[:-1])  # e.g. "encoder_model" -> "encoder"
-            ov_component: Union[OVEncoder, OVDecoder] = getattr(self.model, ov_component_name)
-            models[ov_component_name] = ov_component
-            collected_inputs[ov_component_name] = []
+        for submodel_name in self.model._ov_submodel_names:
+            ov_component: Union[OVEncoder, OVDecoder] = getattr(self.model, submodel_name)
+            models[submodel_name] = ov_component
+            collected_inputs[submodel_name] = []
```
nikita-savelyevv, line 721: 👍

optimum/intel/openvino/modeling_seq2seq.py
echarlaix: > I don't think this will work out because submodel is an instance of openvino.Model but should be an instance of OVEncoder or OVDecoder.

yes you're correct @nikita-savelyevv, we can't  use `self.ov_submodels` directly for this reason unfortunately @IlyasMoutawwakil

rkazants (approved):

nikita-savelyevv (approved):

optimum/intel/openvino/modeling_seq2seq.py
rkazants: Please give some common description in what cases this class is applicable like encoder-decoder arch with text input and output or smth else. So this way I know that I should use this class for some model enabling.
Or please give description that this class presents common/abstract API that child class should implement. And this child class will help to enable some type of models like encoder-decoder arch with text-to-audio.
I am not sure that just having constructor argument description will be helpful to developers.

optimum/intel/openvino/modeling_seq2seq.py
IlyasMoutawwakil: ooh I see thanks !

PR #1279, opened 2025-04-30
Fix minimum diffusers version for LTXPipeline

echarlaix:
LTXPipeline was introduced in diffusers v0.32 resulting in current issuers for diffusers version [0.28.2,0.31]


IlyasMoutawwakil (approved):
LGTM ! Thanks for the fix 👌

eaidova (approved):

PR #1274, opened 2025-04-29
Avoid use of deprecated openvino.runtime

rkazants:
# What does this PR do?

Avoid use of deprecate openvino.runtime namespace in favor of openvino. In 2025.1.0 openvino release, openvino namespace of the OpenVINO Python API has been redesigned, removing the nested openvino.runtime module. The old namespace is now considered deprecated and will be discontinued in 2026.0.

Fixes # TBD

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1280, opened 2025-04-30
Replace deprecated openvino.runtime 

echarlaix:
Following https://github.com/huggingface/optimum-intel/pull/1274


rkazants (approved):

IlyasMoutawwakil (approved):

IlyasMoutawwakil:
Fixed the two notebooks for inference and question answering quantization. For the other two (were never being tested) it seems that the quantization recipe is broken from latest changes (@nikita-savelyevv) and for the stable diffusion hybrid quantization, we will need to probably switch to lcm (takes less time to quantize and run inference with).

echarlaix:
Thanks a lot for the fixes @IlyasMoutawwakil  :raised_hands:

PR #1286, opened 2025-05-05
Fix custom dataset processing for text encoding tasks

nikita-savelyevv:
# What does this PR do?

Add a previously missed support for calibration dataset creation from a custom dataset for text encoding tasks. In particular, this fixes sentence_transformer_quantization notebook.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -833,13 +833,6 @@ class OVCalibrationDatasetBuilder:
                 inference_result_mock=inference_result_mock,
             )
 
-            if isinstance(self.model, OVSentenceTransformer):
-                tokenizer = self.model.tokenizer
-            else:
-                tokenizer = AutoTokenizer.from_pretrained(
-                    quantization_config.tokenizer, trust_remote_code=quantization_config.trust_remote_code
-                )
-
             max_position_embeddings = getattr(self.model.config, "max_position_embeddings", None)
             if max_position_embeddings is not None and max_position_embeddings > 0:
                 seq_len = min(seq_len, max_position_embeddings)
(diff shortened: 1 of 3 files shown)


optimum/intel/openvino/quantization.py
```diff
@@ -849,15 +842,33 @@ def _prepare_text_encoder_model_calibration_data(
...
+                        if isinstance(self.model, OVSentenceTransformer):
+                            tokenizer = self.model.tokenizer
+                        else:
+                            if quantization_config.tokenizer is None:
+                                raise ValueError(
+                                    "Please provide tokenizer for calibration via quantization_config.tokenizer."
+                                )
+                            tokenizer = AutoTokenizer.from_pretrained(
+                                quantization_config.tokenizer, trust_remote_code=quantization_config.trust_remote_code
+                            )
```
IlyasMoutawwakil, line 862: let's keep this outside the loop please

IlyasMoutawwakil (approved):
Thanks ! I left a small suggestion for better separation of concerns. Otherwise LGTM !
It's pretty difficult to do everything for every possible combination of tasks and classes in the same function so I suggest that we make this code more modular in future PRs.

optimum/intel/openvino/quantization.py
nikita-savelyevv: Moved

optimum/intel/openvino/quantization.py
```diff
@@ -849,15 +853,22 @@ def _prepare_text_encoder_model_calibration_data(
                 with numpy_seed(self.seed):
                     random_positions = np.random.randint(0, seq_len, num_samples)
 
+            tokenizer = None
             pbar = tqdm(total=num_samples, desc="Collecting calibration data")
             for item in dataset:
-                inputs = tokenizer(item["text"], truncation=True, max_length=seq_len, return_tensors="np")
-                if inputs["input_ids"].shape[1] < seq_len:
-                    continue
+                if "input_ids" in item:
```
nikita-savelyevv, line 859: I've noticed that similar checks will have to be added to other model type dataset preparation functions which is quite inconvenient. This is shortcoming after #1232 . I hope there is a better solution, at least temporary. I need to analyze it a bit more.

nikita-savelyevv: I have some ideas for improvement, but prototyping is needed. I will have to address this after the upcoming release.

PR #1292, opened 2025-05-09
Add compression tests for phi4mm

nikita-savelyevv:
# What does this PR do?

As in the description.

## Before submitting

### DIFF
--- a/tests/openvino/test_exporters_cli.py
+++ b/tests/openvino/test_exporters_cli.py
@@ -526,6 +526,29 @@ class OVCLIExportTestCase(unittest.TestCase):
             ]
         )
 
+    if is_transformers_version(">=", "4.49.0"):
+        TEST_4BIT_CONFIGURATIONS.extend(
+            [
+                (
+                    "image-text-to-text",
+                    "phi4mm",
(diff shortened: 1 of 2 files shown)


tests/openvino/test_exporters_cli.py
```diff
@@ -526,6 +526,29 @@ class OVCLIExportTestCase(unittest.TestCase):
...
+                    [
+                        {"int8": 8, "int4": 42},
+                        {"int8": 1},
+                        {"int8": 8},
+                        {"int8": 2},
+                        {},
+                        {"int8": 6},
+                        {"int8": 25},
+                        {"int8": 2},
+                        {"int8": 2},
```
echarlaix, line 546: would you mind adding a comment to specify which submodel does each apply to ? (to improve clarity)

echarlaix (approved):
LGTM!

eaidova (approved):

tests/openvino/test_exporters_cli.py
nikita-savelyevv: Thanks for the idea, but let me address this in a separate PR because this will require changes across a lot of test cases.

PR #1293, opened 2025-05-12
fix automatic task detection for phi4-multimodal during data-aware quant

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -88,6 +88,18 @@ def infer_task(
                     token=token,
                     library_name=library_name,
                 )
+                # phi-4-multimodal has more common with multimodal, however pipeline_tag defined as automatic-speech-recognition on hub
+                if "automatic-speech-recognition" in task and library_name == "transformers":
+                    config = AutoConfig.from_pretrained(
+                        model_name_or_path,
+                        subfolder=subfolder,
+                        revision=revision,
+                        cache_dir=cache_dir,
+                        token=token,
+                        trust_remote_code=True,
+                    )
+                    if getattr(config, "model_type", "").replace("_", "-") in ["phi4mm", "phi4-multimodal"]:
+                        task = "image-text-to-text"
             except KeyError as e:
                 raise KeyError(
                     f"The task could not be automatically inferred. Please provide the argument --task with the relevant task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"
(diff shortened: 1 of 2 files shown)


optimum/exporters/openvino/__main__.py
```diff
@@ -88,6 +88,18 @@ def infer_task(
                     token=token,
                     library_name=library_name,
                 )
+                # phi-4-multimodal has more common with multimodal, however pipeline_tag defined as automatic-speech-recognition on hub
+                if "automatic-speech-recognition" in task and library_name == "transformers":
```
echarlaix, line 92: is it needed because for quantization we are checking for the task to match like in :
 https://github.com/huggingface/optimum-intel/blob/3910592c0cdb9f11ce0d773cc3ec14d7713b8522/optimum/exporters/openvino/convert.py#L98

in all cases why not checking for the architecture instead [MULTI_MODAL_TEXT_GENERATION_MODELS](https://github.com/huggingface/optimum-intel/blob/3910592c0cdb9f11ce0d773cc3ec14d7713b8522/optimum/exporters/openvino/utils.py#L220) ?

eaidova: not sure tha tI understand you.
when we use quantization with dataset via optimum-cli, we use task to determine which class should be used for model loading here https://github.com/huggingface/optimum-intel/blob/main/optimum/commands/export/openvino.py#L406

phi-4-multimodal model card contains pipeline_tag as automatic-speech-recognition here https://huggingface.co/microsoft/Phi-4-multimodal-instruct

as the result instead of multimodal model loading we will try to create speech recognition model

nikita-savelyevv (approved):

optimum/exporters/openvino/__main__.py
echarlaix: OK I see this is only needed for the CLI + data aware quantization (it could also make sense to only set it default there)

eaidova: moved to command

optimum/exporters/openvino/__main__.py
```diff
@@ -88,6 +88,18 @@ def infer_task(
...
+                    config = AutoConfig.from_pretrained(
+                        model_name_or_path,
+                        subfolder=subfolder,
+                        revision=revision,
+                        cache_dir=cache_dir,
+                        token=token,
+                        trust_remote_code=True,
```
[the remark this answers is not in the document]
eaidova: moved where user specifiied trust_remote_code known

echarlaix:
thanks a lot @eaidova, merging

Issue #1294, opened 2025-05-12
IPEX test failed


jiqing-feng:
The [IPEXLangchainTest](https://github.com/huggingface/optimum-intel/blob/main/tests/ipex/test_modeling.py#L442) failed from this PR: #1289 . But this PR seems not the reason. 
Hi @echarlaix , could you take a look? Thanks!

PR #1296, opened 2025-05-12
Update tiny model for ipex lanchain tests

echarlaix:
Fix https://github.com/huggingface/optimum-intel/issues/1294

### DIFF
--- a/tests/ipex/test_modeling.py
+++ b/tests/ipex/test_modeling.py
@@ -440,7 +440,7 @@ class IPEXSTModel(unittest.TestCase):
 
 
 class IPEXLangchainTest(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


tests/ipex/test_modeling.py
```diff
@@ -440,7 +440,7 @@ def test_langchain(self, model_arch):
 
 
 class IPEXLangchainTest(unittest.TestCase):
-    SUPPORTED_ARCHITECTURES = ("gpt2",)
```
echarlaix, line 443: model from https://huggingface.co/Intel/tiny-random-gpt2 doesn't generate anything (same issue for original model, not specific to the ipex resulting model), replacing it with an other model

PR #1297, opened 2025-05-12
enable longrope for phi3 model

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


optimum/exporters/openvino/model_patcher.py
```diff
@@ -1608,15 +1608,54 @@ def _phi3_self_attn_sdpa_forward(
     return attn_output, None, past_key_value
 
 
+# @torch.jit.script
```
eaidova, line 1611:
```suggestion
```

PR #1302, opened 2025-05-13
[OV] Introduce default full quantization configs for clip models

nikita-savelyevv:
# What does this PR do?

Add functionally for adding default int8 quantization configs alongside default int4 weight compression configs. The need for such feature came up during enabling of clip models which require a specific `smooth_quant_alpha` parameter to increase accuracy.

| Model ID           | FP32 Accuracy    | INT8 PTQ Accuracy|
|---------------------------------|--------------------|---------------------|
| openai/clip-vit-base-patch16                |  68.33%             |      67.71% (-0.62%)         | 
| openai/clip-vit-base-patch32                 |   63.37%         |        62.52% (-0.85%)      | 
| openai/clip-vit-large-patch14                  |    75.55%       |     74.88% (-0.67%)         | 

INT8 PTQ was applied like this: 
```
optimum-cli export openvino -m openai/clip-vit-base-patch16 --quant-mode int8 ./save_dir
```

Accuracy was measured using https://github.com/l-bat/CLIP_benchmark with a command like this: 
```
python clip_benchmark/cli.py eval --model_type "optimum_intel" --model openai \
  --pretrained clip-vit-base-patch16 --dataset "imagenet1k" --seed 0 --task "zeroshot_classification" \
  --output "eval.json" --dataset_root <dataset_path> 
```

## Before submitting

### DIFF
--- a/docs/source/openvino/export.mdx
+++ b/docs/source/openvino/export.mdx
@@ -183,6 +183,21 @@ Besides weight-only quantization, you can also apply full model quantization inc
 optimum-cli export openvino -m openai/whisper-large-v3-turbo --quant-mode int8 --dataset librispeech --num-samples 32 --smooth-quant-alpha 0.9 ./whisper-large-v3-turbo
 ```
 
+#### Default quantization configs
+
+For some models we maintain a set of default quantization configs. To apply a default 4-bit weight-only quantization one should provide `--weight-format int4` without any additional arguments. For int8 weight & activation quantization it should be `--quant-mode int8`. For example:
+
+```bash
+optimum-cli export openvino -m microsoft/Phi-4-mini-instruct --weight-format int4 ./Phi-4-mini-instruct
+```
+
+or
+
+```bash
+optimum-cli export openvino -m openai/clip-vit-base-patch16 --quant-mode int8 ./clip-vit-base-patch16
+```
+
+
 ### Decoder models
 
 For models with a decoder, we enable the re-use of past keys and values by default. This allows to avoid recomputing the same intermediate activations at each generation step. To export the model without, you will need to remove the `-with-past` suffix when specifying the task.
--- a/docs/source/openvino/export.mdx
+++ b/docs/source/openvino/export.mdx
@@ -185,7 +185,7 @@ optimum-cli export openvino -m openai/whisper-large-v3-turbo --quant-mode int8 -
 
 #### Default quantization configs
 
-For some models we maintain a set of default quantization configs. To apply a default 4-bit weight-only quantization one should provide `--weight-format int4` without any additional arguments. For int8 weight & activation quantization it should be `--quant-mode int8`. For example:
+For some models we maintain a set of default quantization configs ([link](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py)). To apply a default 4-bit weight-only quantization one should provide `--weight-format int4` without any additional arguments. For int8 weight & activation quantization it should be `--quant-mode int8`. For example:
 
 ```bash
 optimum-cli export openvino -m microsoft/Phi-4-mini-instruct --weight-format int4 ./Phi-4-mini-instruct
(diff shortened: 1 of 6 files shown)


tests/openvino/test_exporters_cli.py
```diff
@@ -961,49 +962,93 @@ def test_exporters_cli_full_quantization(
                 expected_fake_nodes_per_model,
             )
 
-    def test_exporters_cli_int4_with_local_model_and_default_config(self):
+    @parameterized.expand(
+        [
+            (
+                "falcon-40b",
+                "tiiuae/falcon-7b-instruct",
```
IlyasMoutawwakil, line 969: no need for the model id, let's continue using MODEL_NAMES[model_type]

alexsu52 (approved):
Good staff!

tests/openvino/test_exporters_cli.py
nikita-savelyevv: Do you have a suggestion how to avoid using `model_id`? In the test we patch `_name_or_path` at `config.json` with the `model_id`. This is needed for the default config to match up and avoids adding a test element to `_DEFAULT_4BIT_WQ_CONFIGS`. 

Unfortunately, for the newly added test, I had to add a separate config for the test model anyways because there is no data-free full quantization default config and collecting 300 samples would take too much time. Also, there is no way to patch the config-retrieval logic because the export is run in a separate process.

docs/source/openvino/export.mdx
```diff
@@ -183,6 +183,21 @@ Besides weight-only quantization, you can also apply full model quantization inc
...
+```bash
+optimum-cli export openvino -m microsoft/Phi-4-mini-instruct --weight-format int4 ./Phi-4-mini-instruct
+```
+
+or
+
+```bash
+optimum-cli export openvino -m openai/clip-vit-base-patch16 --quant-mode int8 ./clip-vit-base-patch16
+```
+
```
eaidova, line 199: not sure if it is a good idea, but maybe for better understanding, it also makes sense to provide a link where these default configs can be found?

eaidova (approved):

docs/source/openvino/export.mdx
nikita-savelyevv: That's a good idea. For now, I've added a link to `configuration.py`. In the future a dedicated page containing all default config would be helpful.

tests/openvino/test_exporters_cli.py
IlyasMoutawwakil: my bad, thought the model_id is the one we're gonna load (was surprised it's a 7b model). I see now that a "model_id" is necessary to trick the quantizer to apply a certain recipe !

optimum/intel/openvino/configuration.py
```diff
@@ -341,38 +342,99 @@ class OVQuantizationMethod(str, Enum):
 }
 
 
-def _check_default_4bit_configs(model_id_or_path: str):
-    if model_id_or_path in _DEFAULT_4BIT_CONFIGS:
-        return _DEFAULT_4BIT_CONFIGS[model_id_or_path]
+# Default configs for int8 full quantization
+_DEFAULT_INT8_FQ_CONFIGS = {
+    "hf-tiny-model-private/tiny-random-CLIPModel": {  # For test purposes
```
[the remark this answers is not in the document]
IlyasMoutawwakil: I see your point about having unnecessary configs only for testing, so I think staying consistent matters here.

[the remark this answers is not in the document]
nikita-savelyevv: For the old tests there were plenty of existing data-free 4bit configs which are fast to apply. For the newly added int8 full quantization case all current configs are data-aware with 300 samples. Applying each of them even on a tiny model would take a couple of minutes because the dataset images need to be downloaded. That's why I've added a test config containing `num_samples=1`. Also, it's not possible to patch the config-retrieval logic before running the test because the export is run in a separate process.

IlyasMoutawwakil (approved):
LGTM !

echarlaix (approved):
Looks great, thanks a lot @nikita-savelyevv !

PR #1303, opened 2025-05-13
raise if converted subcomponent not found

echarlaix:


optimum/exporters/openvino/__main__.py
```diff
@@ -475,6 +475,12 @@ class StoreAttr(object):
 
         for submodel_path in submodel_paths:
             submodel_path = Path(output) / submodel_path
+
+            if not submodel_path.is_file():
+                raise RuntimeError(
+                    f"An issue happenned during export : {submodel_path.name} was not converted and saved as expected."
```
echarlaix, line 481: @helena-intel let me know if that differs with what you had in mind!

helena-intel: Thanks @echarlaix! I think usually when there are issues with this, the model does exist, but it's broken. When I have seen that with LLMs, it resulted in an empty .xml file, so I would suggest for example
```
 if (not submodel_path.is_file()) or (submodel_path.stat().st_size == 0)
```
@eaidova do you have another suggestion?

echarlaix: sure updated in https://github.com/huggingface/optimum-intel/pull/1303/commits/41b8adc5bb511cf52e1e69a2d91362f346f2a6b7

IlyasMoutawwakil (approved):

PR #1308, opened 2025-05-14
Fixes openvino decoder models output

echarlaix:
cc @helena-intel who found this pretty serious issue that impacts decoder, seq2seq and visual_language models, let me know if you still encounter something unexpected. If that works for you we can include it in the release (which can be done either today or tomorrow)

### DIFF
--- a/tests/openvino/test_modeling.py
+++ b/tests/openvino/test_modeling.py
@@ -1450,25 +1450,35 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         model_id = MODEL_NAMES["gpt2"]
         tokenizer = AutoTokenizer.from_pretrained(model_id)
         tokens = tokenizer("This is a sample input", return_tensors="pt")
+
         model_with_pkv = OVModelForCausalLM.from_pretrained(model_id, export=True, use_cache=True, stateful=False)
         outputs_model_with_pkv = model_with_pkv.generate(
             **tokens, min_length=self.GENERATION_LENGTH, max_length=self.GENERATION_LENGTH, num_beams=1
         )
+        del model_with_pkv
+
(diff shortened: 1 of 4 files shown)


tests/openvino/test_modeling.py
```diff
@@ -1450,25 +1450,35 @@ def test_compare_with_and_without_past_key_values(self):
...
         )
         self.assertTrue(torch.equal(outputs_model_without_pkv, outputs_model_stateful))
 
-        del model_with_pkv
-        del model_without_pkv
+        logits = model_stateful(**tokens).logits
+        copy_logits = copy.deepcopy(logits)
+        tokens = tokenizer("Input sample", return_tensors="pt")
+        model_stateful(**tokens).logits
+        self.assertTrue(torch.equal(copy_logits, logits))
```
nikita-savelyevv, line 1480: As I understand, before the changes these two tensor would not be equal, right? How exactly is this connected with the newly added cloning of logits/keys/values?

helena-intel (approved):
Thanks very much for the quick fix @echarlaix ! With this PR, I don't see the changing outputs anymore.  I ran a quick benchmark with llama-3.1-8b  INT4 on Xeon and saw similar performance as before the PR, but slightly higher memory use (10.6 GB before, 11 GB after, with 1024/1024 inputs/outputs). But for now that seems acceptable to me.

tests/openvino/test_modeling.py
echarlaix: yes exactly, here we return a copy of the output to make sure it's not modified during a following inference (would be modified here otherwise https://github.com/huggingface/optimum-intel/blob/94b2d60466624cd4f8128a0943cb59dc15cd0b7c/optimum/intel/openvino/modeling_decoder.py#L559

echarlaix:
Huge catch @helena-intel, thanks a lot!

PR #1310, opened 2025-05-15
Introduce OVPipelineQuantizationConfig

nikita-savelyevv:
# What does this PR do?

Changes:
- Introduced `OVPipelineQuantizationConfig` allowing to specify quantization parameters per model component. Add corresponding tests.
- Introduced a more advanced logic for inferring quantization config type from dictionary. Moved this logic to a separate function: `_quantization_config_from_dict()`.
- Updated default int4 config for phi4-multimodal. [WWB](https://github.com/openvinotoolkit/openvino.genai/tree/master/tools/who_what_benchmark) similarity: 85.30%.

For example the code below applies int8 PTQ to `lm_model`, int8 WC to `text_embeddings_model` and no optimization to `vision_embeddings_model`.
```python
from optimum.intel import OVModelForVisualCausalLM
from optimum.intel import OVPipelineQuantizationConfig, OVQuantizationConfig, OVWeightQuantizationConfig

model_id = "OpenGVLab/InternVL2-1B"
model = OVModelForVisualCausalLM.from_pretrained(
    model_id,
    export=True,
    trust_remote_code=True,
    quantization_config=OVPipelineQuantizationConfig(
        quantization_configs={
            "lm_model": OVQuantizationConfig(bits=8),
            "text_embeddings_model": OVWeightQuantizationConfig(bits=8),
        },
        dataset="contextual",
        trust_remote_code=True,
    )
)
```

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -316,7 +316,7 @@ _DEFAULT_4BIT_WQ_CONFIGS = {
             "lm_model": {
                 "bits": 4,
                 "sym": False,
-                "group_size": 128,
+                "group_size": 64,
                 "dataset": "contextual",
                 "quant_method": OVQuantizationMethod.AWQ,
                 "scale_estimation": True,


nikita-savelyevv:
@l-bat Could you please review this PR?

tests/openvino/test_quantization.py
```diff
@@ -1465,6 +1376,289 @@ def test_ovmodel_4bit_dynamic_with_config(
...
+                        "vae_encoder": {"int8": 42},
+                        "text_encoder": {"int8": 30},
+                        "text_encoder_2": {"int8": 0},
+                        "text_encoder_3": {"int8": 0},
+                    },
+                ),
+            ]
+        )
+
+    if is_transformers_version(">", "4.43.0"):
```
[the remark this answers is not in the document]
nikita-savelyevv: Thanks, fixed.

docs/source/openvino/optimization.mdx
```diff
@@ -636,4 +636,28 @@ To apply mixed quantization through CLI, the `--quant-mode` argument should be u
...
+
+model_id = "OpenGVLab/InternVL2-8B"
+model = OVModelForVisualCausalLM.from_pretrained(
+    model_id,
+    export=True,
+    quantization_config=OVPipelineQuantizationConfig(
+        pipeline_quantization_configs={
+            "lm_model": OVWeightQuantizationConfig(bits=4, group_size=128, dataset="contextual", quant_method="awq"),
+            "text_embeddings_model": OVWeightQuantizationConfig(bits=8),
+        },
```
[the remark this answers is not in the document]
l-bat: Yes, for demo purposes. Maybe add another example with this in the PR description.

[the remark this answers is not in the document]
nikita-savelyevv: Updated the docs too. I've run the sample, the resulting model produces adequate results.

docs/source/openvino/optimization.mdx
```diff
@@ -636,4 +646,28 @@ To apply mixed quantization through CLI, the `--quant-mode` argument should be u
...
+
+For example, the code below quantizes weights and activations of a language model inside InternVL2-1B, compresses weights of a text embedding model and skips any quantization for vision embedding model.
+
+```python
+from optimum.intel import OVModelForVisualCausalLM
+from optimum.intel import OVPipelineQuantizationConfig, OVQuantizationConfig, OVWeightQuantizationConfig
+
+model_id = "OpenGVLab/InternVL2-1B"
+model = OVModelForVisualCausalLM.from_pretrained(
+    model_id,
```
eaidova, line 664: probably this example required trust_remote_code=True?

docs/source/openvino/optimization.mdx
```diff
@@ -636,4 +646,28 @@ To apply mixed quantization through CLI, the `--quant-mode` argument should be u
...
 ```
 
-Don't forget to provide a dataset since it is required for the calibration procedure during full quantization.
\ No newline at end of file
+Don't forget to provide a dataset since it is required for the calibration procedure during full quantization.
+
+
+## Pipeline Quantization
+
+There are multimodal pipelines that consist of multiple components, such as Stable Diffusion or Visual Causal Language models. In these cases, there may be a need to apply different quantization methods to different components of the pipeline. For example, you may want to apply int4 data-aware weight-only quantization to a language model in visual-language pipeline, while applying int8 weight-only quantization to other components. In this case you can use the `OVPipelineQuantizationConfig` class to specify the quantization configuration for each component of the pipeline.
```
eaidova, line 654: VLM, Visual Language Models, it is the more popular name for such model types
```suggestion
There are multimodal pipelines that consist of multiple components, such as Stable Diffusion or Visual Language models. In these cases, there may be a need to apply different quantization methods to different components of the pipeline. For example, you may want to apply int4 data-aware weight-only quantization to a language model in visual-language pipeline, while applying int8 weight-only quantization to other components. In this case you can use the `OVPipelineQuantizationConfig` class to specify the quantization configuration for each component of the pipeline.
```

eaidova (approved):

docs/source/openvino/optimization.mdx
nikita-savelyevv: Thanks, updated

optimum/intel/openvino/configuration.py
```diff
@@ -1176,3 +1174,146 @@ def to_dict(self):
         result["weight_quantization_config"] = self.weight_quantization_config.to_dict()
         result["full_quantization_config"] = self.full_quantization_config.to_dict()
         return result
+
+
+class OVPipelineQuantizationConfig(OVQuantizationConfigBase):
+    def __init__(
+        self,
+        pipeline_quantization_configs: Dict[str, Union[Dict, OVQuantizationConfigBase]],
```
IlyasMoutawwakil, line 1182: would makes more sense to call this argument just quantization_configs because the PR already introduced something called "pipeline quantization config" so it's a bit confusing

optimum/intel/openvino/quantization.py
```diff
@@ -1365,7 +1405,7 @@ def _quantize_torchmodel(
             model = nncf.quantize(
                 model,
                 calibration_datasets["model"],
-                subset_size=quantization_config.num_samples,
+                subset_size=quantization_config.num_samples or 128,
```
IlyasMoutawwakil, line 1408: why is this value hidden here ?

nikita-savelyevv: 128 was previously a default value for `num_samples` argument in `OVQuantizationConfig`. In this PR I've removed it so it is now `None` by default. That's why 128 has appeared here.

Ideally, we should transition to providing arguments via `quantization_config.to_nncf_dict()` here, as it is done for OV case. But I propose to do this in a separate PR.

optimum/intel/openvino/configuration.py
nikita-savelyevv: In my opinion `quantization_configs` is less clear when initializing config as a dictionary, but I agree that `pipeline` part seems redundant. Removed.

nikita-savelyevv:
Hi @echarlaix! Do you mind us merging this or you would like to take a look?

echarlaix:
apologies for the delay @nikita-savelyevv, taking a look right now!

echarlaix (approved):
Looks great thanks @nikita-savelyevv !!

PR #1312, opened 2025-05-16
allow passing ov_config to from_pretrained

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -619,11 +619,13 @@ class OVBaseModel(OptimizedModel):
             )
             compile_only = False
 
-        # If load_in_8bit and quantization_config not specified then ov_config is set to None and will be set by default in convert depending on the model size
-        if load_in_8bit is None and not quantization_config:
-            ov_config = None
-        else:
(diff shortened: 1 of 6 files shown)


optimum/intel/openvino/modeling_base.py
```diff
@@ -619,10 +619,8 @@ def _from_transformers(
             )
             compile_only = False
 
-        # If load_in_8bit and quantization_config not specified then ov_config is set to None and will be set by default in convert depending on the model size
-        if load_in_8bit is None and not quantization_config:
-            ov_config = None
-        else:
+        ov_config = kwargs.get("ov_export_config")
```
rkazants, line 622: can you please add test with passing ov_export_config?
Also, please update text2speech modelling file

rkazants (approved):
May be it makes sense to add this into documentation.
Also, should not we check the content of this config file like presence of `dtype`?
Can it be that it contradicts to quantization_config file if it is also specified?

eaidova:
From beggining, I would like to clarify that I'm still working on this PR do not invited nobody to review and do not asked @rkazants  to do that. Pleae ignore this PR sometime until it will not be ready

PR #1314, opened 2025-05-19
fix whisper with auto language detection

eaidova:
# What does this PR do?

after providing previous workaround for whisper, model starts to behave differently in preference for selection task by default.
Original transformers models have 2 places for forcing decoder ids: [generation_config](https://huggingface.co/openai/whisper-tiny/blob/main/generation_config.json#L35) and [config](https://huggingface.co/openai/whisper-tiny/blob/main/config.json#L26)

as it can see in whisper-tiny, they may be inaligned and as the result in case, when we use from_pretrained, it loads config without provided update and forces decoder ids from it.

For alignment in behavior with transformers, suggested to force decoder ids in config and keep it aligned with transcribe task as described in https://github.com/huggingface/transformers/pull/28687

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -724,10 +724,10 @@ def export_from_model(
                     setattr(model.config, param_name, None)
 
         # workaround for https://github.com/huggingface/transformers/issues/37172
-        if is_transformers_version(">=", "4.50.0") and model_type == "whisper":
-            if hasattr(model, "generation_config") and hasattr(model.generation_config, "forced_decoder_ids"):
-                model.config.forced_decoder_ids = model.generation_config.forced_decoder_ids
-                model.generation_config.forced_decoder_ids = None
+        # if is_transformers_version(">=", "4.50.0") and model_type == "whisper":
+        #     if hasattr(model, "generation_config") and hasattr(model.generation_config, "forced_decoder_ids"):
+        #         model.config.forced_decoder_ids = model.generation_config.forced_decoder_ids
+        #         model.generation_config.forced_decoder_ids = None
         # Saving the model config and preprocessor as this is needed sometimes.
         save_config(model.config, output)
         generation_config = getattr(model, "generation_config", None)
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -724,10 +724,6 @@ def export_from_model(
                     setattr(model.config, param_name, None)
 
         # workaround for https://github.com/huggingface/transformers/issues/37172
-        # if is_transformers_version(">=", "4.50.0") and model_type == "whisper":
-        #     if hasattr(model, "generation_config") and hasattr(model.generation_config, "forced_decoder_ids"):
-        #         model.config.forced_decoder_ids = model.generation_config.forced_decoder_ids
-        #         model.generation_config.forced_decoder_ids = None
         # Saving the model config and preprocessor as this is needed sometimes.
         save_config(model.config, output)
         generation_config = getattr(model, "generation_config", None)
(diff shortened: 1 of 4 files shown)


optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1340,6 +1340,13 @@ class _OVModelForWhisper(OVModelForSpeechSeq2Seq, WhisperForConditionalGeneratio
     # force the use of the WhisperForConditionalGeneration generate and prepare_inputs_for_generation methods
     generate = WhisperForConditionalGeneration.generate
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Restore multilingual model transcribe task by default after model export
+        if is_transformers_version(">=", "4.50"):
+            if self.generation_config.is_multilingual and self.generation_config.forced_decoder_ids is None:
+                self.config.forced_decoder_ids = [[1, None], [2, self.generation_config.task_to_id["transcribe"]]]
```
echarlaix, line 1348: why do we need to set `self.config.forced_decoder_ids` in this specific case ?

eaidova: there is bug in transformers that does not allow serialize and read from disk generation config with forced_decoder_ids. As The result we lost this information. however Whisper generation pipeline uses forced_decoder_ids from generation_config and if it is not provided, it checks model.config for getting them. These ids forcing which task and language model will use by default if user does not specified that (None means language is not provided and allow autodetection and transcribe as task) that is matched with default behavior. 

We modify config and generation config during export for preserving this behavior, while if model export invoked using from_pretrained, there is no config reloading.

Do you think, it will be better to reload it instead of forcing?

tests/openvino/utils_tests.py
```diff
@@ -177,7 +177,7 @@
     "wav2vec2": "anton-l/wav2vec2-random-tiny-classifier",
     "wav2vec2-hf": "hf-internal-testing/tiny-random-Wav2Vec2Model",
     "wav2vec2-conformer": "hf-internal-testing/tiny-random-wav2vec2-conformer",
-    "whisper": "optimum-internal-testing/tiny-random-whisper",
+    "whisper": "katuni4ka/tiny-random-whisper",
```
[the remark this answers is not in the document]
eaidova: that is exact goal of changes in model (I changed generation config to be more aligned with original whisper and make forced_decoder_ids different in config and generation_config

echarlaix (approved):
LGTM, thanks for the fix!

optimum/intel/openvino/modeling_seq2seq.py
echarlaix: thanks a lot @eaidova, looks good to me !

echarlaix:
thanks for the fix in https://github.com/huggingface/optimum-intel/pull/1314/commits/35a051d50cf0b7666bb750d1052108646c4c2a0f, will merge once the tests pass

optimum/exporters/openvino/convert.py
```diff
@@ -724,11 +724,10 @@ def export_from_model(
...
         # workaround for https://github.com/huggingface/transformers/issues/37172
-        if is_transformers_version(">=", "4.50.0") and model_type == "whisper":
-            if hasattr(model.config, "forced_decoder_ids"):
-                model.config.forced_decoder_ids = None
-            if hasattr(model, "generation_config") and hasattr(model.generation_config, "forced_decoder_ids"):
-                model.generation_config.forced_decoder_ids = None
+        # if is_transformers_version(">=", "4.50.0") and model_type == "whisper":
+        #     if hasattr(model, "generation_config") and hasattr(model.generation_config, "forced_decoder_ids"):
+        #         model.config.forced_decoder_ids = model.generation_config.forced_decoder_ids
+        #         model.generation_config.forced_decoder_ids = None
```
echarlaix, line 730: can be removed also 
```suggestion
```

PR #1317, opened 2025-05-22
Upgrade transformers version to 4.52 for ipex patching

jiqing-feng:
Upgrade transformers version to 4.52 for ipex patching


jiqing-feng:
@kaixuanliu . Please check if this change blocks xpu.

kaixuanliu:
LGTM

jiqing-feng:
Hi @eaidova. I upgraded the transformers version to 4.52, but OV has some incompatible issues. Can you take a look? You can either fix it or set a specific transformers version for testing like [ipex](https://github.com/huggingface/optimum-intel/blob/main/.github/workflows/test_ipex.yml#L24)

optimum/exporters/ipex/cache_utils.py
```diff
@@ -67,7 +67,7 @@ def __init__(
         self.max_cache_len = max_cache_len
         self.num_kv_heads = config.num_key_value_heads
         self.num_hidden_layers = config.num_hidden_layers
-        if hasattr(config, "head_dim"):
+        if getattr(config, "head_dim", None) is not None:
```
yao-matrix, line 70: why change this? I suppose they are equivalent.

jiqing-feng: The config has the attr `head_dim` but the value is None

setup.py
```diff
@@ -29,7 +29,7 @@
 INSTALL_REQUIRE = [
     "torch>=1.11",
     "optimum~=1.25",
-    "transformers>=4.36,<4.52",
+    "transformers>=4.36,<4.52.3",
```
echarlaix, line 32: I think we should include patch release by default so would replace with

```suggestion
    "transformers>=4.36,<4.53",
```

setup.py
```diff
@@ -68,7 +68,7 @@
     "nncf": ["nncf>=2.16.0"],
     "openvino": ["nncf>=2.16.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0"],
     "neural-compressor": ["neural-compressor[pt]>3.0", "accelerate", "transformers<4.46"],
-    "ipex": ["intel-extension-for-pytorch>=2.6", "transformers>4.49,<4.52", "accelerate"],
+    "ipex": ["intel-extension-for-pytorch>=2.6", "transformers>4.49,<4.52.3", "accelerate"],
```
echarlaix, line 71:
```suggestion
    "ipex": ["intel-extension-for-pytorch>=2.6", "transformers>4.49,<4.53", "accelerate"],
```

optimum/exporters/ipex/model_patcher.py
```diff
@@ -52,7 +52,7 @@
 
 # Please also update in the setup.py and .github/workflows/test_ipex.yml if you change the transformers version
 _TRANSFORMERS_MIN_VERSION = "4.50.0"
-_TRANSFORMERS_MAX_VERSION = "4.51.3"
+_TRANSFORMERS_MAX_VERSION = "4.52.2"
```
echarlaix, line 55: same what do you think about including patch release by default ?

.github/workflows/test_ipex.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        transformers-version: ["4.51.3"]
+        transformers-version: ["4.52.2"]
```
echarlaix, line 24: what do you think about testing multiple version of trasnformers (min and max version supported) ?

echarlaix (approved):
LGTM!

.github/workflows/test_ipex.yml
jiqing-feng: Done. I also added minimum version tests.

jiqing-feng:
Hi @echarlaix . I get your point, and set the version check to transformers 4.43. Only OV tests issue left, do you have any idea like set a fixed version of transformres for testing_ov?

setup.py
```diff
@@ -68,7 +68,7 @@
     "nncf": ["nncf>=2.16.0"],
     "openvino": ["nncf>=2.16.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0"],
```
echarlaix: adding before official support of openvino is available
```suggestion
    "openvino": ["nncf>=2.16.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0", "transformers<4.52"],
```

PR #1318, opened 2025-05-22
add activation scale factor for qwen2.5vl embeddings merger

eaidova:
# What does this PR do?

this patch adds activation scale factor hint into qwen2.5vl embeddings merger IR. This hint helps to prevent fp16 inference overflow during inference on GPU device 

## Before submitting


eaidova:
no longer required

PR #1319, opened 2025-05-22
fix openvino for compatibility with transformers 4.52

eaidova:
# What does this PR do?

Fixes # (issue)

## Before submitting


eaidova:
@echarlaix this is the first batch of fixes related to seq2seq models changes for compatibility with transformers 4.52.
Unfortunatly due to some issues with my network, Im not able to run tests locally for now, so I'll try to resolve other issues latter

optimum/exporters/openvino/model_configs.py
```diff
@@ -3695,6 +3700,63 @@ class LongT5OpenVINOConfig(T5OpenVINOConfig):
...
+    ) -> ModelPatcher:
+        if getattr(self, "stateful", False) and self._behavior == ConfigBehavior.DECODER:
+            return StatefulSeq2SeqDecoderPatcher(self, model, model_kwargs)
+        return super().patch_model_for_export(model, model_kwargs)
+
+    @property
+    def inputs(self):
+        common_inputs = super().inputs
+        if getattr(self, "stateful", False) and self._behavior == ConfigBehavior.DECODER:
+            common_inputs["decoder_input_ids"] = {0: "batch_size", 1: "decoder_sequence_length"}
```
echarlaix, line 3729: why do we need to change the original behavior for these architectures for non stateful models : https://github.com/huggingface/optimum/blob/e15053d33e60f42bb87389a869c3a9d823ea972f/optimum/exporters/onnx/config.py#L298

echarlaix (approved):
Looks good, thanks a lot for the PR @eaidova

optimum/exporters/openvino/model_configs.py
eaidova: we export model with use_past_in_inputs = True, it hard codes to export model with static sequence_len (equal 1)
https://github.com/huggingface/optimum/blob/e15053d33e60f42bb87389a869c3a9d823ea972f/optimum/exporters/onnx/config.py#L295

eaidova:
@echarlaix , I found that perciver-vision model can not properly prepare inputs. It loads Tokenizer instead of ImageProcessor in dummy input generator here https://github.com/huggingface/optimum/blob/main/optimum/utils/input_generators.py#L1649 with new transformers version, 
Do you have some solution how it should be fixed or I need to provide workaround for openvino only?

echarlaix:
> @echarlaix , I found that perciver-vision model can not properly prepare inputs. It loads Tokenizer instead of ImageProcessor in dummy input generator here https://github.com/huggingface/optimum/blob/main/optimum/utils/input_generators.py#L1649 with new transformers version and as the result instead of preparing pixel values for 224 x 224, model obtain input 56 x 56,
> Do you have some solution how it should be fixed or I need to provide workaround for openvino only?

Yes you're correct, this issue should be fixed in https://github.com/huggingface/optimum/pull/2270 (a fix can be added directly in the optimum input generator directly) so no need to add a fix in optimum-intel @eaidova thanks for suggesting.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -5156,6 +5156,108 @@ def _blenderbot_attn_forward(
...
+            # set flag that curr layer for cross-attn is already updated so we can re-use in subsequent calls
+            if is_cross_attention:
+                past_key_value.is_updated[self.layer_idx] = True
+
+    proj_shape = (bsz, self.num_heads, -1, self.head_dim)
+    query_states = query_states.reshape(*proj_shape)
+    key_states = key_states.reshape(*proj_shape)
+    value_states = value_states.reshape(*proj_shape)
+
+    attn_output = torch.nn.functional.scaled_dot_product_attention(
```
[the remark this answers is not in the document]
eaidova: done

optimum/exporters/openvino/model_patcher.py
```diff
@@ -5156,6 +5156,108 @@ def _blenderbot_attn_forward(
     return attn_output, None, past_key_value
 
 
+def _blenderbot_attn_forward_new(
```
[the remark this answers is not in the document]
eaidova: done

eaidova:
@echarlaix now all issues except discussed above perceiver-vision should be resolved. Please take a look one more time

optimum/exporters/openvino/convert.py
```diff
@@ -724,11 +724,6 @@ def export_from_model(
                     setattr(model.config, param_name, None)
 
         # workaround for https://github.com/huggingface/transformers/issues/37172
```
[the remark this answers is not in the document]
eaidova: @echarlaix  I think it is too early to remove that, fix exists only on main, but not in 4.52 https://github.com/huggingface/transformers/commit/f8630c778c9220defecf1e3026d3438108b0baba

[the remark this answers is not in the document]
echarlaix: apologies @eaidova I've already pushed the change, the comment I removed is referencing to a change that is removed in this PR so wondering why we want to keep it ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -5249,6 +5385,73 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+class Qwen2MoEPatcher(UpdateCausalMaskModelPatcher):
+    def __enter__(self):
+        super().__enter__()
+        if is_transformers_version(">=", "4.52.0"):
+            from transformers.models.qwen2_moe.modeling_qwen2_moe import Qwen2MoeSparseMoeBlock
+
+            modulewise_patch(self._model, Qwen2MoeSparseMoeBlock, _qwen2moe_sparse_block_forward)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if is_transformers_version(">=", "4.52.0"):
```
echarlaix, line 5446: failing arctic tests resulting from missing unpatching (resulting from previous model tested qwen2-moe) 
```suggestion
        super().__exit__(exc_type, exc_value, traceback)
        if is_transformers_version(">=", "4.52.0"):
```

tests/openvino/test_modeling.py
```diff
@@ -1326,9 +1326,11 @@ def test_compare_to_transformers(self, model_arch):
                 transformers_outputs = transformers_model(**tokens)
 
         # Compare tensor outputs
-        atol = 1e-3 if model_arch == "minicpm" else 1e-4
+        atol = 3e-3 if model_arch in ["minicpm", "qwen2-moe"] else 1e-4
         # quantized models have different logits value range
         if "awq" not in model_arch and "gptq" not in model_arch:
+            print(torch.abs(ov_outputs.logits - transformers_outputs.logits).max())
+            print(atol)
```
echarlaix:
```suggestion
```

tests/openvino/test_modeling.py
```diff
@@ -1976,8 +1981,8 @@ def test_compare_to_transformers(self, model_arch):
         gc.collect()
 
     @parameterized.expand(SUPPORTED_ARCHITECTURES)
-    @pytest.mark.run_slow
-    @slow
+    # @pytest.mark.run_slow
+    # @slow
```
echarlaix:
```suggestion
    @pytest.mark.run_slow
    @slow
```

echarlaix:
failing perceiver tests can be ignored (fixed in https://github.com/huggingface/optimum/pull/2270 so available in v1.26) good to merge cc @eaidova

echarlaix:
perceiver tests will be fixed in https://github.com/huggingface/optimum-intel/pull/1352

optimum/exporters/openvino/model_configs.py
```diff
@@ -1869,10 +1879,14 @@ def get_model_for_behavior(self, model, behavior: Union[str, LlavaNextVideoConfi
             behavior = LlavaNextVideoConfigBehavior(behavior)
 
         if behavior == LlavaNextVideoConfigBehavior.MULTI_MODAL_PROJECTOR:
-            return model.multi_modal_projector
+            return (
+                model.multi_modal_projector
+                if hasattr(model, "multi_model_projector")
+                else model.model.multi_modal_projector
+            )
```
mitruska, line 1886: Hi @echarlaix, looks like this change has a small typo, and "multi_mod**e**l_projector" should be "multi_mod**a**l_projector" in `hasattr` check.

The following error is raised for transformers==4.51.3 
```"AttributeError: 'LlavaNextVideoForConditionalGeneration' object has no attribute 'model'"```

I've prepared PR with fix: 
- https://github.com/huggingface/optimum-intel/pull/1375

PR #1339, opened 2025-06-12
Fix setup.py to support INC latest version 3.4.1

changwangss:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ QUALITY_REQUIRE = ["black~=23.1", "ruff==0.4.4"]
 EXTRAS_REQUIRE = {
     "nncf": ["nncf>=2.16.0"],
     "openvino": ["nncf>=2.16.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0"],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -67,7 +67,7 @@
 EXTRAS_REQUIRE = {
     "nncf": ["nncf>=2.16.0"],
     "openvino": ["nncf>=2.16.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0"],
-    "neural-compressor": ["neural-compressor[pt]>3.0,<=3.3.1", "accelerate", "transformers<4.46"],
+    "neural-compressor": ["neural-compressor[pt]>3.0,!=3.4.0", "accelerate", "transformers<4.46"],
```
IlyasMoutawwakil: why not >=3.4.1 ?

changwangss: INC 3.1, 3.2, 3.3 also works, only 3.4 has issue. if you prefer the latest, I can improve it to >=3.4.1.

IlyasMoutawwakil: we only test the latest so good thing to update it

PR #1349, opened 2025-06-16
[OV] Add support for data-free AWQ

nikita-savelyevv:
# What does this PR do?

Updated tests and docs to account for the newly added data-free AWQ.

## Before submitting


nikita-savelyevv:
Waiting for NNCF 2.17 release to happen to mark this PR as ready for review.

docs/source/openvino/optimization.mdx
```diff
@@ -507,7 +507,7 @@ For 4-bit quantization you can also specify the following arguments in the quant
 Smaller `group_size` and `ratio` values usually improve accuracy at the sacrifice of the model size and inference latency.
 
 Quality of 4-bit weight compressed model can further be improved by employing one of the following data-dependent methods:
-* **AWQ** which stands for Activation Aware Quantization is an algorithm that tunes model weights for more accurate 4-bit compression. It slightly improves generation quality of compressed LLMs, but requires significant additional time and memory for tuning weights on a calibration dataset. Please note that it is possible that there will be no matching patterns in the model to apply AWQ, in such case it will be skipped.
+* **AWQ** which stands for Activation Aware Quantization is an algorithm that tunes model weights for more accurate 4-bit compression. It slightly improves generation quality of compressed LLMs, but requires significant additional time and memory for tuning weights on a calibration dataset. Please note that it is possible that there will be no matching patterns in the model to apply AWQ, in such case it will be skipped. There is also a data-free version of AWQ available.
```
echarlaix: can you extend on how does the data free awq strategy works in the documentation ? Also should we specify that this is available on with nncf >= 2.17

echarlaix (approved):
LGTM! Thanks a lot @nikita-savelyevv

docs/source/openvino/optimization.mdx
nikita-savelyevv: @echarlaix thank you for the suggestions! 

> can you extend on how does the data free awq strategy works in the documentation ?

I've added a note about that.

> Also should we specify that this is available on with nncf >= 2.17 

I think it's ok to assume the latest version of nncf in the documentation. Otherwise there will be a lot of such notes. Instead I've added an explicit error message for those who will try to run data-free AWQ with NNCF < 2.17. What do you think?

echarlaix: thanks a lot! yes that makes sense

PR #1355, opened 2025-06-20
Compression config for Qwen/Qwen2.5-Coder-3B-Instruct

MaximProshin:
# What does this PR do?

Improves the similarity of the compressed int4 model from 0.405464 to 0.903016 (0.897232 with GenAI) on CPU.

Fixes #169307


MaximProshin:
@nikita-savelyevv , please take a look.

PR #1359, opened 2025-06-25
[OV] Add quantization for text2text-generation models

nikita-savelyevv:
# What does this PR do?

Added post training quantization support for text2text-generation models.

Example metrics are below. Accuracy is computed with https://github.com/EleutherAI/lm-evaluation-harness/pull/3101 on HellaSwag [dataset](https://github.com/EleutherAI/lm-evaluation-harness/tree/main/lm_eval/tasks/hellaswag).

| Model ID           | FP32 Accuracy | INT8 WC Accuracy | INT8 PTQ Accuracy | INT8 WC Speedup | INT8 PTQ Speedup |
|--------------------|---------------|------------------|-------------------|------------------|-------------------|
| google-t5/t5-small |    26.72%     |   26.78%         |    26.96%        | 1.06x            | 1.14x             |
| google-t5/t5-large |    34.84%    |   34.82%         |    34.80%        | 1.70x            | 1.81x             |
| google-t5/t5-3b |    38.52%     |   38.41%         |    38.79%        | 2.46x            | 2.38x             |

WC stands for weight compression, PTQ stands for Post-Training-Quantization.

Ticket 165770

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_seq2seq.py
+++ b/optimum/intel/openvino/modeling_seq2seq.py
@@ -1377,6 +1377,15 @@ class _OVModelForWhisper(OVModelForSpeechSeq2Seq, WhisperForConditionalGeneratio
     # force the use of the WhisperForConditionalGeneration generate and prepare_inputs_for_generation methods
     generate = WhisperForConditionalGeneration.generate
 
+    @classmethod
+    def _from_pretrained(
+        cls,
+        model_id: Union[str, Path],
+        config: "PretrainedConfig",
+        **kwargs,
+    ):
+        return super(OVModelForSpeechSeq2Seq, cls)._from_pretrained(model_id, config, **kwargs)
+
     class DummyWhisperModel:
         def __init__(self):
             self.encoder = self.Encoder()


optimum/commands/export/openvino.py
```diff
@@ -491,6 +492,10 @@ def run(self):
                 from optimum.intel import OVModelForCausalLM
 
                 model_cls = OVModelForCausalLM
+            elif task.startswith("text2text-generation"):
+                from optimum.intel import OVModelForSeq2SeqLM
+
+                model_cls = OVModelForSeq2SeqLM
```
rkazants, line 498: why do we need it?

nikita-savelyevv: This is our approach to apply data-aware quantization during export from CLI. This differs from data-free approach applied during export by the fact that in order to collect quantization calibration data, we require a created instance of the model class. In this case this is an instance of `OVModelForSeq2SeqLM`.

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -786,6 +801,27 @@ def compile(self):
...
             getattr(self, submodel_name)._compile()
 
+    def _shift_right(self, input_ids):
+        decoder_start_token_id = self.config.decoder_start_token_id
+        pad_token_id = self.config.pad_token_id
+
+        if decoder_start_token_id is None:
+            raise ValueError(
+                "self.model.config.decoder_start_token_id has to be defined. In T5 it is usually set to the pad_token_id. "
+                "See T5 docs for more information."
```
rkazants, line 812: may be it makes sense to give a link to useful documentation

optimum/intel/openvino/quantization.py
```diff
@@ -770,6 +777,56 @@ def _prepare_speech_to_text_calibration_data(
...
+            num_samples = config.num_samples or 128
+            dataset = list(tqdm(dataset.take(num_samples), desc="Downloading dataset", total=num_samples))
+
+            tokenizer = None
+            for item in tqdm(dataset, desc="Collecting calibration data"):
+                if "input_ids" in item:
+                    # Assuming that dataset contains already preprocessed text
+                    inputs = self._wrap_sample_as_array(item, add_batch_dim=True)
+                else:
+                    tokenizer = tokenizer or get_tokenizer()
```
rkazants, line 817: is it fine to get a tokenizer each iteration? Can we get it once before for-loop?

optimum/intel/openvino/modeling_seq2seq.py
nikita-savelyevv: Thanks for the suggestion. I've added a link to the original code location.

optimum/intel/openvino/quantization.py
nikita-savelyevv: `tokenizer` variable is set on the first iteration and starting from the second iteration `get_tokenizer()` won't be called.

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1345,27 +1382,9 @@ def _from_pretrained(
...
-            quantization_config_copy = copy.deepcopy(quantization_config)
-            quantization_config_copy.processor = quantization_config.processor or model_id
-            OVQuantizer(model).quantize(ov_config=OVConfig(quantization_config=quantization_config_copy))
-        else:
-            model = super(OVModelForSpeechSeq2Seq, cls)._from_pretrained(
-                model_id, config, load_in_8bit=load_in_8bit, quantization_config=quantization_config, **kwargs
-            )
-
-        return model
+        return super(OVModelForSpeechSeq2Seq, cls)._from_pretrained(model_id, config, **kwargs)
```
echarlaix, line 1387: with this update, no need to overwrite `_from_pretrained` here so we can remove no ?

nikita-savelyevv: Seems like yes. I've just reverted it to the state before introduction of quantization for whisper pipelines: https://github.com/huggingface/optimum-intel/pull/1040 . For some reason, it was defined like this before.

echarlaix: apologies @nikita-savelyevv looks like it was indeed needed, likely coming from https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_seq2seq.py#L1321 
reverted in https://github.com/huggingface/optimum-intel/pull/1359/commits/4a1802831252154ea20823d651f5acfe2adc0f5e

IlyasMoutawwakil: it's needed because the whisper class inherits from OVModelForSpeechSeq2Seq so if not overwritten the OVModelForSpeechSeq2Seq keeps calling is own method recursively

PR #1329, opened 2025-05-27
Original model types

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1360, opened 2025-06-25
[OpenVINO] Add support for Mamba and Falcon-mamba

rkazants:
# What does this PR do?

Add support for Mamba and Falcon-mamba

Fixes # (issue)

## Before submitting


optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4368,144 @@ def patch_model_for_export(
         if self._behavior != VLMConfigBehavior.VISION_EMBEDDINGS:
             return super().patch_model_for_export(model, model_kwargs)
         return Llama4ImageEmbeddingsModelPatcher(self, model, model_kwargs)
+
+
```
rkazants, line 4369:
```suggestion
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4368,144 @@ def patch_model_for_export(
...
+@register_in_tasks_manager(
+    "falcon-mamba", *["text-generation", "text-generation-with-past"], library_name="transformers"
+)
+class FalconMambaOpenVINOConfig(MambaOpenVINOConfig):
+    def patch_model_for_export(
+        self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ):
+
```
```suggestion
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,411 @@ def __exit__(self, exc_type, exc_value, traceback):
             if layer.is_moe_layer:
                 layer.feed_forward.forward = layer.feed_forward._orig_forward
             layer.self_attn.forward = layer.self_attn._orig_forward
+
+
```
rkazants, line 6460:
```suggestion
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,411 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            layer.mixer.forward = types.MethodType(falcon_mamba_mixer_forward, layer.mixer)
+            conv_transform = ConvSequenceTransform(
+                layer.mixer.conv_kernel_size,
+                layer.mixer.use_conv_bias,
+                layer.mixer.conv1d,
+                layer.mixer.act,
+                layer.mixer.conv1d.bias,
+            )
+            layer.mixer.conv_sequence_transform = torch.jit.script(conv_transform)
+
```
```suggestion
```

optimum/exporters/openvino/model_patcher.py
rkazants, line 6460:
```suggestion
```

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -100,6 +107,15 @@
 # during dataset preparatioon nncf checks isinstance(model, PreTrainedModel.__bases__)
 # in transformers 4.52.0 PreTrainedModel does not include GenerationMixin and this check failed for OVModelForCausalLM
 # TO DO: remove it after migration on new nncf
+def has_cache_inputs(model):
+    return any(
+        "past_key_values" in key.get_any_name()
+        or "past_ssm" in key.get_any_name()
+        or "past_conv" in key.get_any_name()
+        for key in model.inputs
+    )
```
nikita-savelyevv: Do we really need a separate function for this? I see that there is only one usage.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -144,13 +160,17 @@ def __init__(
...
         model_has_sinks = model_has_state(self.model)
-        self.use_cache = any("past_key_values" in key.get_any_name() for key in model.inputs) or model_has_sinks
+        self.use_cache = has_cache_inputs(model) or model_has_sinks
         stateful = kwargs.pop("stateful", None)  # stateful model only if it is converted with stateful=True
         self.stateful = model_has_sinks
         self.main_input_name = "input_ids"
         self.num_pkv = 2
         self.key_value_input_names = [key for key in self.input_names if "key_values" in key]
-        self.key_value_output_names = [key for key in self.output_names if "present" in key]
+        self.key_value_output_names = [key for key in self.output_names if "present." in key]
```
nikita-savelyevv: Why it is "present." instead of the previous "present"?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -144,13 +160,17 @@ def __init__(
...
+        self.ssm_cache_input_names = [key for key in self.input_names if "past_ssm_states" in key]
+        self.conv_cache_input_names = [key for key in self.input_names if "past_conv_states" in key]
+        self.ssm_cache_output_names = [key for key in self.output_names if "present_ssm_states" in key]
+        self.conv_cache_output_names = [key for key in self.output_names if "present_conv_states" in key]
```
nikita-savelyevv: Perhaps it would be better to define these only for OVMambaForCausalLM? That is, override the __init__ method for this class. As I understand these fields are not used by any other model type.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -1031,3 +1053,188 @@ def _reorder_cache(
...
+                self.ssm_states.append(ssm_state)
+
+
+@dataclass
+class MambaOutput(ModelOutput):
+    """
+    Class for the MAMBA model outputs.
+
+    Args:
+        last_hidden_state (`torch.FloatTensor` of shape `(batch_size, sequence_length, hidden_size)`):
```
nikita-savelyevv:
```suggestion
        logits (`torch.FloatTensor` of shape `(batch_size, sequence_length, hidden_size)`):
```

optimum/exporters/openvino/stateful.py
```diff
@@ -272,9 +272,39 @@ def insert_state_for_nodes(model: ov.Model, nodes):
         model.add_sinks([assign])
 
 
+def patch_stateful_ssm(config, ov_model):
```
nikita-savelyevv: `config` argument is not used

optimum/exporters/openvino/stateful.py
```diff
@@ -272,9 +272,39 @@ def insert_state_for_nodes(model: ov.Model, nodes):
...
+    cache_input_names = [key_name for key in ov_model.inputs for key_name in key.get_names() if "past_" in key_name]
+    cache_output_names = [
+        key_name for key in ov_model.outputs for key_name in key.get_names() if "present" in key_name
+    ]
+
+    print(cache_output_names)
+    print(ov_model.outputs)
```
nikita-savelyevv:
```suggestion
```

optimum/exporters/openvino/stateful.py
```diff
@@ -272,9 +272,39 @@ def insert_state_for_nodes(model: ov.Model, nodes):
...
+
+    batch_dim = 0
+
+    from openvino._offline_transformations import apply_make_stateful_transformation
+
+    input_output_map = {}
+    for cache_name_pair in zip(cache_input_names, cache_output_names):
+        input_output_map[cache_name_pair[0]] = cache_name_pair[1]
+
+    print(input_output_map)
```
nikita-savelyevv:
```suggestion
```

optimum/exporters/openvino/stateful.py
```diff
@@ -272,9 +272,39 @@ def insert_state_for_nodes(model: ov.Model, nodes):
...
+
+    apply_make_stateful_transformation(ov_model, input_output_map)
+    build_state_initializer(ov_model, batch_dim)
+
+
+SSM_MODELS = ["mamba", "falcon-mamba"]
```
nikita-savelyevv: I see that this variable is defined twice: here and at optimum/exporters/openvino/utils.py . Can we keep it only at utils.py?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -1031,3 +1053,188 @@ def _reorder_cache(
...
+
+            Hidden-states of the model at the output of each layer plus the optional initial embedding outputs.
+    """
+
+    logits: Optional[torch.FloatTensor] = None
+    cache_params: Optional[OVMambaCache] = None
+    hidden_states: Optional[Tuple[torch.FloatTensor]] = None
+
+
+class OVMambaForCausalLM(OVModelForCausalLM):
```
nikita-savelyevv, line 1147: Some comment description of what this model type is and what is its major difference from a regular OVModelForCausalLM would be helpful.

tests/openvino/test_modeling.py
```diff
@@ -1328,7 +1336,7 @@ def test_compare_to_transformers(self, model_arch):
         # Compare tensor outputs
         atol = 3e-3 if model_arch in ["minicpm", "qwen2-moe"] else 1e-4
         # quantized models have different logits value range
-        if "awq" not in model_arch and "gptq" not in model_arch:
```
nikita-savelyevv: Why don't we expect the logits to be equal for SSM models?

tests/openvino/utils_tests.py
```diff
@@ -103,6 +103,7 @@
     "opt": "hf-internal-testing/tiny-random-OPTModel",
     "opt125m": "facebook/opt-125m",
     "opt_gptq": "ybelkada/opt-125m-gptq-4bit",
+    "mamba": "rkazants/tiny-mamba",
```
nikita-savelyevv, line 107: I see that the test model has 159M parameters. Is it possible to make it smaller? We're trying to conserve disk size of GHA test hosts.

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4368,144 @@ def patch_model_for_export(
...
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, MambaCacheDummyInputGenerator)
+    DUMMY_PKV_GENERATOR_CLASS = MambaCacheDummyInputGenerator
+    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
+
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        if self.use_past_in_inputs:
+            common_inputs = {"input_ids": {0: "batch_size", 1: "sequence_length"}}
+            self.add_past_key_values(common_inputs, direction="inputs")
+            # common_inputs["attention_mask"] = {0: "batch_size", 1: "past_sequence_length + 1"}
```
nikita-savelyevv: Is this commented line needed?

optimum/intel/openvino/modeling_decoder.py
rkazants: Fixed.

optimum/intel/openvino/modeling_decoder.py
rkazants: Fixed.

optimum/exporters/openvino/stateful.py
rkazants: Removed

optimum/exporters/openvino/stateful.py
rkazants: Removed.

optimum/exporters/openvino/stateful.py
rkazants: Done. Use from utils

optimum/intel/openvino/modeling_decoder.py
rkazants: Added description

optimum/exporters/openvino/stateful.py
rkazants: Done.

optimum/exporters/openvino/model_configs.py
rkazants: Removed

tests/openvino/utils_tests.py
rkazants: Added smaller one

docs/source/openvino/models.mdx
```diff
@@ -52,6 +52,7 @@ Here is the list of the supported architectures :
 - ESM
 - Exaone
 - Falcon
+- Falcon-mamba
```
```suggestion
- Falcon-Mamba
```

optimum/intel/openvino/modeling_decoder.py
rkazants: Removed

optimum/intel/openvino/modeling_decoder.py
rkazants: Done

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,379 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+                conv_states: Optional[List[torch.Tensor]] = None,
+                ssm_states: Optional[List[torch.Tensor]] = None,
+            ):
+                self.dtype = dtype
+                self.max_batch_size = batch_size or max_batch_size
+                self.intermediate_size = config.intermediate_size
+                self.ssm_state_size = config.state_size
+                self.conv_kernel_size = config.conv_kernel
+                self.device = torch.device(device) if device is not None else torch.device("cpu")
+                # print(config.num_hidden_layers)
```
```suggestion
```

tests/openvino/utils_tests.py
echarlaix, line 107: would you mind adding a test for falcon mamba as well like for example https://huggingface.co/trl-internal-testing/tiny-FalconMambaForCausalLM

echarlaix: also couple can be used from the [hf-internal-testing](https://huggingface.co/hf-internal-testing) org like  https://huggingface.co/hf-internal-testing/tiny-random-MambaForCausalLM if needed

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4368,142 @@ def patch_model_for_export(
...
+                    f'Could not generate dummy input for "{input_name}". Try adding a proper dummy input generator to the model ONNX config.'
+                )
+
+        return dummy_inputs
+
+
+@register_in_tasks_manager(
+    "falcon-mamba", *["text-generation", "text-generation-with-past"], library_name="transformers"
+)
+class FalconMambaOpenVINOConfig(MambaOpenVINOConfig):
```
echarlaix: would you mind adding min transformers version

```suggestion
class FalconMambaOpenVINOConfig(MambaOpenVINOConfig):
    MIN_TRANSFORMERS_VERSION = version.parse("4.45.0")
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4368,142 @@ def patch_model_for_export(
...
+            )
+
+        raise ValueError(f"Unsupported input name {input_name}")
+
+
+@register_in_tasks_manager("mamba", *["text-generation", "text-generation-with-past"], library_name="transformers")
+class MambaOpenVINOConfig(TextDecoderOnnxConfig):
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, MambaCacheDummyInputGenerator)
+    DUMMY_PKV_GENERATOR_CLASS = MambaCacheDummyInputGenerator
+    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
```
echarlaix, line 4421: same can also be added
```suggestion
    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
    MIN_TRANSFORMERS_VERSION = version.parse("4.43.0")
```

(setting 4.43 as MambaCache was introduced there)

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -1031,3 +1040,235 @@ def _reorder_cache(
...
+                raise ValueError(
+                    "`cache_position` should not be None as it should have been initialized in "
+                    "`model.generate`, you are responsible for passing in a valid `cache_position` if "
+                    "you are calling `prepare_inputs_for_generation` directly with `use_cache=True`"
+                )
+            if cache_position[0] > 0:
+                input_ids = input_ids[:, -1].unsqueeze(-1)
+
+                if attention_mask is not None:
+                    attention_mask = None
```
echarlaix: why is it needed?

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4368,142 @@ def patch_model_for_export(
...
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        if self.use_past_in_inputs:
+            common_inputs = {"input_ids": {0: "batch_size", 1: "sequence_length"}}
+            self.add_past_key_values(common_inputs, direction="inputs")
+            common_inputs["cache_position"] = {0: "cache_sequence_length"}
+        else:
+            common_inputs = {
+                "input_ids": {0: "batch_size", 1: "sequence_length"},
+                "cache_position": {0: "cache_sequence_length"},
+            }
```
echarlaix: why not instead ?
```suggestion
        common_inputs = {
            "input_ids": {0: "batch_size", 1: "sequence_length"},
            "cache_position": {0: "cache_sequence_length"},
        }
        if self.use_past_in_inputs:
            self.add_past_key_values(common_inputs, direction="inputs")
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4368,142 @@ def patch_model_for_export(
...
+        self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ):
+        return MambaPatcher(self, model, model_kwargs)
+
+    def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
+        dummy_inputs_generators = self._create_dummy_input_generator_classes(**kwargs)
+
+        dummy_inputs = {}
+        input_names = [key for key in self.inputs.keys() if not key.startswith("past_")]
+        if self.use_past_in_inputs and self.use_cache_branch is not False:
```
echarlaix: shouldn't we ignore `self.use_cache_branch`here (I imagine that the case for which it's set to True is not compatible with the openvino export)

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4368,142 @@ def patch_model_for_export(
...
+            conv_name = "past_conv_states"
+        else:
+            ssm_name = "present_ssm_states"
+            conv_name = "present_conv_states"
+
+        for i in range(self._normalized_config.num_layers):
+            # [batch_size, d_state, d_model]
+            inputs_or_outputs[f"{ssm_name}.{i}"] = {0: "batch_size"}
+
+        for i in range(self._normalized_config.num_layers):
```
echarlaix: why not do everything in one loop instead ?
```suggestion
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,378 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        dA = torch.einsum("bld,dn->bldn", dt, A)
+        dB_u = torch.einsum("bld,bld,bln->bldn", dt, u, B)
+        dA_cumsum = torch.nn.functional.pad(dA[:, 1:], (0, 0, 0, 0, 0, 1)).flip(1).cumsum(1).exp().flip(1)
+        x = dB_u * dA_cumsum + (ssm.unsqueeze(1) * dA[:, :1].exp())
+        x = x.cumsum(1) / (dA_cumsum + 1e-12)
+        y = torch.einsum("bldn,bln->bld", x, C)
+        return y + u * D, x[:, -1, :, :]
+
+
+def mamba_mixer_forward(
```
echarlaix, line 6541: would you mind adding a link to the original code as well as highlight which part was modified and why it's needed)
https://github.com/huggingface/transformers/blob/v4.53.0/src/transformers/models/mamba/modeling_mamba.py#L233

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,378 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+    if cache_params is not None:
+        cache_params.ssm_states[self.layer_idx].copy_(ssm_state)
+
+    # 4. Final linear projection
+    contextualized_states = self.out_proj(scan_output.transpose(1, 2))  # [batch, seq_len, hidden_size]
+    return contextualized_states
+
+
+def falcon_mamba_mixer_forward(
```
echarlaix: same if possible (link + highlight differences

optimum/exporters/openvino/stateful.py
```diff
@@ -272,9 +272,32 @@ def insert_state_for_nodes(model: ov.Model, nodes):
         model.add_sinks([assign])
 
 
+def patch_stateful_ssm(ov_model):
```
echarlaix:
```suggestion
def patch_stateful_ssm(ov_model: ov.Model):
```

tests/openvino/test_modeling.py
```diff
@@ -1134,6 +1134,10 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "decilm",
     )
 
+    SUPPORTED_SSM_ARCHITECTURES = ("mamba",)
+    if is_transformers_version(">=", "4.39"):
+        SUPPORTED_ARCHITECTURES += SUPPORTED_SSM_ARCHITECTURES
+
```
echarlaix, line 1140: question : are we testing all cases in addition of stateful mode, if not would you mind adding a test for this as well to make sure all different cases are supported ? other possibility : could also make sense to only support stateful mode for newly added architectures so we could also restrict the export to stateful mode only

echarlaix (commented):
Thanks a lot for the new addition @rkazants

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4368,141 @@ def patch_model_for_export(
...
+        self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ):
+        return MambaPatcher(self, model, model_kwargs)
+
+    def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
+        dummy_inputs_generators = self._create_dummy_input_generator_classes(**kwargs)
+
+        dummy_inputs = {}
+        input_names = [key for key in self.inputs.keys() if not key.startswith("past_")]
+        if self.use_past_in_inputs and self.use_cache_branch is not False:
```
```suggestion
        if self.use_past_in_inputs:
```

optimum/exporters/openvino/model_configs.py
rkazants: Removed

tests/openvino/test_modeling.py
rkazants: yes, let us support only stateful mode, i.e. `text-generation-with-past`

tests/openvino/utils_tests.py
rkazants: Added

tests/openvino/test_modeling.py
rkazants: Now we have a problem with unstable computation for tiny mamba models that leads to nan values in output. For those two real models, outputs match the reference.

nikita-savelyevv: Is the exact reason behind nan values understood? Can the tiny model be changed to avoid nans? My concern is that without comparing ov outputs to the reference ones, at some point we risk missing the fact that they started to differ for real models too.

tests/openvino/test_modeling.py
```diff
@@ -1264,6 +1268,7 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "glm4": 2,
         "qwen3": 2,
         "qwen3-moe": 2,
+        "mamba": 0,
```
rkazants, line 1271:
```suggestion
        "mamba": 0,
        "falcon-mamba": 0,
```

tests/openvino/test_modeling.py
echarlaix: then would you mind updating the supported task from `MambaOpenVINOConfig` (removing text-generation support), as well as `OVMambaForCausalLM` (we can add a check there that the loaded model is stateful) + a check during export to make sure the model is always exported that way (so removing the case where the decoder is exported with use_cache=False + use_cache=False and stateful=False). We could also think of only adding support of stateful models from now on for new architectures

tests/openvino/utils_tests.py
echarlaix: thanks ! just wondering why not using the models from the [hf-internal-testing](https://huggingface.co/hf-internal-testing) org, is there some modifications that was needed to add or just because it's easier for you that way?

tests/openvino/test_modeling.py
```diff
@@ -1554,6 +1567,9 @@ def test_default_filling_attention_mask_and_position_ids(self):
     @pytest.mark.run_slow
     @slow
     def test_beam_search(self, model_arch):
+        if model_arch in self.SUPPORTED_SSM_ARCHITECTURES:
+            pytest.skip(f"Skipping unsupported architecture: {model_arch}")
```
echarlaix: why ?

tests/openvino/test_modeling.py
```diff
@@ -1359,6 +1368,10 @@ def test_compare_to_transformers(self, model_arch):
 
         ov_outputs = ov_model.generate(**tokens, generation_config=gen_config)
 
+        if model_arch in self.SUPPORTED_SSM_ARCHITECTURES:
+            # does not match reference output due to unstable computation for tiny mamba models
+            return
```
echarlaix: could you describe in more detail what is the exact issue here ? any possibility this comes from the export phase or directly from `OVMambaForCausalLM` ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,306 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+class SelectiveScan(torch.nn.Module):
+    def forward(self, ssm, u, dt, A, B, C, D):
+        # dt - [batch, seq_len, intermediate_size]
+        # A - [intermediate_size, ssm_state_size]
+        # B [batch, seq_len, ssm_state_size]
+        # u, hidden states - [batch, seq_len, intermediate_size]
+        dA = torch.einsum("bld,dn->bldn", dt, A)
+        dB_u = torch.einsum("bld,bld,bln->bldn", dt, u, B)
+        dA_cumsum = torch.nn.functional.pad(dA[:, 1:], (0, 0, 0, 0, 0, 1)).flip(1).cumsum(1).exp().flip(1)
+        x = dB_u * dA_cumsum + (ssm.unsqueeze(1) * dA[:, :1].exp())
```
rkazants, line 6531:
```suggestion
        x = dB_u * dA_cumsum + (ssm.unsqueeze(1) * dA[:, :1].exp())
        x = x.cumsum(1) / (dA_cumsum + 1e-12)
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4367,139 @@ def patch_model_for_export(
...
+
+    def patch_model_for_export(
+        self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ):
+        model_type = getattr(model.config, "model_type", None)
+        ssm_rms_normalization = None
+        if model_type == "falcon_mamba":
+            from transformers.models.falcon_mamba.modeling_falcon_mamba import rms_forward
+
+            ssm_rms_normalization = rms_forward
```
IlyasMoutawwakil: this check could be done directly in `ModelPatcher.__init__`, we should avoid segmenting the patching code.

optimum/exporters/openvino/model_configs.py
```diff
@@ -4366,3 +4367,139 @@ def patch_model_for_export(
...
+
+        return MambaPatcher(self, model, model_kwargs, ssm_rms_normalization)
+
+    def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
```
IlyasMoutawwakil, line 4465: there's no need to override this method if you have correct `self.inputs`, which is the case since you do check `if self.use_past_in_inputs:` there

optimum/exporters/openvino/model_patcher.py
rkazants: Done

optimum/exporters/openvino/model_configs.py
rkazants: Fixed.

optimum/exporters/openvino/model_configs.py
rkazants: yes, we need to override this method. That is because mamba model has a different structure of states (ssm_states and conv_state) in contrast to transformers. It prepares input names in a different name before calling special generator for mamba past states values.

I added a comment about this.

optimum/exporters/openvino/model_configs.py
IlyasMoutawwakil: you moved it to the model config, not the `ModelPatcher`

rkazants: Yeah, sorry for misunderstanding. Now it is moved there.

tests/openvino/test_modeling.py
rkazants: let us discuss in below comment from Ella. Now I am trying to fix and have some progress.

tests/openvino/test_modeling.py
IlyasMoutawwakil: @rkazants

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -1036,3 +1045,235 @@ def _reorder_cache(
...
+        **kwargs,
+    ):
+        inputs = {"input_ids": input_ids.cpu().numpy()}
+        if "cache_position" in self.input_names:
+            if cache_position is None:
+                # initialize it as for prefill stage
+                cache_position = torch.arange(0, self.config.conv_kernel, device=input_ids.device)
+            inputs["cache_position"] = cache_position.cpu().numpy()
+        if "attention_mask" in self.input_names:
+            inputs["attention_mask"] = cache_position.cpu().numpy()
```
IlyasMoutawwakil: why is the numpy() or cpu() needed here ? we usually don't do that for other models because ov accepts torch tensors anyways.

tests/openvino/test_modeling.py
rkazants: @nikita-savelyevv, managed to fix this skip. Avoiding jit scripting for ConvSentenceTransform helps to fix it.

tests/openvino/utils_tests.py
```diff
@@ -318,6 +320,8 @@
         "vocoder": 80,
     },
     "clip": {"model": 130},
+    "mamba": {"model": 258},
+    "falcon-mamba": {"model": 258},
```
```suggestion
    "mamba": {"model": 322},
    "falcon-mamba": {"model": 322},
```

optimum/intel/openvino/modeling_decoder.py
rkazants: Fixed. Indeed, torch is handled by OV.

optimum/exporters/openvino/stateful.py
```diff
@@ -272,9 +272,34 @@ def insert_state_for_nodes(model: ov.Model, nodes):
...
+        input_output_map[cache_name_pair[0]] = cache_name_pair[1]
+
+    apply_make_stateful_transformation(ov_model, input_output_map)
+    build_state_initializer(ov_model, batch_dim)
+
+
 def patch_stateful(config: PretrainedConfig, ov_model: ov.Model):
     if config.is_encoder_decoder and model_has_input_output_name(ov_model, "encoder_hidden_states"):
         return patch_stateful_encoder_decoder(config, ov_model)
+    if config.model_type.replace("_", "-") in SSM_MODELS:
```
IlyasMoutawwakil: the `replace("_", "-")` is no longer needed since #1329

rkazants: Hi @IlyasMoutawwakil, I will fix. Good catch!

Could you please do the full review for PR? Just avoiding the single comments every day:)
Thanks in advance!

IlyasMoutawwakil: Hi, normally it's good for me (except for the two comments from Ella that I mentioned you in)
I only re-checked the code because you made new changes after the last review :)

IlyasMoutawwakil: I will leave an LGTM once all tests are passing (not skipped)

rkazants: the replace("_", "-") was fixed. Now I am working on these two skips.

optimum/intel/openvino/modeling_decoder.py
rkazants: replace with a concrete value of attention_mask for decoding step. It must be always a mask of ones for decoding step since this is Mamba model.

optimum/exporters/openvino/stateful.py
rkazants: @IlyasMoutawwakil, it turned out that replacement of "_" in model type is still needed because of this replacement:
https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_decoder.py#L856

and PR you listed is not merged.

tests/openvino/test_modeling.py
rkazants: Fixed.

tests/openvino/test_modeling.py
rkazants: Fixed.

tests/openvino/utils_tests.py
```diff
@@ -318,6 +320,8 @@
         "vocoder": 80,
     },
     "clip": {"model": 130},
+    "mamba": {"model": 322},
+    "falcon-mamba": {"model": 322},
```
```suggestion
    "mamba": {"model": 386},
    "falcon-mamba": {"model": 386},
```

nikita-savelyevv (approved):
LGTM!

tests/openvino/utils_tests.py
```diff
@@ -318,6 +320,8 @@
         "vocoder": 80,
     },
     "clip": {"model": 130},
+    "mamba": {"model": 386},
+    "falcon-mamba": {"model": 386},
```
```suggestion
    "falcon-mamba": {"model": 194},
```

tests/openvino/test_modeling.py
```diff
@@ -1296,12 +1302,15 @@ def test_compare_to_transformers(self, model_arch):
...
         self.assertTrue("logits" in ov_outputs)
         self.assertIsInstance(ov_outputs.logits, torch.Tensor)
-        self.assertTrue("past_key_values" in ov_outputs)
-        self.assertIsInstance(ov_outputs.past_key_values, tuple)
-        is_stateful = ov_model.config.model_type not in not_stateful
-        self.assertEqual(ov_model.stateful, is_stateful)
-        if is_stateful:
-            self.assertTrue(len(ov_outputs.past_key_values) == 1 and len(ov_outputs.past_key_values[0]) == 0)
+        if model_arch in self.SUPPORTED_SSM_ARCHITECTURES:
+            self.assertTrue("cache_params" in ov_outputs)
```
IlyasMoutawwakil, line 1308: shouldn't we do the same checks in the else branch here as well ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,312 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            ) -> torch.Tensor:
+                conv_state = self.conv_states[layer_idx]
+                cache_position = cache_position.clamp(0, self.conv_kernel_size - 1)
+
+                conv_state = conv_state.roll(shifts=-1, dims=-1)
+                conv_state = conv_state.scatter(
+                    2, cache_position.expand(conv_state.shape[0], conv_state.shape[1], -1), new_conv_state
+                )
+                self.conv_states[layer_idx] = conv_state
+                return self.conv_states[layer_idx]
```
IlyasMoutawwakil: please add a comment about why we override this method

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,312 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        model: Union["PreTrainedModel", "TFPreTrainedModel"],
+        model_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        from transformers import PretrainedConfig
+        from transformers.cache_utils import MambaCache
+
+        super().__init__(config, model, model_kwargs)
+
+        class MambaCacheWrap(MambaCache):
+            def __init__(
```
IlyasMoutawwakil, line 6630: just a nit, but there is actually a method that's usually used to create Caches from legacy format, 
it's not implemented for MambaCache, but could be better to implement it rather than init to avoid rewriting init
https://github.com/huggingface/transformers/blob/0b724114cf8475f146ca2fd644c4e31f395441eb/src/transformers/cache_utils.py#L569C1-L580C21

IlyasMoutawwakil (approved):
LGTM thanks for iterating on this ! left one comment and a nit suggestion

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,312 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        dB_u = dB_u.unsqueeze(-1) * B.unsqueeze(2)  # (b, l, d, 1) * (b, l, 1, n) => (b, l, d, n)
+        dA_cumsum = torch.nn.functional.pad(dA[:, 1:], (0, 0, 0, 0, 0, 1)).flip(1).cumsum(1).exp().flip(1)
+        x = dB_u * dA_cumsum + (ssm.unsqueeze(1) * dA[:, :1].exp())
+        x = x.cumsum(1) / (dA_cumsum + 1e-12)
+        y = (x * C.unsqueeze(2)).sum(dim=-1)
+        return y + u * D, x[:, -1, :, :]
+
+
+# The original implementation of this forward method can be found at:
+# https://github.com/huggingface/transformers/blob/main/src/transformers/models/mamba/modeling_mamba.py#L233
```
echarlaix: easier to use a specific version as main will be updated (line number will not correspond in the future)
```suggestion
# https://github.com/huggingface/transformers/blob/v4.53.2/src/transformers/models/mamba/modeling_mamba.py#L233
```

tests/openvino/test_modeling.py
```diff
@@ -1134,6 +1134,10 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "decilm",
     )
 
+    SUPPORTED_SSM_ARCHITECTURES = ("mamba", "falcon-mamba")
+    if is_transformers_version(">=", "4.39"):
```
echarlaix: should be 4.43 no ?

echarlaix (commented):
Looks great!

tests/openvino/test_modeling.py
rkazants: No, we should not. That is because we have transformers in else-branch. The name of KV cache outputs are different.
For Mamba based models cache output name is "cache_params".

optimum/exporters/openvino/model_patcher.py
rkazants: Let us do it in separate PR when, for example, adding support for other Mamba model. `DynamicCache` class does not fit to represent Mamba cache because it is always static and have two parts (conv and ssm, both are of static size).
Moreover, there are models that have a mix of caches Mamba cache+KV cache. So we need to think which class to create or this one to extend.

tests/openvino/test_modeling.py
rkazants:
```suggestion
    if is_transformers_version(">=", "4.43"):
```
changed sinсe there is an (transformer) issue with Autotokenizer for mamba models in early version that is reproduced in slow tests.

rkazants: Done

optimum/exporters/openvino/model_patcher.py
rkazants: Great catch - no need to override. It works without this modification. So I removed it

tests/openvino/test_modeling.py
IlyasMoutawwakil: yes and we should do the same checks for that as well 😅
Something along the lines of:
```suggestion
            self.assertTrue("cache_params" in ov_outputs)
            self.assertIsInstance(ov_outputs.cache_params, tuple)
            is_stateful = ov_model.config.model_type not in not_stateful
            self.assertEqual(ov_model.stateful, is_stateful)
            if is_stateful:
                self.assertTrue(len(ov_outputs.cache_params) == 1 and len(ov_outputs.cache_params[0]) == 0)
```
because ssm are also stateful in this case no ?

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: Yes DynamicCache doesn't fit here, that's why I didn't suggest using it, I'm suggesting taking inspiration from the method `from_legacy_cache` method.

tests/openvino/test_modeling.py
rkazants: Thank you for the explanation. Added more checks.

IlyasMoutawwakil (approved):
Great work ! will merge once tests pass.

PR #1366, opened 2025-06-30
Add Ernie 4.5 for OpenVINO

openvino-dev-samples:
This patcher is for supporting Baidu Ernie 4.5 LLM:

https://huggingface.co/baidu/ERNIE-4.5-0.3B-PT


openvino-dev-samples:
@nikita-savelyevv

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6456,3 +6456,375 @@ def __exit__(self, exc_type, exc_value, traceback):
             if layer.is_moe_layer:
                 layer.feed_forward.forward = layer.feed_forward._orig_forward
             layer.self_attn.forward = layer.self_attn._orig_forward
+
+
```
rkazants: can you please give a link to original implementations for each of these functions and elaborate why is patching needed?

rkazants (commented):
please add tests for inference (test_modeling) and for export

rkazants (commented):
also, documentation update is needed

optimum/exporters/openvino/model_patcher.py
openvino-dev-samples: Hi, These patches are for 3 purposes:
1. Switch to standard SDPA OP to optimization in Plugin. [original implementation](https://huggingface.co/baidu/ERNIE-4.5-0.3B-PT/blob/main/modeling_ernie4_5.py#L484)
2. Reshape the QKV and kvcache from [batch, seq_len, heads, dim] to  [batch, heads, seq_len, dim] for better performance. [original implementation](https://huggingface.co/baidu/ERNIE-4.5-0.3B-PT/blob/main/modeling_ernie4_5.py#L575)
3. replace with a general ROPE implementation for graph fusion.[original implementation](https://huggingface.co/baidu/ERNIE-4.5-0.3B-PT/blob/main/modeling_ernie4_5.py#L82)
4. trigger 'use_cache=true' for stateful model. [original implementation](https://huggingface.co/baidu/ERNIE-4.5-0.3B-PT/blob/main/config.json#L23)

rkazants: please add this in a comment and add a link to original implementation

tests/openvino/utils_tests.py
```diff
@@ -63,6 +63,7 @@
     "donut": "fxmarty/tiny-doc-qa-vision-encoder-decoder",
     "donut-swin": "hf-internal-testing/tiny-random-DonutSwinModel",
     "detr": "hf-internal-testing/tiny-random-DetrModel",
+    "ernie4_5": "snake7gun/tiny-random-ernie4_5",
```
rkazants: please update test files: `test_export.py`, `test_export_cli.py` and `test_modeling.py`

openvino-dev-samples: I just wonder why 'test_export.py, test_export_cli.py' are needed in this case, looks they are only designed for task oriented test ?

IlyasMoutawwakil: @openvino-dev-samples the name ernie4_5 need to be added in the causal LM supported architectures so that it can be tested

openvino-dev-samples: done

IlyasMoutawwakil:
@openvino-dev-samples the model is failing its modeling test

tests/openvino/test_modeling.py
```diff
@@ -1292,8 +1297,12 @@ def test_compare_to_transformers(self, model_arch):
...
         tokenizer = AutoTokenizer.from_pretrained(model_id, trust_remote_code=model_arch in self.REMOTE_CODE_MODELS)
         tokens = tokenizer("This is a sample output", return_tensors="pt")
-
-        ov_outputs = ov_model(**tokens)
+        if model_arch == "ernie4_5":
+            # As the offical example, ernie4_5 can only accept input_ids
+            # https://huggingface.co/baidu/ERNIE-4.5-0.3B-PT/blob/main/README.md?code=true#L92
+            ov_outputs = ov_model(tokens.input_ids)
+        else:
+            ov_outputs = ov_model(**tokens)
```
IlyasMoutawwakil: I'm not sure this is true, checking the code for ernie4.5 causal lm, the forward pass does accept attention mask and other stuff.

IlyasMoutawwakil: also the export uses text encoder with position ids, so how come they shouldn't be passed here ?

openvino-dev-samples: I think its a mistake in ernie's modeling design, because if you get a chance to try its `AutoModelForCausalLM` .generate pipeline, there will be error reports:
`ValueError: `attention_mask` passed to `generate` must be 2D.`

Also  [`attention mask`](https://huggingface.co/baidu/ERNIE-4.5-0.3B-PT/blob/main/modeling_ernie4_5.py#L448
) and [`position ids`](https://huggingface.co/baidu/ERNIE-4.5-0.3B-PT/blob/main/modeling_ernie4_5.py#L524) actually do not working in is modeling.

IlyasMoutawwakil: hmm yeah you're correct it's never used in the internal logic, so the torchscript tracer does not pick it up as input.

IlyasMoutawwakil: honestly this is why I prefer if remote code integrations stay as examples that aren't part of the core library.
the remote code in many cases has issues and is not tested enough if not at all (because it's not part of transformers).

optimum/exporters/openvino/model_patcher.py
echarlaix: @openvino-dev-samples looks like we can remove all patches when using the transformers integration https://github.com/huggingface/transformers/blob/v4.53.2-Ernie-4.5-preview/src/transformers/models/ernie4_5/modeling_ernie4_5.py#L402 Let me know if I can help on this! 

Also looks like all the modeling code has been removed from the models repo (for example in https://huggingface.co/baidu/ERNIE-4.5-0.3B-PT/commit/54cd03c1f4dc38d601910726156d4b685c4fd6d6)

IlyasMoutawwakil:
@openvino-dev-samples please fix style

echarlaix:
@bot /style

optimum/exporters/openvino/model_configs.py
```diff
@@ -4490,3 +4490,17 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
...
+    MIN_TRANSFORMERS_VERSION = "4.54.0"
+
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, GemmaDummyPastKeyValuesGenerator)
+    DUMMY_PKV_GENERATOR_CLASS = GemmaDummyPastKeyValuesGenerator
+    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
+
+    def patch_model_for_export(
+        self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ) -> "ModelPatcher":
+        return UpdateCausalMaskModelPatcher(self, model, model_kwargs=model_kwargs)
```
echarlaix: @openvino-dev-samples `UpdateCausalMaskModelPatcher` was removed in https://github.com/huggingface/optimum-intel/pull/1377, can be replaced with `OVDecoderModelPatcher` cc @IlyasMoutawwakil

IlyasMoutawwakil: ah yes true ! thanks for the catch !

PR #1370, opened 2025-07-01
fix ipex model loading for sentence transformers v5

echarlaix:
IPEX Model loading is broken in v5.0 coming from https://github.com/UKPLab/sentence-transformers/blob/v5.0.0/sentence_transformers/SentenceTransformer.py#L338 which results in model now being loaded with `_load_auto_model`


IlyasMoutawwakil (approved):
LGTM, Thanks !

jiqing-feng:
It seem that the IPEXSTModel test failed.

PR #1374, opened 2025-07-07
[OpenVINO]Add Kontext support

openvino-dev-samples:
Depends on this [PR for Diffusers](https://github.com/huggingface/diffusers/pull/11875)


openvino-dev-samples:
@nikita-savelyevv could help to review ?

nikita-savelyevv:
@bot /style

tests/openvino/utils_tests.py
```diff
@@ -76,6 +76,7 @@
     "flaubert": "hf-internal-testing/tiny-random-flaubert",
     "flux": "katuni4ka/tiny-random-flux",
     "flux-fill": "katuni4ka/tiny-random-flux-fill",
+    "flux-kontext": "snake7gun/flux-kontext-random",
```
nikita-savelyevv, line 82: I believe adding this here is not enough for the test for this model to be run. Could you please add changes to `test_diffusion.py`, `test_exporters_cli.py` and `test_quantization.py` similar to how it was done in this PR for example https://github.com/huggingface/optimum-intel/pull/1106?

nikita-savelyevv (changes requested):

tests/openvino/utils_tests.py
openvino-dev-samples: Hi @IlyasMoutawwakil my patch depends on the fix for diffusers lib, could you help to check ?
https://github.com/huggingface/diffusers/pull/11875

IlyasMoutawwakil: you don't need to wait for diffusers to do an entire release to test this. You can do the same as sana by using DiffusionPipeline instead of AutoPipelineForxxx https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_diffusion.py#L146 for now

openvino-dev-samples: > you don't need to wait for diffusers to do an entire release to test this. You can do the same as sana by using DiffusionPipeline instead of AutoPipelineForxxx https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_diffusion.py#L146 for now

yes, but how about in CLI export test case ?

IlyasMoutawwakil: cli export test should pass as it uses the correct class name https://github.com/huggingface/optimum/blob/ec756fdd00543d3a93458fafbcdf5abbb17b61ba/optimum/exporters/tasks.py#L2159 🙂

openvino-dev-samples: > cli export test should pass as it uses the correct class name https://github.com/huggingface/optimum/blob/ec756fdd00543d3a93458fafbcdf5abbb17b61ba/optimum/exporters/tasks.py#L2159 🙂

1. CLI export should not work without my patch on diffusers
2. another question is if there is a method to update the version of diffusers in pytest?
`E   ImportError: cannot import name 'FluxKontextPipeline' from 'diffusers' `

IlyasMoutawwakil: 1. no it should work without it. your patch is only rleevant to the AutoPipelineForxxx class, not the DiffusionPipeline class.
2. update it to what version ? I just checked the latest diffusers and this model is not even released in diffusers yet https://github.com/huggingface/diffusers/blob/v0.34.0/src/diffusers/pipelines/flux/pipeline_flux_kontext.py
what's exactly the motive for adding this model anyway ?

openvino-dev-samples: So your suggestion is adding the model after 0.34.0 version of diffuser released, am i right ?

IlyasMoutawwakil: or install diffusers from source.. for testing sake.
but bear in mind that python packages can't be released with a "git+https...." dependency, so we will have to wait for a diffusers release if you want this in the next optimum-intel release.

PR #1183, opened 2025-03-03
add support llava-next-video

eaidova:
# What does this PR do?

```
from transformers import AutoProcessor, TextStreamer
from transformers.image_utils import load_video
from optimum.intel.openvino import OVModelForVisualCausalLM
from huggingface_hub import hf_hub_download



model_id = "llava-hf/LLaVA-NeXT-Video-7B-hf"

model = OVModelForVisualCausalLM.from_pretrained(model_id)
processor = AutoProcessor.from_pretrained(model_id)


video_path = hf_hub_download(
    repo_id="raushan-testing-hf/videos-test", filename="sample_demo_1.mp4", repo_type="dataset"
)

question = "Why is this video funny?"


video, video_meta = load_video(video_path, num_frames=8)

inputs = model.preprocess_inputs(question, video=video, processor=processor)

model.generate(**inputs, max_new_tokens=100, do_sample=False, streamer=TextStreamer(processor.tokenizer, skip_special_tokens=True, skip_prompt=True))
```

Result:

> The humor in this video comes from the unexpected and endearing nature of the situation. The baby is wearing glasses and appears to be reading a book, which is a humorous and endearing sight because babies are typically not expected to be able to read at such a young age. The glasses add a touch of whimsy and the baby's focused expression while reading adds to the comedic effect. The scene is also amusing because it's a playful

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

PR #1375, opened 2025-07-07
Fix attr name typo in model_configs for llava-next compatibility with transformers 4.51.3

mitruska:
# What does this PR do?

Fixes # (issue)
This PR fixes a typo in optimum-intel code, replacing "mod**e**l" with "mod**a**l" for "multi_modal_projector" attr name here: https://github.com/huggingface/optimum-intel/blob/8f127ce3a2c17fca929ab2d38843dbfd16b1d4c0/optimum/exporters/openvino/model_configs.py#L1905-L1910
Without the fix: The following error is raised for transformers==4.51.3 
```"AttributeError: 'LlavaNextVideoForConditionalGeneration' object has no attribute 'model'"```

The issue has been introduced with a PR #1319 to optimum-intel with changes for compatibility with transformers 4.52
(working correctly).

## Before submitting
- [X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
   - **Please guide me if there is any appropriate test to be added for this issue**

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -1905,7 +1905,7 @@ class LlavaNextVideoOpenVINOConfig(LlavaOpenVINOConfig):
         if behavior == LlavaNextVideoConfigBehavior.MULTI_MODAL_PROJECTOR:
             return (
                 model.multi_modal_projector
-                if hasattr(model, "multi_model_projector")
+                if hasattr(model, "multi_modal_projector")
                 else model.model.multi_modal_projector
             )
 

### CODE REVIEW

File: optimum/exporters/openvino/model_configs.py [hunk 1]
```diff
@@ -1905,7 +1905,7 @@ def get_model_for_behavior(self, model, behavior: Union[str, LlavaNextVideoConfi
         if behavior == LlavaNextVideoConfigBehavior.MULTI_MODAL_PROJECTOR:
             return (
                 model.multi_modal_projector
-                if hasattr(model, "multi_model_projector")
+                if hasattr(model, "multi_modal_projector")
```
rkazants, line 1908: is it possible to add test(s) to avoid such issues in the future?:)
https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_export.py
https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_exporters_cli.py

mitruska: Actually there are existing tests for this model in the mentioned files, added with PR:
- https://github.com/huggingface/optimum-intel/pull/1183

I would expect this issue to be detected when tested with transformers==4.51.3 and original model.

IlyasMoutawwakil: yeah testing these things will require testing optimum-intel against all versions of transformers, although most ppl will just use the latest

rkazants:
@echarlaix, @IlyasMoutawwakil, could you please take a look? I think we need to merge it since it is fixing a typo introduced in other PR and we have a customer who anticipated this fix.
In separate PR we can add a test if needed and any idea.

Best regards,
Roman

PR #1377, opened 2025-07-10
Transformers 4.53 support

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting


setup.py
```diff
@@ -29,7 +29,7 @@
 INSTALL_REQUIRE = [
     "torch>=1.11",
     "optimum@git+https://github.com/huggingface/optimum.git",
```
IlyasMoutawwakil:
```suggestion
    "optimum@git+https://github.com/huggingface/optimum.git@transformers-4.53",
```

tests/openvino/utils_tests.py
```diff
@@ -93,8 +93,8 @@
     "jais": "katuni4ka/tiny-random-jais",
     "levit": "hf-internal-testing/tiny-random-LevitModel",
     "longt5": "hf-internal-testing/tiny-random-longt5",
-    "llama": "HuggingFaceM4/tiny-random-LlamaForCausalLM",
-    "llama_awq": "HuggingFaceH4/tiny-random-LlamaForCausalLM",
+    "llama": "hf-internal-testing/tiny-random-LlamaForCausalLM",
+    "llama_awq": "hf-internal-testing/tiny-random-LlamaForCausalLM",
```
IlyasMoutawwakil:
```suggestion
    "llama": "HuggingFaceH4/tiny-random-LlamaForCausalLM",
    "llama_awq": "HuggingFaceH4/tiny-random-LlamaForCausalLM",
```

tests/openvino/utils_tests.py
```diff
@@ -93,7 +93,7 @@
     "jais": "katuni4ka/tiny-random-jais",
     "levit": "hf-internal-testing/tiny-random-LevitModel",
     "longt5": "hf-internal-testing/tiny-random-longt5",
-    "llama": "HuggingFaceM4/tiny-random-LlamaForCausalLM",
+    "llama": "HuggingFaceH4/tiny-random-LlamaForCausalLM",
```
IlyasMoutawwakil:
```suggestion
    "llama": "HuggingFaceM4/tiny-random-LlamaForCausalLM",
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -107,25 +85,13 @@ def patch_model_with_bettertransformer(model):
     if hasattr(model, "use_bettertransformer") and model.use_bettertransformer is True:
         return model
 
-    if is_transformers_version("<", "4.36") or is_torch_version("<", "2.1.1"):
```
echarlaix, line 110: good catch !

optimum/exporters/openvino/model_patcher.py
```diff
@@ -161,27 +127,255 @@ def patch_update_causal_mask(
...
+            # for non-stateful decoder models, we use eager mask without vmap for both eager and sdpa
+            # to avoid a nan output issue in OpenVINO that only happens in case of non-stateful models
+            if not getattr(self.real_config, "stateful", False):
+                logger.warning(
+                    "Exporting a non-stateful decoder model currently results in a nan output in OpenVINO. "
+                    "There might be a performance impact due to the use of eager mask (floats) instead of sdpa mask (bools). "
+                )
+                # for non-stateful models, we have to use eager mask without vmap for both eager and sdpa
+                ALL_MASK_ATTENTION_FUNCTIONS.register("sdpa", eager_mask_without_vmap)
+                ALL_MASK_ATTENTION_FUNCTIONS.register("eager", eager_mask_without_vmap)
```
echarlaix: question : for eager shouldn't it be done even for stateful models as well (to update min dtype) or isn't it needed ?

echarlaix (approved):
Huge work @IlyasMoutawwakil :fire: thanks a lot!

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: yess you're right !

tests/openvino/test_modeling.py
```diff
@@ -2038,29 +2040,6 @@ def test_pipeline(self, model_arch):
...
-
-        # With input ids
-        outputs = model.generate(input_ids=tokens["input_ids"])
-        outputs = tokenizer.batch_decode(outputs, skip_special_tokens=True)
-        self.assertIsInstance(outputs[0], str)
-        del model
-
-        gc.collect()
-
     def test_compare_with_and_without_past_key_values(self):
```
rkazants: why removed?

tests/openvino/test_modeling.py
```diff
@@ -2370,22 +2349,6 @@ def test_compare_to_transformers(self, model_arch):
...
-        inputs = preprocessor(images=self.IMAGE, text=question, return_tensors="pt")
-
-        # General case
-        outputs = model.generate(**inputs)
-        outputs = preprocessor.batch_decode(outputs, skip_special_tokens=True)
-        self.assertIsInstance(outputs[0], str)
-        del model
-
-        gc.collect()
-
```
rkazants: why removed?

IlyasMoutawwakil: bcz the testing suite is extremely slow and redundant, these tests used to be relevant but they're not anymore because most of what is done is done and more in compare_to_transformers

setup.py
```diff
@@ -28,8 +28,8 @@
 
 INSTALL_REQUIRE = [
     "torch>=1.11",
-    "optimum@git+https://github.com/huggingface/optimum.git",
-    "transformers>=4.36,<4.53",
+    "optimum@git+https://github.com/huggingface/optimum.git@transformers-4.53",
```
rkazants: does it mean that it will no longer work with older transformers version?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -779,6 +782,7 @@ def __exit__(self, exc_type, exc_value, traceback):
 SUPPORT_SDPA = is_torch_version(">", "2.1.0")
 
 
+# TODO: why
```
rkazants, line 785: please add more detailed comment

setup.py
IlyasMoutawwakil: no this is just an optimum branch that's compatible with the transformers update, used as long as the branch is not yet merged, then we use optimum from main and then when optimum is released we use the new version.

tests/openvino/test_modeling.py
IlyasMoutawwakil: redundant, each time one of these tests run, we re export the model which takes a lot of time.
i suggest you check optimum/onnxruntime/test_decoder.py
I will be refactoring and isolating the decoder testing suite because in its current states it has too much skips and undocumented special cases.

tests/openvino/test_modeling.py
rkazants: they are useful. At least for Mamba model, I found an issue using these tests. Sometimes we run these tests for some PRs by adding the corresponding tag.

IlyasMoutawwakil: can you please link to the commit id where this test failed while test_compare_to_transfroemrs passed ? 😊

rkazants: {33D9F807-7660-4FA1-9AE8-5DCE9FC4012B}

https://github.com/huggingface/optimum-intel/pull/1360/commits/fb5b926ab979d69cf9f3b7abefa7e4cb4e072034

IlyasMoutawwakil: please read the names of the tests that failed in https://github.com/huggingface/optimum-intel/actions/runs/16270476237/job/45936556506 and https://github.com/huggingface/optimum-intel/actions/runs/16270476237/job/45936556555 they're beam search tests and pipeline tests, no 
test_generate_utils 🙃

rkazants: Do you have results of slow tests for your PR? Can you please share?

IlyasMoutawwakil: no worries, if it's not obvious I'll just revert them back.

optimum/exporters/openvino/convert.py
```diff
@@ -381,6 +381,9 @@ def export_pytorch(
     logger.info(f"Using framework PyTorch: {torch.__version__}")
     output = Path(output)
 
+    # TODO: temporary solution but statefulness should be added to the export config earlier
+    config.stateful = stateful
```
rkazants, line 385: what is this WA needed?
Do we still normally support stateful and stateless models?

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: that's exactly what I'm asking for, these methods are patched but there's no comment as to why they're patched

optimum/exporters/openvino/model_patcher.py
```diff
@@ -2935,28 +2949,14 @@ def __enter__(self):
...
-        unpatch_update_causal_mask(self._model, "transformer")
+
         if is_transformers_version(">=", "4.49"):
             self._model.config._attn_implementation = self._model.config._orig_attn_implementation
             for block in self._model.transformer.h:
                 block.attn.forward = block.attn._orig_forward
                 block.attn._attn = block.attn._orig_attn
 
 
-class GptNeoxJapaneseModelPatcher(DecoderModelPatcher):
```
rkazants, line 2946: you are removing a lot of patches. Why is it so, please elaborate.
I am afraid that a lot of model IRs will be changed so it affects OpenVINO GenAI.

IlyasMoutawwakil: can you please check just below it, this patch was simply moved next to GptNeoxModelPatcher

IlyasMoutawwakil: > I am afraid that a lot of model IRs will be changed so it affects OpenVINO GenAI.

can you elaborate

optimum/exporters/openvino/convert.py
IlyasMoutawwakil: it's for this https://github.com/huggingface/optimum-intel/pull/1377/files#diff-3cb374575ada21e17fb8db07d59e4f779d3467adaf91d3c9db9e49b779bd4a62R361 I explain the issue in details there

nikita-savelyevv (commented):
LGTM. My only concern is merging while `OVModelForCausalLMIntegrationTest::test_beam_search_66_smollm3` fails on the latest OpenVINO release. @IlyasMoutawwakil is the problem understood?

IlyasMoutawwakil:
@nikita-savelyevv still investigating, the problem seems to be obscure, the model only fails with openvino export+inference, and works perfectly with onnx export + onnxruntime inference.
is there a way to debug the exported model graph ? like onnx netron ?

setup.py
IlyasMoutawwakil, line 31:
```suggestion
    "optimum@git+https://github.com/huggingface/optimum.git@fix-ort-batched-generation",
```

setup.py
echarlaix: could be set to 
```suggestion
    "optimum==1.27.*",
```

echarlaix (approved):
Excellent work @IlyasMoutawwakil!

optimum/exporters/openvino/model_patcher.py
```diff
@@ -161,27 +123,261 @@ def patch_update_causal_mask(
...
+            if layer.self_attn.rotary_emb.dtype != torch.float32:
+                layer.self_attn.rotary_emb._set_cos_sin_cache(
+                    seq_len=layer.self_attn.rotary_emb.max_position_embeddings,
+                    device=layer.self_attn.rotary_emb.inv_freq.device,
+                    dtype=torch.float32,
+                )
+
+
+# Adapted from https://github.com/huggingface/transformers/blob/v4.53.0/src/transformers/masking_utils.py#L433
+# Specifically for OpenVINO, we use torch.finfo(torch.float16).min instead of torch.finfo(dtype).min
```
beleiuandrei, line 323: @IlyasMoutawwakil can I find somewhere documented why OpenVINO expects torch.finfo(torch.float16).min instead of torch.finfo(dtype).min.

This approach is dangerous for 4D masks that has lines completely masked. This mask is used inside SDPA as input for an Add operation that adds the mask to some activation tensor which can have various range for example [-300;+300]. This means that output of the Add will have range [-65504-300: 0+300]. But the problem is that the line full of mask (-65504) will no longer be represent a masked value but some random value around -65504.

The problem appears after when the output of the Add goes to a Softmax operation. For a line that supposed to be fully maked, Softmax will have to do with values in range [-65504 - 300: -65504 + 300]. Softmax inside is doing a smart thing in order to better match the range of the values: it does a max subtraction, because Softmax(x) = Softmax(x - max(x)). So Softmax will apply on  range [-600:0] instead of  [-65504 - 300: -65504 + 300] and this will produce strange results after Softmax.

Using torch.finfo(torch.float32).min would better in my understanding. Because the difference between torch.finfo(torch.float32).min and -torch.finfo(torch.float32).inf is very large > 10^30, so it would be hard to produce -infinity after Add.

IlyasMoutawwakil: short answer is its for the npu's half precision 🥲 @rkazants 
at export time we dont know who/which device will use the exported model so we make this compromise for the npu

IlyasMoutawwakil: we have requested the openvino attention op supports inf masks directly because that's the most natural and closest to current torch sdpa but i think the different devices numerics make it hard to achieve globally.

beleiuandrei: SoftMax implementation of OpenVino CPU provides indeed nan output if the input contains -inf values. It produces good results if the input contains FP32::min.
On OpenVino NPU seems like we would be able to work with FP32::min.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -412,176 +603,6 @@ def __exit__(self, exc_type, exc_value, traceback):
...
-            attention_mask, inputs_embeds=input_tensor, past_key_values_length=past_seen_tokens
-        ):
-            return None
-
-    dtype, device = input_tensor.dtype, input_tensor.device
-
-    # difference with original modeling
-    # using minimum from dtype with larger bandwith (floa32) may lead to overflow
-    # during execution on platforms with default lower precision (bfloat16, float16)
-    min_dtype = torch.finfo(torch.float16).min
```
IlyasMoutawwakil, line 434: @beleiuandrei

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: then i'm really not sure why that decision was made, it is not part of this pr though, i only kept what was already there, i tagged you in the original comment which was -probably- added by @eaidova long before

beleiuandrei: Unfortunately I didn't get any notification about being tagged in the original comment. Can you please share me a link to it?

IlyasMoutawwakil: https://github.com/huggingface/optimum-intel/pull/1377#discussion_r2988839272

PR #1381, opened 2025-07-11
[OV] Prepare quantization dataset collection logic to transition to datasets v4.0

nikita-savelyevv:
# What does this PR do?

Prepare quantization dataset collection logic to transition to datasets v4.0:
- Set revision for librispeech dataset (points to parquet-converted dataset)
- Set revision for squadshifts dataset (points to parquet-converted dataset)
- Remove `trust_remote_code` argument from quantization-dataset-related logic

## Before submitting


setup.py
```diff
@@ -55,7 +55,7 @@
     "sentence-transformers",
     "open_clip_torch>=2.26.1",
     "peft",
-    "datasets[audio]>=1.4.0,<4.0.0",
+    "datasets[audio]>=1.4.0",
```
nikita-savelyevv: Seems like there are other issues caused by datasets 4.0, namely torch 2.7 being required.
```suggestion
    "datasets[audio]>=1.4.0,<4.0.0",
```

optimum/intel/openvino/quantization.py
```diff
@@ -355,7 +352,6 @@ def build_from_quantization_config(self, config: OVQuantizationConfigBase) -> OV
                     num_samples=None,
                     dataset_config_name=dataset_metadata["name"],
```
echarlaix, line 353: shouldn't it be updated here as well ?

echarlaix (approved):
Thanks a lor @nikita-savelyevv !!

optimum/intel/openvino/quantization.py
nikita-savelyevv: The `PREDEFINED_LANGUAGE_DATASETS` are kept the same as before because wikitext and c4 are already in parquet format.

PR #1386, opened 2025-07-17
[OV] Update the reference number of int8 nodes for SANA model

nikita-savelyevv:
# What does this PR do?

RoPE is no longer quantized.

## Before submitting


echarlaix (approved):
LGTM!

PR #1387, opened 2025-07-18
[OpenVINO][Draft] Add support for GLM-4.1V-9B-Thinkin

openvino-dev-samples:


optimum/exporters/openvino/model_patcher.py
```diff
@@ -6754,3 +6754,284 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            padding_mask = causal_mask[:, :, :, :mask_length] + attention_mask[:, None, None, :].to(causal_mask.device)
+            padding_mask = padding_mask == 0
+            causal_mask[:, :, :, :mask_length] = causal_mask[:, :, :, :mask_length].masked_fill(
+                padding_mask, min_dtype
+            )
+
+    return causal_mask
+
+
+def _glm4v_update_causal_mask(
```
echarlaix: why is this needed? I was thinking that we can remove `_glm4v_update_causal_mask` and `_glm4v_prepare_4d_causal_attention_mask_with_cache_position` as we should be compatible with `create_causal_mask` https://github.com/huggingface/transformers/blob/v4.53.0/src/transformers/models/glm4v/modular_glm4v.py#L982  (https://github.com/huggingface/optimum-intel/pull/1377 will soon be merged)

optimum/exporters/openvino/model_configs.py
```diff
@@ -154,6 +157,10 @@
 def init_model_configs():
     if "open_clip" not in TasksManager._LIBRARY_TO_SUPPORTED_MODEL_TYPES:
         TasksManager._LIBRARY_TO_SUPPORTED_MODEL_TYPES["open_clip"] = {}
+    TasksManager._CUSTOM_CLASSES[("pt", "glm4v", "image-text-to-text")] = (
```
echarlaix, line 153: here why not used `AutoModelForImageTextToText` directly to load all the image-text-to-text task models? 
https://github.com/huggingface/transformers/blob/5dba4bc7b2c1ef517ed44bba76bb70b59001c737/src/transformers/models/auto/modeling_auto.py#L941

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4367,4 +4679,5 @@ def preprocess_inputs(
     "phi4mm": _OVPhi4MMForCausalLM,
     "phi4_multimodal": _OVPhi4MMForCausalLM,
     "llama4": _OVLlama4ForCausalLM,
+    "glm4v": _OVGlm4vForCausalLM,
 }
```
echarlaix, line 4672: would you mind adding a test with a tiny random model ?

optimum/exporters/openvino/model_configs.py
```diff
@@ -4490,3 +4497,210 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
...
+        grid_t = self.batch_size
+        import torch
+
+        if input_name == "hidden_states":
+            return self.random_float_tensor(
+                [grid_t * grid_h * grid_w, self.embed_dim], framework=framework, dtype=float_dtype
+            )
+
+        if input_name == "seqlens":
+            return torch.tensor([grid_t * grid_h * grid_w], dtype=torch.int64)
```
echarlaix, line 4584: question : do we need to generate `seqlens` in the input generator or should we infer it directly in the patch instead ? (given `hidden_states` for example) ?

optimum/exporters/openvino/model_configs.py
```diff
@@ -4490,3 +4497,210 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
...
+        if input_name == "rotary_pos_emb":
+            dim = self.embed_dim // self.num_heads // 2
+            return self.random_float_tensor([grid_h * grid_t * grid_w, dim], framework=framework, dtype=float_dtype)
+
+        if input_name == "image_type_ids":
+            return self.random_int_tensor(
+                [grid_t * grid_h * grid_w, 2], max_value=grid_h, framework=framework, dtype=int_dtype
+            )
+
+        if input_name == "grid_thw":
```
echarlaix, line 4600: same question shouldn't it be inferred in the patch ?

optimum/exporters/openvino/model_configs.py
```diff
@@ -4490,3 +4497,210 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
...
+        if input_name in ["attention_mask", "window_attention_mask"]:
+            return self.random_mask_tensor(
+                [1, grid_t * grid_h * grid_w, grid_t * grid_h * grid_w], framework=framework, dtype=float_dtype
+            )
+
+        if input_name == "rotary_pos_emb":
+            dim = self.embed_dim // self.num_heads // 2
+            return self.random_float_tensor([grid_h * grid_t * grid_w, dim], framework=framework, dtype=float_dtype)
+
+        if input_name == "image_type_ids":
```
echarlaix, line 4595: looks like `DummyGlm4vVisionEmbedInputGenerator` could inherit from `DummyQwen2VLVisionEmbedInputGenerator` as they are both very close (just need to overwrite `generate` and just add `image_type_ids` ) what do you think ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4349,6 +4349,318 @@ def preprocess_inputs(
...
+
+            self._rotary_pos_emb = Glm4vVisionRotaryEmbedding(
+                self.config.vision_config.hidden_size // self.config.vision_config.num_heads // 2
+            )
+        else:
+            raise ValueError(
+                f"Initialization model for {self.config.model_type} required at least transformers >= 4.45"
+            )
+
+    def get_rope_index(
```
echarlaix, line 4384: would you mind adding a link to the original code https://github.com/huggingface/transformers/blob/v4.53.3/src/transformers/models/glm4v/modular_glm4v.py#L1014 (and highlight any modifications if any)

optimum/exporters/openvino/model_patcher.py
openvino-dev-samples: currently, there is accuracy issue on conversion [torch.vmap](https://github.com/huggingface/transformers/blob/6017f5e8ed33d48096cdf8630d1cc7cbf2550c90/src/transformers/masking_utils.py#L174), which is included in create_casual_mask.

optimum/exporters/openvino/model_configs.py
openvino-dev-samples: its directly from [forward function](https://github.com/huggingface/transformers/blob/5dba4bc7b2c1ef517ed44bba76bb70b59001c737/src/transformers/models/glm4v/modeling_glm4v.py#L503C56-L503C63)

optimum/exporters/openvino/model_configs.py
openvino-dev-samples: @echarlaix Sorry, I dont understand what "infer it directly in the patch instead" means ? could you show me a link of example code ? thanks

PR #1391, opened 2025-07-23
fix auto_model_class for OVModelForVisualCausalLM

echarlaix:


optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -44,15 +42,14 @@
...
 
-try:
-    from transformers import LlavaNextForConditionalGeneration
-except ImportError:
-    LlavaNextForConditionalGeneration = None
+    transformers_auto_class = AutoModelForImageTextToText
+else:
+    from transformers import AutoModelForVision2Seq
+
+    transformers_auto_class = AutoModelForVision2Seq
```
echarlaix, line 52: `AutoModelForVision2Seq` will be deprecated in v5 https://github.com/huggingface/transformers/blob/main/src/transformers/models/auto/modeling_auto.py#L2135 we should use `AutoModelForImageTextToText` instead (when possible)

helena-intel (approved):
Thanks @echarlaix ! The pipeline now works as expected in my code.

IlyasMoutawwakil (approved):

PR #1395, opened 2025-07-31
Install latest OpenVINO 2025.2 and NNCF 2.17

rkazants:
# What does this PR do?

Install latest OpenVINO 2025.2 and NNCF 2.17 for optimum-intel

Fixes # N/A

## Before submitting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,8 +65,8 @@ TESTS_REQUIRE = [
 QUALITY_REQUIRE = ["black~=23.1", "ruff==0.4.4"]
 
 EXTRAS_REQUIRE = {
-    "nncf": ["nncf>=2.16.0"],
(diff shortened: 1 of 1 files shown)


IlyasMoutawwakil:
By rationale I mean reasoning, what's the reasoning for dropping support of 2025.1 ?

> Also, I would like optimum-intel to be validated with latest OpenVINO versions in optimum-intel precommit.

It is the case already, please check the result of the pip install in percommit CI https://github.com/huggingface/optimum-intel/actions/runs/16649511615/job/47118231146?pr=1377#step:4:575

What you are changing in this PR is the not the version that will be installed, but rather the minimum version the dependency resolver can accept.

Overall I have no issue with updating the minimum version, I just wanna make sure we are on the same page of what this does.

PR #1402, opened 2025-08-04
[OpenVINO] Handle deprecation of forced_decoder_ids in transformers generation_config

aleksandr-mokrov:
# What does this PR do?

Since transformers 4.53.0, forced_decoder_ids is deprecated: https://github.com/huggingface/transformers/pull/38232
Therefore, there is no `forced_decoder_ids` attribute in  `generation_config`, and `forced_decoder_ids = generation_config.forced_decoder_ids` causes an error. However, for transformer versions >= 4.50.0 and <  4.53.0,  the fix in `_get_logits_processor` is still required.

### DIFF
--- a/optimum/intel/openvino/modeling_seq2seq.py
+++ b/optimum/intel/openvino/modeling_seq2seq.py
@@ -1465,12 +1465,8 @@ class _OVModelForWhisper(OVModelForSpeechSeq2Seq, WhisperForConditionalGeneratio
     def _get_logits_processor(self, generation_config: GenerationConfig, *args, **kwargs):
         # Whisper uses forced_decoder_ids for default task and language specification, while original _get_logits_processor does not allow it
         # see for details https://github.com/huggingface/transformers/issues/37172
-        if not hasattr(generation_config, "forced_decoder_ids"):
+        if not hasattr(generation_config, "forced_decoder_ids") or is_transformers_version(">=", "4.53.0"):
             # since transformers 4.53.0, forced_decoder_ids is deprecated: https://github.com/huggingface/transformers/pull/38232
-            logits_processor = super()._get_logits_processor(generation_config, *args, **kwargs)
-        elif is_transformers_version(">=", "4.53.0"):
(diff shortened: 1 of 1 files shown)


nikita-savelyevv:
I wonder why this was not caught by tests 🤔

echarlaix (approved):
LGTM, thanks for the fix!

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1463,11 +1463,16 @@ def prepare_inputs_for_generation(
...
 
     def _get_logits_processor(self, generation_config: GenerationConfig, *args, **kwargs):
-        forced_decoder_ids = generation_config.forced_decoder_ids
         # Whisper uses forced_decoder_ids for default task and language specification, while original _get_logits_processor does not allow it
         # see for details https://github.com/huggingface/transformers/issues/37172
-        if is_transformers_version(">=", "4.50.0"):
-            generation_config.forced_decoder_ids = None
-        logits_processor = super()._get_logits_processor(generation_config, *args, **kwargs)
-        generation_config.forced_decoder_ids = forced_decoder_ids
+        if not hasattr(generation_config, "forced_decoder_ids"):
```
echarlaix: should also be enforced for transformers >= 4.53 no ? (even in the case where the generation config was saved with a forced_decoder_ids)

aleksandr-mokrov: In that case, we should delete the attribute. The handling has been added.

echarlaix: I don't think we should modify the model's config in optimum (should be done in transformers if this is something expected by transformers)

aleksandr-mokrov: Deleting the attribute has been removed

echarlaix: thanks a lot @aleksandr-mokrov, will merge once the tests pass

PR #1412, opened 2025-08-07
add support for minicpm4v

openvino-dev-samples:
Depends on [PR](https://github.com/huggingface/optimum/pull/2335)

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -2822,10 +2822,10 @@ class MiniCPMVOpenVINOConfig(BaseVLMOpenVINOConfig):
             behavior = MiniCPMVConfigBehavior(behavior)
 
         if behavior == MiniCPMVConfigBehavior.TEXT_EMBEDDINGS:
-            return get_vlm_text_embeddings_config("qwen2", self._orig_config, self.int_dtype, self.float_dtype)
+            return get_vlm_text_embeddings_config("qwen2" if self._orig_config.version == 2.6 else "llama", self._orig_config, self.int_dtype, self.float_dtype)
 
         if behavior == MiniCPMVConfigBehavior.LANGUAGE:
-            return get_vlm_text_generation_config("qwen2", self._orig_config, self.int_dtype, self.float_dtype)
+            return get_vlm_text_generation_config("qwen2" if self._orig_config.version == 2.6 else "llama", self._orig_config, self.int_dtype, self.float_dtype)
 
         if behavior == MiniCPMVConfigBehavior.VISION_EMBEDDINGS:
             return self.__class__(


openvino-dev-samples:
@IlyasMoutawwakil

IlyasMoutawwakil:
Thanks for the fix ! let's create a tiny random model with llama as the decoder to test this 🤗 tell me you need help with that !

openvino-dev-samples:
> Thanks for the fix ! let's create a tiny random model with llama as the decoder to test this 🤗 tell me you need help with that !

But i guess we need merge this [PR](https://github.com/huggingface/optimum/pull/2335) first ? otherwise test case will not work

IlyasMoutawwakil:
@openvino-dev-samples no need to merge it now, you can simply pin that PR in setup.py so that the tests would run with it 🤗
We will merge both PRs once everything works together.

openvino-dev-samples:
> @openvino-dev-samples no need to merge it now, you can simply pin that PR in setup.py so that the tests would run with it 🤗 We will merge both PRs once everything works together.

Hi since minicpmv4 and minicpmv share a same model [type](https://huggingface.co/openbmb/MiniCPM-V-4/blob/main/config.json#L27), but different LLM.  It is possible to add both of them in utils_tests.py ?

openvino-dev-samples:
> Hi @openvino-dev-samples it would be faster if you made sure the minicpmv4 tests pass locally, the ci is slow and shouldn't be used as a testing mechanism, only use it for validating when local tests are already passing.

Sorry for that, and i fully understand, but i always met connection issue in local test case, e.g

`huggingface_hub.errors.HfHubHTTPError: 429 Client Error: Too Many Requests for url: https://huggingface.co/api/models/katuni4ka/tiny-random-qwen2.5-vl/tree/main?recursive=True&expand=False`

IlyasMoutawwakil:
> but i always met connection issue in local test case

you can target minicpmv tests specifically to avoid this issue with pytest -k "minicpmv"

tests/openvino/utils_tests.py
```diff
@@ -116,6 +116,8 @@
     "minicpm": "katuni4ka/tiny-random-minicpm",
     "minicpm3": "katuni4ka/tiny-random-minicpm3",
     "minicpmv": "katuni4ka/tiny-random-minicpmv-2_6",
+    "minicpmv4": "snake7gun/minicpm-v-4-tiny",
+    "minicpmv4_5": "snake7gun/tiny-minicpmv-4_5",
```
rkazants: 158M model size, it makes sense to try to reduce the size

openvino-dev-samples:
@IlyasMoutawwakil could you help to trigger CI, thanks

optimum/exporters/openvino/model_configs.py
```diff
@@ -2625,12 +2625,22 @@ def with_behavior(
         """
         if isinstance(behavior, str) and not isinstance(behavior, MiniCPMVConfigBehavior):
             behavior = MiniCPMVConfigBehavior(behavior)
-
+        model_mapping = {2.6: "llama", 4.0: "qwen2", 4.5: "qwen3"}
```
IlyasMoutawwakil: should use str for versions

openvino-dev-samples: may i understand why？the version in model's config is a number:
https://huggingface.co/openbmb/MiniCPM-V-4_5/blob/main/config.json#L3

IlyasMoutawwakil: ah okay I see ! thanks for the clarification.
(it's generally a bad idea to use numbers for versions: 4.0 becomes 4 and 4.10 and 4.1 are the same version 😅)

rkazants: I think it is a bad idea to make decision about architecture based on model version in general.
I think you should parse class model object and use `isinstance` for inner objects to make decision.

tests/openvino/test_seq2seq.py
```diff
@@ -837,6 +847,9 @@ def test_generate_utils(self, model_arch):
             input_audio = self._generate_random_audio_data()
             question = "Translate this audio to French"
             inputs = model.preprocess_inputs(**preprocessors, text=question, audio=[input_audio])
+            # skip the temporal_ids which makes the number of loop inconstant:
+            # https://huggingface.co/openbmb/MiniCPM-V-4_5/blob/main/resampler.py#L261
+            inputs.pop("temporal_ids", None)
```
rkazants: don't quite understand. To me it degrades user experience that we make users to skip this parameter after preprocessing to have matched results with the reference.

optimum/exporters/openvino/model_configs.py
openvino-dev-samples: Yes, its a better approach in this case, but I dont know if we can access the modeling file at this stage.

tests/openvino/test_seq2seq.py
openvino-dev-samples: Yes, it will impact the user experience in some cases. I think its one of the limitation for torchscipt. I will try to find a better solution handle this parameter if possible.

nikita-savelyevv:
@openvino-dev-samples Please fix the failed tests

tests/openvino/test_seq2seq.py
rkazants: still not fixed

openvino-dev-samples: sorry, I just forget to remove one of them in test case, but the implementation is done.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -285,11 +285,21 @@ def __init__(self, model: ov.Model, parent_model: OVBaseModel) -> None:
...
+                    "image_feature": image_feature,
+                    "pos_embed": pos_embed,
+                    "key_padding_mask": key_padding_mask,
+                    "temporal_embed": temporal_embed,
+                }
+            )[0]
+        else:
+            result = self.request(
+                {"image_feature": image_feature, "pos_embed": pos_embed, "key_padding_mask": key_padding_mask}
+            )[0]
```
nikita-savelyevv:
```suggestion
        inputs = {"image_feature": image_feature, "pos_embed": pos_embed, "key_padding_mask": key_padding_mask}
        if temporal_embed is not None:
            inputs["temporal_embed"] = temporal_embed
        result = self.request(inputs)[0]
```

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -1923,10 +1936,18 @@ def __init__(
         max_size = self.config.vision_config.image_size // self.config.vision_config.patch_size
         self._pos_embeds = torch.from_numpy(self._get_2d_sincos_pos_embed(self.embed_dim, max_size)).float()
         self.max_size = (max_size, max_size)
+        self.max_temporal_size = 72000
```
nikita-savelyevv, line 1930: Why 72000? Should this value be loaded from the config?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -1923,10 +1936,18 @@ def __init__(
...
-    def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
+    def get_vision_embeddings(self, pixel_values, input_ids=None, temporal_ids=None, **kwargs):
         if input_ids is not None and input_ids.shape[1] == 1:
             return None
+
+        all_temporal_ids = None
+        if temporal_ids is not None:
+            all_temporal_ids = []
+            for t in temporal_ids:
+                all_temporal_ids.extend(t)
```
nikita-savelyevv, line 1940:
```suggestion
        all_temporal_ids = [t for seq_t in temporal_ids for t in seq_t] if temporal_ids is not None else None
```

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -1979,26 +2000,57 @@ def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
...
 
         patch_len = tgt_sizes[:, 0] * tgt_sizes[:, 1]
 
         self._adjust_pos_cache(tgt_sizes)
 
+        temporal_pos_emb = False
+        temporal_ids_flatten = None
+        if temporal_ids is not None:
+            # example: [[-1], [-1], [2, 6, 9]]
+            temporal_ids_flatten = list(chain.from_iterable(temporal_ids))
```
nikita-savelyevv, line 2041: Do we actually need to do an additional flattening pass here? As I understand `all_temporal_ids` is prepared already flattened inside `get_vision_embeddings()`. If not needed, I'd remove flattening logic from `get_vision_embeddings()` and keep it only here.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -1979,26 +2000,57 @@ def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
...
+            max_temporal_size = max(temporal_ids_flatten) + 1
+            if max_temporal_size > -1:
+                temporal_pos_emb = True
+            if max_temporal_size > self.max_temporal_size:
+                self._adjust_temporal_pos_cache(max_temporal_size, "cpu")
```
nikita-savelyevv, line 2046: I don't see a definition of `self._adjust_temporal_pos_cache()`. Since the tests pass, this means the code does not reach this point in any of existing tests. Please clarify this. Ideally, every scenario should be tested.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -1979,26 +2000,57 @@ def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
...
         pos_embed = []
+        pos_embed_temporal = []
         for i in range(bs):
             tgt_h, tgt_w = tgt_sizes[i]
+
+            if temporal_pos_emb:
+                if temporal_ids_flatten[i] == -1:
+                    pos_embed_temporal.append(torch.zeros(self.embed_dim, dtype=torch.float32, device="cpu"))
+                else:
+                    pos_embed_temporal.append(self.temporal_pos_embed[temporal_ids_flatten[i]].to(torch.float32))  # D
```
nikita-savelyevv, line 2061: Where is `self.temporal_pos_embed` defined?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -1979,26 +2000,57 @@ def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
                 vision_hidden_states.append(dummy_feature)
         return vision_hidden_states
 
-    def resampling(self, x, tgt_sizes):
+    def resampling(self, x, tgt_sizes, temporal_ids=None):
+        from itertools import chain
```
nikita-savelyevv, line 2029: Should be imported at the top of the file.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -1979,26 +2000,57 @@ def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
...
+
         bs = x.shape[0]
 
         patch_len = tgt_sizes[:, 0] * tgt_sizes[:, 1]
 
         self._adjust_pos_cache(tgt_sizes)
 
+        temporal_pos_emb = False
```
nikita-savelyevv, line 2037: For me these names are a bit confusing: `temporal_pos_emb`, `pos_embed_temporal`, `self.temporal_pos_embed`, `temporal_embed`. I would suggest to rename these variable to something more meaningful. For example, `use_temporal_pos_embed` instead of `temporal_pos_emb`.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -1979,26 +2000,57 @@ def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
...
+                    pos_embed_temporal.append(self.temporal_pos_embed[temporal_ids_flatten[i]].to(torch.float32))  # D
+
             pos_embed.append(self._pos_embeds[:tgt_h, :tgt_w, :].reshape((tgt_h * tgt_w, -1)))  # patches * D
             key_padding_mask[i, patch_len[i] :] = True
 
         pos_embed = torch.nn.utils.rnn.pad_sequence(pos_embed, batch_first=True, padding_value=0.0).permute(
             1, 0, 2
         )  # BLD => L * B * D
-        res = torch.from_numpy(self.resampler(image_feature=x, pos_embed=pos_embed, key_padding_mask=key_padding_mask))
+        if temporal_pos_emb:
```
nikita-savelyevv, line 2069:
```suggestion
        if len(pos_embed_temporal) > 0:
```

tests/openvino/test_exporters_cli.py
```diff
@@ -743,13 +765,13 @@ def _openvino_export(self, model_name: str, task: str, model_kwargs: Dict = None
...
         if is_transformers_version("<", "4.49"):
-            expected = {"llama4", "qwen2_5_vl", "phi4mm"}
+            expected = {"llama4", "qwen2_5_vl", "phi4mm", "minicpmv4", "minicpmv4_5"}
         elif is_transformers_version("<", "4.51"):
             expected = {"llama4", "phi4mm"}
         elif is_transformers_version("<", "4.52"):
             expected = set()
         else:
-            expected = {"llava-qwen2", "phi3_v", "phi4mm", "minicpmo"}
+            expected = {"llava-qwen2", "phi3_v", "phi4mm", "minicpmo", "minicpmv4", "minicpmv4_5"}
```
nikita-savelyevv, line 774: From this, I get an understanding that minicpmv4/minicpmv4_5 are supported for `transformers 4.49 .. 4.51`. Is this correct? If so, please set `MIN_TRANSFORMERS_VERSION = "4.49.0"` and `MAX_TRANSFORMERS_VERSION = "4.51.3"` for  `MiniCPMVOpenVINOConfig`.

optimum/intel/openvino/modeling_visual_language.py
openvino-dev-samples: Its default for this model: https://huggingface.co/openbmb/MiniCPM-V-4_5/blob/main/resampler.py#L100

and not initialized from config: https://huggingface.co/openbmb/MiniCPM-V-4_5/blob/main/modeling_minicpmv.py#L50

optimum/intel/openvino/modeling_visual_language.py
openvino-dev-samples: its copied from original model: https://huggingface.co/openbmb/MiniCPM-V-4_5/blob/main/resampler.py#L216

optimum/intel/openvino/modeling_visual_language.py
openvino-dev-samples: i only created `temporal_embed`, and other of them are from original modeling file directly.

optimum/intel/openvino/modeling_visual_language.py
openvino-dev-samples: this imports is used by minicpmv only, so i think it can be left here. e.g https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_visual_language.py#L1229

tests/openvino/test_exporters_cli.py
openvino-dev-samples: I dont see any limitation on these 2 models. They can share same same version of transformers with minicpm-v-2.6

optimum/intel/openvino/modeling_visual_language.py
openvino-dev-samples: its copied from original model: https://huggingface.co/openbmb/MiniCPM-V-4_5/blob/main/modeling_minicpmv.py#L94

optimum/intel/openvino/modeling_visual_language.py
openvino-dev-samples: Fixed

optimum/intel/openvino/modeling_visual_language.py
openvino-dev-samples: Updated to align with original model.

optimum/intel/openvino/modeling_visual_language.py
openvino-dev-samples: done

Issue #1409, opened 2025-08-06
Model mistralai/Mistral-7B-Instruct-v0.3 doesn't have defined tasks


dtrawins:
While exporting model mistralai/Mistral-7B-Instruct-v0.3 there is not default task defined:
`optimum-cli export openvino -m mistralai/Mistral-7B-Instruct-v0.3 output`
```
Traceback (most recent call last):
  File "/home/dtrawins/model_server/demos/common/export_models/.venv/lib/python3.12/site-packages/optimum/exporters/openvino/__main__.py", line 83, in infer_task
    task = TasksManager._infer_task_from_model_name_or_path(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dtrawins/model_server/demos/common/export_models/.venv/lib/python3.12/site-packages/optimum/exporters/tasks.py", line 1780, in _infer_task_from_model_name_or_path
    raise KeyError(f"Could not find the proper task name for the model {model_name_or_path}.")
KeyError: 'Could not find the proper task name for the model mistralai/Mistral-7B-Instruct-v0.3.'
```

Could it be defined like for other public models list meta-llama/Llama-3.2-3B-Instruct ?

PR #1413, opened 2025-08-07
Add custom task inferring logic for mistralai/Mistral-7B-Instruct-v0.3

nikita-savelyevv:
# What does this PR do?

The reason behind this change is that the task for `mistralai/Mistral-7B-Instruct-v0.3` can't be automatically inferred.

Fixes https://github.com/huggingface/optimum-intel/issues/1409

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -89,9 +89,14 @@ def infer_task(
                     library_name=library_name,
                 )
             except KeyError as e:
-                raise KeyError(
-                    f"The task could not be automatically inferred. Please provide the argument --task with the relevant task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"
-                )
+                try:
+                    config = AutoConfig.from_pretrained(model_name_or_path)
+                    if "MistralForCausalLM" in config.architectures:
+                        task = "text-generation-with-past"
+                except Exception:
+                    raise KeyError(
+                        f"The task could not be automatically inferred. Please provide the argument --task with the relevant task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"
+                    )
             except RequestsConnectionError as e:
                 raise RequestsConnectionError(
                     f"The task could not be automatically inferred as this is available only for models hosted on the Hugging Face Hub. Please provide the argument --task with the relevant task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"


optimum/exporters/openvino/__main__.py
```diff
@@ -89,9 +89,14 @@ def infer_task(
                     library_name=library_name,
                 )
             except KeyError as e:
-                raise KeyError(
-                    f"The task could not be automatically inferred. Please provide the argument --task with the relevant task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"
-                )
+                try:
+                    config = AutoConfig.from_pretrained(model_name_or_path)
+                    if "MistralForCausalLM" in config.architectures:
```
echarlaix, line 94: why not do this by default for all models (the way to check for this will be different depending of `library_name`), could also make sense to deprecate inferring task

nikita-savelyevv: > why not do this by default for all models (the way to check for this will be different depending of library_name), could also make sense to deprecate inferring task

Do you suggest to maintain a list of all supported architecture names like "MistralForCausalLM" and check for those?

The problem with Mistral-7B-Instruct-v0.3 is that its library_name property is set to "vllm" https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.3/blob/main/README.md?code=true#L2. That's why the regular way does not work for it. So I would consider it a special case. For other models the current approach works fine so I don't think there is a strong need to change things in general. What do you think?

echarlaix: Thanks for the explanation!

> Do you suggest to maintain a list of all supported architecture names like "MistralForCausalLM" and check for those?

I was thinking to check for the `config.architectures` directly for all models in the future and load the model using this class (removing the need to infer the task). So for Mistral-7B-Instruct-v0.3  we would load the model using `MistralForCausalLM`
https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.3/blob/main/config.json#L3, what do you think ?

nikita-savelyevv: Thanks for the explanation, probably this indeed would be more convenient. However the `task` variable is used in quite a few places, so as I understand such transition would take some significant effort.. Unfortunately, I can't dive into it right now 😞 What about adding a note here that if we transition to `config.architectures` approach then such workaround as added in this PR won't be needed?

echarlaix: @nikita-savelyevv no worries, we can merge if no time, feel free to merge!

Issue #1410, opened 2025-08-06
Exporting model NousResearch/Hermes-3-Llama-3.1-8B doesn't create tokenizer and detokenizers openvino models


dtrawins:
While exporting the model via:
`optimum-cli export openvino -m NousResearch/Hermes-3-Llama-3.1-8B output`

there is not tokenizer and detokenizer created:
```
ls -1
additional_chat_templates
chat_template.jinja
config.json
generation_config.json
openvino_model.bin
openvino_model.xml
special_tokens_map.json
tokenizer_config.json
tokenizer.json
```
It is required to run also
`convert_tokenizer -o output NousResearch/Hermes-3-Llama-3.1-8B --with-detokenizer`

Could then be exported automatically by optimum-cli?

PR #1414, opened 2025-08-07
Fix tokenizer conversion

nikita-savelyevv:
# What does this PR do?

Fix tokenizer not being exported to OpenVINO format for some models.

Fixes https://github.com/huggingface/optimum-intel/issues/1410

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/utils.py
+++ b/optimum/exporters/openvino/utils.py
@@ -357,11 +357,10 @@ COMPLEX_CHAT_TEMPLATES = {
 
 def set_simplified_chat_template(ov_tokenizer_model, processor_chat_template=None):
     tokenizer_chat_template = None
+    if processor_chat_template is not None:
+        ov_tokenizer_model.set_rt_info(processor_chat_template, "chat_template")
     if ov_tokenizer_model.has_rt_info("chat_template"):
         tokenizer_chat_template = ov_tokenizer_model.get_rt_info("chat_template")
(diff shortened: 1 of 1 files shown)


nikita-savelyevv:
To be honest I don't fully understand what the problem is. It fails on the expression `tokenizer_chat_template in COMPLEX_CHAT_TEMPLATES` with `'NoneType' object is not callable`. The type of `tokenizer_chat_template` is `OVAny` and `COMPLEX_CHAT_TEMPLATES` is a dict. Perhaps, it fails somewhere in the overloaded `__eq__` for `OVAny`, but at the same time it does not fail if dict keys are casted to a tuple or a list 🤔 

@apaniukov Do you have some ideas regarding what's going on here?

optimum/exporters/openvino/utils.py
```diff
@@ -362,8 +362,15 @@ def set_simplified_chat_template(ov_tokenizer_model, processor_chat_template=Non
     if processor_chat_template is not None:
         tokenizer_chat_template = processor_chat_template
         ov_tokenizer_model.set_rt_info(processor_chat_template, "chat_template")
```
apaniukov, line 364:
```suggestion
    if processor_chat_template is not None:
        ov_tokenizer_model.set_rt_info(processor_chat_template, "chat_template")
        tokenizer_chat_template = ov_tokenizer_model.get_rt_info("chat_template")
```

PR #1415, opened 2025-08-07
[OpenVINO]to support qwen3 embedding/rerank

openvino-dev-samples:
1. to support Qwen3 embedding
2. to support Qwen3 rerank

### DIFF
--- a/tests/openvino/test_exporters_cli.py
+++ b/tests/openvino/test_exporters_cli.py
@@ -115,13 +115,6 @@ class OVCLIExportTestCase(unittest.TestCase):
             ]
         )
 
-    if is_transformers_version(">=", "4.51"):
-        SUPPORTED_ARCHITECTURES.extend(
-            [
-                ("feature-extraction", "qwen3"),
-            ]
-        )
-
     if is_transformers_version(">=", "4.54"):
         SUPPORTED_ARCHITECTURES.extend(
             [


openvino-dev-samples:
@IlyasMoutawwakil could you help to check, thanks

nikita-savelyevv (commented):
Thanks for the contribution! I suppose some tests are needed.

optimum/exporters/openvino/model_configs.py
```diff
@@ -202,6 +201,14 @@ def init_model_configs():
...
         "AutoModelForImageTextToText",
     )
+    TasksManager._CUSTOM_CLASSES[("pt", "qwen3", "text-classification")] = (
+        "transformers",
+        "AutoModelForCausalLM",
+    )
+    TasksManager._CUSTOM_CLASSES[("pt", "qwen3_moe", "text-classification")] = (
+        "transformers",
+        "AutoModelForCausalLM",
+    )
```
IlyasMoutawwakil: why using CausalLM for text classification ?

IlyasMoutawwakil (approved):
thanks for the addition, just one question and let's add some tests please 🤗

optimum/exporters/openvino/model_configs.py
openvino-dev-samples: https://huggingface.co/Qwen/Qwen3-Reranker-0.6B#:~:text=model%20%3D%20AutoModelForCausalLM.from_pretrained(%22Qwen/Qwen3%2DReranker%2D0.6B%22).eval()

IlyasMoutawwakil: but that is not what we mean with "task" in optimum (or transformers) 😅 that is more of how to use the model on the frontend if you see what I mean, the underlying "task" is still text generation. You can do the same thing (use it for classification) for almost any text generation or text2text generation.

openvino-dev-samples: > but that is not what we mean with "task" in optimum (or transformers) 😅 that is more of how to use the model on the frontend if you see what I mean, the underlying "task" is still text generation. You can do the same thing (use it for classification) for almost any text generation or text2text generation.

so what class do you suggest to use for text classification in this case? This implementation bring me the correct result for rerank task

IlyasMoutawwakil: let me re explain, basically any decoder or encoder-decoder model can be used for "text classification" while using it as a text generation (you simply prompt the model to reply with one of the options), for example the same was being done with t5 like models when it was first released. But for the exporter this model is just a normal decoder model, that can be used for "text calssification" with special prompting and chat templating.
The distinction here is between what the exporter cares about about and calls text classification, and how the user can use that model.
Another example is using Llama for translation, you dont actually export llama with text2text generation task just because you are gonna use it for translation (which is a text2text synonym).

IlyasMoutawwakil: simply remove the custom task registration and replace AutoModelFprCausalLM with OVMofemForCausalLM in the snippet you shared with me and notice that it will still work the same way (without these line 😅)

openvino-dev-samples: > simply remove the custom task registration and replace AutoModelFprCausalLM with OVMofemForCausalLM in the snippet you shared with me and notice that it will still work the same way (without these line 😅)

As my test, it wouldn't. My purpose is to export model with this ` optimum-cli export openvino --model Qwen/Qwen3-Reranker-0.6B --trust-remote-code --task text-classification --weight-format fp16 qwen3_reranker_0.6B_ov ` and the error messages with your suggestion are 
```
batch_scores = model(**inputs).logits[:, -1, :]
IndexError: too many indices for tensor of dimension 2
```

IlyasMoutawwakil: yes because the model should be exported with the text-generation task 😅
it is ultimately a text generation model... just read the tags on the model page.

IlyasMoutawwakil: Let me ask you a simple question and let's see where this method takes us
so by registering
```
    TasksManager._CUSTOM_CLASSES[("pt", "qwen3", "text-classification")] = (
        "transformers",
        "AutoModelForCausalLM",
    )
```
we can use text-classification with models such as https://huggingface.co/Qwen/Qwen3-Reranker-0.6B
okay now what if someone trained/finetuned a Qwen3ModelForSequenceClassification, which is defined [here](https://github.com/huggingface/transformers/blob/b347e93567a1770fe1ae588ac2fd51898028e6ed/src/transformers/models/qwen3/modeling_qwen3.py#L512), what will happen when  they do `optimum-cli export -m finetuned-model --task text-classification` ? it will be loaded with AutoModelForCausalLM 😅 and exported with a next token prediction head 😅 do you see the problem there ?

IlyasMoutawwakil: Here is one example https://huggingface.co/Skywork/Skywork-Reward-V2-Qwen3-0.6B
This is a "true" text classification qwen3 model, what will happen when you export it on your branch ? what AutoModelForxxx class will load it ?

nikita-savelyevv (approved):

IlyasMoutawwakil:
failing error is probably related to diffusers 0.35

optimum/exporters/openvino/model_configs.py
```diff
@@ -304,21 +303,46 @@ def patch_model_for_export(
...
     DUMMY_PKV_GENERATOR_CLASS = GemmaDummyPastKeyValuesGenerator
     NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
 
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        if self.task in ["feature-extraction"]:
+            common_inputs = {
+                "input_ids": {0: "batch_size", 1: "sequence_length"},
+                "attention_mask": {0: "batch_size", 1: "sequence_length"},
+            }
```
IlyasMoutawwakil, line 330: does this mean position ids can't be passed with feature extraction ?

openvino-dev-samples: yes, OVModelForFeatureExtraction do not support position ids as inputs

IlyasMoutawwakil: I think we should add it, because looking at the base model (Qwen3Model) it does take position ids and it generates them when not given, so if we export without position ids (even for feature extraction), the model ends being exported with this formula for position ids https://github.com/huggingface/transformers/blob/41d17178827455e7b6553a7026d71d3036ef719c/src/transformers/models/qwen3/modeling_qwen3.py#L383 which is not always correct (with batched and masked inputs).

IlyasMoutawwakil: @openvino-dev-samples wdyt ?

openvino-dev-samples: > @openvino-dev-samples wdyt ?

Hi, the stage of qwen3-embedding's tokenization do not has position_ids as its output.

IlyasMoutawwakil: thanks for investigating @openvino-dev-samples 
let's go ahead and merge, we can go back to this if ever some inconsistency with transformers is observed.

PR #1406, opened 2025-08-06
Add support transformers v4.54 v4.55

echarlaix:
Fix https://github.com/huggingface/optimum-intel/issues/1404

PR #1428, opened 2025-08-19
Add OpenVINO support for gpt-oss

echarlaix:
Needs for https://github.com/huggingface/optimum-intel/pull/1406 before it can be merged


echarlaix:
Failing tests (nightly tests from openvino nightly and slow tests from transformers main) not related to this PR

helena-intel:
Thanks @echarlaix ! Really happy to see this! I tried `optimum-cli export openvino -m openai/gpt-oss-20b gpt-oss-20b-ov` with this PR on both LNL and SPR and got this error (log from SPR). I updated transformers to 4.55. I have PyTorch 2.7.1+cpu. On SPR I tried in a completely clean Python environment.

```
  File "/home/helena/venv/lib/python3.10/site-packages/transformers/models/gpt_oss/modeling_gpt_oss.py", line 239, in eager_attention_forward
    attn_weights = torch.matmul(query, key_states.transpose(2, 3)) * scaling
RuntimeError: expected scalar type Float but found BFloat16

The above exception was the direct cause of the following exception:

  File "/home/helena/venv/bin/optimum-cli", line 8, in <module>
    sys.exit(main())
  File "/home/helena/venv/lib/python3.10/site-packages/optimum/commands/optimum_cli.py", line 208, in main
    service.run()
  File "/home/helena/venv/lib/python3.10/site-packages/optimum/commands/export/openvino.py", line 583, in run
    main_export(
  File "/home/helena/venv/lib/python3.10/site-packages/optimum/exporters/openvino/__main__.py", line 453, in main_export
    submodel_paths = export_from_model(
  File "/home/helena/venv/lib/python3.10/site-packages/optimum/exporters/openvino/convert.py", line 812, in export_from_model
    export_models(
  File "/home/helena/venv/lib/python3.10/site-packages/optimum/exporters/openvino/convert.py", line 579, in export_models
    export(
  File "/home/helena/venv/lib/python3.10/site-packages/optimum/exporters/openvino/convert.py", line 217, in export
    return export_pytorch(
  File "/home/helena/venv/lib/python3.10/site-packages/optimum/exporters/openvino/convert.py", line 486, in export_pytorch
    ts_decoder = TorchScriptPythonDecoder(model, example_input=dummy_inputs, **ts_decoder_kwargs)
  File "/home/helena/venv/lib/python3.10/site-packages/openvino/frontend/pytorch/ts_decoder.py", line 84, in __init__
    raise RuntimeError(
RuntimeError: Couldn't get TorchScript module by tracing.
Exception:
expected scalar type Float but found BFloat16
Please check correctness of provided 'example_input'. Sometimes models can be converted in scripted mode, please try running conversion without 'example_input'.
```

setup.py
```diff
@@ -52,7 +52,7 @@
     "transformers_stream_generator",
     "einops",
     "tiktoken",
-    "sentence-transformers",
+    "sentence-transformers<5.1.1",
```
helena-intel: If optimum-intel now breaks with latest sentence-transformers we should look into that (not for this PR, but if you have any insights because of this PR that would be useful).

helena-intel (commented):
Thanks @echarlaix ! I could now export the model on SPR. Please note that there are still some known issues with performance and device support for OpenVINO, so this PR functionally enables the model, but it is not yet recommended for actual use with OpenVINO. We should see better performance in an upcoming OpenVINO release.

I noticed an issue with INT4 export. With the default config many layers were ignored because of incompatible group size, with group size 64 and AWQ there was a different error. I am checking that.

setup.py
echarlaix: fixed in https://github.com/huggingface/optimum-intel/pull/1449 so not needed anymore! thanks a lot for checking @helena-intel !! will merge once the test pass

optimum/exporters/openvino/__main__.py
```diff
@@ -255,11 +259,19 @@ def main_export(
             trust_remote_code=trust_remote_code,
         )
         quantization_config = getattr(config, "quantization_config", None)
+        quant_method = quantization_config.get("quant_method", None) if quantization_config else None
+
+        # mxfp4 quantized model will be dequantized to bf16
+        if quant_method == "mxfp4" and is_transformers_version(">=", "4.55"):
+            dtype = torch.bfloat16
+            loading_kwargs["quantization_config"] = Mxfp4Config(dequantize=True)
```
echarlaix, line 267: Added to ensure the model will be dequantized to bf16

optimum/exporters/openvino/model_configs.py
```diff
@@ -627,6 +627,23 @@ def patch_model_for_export(
...
+        "text-generation",
+        "text-generation-with-past",
+        "text-classification",
+    ],
+    library_name="transformers",
+)
+class GptOssOpenVINOConfig(LlamaOpenVINOConfig):
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, GemmaDummyPastKeyValuesGenerator)
+    DUMMY_PKV_GENERATOR_CLASS = GemmaDummyPastKeyValuesGenerator
+    MIN_TRANSFORMERS_VERSION = "4.55.1"
```
echarlaix, line 644: needs v4.55.1 as we need this fix https://github.com/huggingface/transformers/commit/99404c7098335fd7c50fc5a6b79fd39e253f26e5

setup.py
ljaljushkin: @helena-intel, I was checking accuracy for different INT4 modes and found that to preserve accuracy only experts should be quantized, as it's defined in the original gpt-oss model (https://huggingface.co/openai/gpt-oss-20b/blob/main/config.json#L51-L55).

image

We are considering adding a default config for gpt-oss model, so I guess there won't be issue with incompatible group_size:
```
        dtype='int4', sym=True, group_size=32,
        ignored_scope=nncf.IgnoredScope(patterns=[r".*self_attn.*", r".*router.*"]),
        backup_mode=nncf.BackupMode.NONE 
```

helena-intel: Thanks @ljaljushkin ! It would be great if that can be added as default config. It can always be changed later if needed, for now it is much better than breaking.

nikita-savelyevv: We can specify the the default int4 config right in this PR. The entry below should be added to [_DEFAULT_4BIT_WQ_CONFIGS](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py#L54) 

```
"openai/gpt-oss-20b": {
    "bits": 4,
    "sym": True,
    "group_size": 32,
    "ignored_scope": {
        "patterns": [".*self_attn.*", ".*router.*"],
    },
    "backup_precision": "none",
}
```

@echarlaix could you please add this?

echarlaix: added in https://github.com/huggingface/optimum-intel/pull/1428/commits/5613f9b50472b803e60a34da4f425232eec176aa

echarlaix:
Thanks for taking a look @ljaljushkin! Could you share more info on what's the issue is ? Looks that there is an issue during export https://github.com/huggingface/optimum-intel/blob/b60e4d4866509a1aeea2b7a3f26f2a70bc464354/optimum/exporters/openvino/__main__.py#L481 Do you have this issue only when setting `quantization_config={"bits": 4})` ? I'm not able to reproduce with a smaller model. I'll merge this PR in the meantime but let me know what's the issue is when you have more info and happy to take a look

PR #1394, opened 2025-07-28
VLM Vision Encoder full quantization

nikita-savelyevv:
# What does this PR do?

Collection of calibration data is implemented for vision encoders within VLM pipelines. This allows to run static quantization on vision encoders which demonstrates noticeable performance improvement for some models without a significant generation quality drop. Please see the validation metrics below.

Code example:
```python
from optimum.intel import OVModelForVisualCausalLM, OVPipelineQuantizationConfig, OVWeightQuantizationConfig, \
    OVQuantizationConfig

model = OVModelForVisualCausalLM.from_pretrained(
    "OpenGVLab/InternVL2-1B",
    quantization_config=OVPipelineQuantizationConfig(
        {
            "lm_model": OVWeightQuantizationConfig(bits=4, scale_estimation=True, num_samples=128),
            "vision_embeddings_model": OVQuantizationConfig(bits=8, num_samples=32),
        },
        default_config=OVWeightQuantizationConfig(bits=8, sym=True),
        dataset="contextual",
        trust_remote_code=True,
    ),
    trust_remote_code=True,
)
```
The new argument `default_config` is used to specify quantization for other submodels other than which were given explicitly. 

From now on, when `--quant-mode` is selected during export through CLI, besides a language model a vision encoder will also be quantized. A more reliable setup is combining quantization of vision encoder with 4-bit weight only quantization of a language model. The latter is currently not available through CLI export though.

# Validation

| Model                     | LM Precision | VE Precision            | Other Precision | WWB Similarity  | MME[scene] Acc. | Inference Time (sec.) |
| ------------------------- | ------------ | ----------------------- | --------------- | --------------- | ------------------- | -------------- |
| InternVL2-1B              | bf16         | bf16                    | bf16            | 99.21%          | 80.00%              | 11.61          |
| InternVL2-1B              | w: int4_asym | w: int8_sym             | w: int8_sym     | 82.45%          | 80.50%              | 8.08           |
| InternVL2-1B              | w: int4_asym | w: int8_sym a:int8_asym | w: int8_sym     | 81.95% (-0.50%) | 81.00% (-0.50%)     | 6.77 (1.19x)   |
|                           |              |                         |                 |                 |                     |                |
| InternVL2-4B              | bf16         | bf16                    | bf16            | 99.66%          | 85.50%              | 51.98          |
| InternVL2-4B              | w: int4_asym | w: int8_sym             | w: int8_sym     | 89.10%          | 84.50%              | 33.56          |
[...]

PR #1433, opened 2025-08-21
[OV] Remove local imports from quantization.py

nikita-savelyevv:
# What does this PR do?

Remove local imports from openvino quantization.py.

As discussed in https://github.com/huggingface/optimum-intel/pull/1394#discussion_r2285332499.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -337,7 +337,7 @@ class OVCalibrationDatasetBuilder:
                 )
             else:
                 raise Exception
-        elif _is_diffuser_instance(self.model):
+        elif is_diffusers_available() and isinstance(self.model, OVDiffusionPipeline):
             if isinstance(config.dataset, str):
                 dataset_name = config.dataset
                 dataset_metadata = PREDEFINED_SD_DATASETS[dataset_name]
@@ -356,9 +356,11 @@ class OVCalibrationDatasetBuilder:
                 )
 
             return self.build_from_dataset(config, dataset)
-        elif isinstance(
-            self.model, (OVModelForFeatureExtraction, OVModelForMaskedLM, OVModelForSeq2SeqLM)
-        ) or _is_sentence_transformer_instance(self.model):
+        elif (
+            isinstance(self.model, (OVModelForFeatureExtraction, OVModelForMaskedLM, OVModelForSeq2SeqLM))
+            or is_sentence_transformers_available()
+            and isinstance(self.model, OVSentenceTransformer)
+        ):
             if isinstance(config.dataset, str):
                 dataset_metadata = PREDEFINED_LANGUAGE_DATASETS[config.dataset]
                 dataset = self.load_dataset(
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/quantization.py
```diff
@@ -336,7 +337,7 @@ def build_from_quantization_config(self, config: OVQuantizationConfigBase) -> OV
                 )
             else:
                 raise Exception
-        elif is_diffusers_available() and isinstance(self.model, OVDiffusionPipeline):
+        elif _is_diffuser_instance(self.model):
```
IlyasMoutawwakil: I would leave `is_diffusers_available() and isinstance(self.model, OVDiffusionPipeline):`, notice that this function would fail if OVDiffusionPipeline is not imported beforehand, and it checks for OVDiffusionPipeline and not DiffusionPipeline

nikita-savelyevv: I believe `is_diffusers_available() and isinstance(self.model, OVDiffusionPipeline)` and `_is_diffuser_instance(self.model)` are equivalent, or am I missing something? It's not possible to call `_is_diffuser_instance()` before `OVDiffusionPipeline` is imported in the beginning of the file, because it happens during import pass.

IlyasMoutawwakil: They are functionally equivalent, since the helper is just wrapping the inline check.
My concern is less about correctness and more about readability and maintainability:
- The function name _is_diffuser_instance suggests it’s checking for a general diffusers.DiffusionPipeline, but in reality it checks for OVDiffusionPipeline (an optimum-intel defined class). That can be surprising for someone reading the code (it was for me).
- By moving the logic into a function, we also hide the fact that it depends on a conditionally imported class. In the inline version, it’s obvious why the is_diffusers_available() check is there. Inside a function, it’s less clear and could mislead someone into thinking the function is completely self-contained (generally functions are).

So the tradeoff here isn’t correctness vs. bugs, but brevity vs. clarity. Keeping the check inline makes it easier for anyone jumping into this file to quickly understand what’s happening, without needing to trace through helper functions.

optimum/intel/openvino/quantization.py
```diff
@@ -1923,6 +1894,14 @@ def _add_nncf_version_flag(model: openvino.Model) -> openvino.Model:
     return model
 
 
+def _is_diffuser_instance(model: Any) -> bool:
+    return is_diffusers_available() and isinstance(model, OVDiffusionPipeline)
+
+
+def _is_sentence_transformer_instance(model: Any) -> bool:
+    return is_sentence_transformers_available() and isinstance(model, OVSentenceTransformer)
```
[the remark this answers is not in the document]
IlyasMoutawwakil: @nikita-savelyevv copilot has the same concern 🤗

optimum/intel/openvino/quantization.py
nikita-savelyevv: I see your point. Thanks for the explanation! I suppose the same holds for `_is_sentence_transformer_instance()`?

IlyasMoutawwakil: yes

optimum/intel/openvino/quantization.py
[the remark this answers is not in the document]
nikita-savelyevv: @IlyasMoutawwakil In my opinion it is not the same 🙂

optimum/intel/openvino/quantization.py
nikita-savelyevv: I've applied the changes

optimum/intel/openvino/quantization.py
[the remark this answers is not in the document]
IlyasMoutawwakil: it is the same conceptual issue (scopes), but copilot suggests a dirty hack to make it less problematic

PR #1434, opened 2025-08-26
fix ernie 4.5 accuracy issue in test case

openvino-dev-samples:
To turn off the KVcache compression in test case


echarlaix (commented):
Thanks a lot for taking the time to look into this @openvino-dev-samples ! Do you have more information on why this is needed ? Also shouldn't we ensure this parameter is always set for ernie 4.5 (during inference as well) ?

openvino-dev-samples:
> Thanks a lot for taking the time to look into this @openvino-dev-samples ! Do you have more information on why this is needed ? Also shouldn't we ensure this parameter is always set for ernie 4.5 (during inference as well) ?

By

> Thanks a lot for taking the time to look into this @openvino-dev-samples ! Do you have more information on why this is needed ? Also shouldn't we ensure this parameter is always set for ernie 4.5 (during inference as well) ?

By default, ov is using int8 precious to keep its kvcache in memory. which could lead accuracy drop during decoding stage.

echarlaix:
If that works with you let's remove `ernie4_5` support in https://github.com/huggingface/optimum-intel/pull/1406 (was never officially supported as for now transformers is limited to < v4.54) and add back support later on once fixed

PR #1438, opened 2025-09-02
Fix disk by distributing tests

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting


IlyasMoutawwakil:
@nikita-savelyevv the reason is probably because it downloads all models in our testing suite, which is a lot.
we will have to dismantele it (it takes 4 hours)

nikita-savelyevv:
> @nikita-savelyevv the reason is probably because it downloads all models in our testing suite, which is a lot

But they do work fine on the main branch, and as I understand in this PR no new models have been added, just moved to other files. So I would expect it to pass here too.

> we will have to dismantele it (it takes 4 hours)

Yeah, some kind of a similar splitting into parts would be great. Not critical though in my opinion.

IlyasMoutawwakil:
Adding 'openvino-test' triggers very long, nightly and slow runs of our CI
I have created a new label for nightly tests and distributed and specified them clearly as well to not explode the ci runners.
There are openvino nightly tests, runnigneach file in tests/openvino, and nncf develop tests, only targeting export and quantization.
I also created a new label for slow tests.
@nikita-savelyevv @echarlaix wdyt ?

IlyasMoutawwakil:
I added the missing combinations to nightly tests, there's now some overlap between the fast+slow ci and nightly but that shouldn't be an issue as now there's a specific label for nightly and one can be triggered without the other.
let's only add labels that make sense in our PR, for example adding a new model export shouldn't require running nightly tests.
we can trigger nightly tests on a PR when something is being implemented that's only available in nightly, or when fixing some failing nightly tests. slow tests are mostly for generative tasks.

nikita-savelyevv (approved):

PR #1439, opened 2025-09-02
Benchmarks updates

ezelanza:
# What does this PR do?

Fixes # 

## Before submitting


notebooks/openvino/vision_language_quantization.ipynb
```diff
@@ -0,0 +1,842 @@
...
+      "What is on the flower?\n",
+      "Assistant: A bee is on the flower.\n"
+     ]
+    }
+   ],
+   "source": [
+    "from optimum.intel import OVModelForVisualCausalLM\n",
+    "\n",
+    "\n",
+    "model_ov = OVModelForVisualCausalLM.from_pretrained(model_id)\n",
```
nikita-savelyevv:
```suggestion
    "model_ov = OVModelForVisualCausalLM.from_pretrained(model_id, load_in_8bit=False)\n",
```
This is more for future-proofing. In case somebody wants to try it out with a bigger model.

notebooks/openvino/vision_language_quantization.ipynb
```diff
@@ -0,0 +1,842 @@
...
+      "Benchmarking SmolVLM2-256M-int8 on GPU.2...\n",
+      "Latency: 0.3563s | Images/sec: 2.81 | Tokens/sec: 2481.35 | Model size: 248.86 MB\n"
+     ]
+    }
+   ],
+   "source": [
+    "# --- Local models ---\n",
+    "models = {\n",
+    "    \"SmolVLM2-256M (full)\": \"smolvlm_ov\",\n",
+    "    \"SmolVLM2-256M-int8\": \"smolvlm_int8\"\n",
```
nikita-savelyevv:
```suggestion
    "    \"SmolVLM2-256M (full)\": fp32_model_path,\n",
    "    \"SmolVLM2-256M-int8\": int8_model_path\n",
```

nikita-savelyevv (commented):
Benchmarking looks good 👍 Below are some minor suggestions I've found while experimenting with this notebook.

notebooks/openvino/vision_language_quantization.ipynb
```diff
@@ -0,0 +1,842 @@
...
+    "    end = time.time()\n",
+    "\n",
+    "    avg_latency = (end - start) / nb_pass\n",
+    "\n",
+    "    # --- Throughput calculations ---\n",
+    "    batch_size = inputs[\"pixel_values\"].shape[0] if \"pixel_values\" in inputs else 1\n",
+    "    num_tokens = outputs.shape[-1]  # sequence length\n",
+    "\n",
+    "    images_per_sec = batch_size / avg_latency\n",
+    "    tokens_per_sec = num_tokens / avg_latency\n",
```
nikita-savelyevv: This is a bit of oversimplification, because such measurement does not take into account how long each actual model took to infer. For example if we increase the speed of vision encoder only, it will also influence the `token_per_sec` value. Perhaps it is ok for demo purposes though.

ezelanza: Agree that this benchmark isn't showing the separate parts improvements. I was thinking more about showing the overall impact since this is what ends up affecting the experience of the demo. I'm open to modify it if you think it doesn't make sense or if we can go deep and show specific benefits, but again my idea is to show overall performance.

nikita-savelyevv: I guess we have here a tradeoff between granularity of the results we want to demonstrate and the effort it takes to compute them. In general, the latency solely shows the overall impact, so we could state it only.

When higher granularity is introduced, I personally would expect it to reflect the changes in the individual pipeline components since we have a clear distinction here: vision embedding model processes images and language model generates tokens.

If you interested, I could jot down some logic that would allow to extract inference time of individual components. It would come with a price of some amount of hacky code though 🙂

ezelanza: That would be great! Appreciate your help here

nikita-savelyevv: @ezelanza please merge this one into your branch https://github.com/ezelanza/optimum-intel/pull/1

ezelanza: Done

notebooks/openvino/vision_language_quantization.ipynb
```diff
@@ -0,0 +1,904 @@
...
+    "\n",
+    "\n",
+    "def benchmark(model, inputs, model_dir: str, nb_pass=10, warmup=4,max_tokens=50):\n",
+    "    \"\"\"\n",
+    "    Benchmark an OV visual causal LM model.\n",
+    "\n",
+    "    Returns a dict with:\n",
+    "    - avg_latency_sec\n",
+    "    - images_per_sec\n",
+    "    - tokens_per_sec\n",
```
echarlaix: needs to be updated

notebooks/openvino/vision_language_quantization.ipynb
```diff
@@ -0,0 +1,904 @@
...
+      "Benchmarking SmolVLM2-256M-int8 on GPU.2...\n",
+      "Latency: 0.3563s | Images/sec: 2.81 | Tokens/sec: 2481.35 | Model size: 248.86 MB\n"
+     ]
+    }
+   ],
+   "source": [
+    "# --- Local models ---\n",
+    "models = {\n",
+    "    \"SmolVLM2-256M (full)\": fp32_model_path,\n",
+    "    \"SmolVLM2-256M-int8\": int8_model_path\n",
```
echarlaix, line 483: for the benchmark wouldn't it make sense to use a model with quantized activation as well ? (so loaded with `static_8bit` or `ppl_q` quantization config in our case)

echarlaix (approved):
LGTM, thanks a lot @ezelanza and @nikita-savelyevv

notebooks/openvino/vision_language_quantization.ipynb
echarlaix:
```suggestion
    "    - image_throughput\n",
    "    - first_token_throughput\n",
    "    - second_token_throughput\n",
```

notebooks/openvino/vision_language_quantization.ipynb
```diff
@@ -397,6 +481,44 @@
...
+      "Average Inference latency: 2.1692 seconds\n"
+     ]
+    }
+   ],
+   "source": [
+    "# 2 rounds of warm-up and 5 rounds of inference (to measure the average)\n",
+    "avg_latency = elapsed_time(q_model, inputs, nb_pass=5, warmup=2)\n",
+    "print(f\"Average Inference latency: {avg_latency:.4f} seconds\")"
+   ]
+  },
```
echarlaix: failing tests coming from `get_model_size` and `elapsed_time`  being defined later, I think we can remove @ezelanza 

```suggestion
```

notebooks/openvino/vision_language_quantization.ipynb
```diff
@@ -507,6 +629,237 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "# --- Local models ---\n",
+    "models = {\n",
+    "    \"SmolVLM2-256M (full)\": fp32_model_path,\n",
+    "    \"SmolVLM2-256M-int8\": int8_model_path\n",
+    "}\n",
+    "devices = [\"CPU\", \"GPU.0\",\"GPU.1\",\"GPU.2\"]\n",
```
echarlaix: shouldn't we iterate on the `core.available_devices` instead ? (otherwise will fail for cases where these devices are note available for example)

notebooks/openvino/vision_language_quantization.ipynb
ezelanza: Sure!

notebooks/openvino/vision_language_quantization.ipynb
ezelanza: That's correct. Updated!

PR #1442, opened 2025-09-03
Fix regexp error with searching onnx model

sbalandi:
Fix issue for onnx models with:
```
Could not infer whether the model was already converted or not to the OpenVINO IR, keeping `export=False`.
nothing to repeat at position 0
```

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -479,7 +479,7 @@ class OVBaseModel(OptimizedModel):
 
             ov_files = _find_files_matching_pattern(
                 model_dir,
-                pattern=cls._search_pattern if not kwargs.get("from_onnx", False) else "*.onnx",
+                pattern=cls._search_pattern if not kwargs.get("from_onnx", False) else ".*\.onnx$",
                 subfolder=subfolder,
                 use_auth_token=token,
                 revision=revision,


nikita-savelyevv:
@sbalandi thanks for the fix! Which model does this reproduce on?

sbalandi:
> @sbalandi thanks for the fix! Which model does this reproduce on?

models: phi-3.5-mini-instruct, llama-3.2-1b-instruct
task CVS-172956

optimum/intel/openvino/modeling_base.py
```diff
@@ -479,7 +479,7 @@ def from_pretrained(
 
             ov_files = _find_files_matching_pattern(
                 model_dir,
-                pattern=cls._search_pattern if not kwargs.get("from_onnx", False) else "*.onnx",
+                pattern=cls._search_pattern if not kwargs.get("from_onnx", False) else ".*.onnx",
```
nikita-savelyevv:
```suggestion
                pattern=cls._search_pattern if not kwargs.get("from_onnx", False) else ".*\.onnx$",
```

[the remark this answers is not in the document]
nikita-savelyevv: The motivation is right, the code suggestion is not.

sbalandi: updated

nikita-savelyevv (approved):

PR #1451, opened 2025-09-25
Remove legacy code needed for transformers < v4.45 

echarlaix:
should have been included in https://github.com/huggingface/optimum-intel/pull/1406 when the min required transformers version was set to v4.45

### DIFF
--- a/tests/openvino/test_exporters_cli.py
+++ b/tests/openvino/test_exporters_cli.py
@@ -164,15 +164,23 @@ class OVCLIExportTestCase(unittest.TestCase):
             "expected_chat_template": True,
             "remote_code": False,
         },
-        "glm": {  # transformers, no processor, no simplified chat template
-            "num_tokenizers": 2 if is_tokenizers_version("<", "0.20") or is_openvino_version(">=", "2024.5") else 0,
-            "task": "text-generation-with-past",
-            "expected_chat_template": True,
-            "simplified_chat_template": False,
-            "processor_chat_template": False,
-            "remote_code": True,
-        },
     }
+
+    if is_transformers_version(">=", "4.46"):
+        TOKENIZER_CHAT_TEMPLATE_TESTS_MODELS.update(
+            {
+                "glm": {  # transformers, no processor, no simplified chat template
+                    "num_tokenizers": 2
+                    if is_tokenizers_version("<", "0.20") or is_openvino_version(">=", "2024.5")
+                    else 0,
+                    "task": "text-generation-with-past",
+                    "expected_chat_template": True,
+                    "simplified_chat_template": False,
+                    "processor_chat_template": False,
+                    "remote_code": True,
+                },
+            }
+        )
 
     if is_transformers_version("<", "4.54"):
         TOKENIZER_CHAT_TEMPLATE_TESTS_MODELS.update(


tests/openvino/test_exporters_cli.py
```diff
@@ -178,7 +166,7 @@ class OVCLIExportTestCase(unittest.TestCase):
         },
     }
 
-    if is_transformers_version("<", "4.45"):
+    if is_transformers_version(">=", "4.46"):
```
echarlaix, line 169: mistake introduced in https://github.com/huggingface/optimum-intel/pull/1406 glm is supported for >= v4.46 https://github.com/huggingface/optimum-intel/blob/b60e4d4866509a1aeea2b7a3f26f2a70bc464354/optimum/exporters/openvino/model_configs.py#L3701
todo : add test automatically depending on min / max version to avoid this

optimum/exporters/openvino/__main__.py
```diff
@@ -281,15 +279,8 @@ def main_export(
                 f"Asked to export a {model_type} model for the task {task}{autodetected_message}, but the Optimum OpenVINO exporter only supports the tasks {', '.join(model_tasks.keys())} for {model_type}. Please use a supported task. Please open an issue at https://github.com/huggingface/optimum/issues if you would like the task {task} to be supported in the ONNX export for {model_type}."
             )
 
-        if (
-            is_transformers_version(">=", "4.36")
-            and is_transformers_version("<=", "4.45.0")
```
echarlaix, line 286: looks like this is an error (for transformers <= v4.45.0 so patch release not included) also looks like it should be v4.42 https://github.com/huggingface/optimum-onnx/blob/fafa2a764401348ee66f17f5b472d58bc67ed520/optimum/exporters/onnx/__main__.py#L335
removing

optimum/exporters/openvino/model_patcher.py
```diff
@@ -83,10 +82,8 @@ def patch_model_with_bettertransformer(model):
     if hasattr(model, "use_bettertransformer") and model.use_bettertransformer is True:
         return model
 
-    if (
-        getattr(model.config, "model_type") in {"gpt_bigcode", "llama", "gemma"}
-        and is_transformers_version(">=", "4.38")
-        and is_openvino_version("<", "2024.1.0-14612")
+    if getattr(model.config, "model_type") in {"gpt_bigcode", "llama", "gemma"} and is_openvino_version(
+        "<", "2024.1.0-14612"
```
echarlaix, line 86: todo : do the same for old openvino versions (for < v2025.1.0)

https://github.com/huggingface/optimum-intel/blob/v1.25.2/setup.py#L69

optimum/exporters/openvino/stateful.py
```diff
@@ -164,7 +161,7 @@ def make_stateful(
         if num_beams_and_batch is not None:
             input = ov_model.input(kv_name_pair[0])
             shape = input.get_partial_shape()
-            shape[batch_dim] = num_beams_and_batch * num_attention_heads
```
echarlaix, line 167: `num_attention_heads` will always be 1 so removing

IlyasMoutawwakil (approved):
LGTM ! thanks Ella

PR #1454, opened 2025-09-29
[OpenVINO] Support openbmb/MiniCPM-o-2_6 for image-text-to-text task

rkazants:
# What does this PR do?

Command to export the model:
```sh
optimum-cli export openvino -m openbmb/MiniCPM-o-2_6 MiniCPM-o-2_6 --task=image-text-to-text --trust-remote-code
```

Example of inference:
```py
from PIL import Image
import requests

model_id="openbmb/MiniCPM-o-2_6"

prompt= "<|im_start|>user\n(<image>./</image>)\nWhat is in the image?<|im_end|>\n<|im_start|>assistant\n"

inputs = processor([prompt], [image], return_tensors="pt")

result  = model.generate(**inputs, max_new_tokens=20)

```

## Before submitting

### DIFF
--- a/docs/source/openvino/models.mdx
+++ b/docs/source/openvino/models.mdx
@@ -93,7 +93,7 @@ Here is the list of the supported architectures :
 - Marian
 - MiniCPM
 - MiniCPM3
-- MiniCPMO (image-text-to-text)
+- MiniCPM-o
 - MiniCPMV
 - Mistral
 - Mixtral
--- a/docs/source/openvino/models.mdx
+++ b/docs/source/openvino/models.mdx
@@ -93,7 +93,7 @@ Here is the list of the supported architectures :
 - Marian
 - MiniCPM
 - MiniCPM3
-- MiniCPM-o
+- MiniCPM-o 
 - MiniCPMV
 - Mistral
 - Mixtral
--- a/docs/source/openvino/models.mdx
+++ b/docs/source/openvino/models.mdx
@@ -93,6 +93,7 @@ Here is the list of the supported architectures :
 - Marian
 - MiniCPM
 - MiniCPM3
+- MiniCPMO (image-text-to-text)
 - MiniCPMV
 - Mistral
 - Mixtral
--- a/docs/source/openvino/models.mdx
+++ b/docs/source/openvino/models.mdx
@@ -93,7 +93,7 @@ Here is the list of the supported architectures :
 - Marian
 - MiniCPM
 - MiniCPM3
-- MiniCPM-o 
+- MiniCPM-o
 - MiniCPMV
 - Mistral
 - Mixtral
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -2768,7 +2768,6 @@ class MiniCPMVConfigBehavior(str, enum.Enum):
 
 
 @register_in_tasks_manager("minicpmv", *["image-text-to-text"], library_name="transformers")
-@register_in_tasks_manager("minicpmo", *["image-text-to-text"], library_name="transformers")
 class MiniCPMVOpenVINOConfig(BaseVLMOpenVINOConfig):
     SUPPORTED_BEHAVIORS = [model_type.value for model_type in MiniCPMVConfigBehavior]
     NORMALIZED_CONFIG_CLASS = NormalizedVisionConfig
@@ -2893,6 +2892,11 @@ class MiniCPMVOpenVINOConfig(BaseVLMOpenVINOConfig):
             return MiniCPMVResamplerModelPatcher(self, model, model_kwargs)
 
         return super().patch_model_for_export(model, model_kwargs)
+
+
+@register_in_tasks_manager("minicpmo", *["image-text-to-text"], library_name="transformers")
+class MiniCPMOOpenVINOConfig(MiniCPMVOpenVINOConfig):
+    MAX_TRANSFORMERS_VERSION = "4.51.0"
 
 
 class Phi3VisionConfigBehavior(str, enum.Enum):
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -2896,6 +2896,7 @@ class MiniCPMVOpenVINOConfig(BaseVLMOpenVINOConfig):
 
 @register_in_tasks_manager("minicpmo", *["image-text-to-text"], library_name="transformers")
 class MiniCPMOOpenVINOConfig(MiniCPMVOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.43.0"
     MAX_TRANSFORMERS_VERSION = "4.51.0"
 
 
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -2685,7 +2685,7 @@ class MiniCPMVOpenVINOConfig(BaseVLMOpenVINOConfig):
 @register_in_tasks_manager("minicpmo", *["image-text-to-text"], library_name="transformers")
 class MiniCPMOOpenVINOConfig(MiniCPMVOpenVINOConfig):
     MIN_TRANSFORMERS_VERSION = "4.43.0"
-    MAX_TRANSFORMERS_VERSION = "4.52.0"
+    MAX_TRANSFORMERS_VERSION = "4.51.99"
 
 
 class Phi3VisionConfigBehavior(str, enum.Enum):
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -2768,6 +2768,7 @@ class MiniCPMVConfigBehavior(str, enum.Enum):
 
 
 @register_in_tasks_manager("minicpmv", *["image-text-to-text"], library_name="transformers")
+@register_in_tasks_manager("minicpmo", *["image-text-to-text"], library_name="transformers")
 class MiniCPMVOpenVINOConfig(BaseVLMOpenVINOConfig):
     SUPPORTED_BEHAVIORS = [model_type.value for model_type in MiniCPMVConfigBehavior]
     NORMALIZED_CONFIG_CLASS = NormalizedVisionConfig
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -2897,7 +2897,7 @@ class MiniCPMVOpenVINOConfig(BaseVLMOpenVINOConfig):
 @register_in_tasks_manager("minicpmo", *["image-text-to-text"], library_name="transformers")
 class MiniCPMOOpenVINOConfig(MiniCPMVOpenVINOConfig):
     MIN_TRANSFORMERS_VERSION = "4.43.0"
-    MAX_TRANSFORMERS_VERSION = "4.51.0"
+    MAX_TRANSFORMERS_VERSION = "4.52.0"
 
 
 class Phi3VisionConfigBehavior(str, enum.Enum):
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -25,6 +25,7 @@ from transformers import (
     PreTrainedTokenizer,
 )
 from transformers.modeling_outputs import BaseModelOutputWithPooling
+from transformers.models.qwen2_vl.modeling_qwen2_vl import VisionRotaryEmbedding
 from transformers.utils import ModelOutput
 
 from ...exporters.openvino import main_export
@@ -44,7 +45,6 @@ from .utils import (
 
 if is_transformers_version(">=", "4.46.0"):
     from transformers import AutoModelForImageTextToText
-    from transformers.models.qwen2_vl.modeling_qwen2_vl import VisionRotaryEmbedding
 
     transformers_auto_class = AutoModelForImageTextToText
 else:
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -41,9 +41,10 @@ from .utils import (
     TemporaryDirectory,
 )
 
+
 if is_transformers_version(">=", "4.46.0"):
-    from transformers.models.qwen2_vl.modeling_qwen2_vl import VisionRotaryEmbedding
     from transformers import AutoModelForImageTextToText
+    from transformers.models.qwen2_vl.modeling_qwen2_vl import VisionRotaryEmbedding
 
     transformers_auto_class = AutoModelForImageTextToText
 else:
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -25,7 +25,6 @@ from transformers import (
     PreTrainedTokenizer,
 )
 from transformers.modeling_outputs import BaseModelOutputWithPooling
-from transformers.models.qwen2_vl.modeling_qwen2_vl import VisionRotaryEmbedding
 from transformers.utils import ModelOutput
 
 from ...exporters.openvino import main_export
@@ -42,8 +41,8 @@ from .utils import (
     TemporaryDirectory,
 )
 
-
 if is_transformers_version(">=", "4.46.0"):
+    from transformers.models.qwen2_vl.modeling_qwen2_vl import VisionRotaryEmbedding
     from transformers import AutoModelForImageTextToText
 
     transformers_auto_class = AutoModelForImageTextToText
@@ -4355,4 +4354,5 @@ MODEL_TYPE_TO_CLS_MAPPING = {
     "phi4mm": _OVPhi4MMForCausalLM,
     "phi4_multimodal": _OVPhi4MMForCausalLM,
     "llama4": _OVLlama4ForCausalLM,
+    "minicpmo": _OVMiniCPMVForCausalLM,
 }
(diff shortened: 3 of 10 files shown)


optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -45,6 +44,7 @@
 
 if is_transformers_version(">=", "4.46.0"):
     from transformers import AutoModelForImageTextToText
+    from transformers.models.qwen2_vl.modeling_qwen2_vl import VisionRotaryEmbedding
```
echarlaix: why is this needed ? `VisionRotaryEmbedding` was introduced in v4.45.0 https://github.com/huggingface/transformers/blob/v4.45.0/src/transformers/models/qwen2_vl/modeling_qwen2_vl.py#L260 which is the min required transformers version

docs/source/openvino/models.mdx
```diff
@@ -93,6 +93,7 @@ Here is the list of the supported architectures :
 - Marian
 - MiniCPM
 - MiniCPM3
+- MiniCPMO (image-text-to-text)
```
```suggestion
- MiniCPM-o
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -2894,6 +2894,11 @@ def patch_model_for_export(
         return super().patch_model_for_export(model, model_kwargs)
 
 
+@register_in_tasks_manager("minicpmo", *["image-text-to-text"], library_name="transformers")
+class MiniCPMOOpenVINOConfig(MiniCPMVOpenVINOConfig):
+    MAX_TRANSFORMERS_VERSION = "4.51.0"
```
echarlaix: Usually against remote code modeling support but because we don't have any transformers support and have a `MAX_TRANSFORMERS_VERSION` this sounds reasonable. Because it's a remote code models support, can you add a `MIN_TRANSFORMERS_VERSION` as well and ensure that support works for every versions in between (we don't have a test for v4.51 anymore and it was never tested for this architecture)

optimum/intel/openvino/modeling_visual_language.py
rkazants: optimum-intel does not work with transformers with lower 4.45 due to import error. I can use `if is_transformers_version(">=", "4.45.0"):` specifically for import of `VisionRotaryEmbedding`

optimum/exporters/openvino/model_configs.py
rkazants: 4.43 is a minimal version. I will set it. thanks.

optimum/intel/openvino/modeling_visual_language.py
echarlaix: I'm not sure to understand, `is_transformers_version(">=", "4.45.0")` will always be True since now the min required transformers version is v4.45 https://github.com/huggingface/optimum-intel/blob/main/setup.py#L32 so why do we need this change ? (`VisionRotaryEmbedding` was introduced in 4.45 so we shouldn't need to check the transformers version to see if we can import)

optimum/exporters/openvino/model_configs.py

optimum/intel/openvino/modeling_visual_language.py
rkazants: Do you mean that "4.45.0" is a minimal by definition so that we don't need explicit check here using "is_transformers_version(">=", "4.45.0")"?

rkazants: reverted

tests/openvino/test_seq2seq.py
```diff
@@ -652,12 +654,21 @@ def test_compare_to_transformers(self, model_arch):
             transformers_inputs["past_key_values"] = DynamicCache()
 
         with torch.no_grad():
+            tokenizer = None
+            if model_arch in ["minicpmo"]:
+                # `generate` method for minicpmo requires tokenizer
+                tokenizer = AutoTokenizer.from_pretrained(
```
IlyasMoutawwakil, line 661: let's rather add it to `additional_inputs`

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -2114,6 +2114,36 @@ def preprocess_inputs(
...
+        attention_mask=None,
+        audio_bounds=None,
+        spk_bounds=None,
+        audio_features=None,
+        audio_feature_lens=None,
+        **kwargs,
+    ):
+        # Audio modality is not supported for MiniCPMO
+        if audio_features is not None and len(audio_features) > 0:
+            raise ValueError("Audio input is not supported for MiniCPMO")
```
IlyasMoutawwakil, line 2134: in general or only with the OV Model ?

IlyasMoutawwakil (approved):
LGTM ! left a question and a nit suggestion

optimum/intel/openvino/modeling_visual_language.py
rkazants: With OV model now it is not supported.

tests/openvino/test_seq2seq.py
rkazants: Done

tests/openvino/utils_tests.py
```diff
@@ -116,6 +116,7 @@
     "minicpm": "katuni4ka/tiny-random-minicpm",
     "minicpm3": "katuni4ka/tiny-random-minicpm3",
     "minicpmv": "katuni4ka/tiny-random-minicpmv-2_6",
+    "minicpmo": "rkazants/tiny-random-MiniCPM-o-2_6",
```
IlyasMoutawwakil, line 119: this model will slow down our ci greatly, it is 400MB 🫨
https://huggingface.co/rkazants/tiny-random-MiniCPM-o-2_6/tree/main

tests/openvino/test_seq2seq.py
```diff
@@ -496,13 +496,15 @@ class OVModelForVisualCausalLMIntegrationTest(unittest.TestCase):
         SUPPORTED_ARCHITECTURES += ["gemma3", "smolvlm"]
     if is_transformers_version(">=", "4.51"):
         SUPPORTED_ARCHITECTURES += ["llama4"]
+    if is_transformers_version("<", "4.51"):
+        SUPPORTED_ARCHITECTURES += ["minicpmo"]
```
nikita-savelyevv, line 502: Can you please also add tests cases similar to these ones?

https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_exporters_cli.py#L725-L734
https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_quantization.py#L991-L1008

optimum/exporters/openvino/model_configs.py
```diff
@@ -2894,6 +2894,12 @@ def patch_model_for_export(
         return super().patch_model_for_export(model, model_kwargs)
 
 
+@register_in_tasks_manager("minicpmo", *["image-text-to-text"], library_name="transformers")
+class MiniCPMOOpenVINOConfig(MiniCPMVOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.43.0"
+    MAX_TRANSFORMERS_VERSION = "4.51.0"
```
nikita-savelyevv:
```suggestion
    MAX_TRANSFORMERS_VERSION = "4.52.0"
```

I was able to export model having `transformers==4.51.3`. But in order for it to work `MAX_TRANSFORMERS_VERSION = "4.52.0"` needs to be set because there is a `>=` condition for raising an exception at https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/convert.py#L206 .

@echarlaix was this logic made this way on purpose, meaning that `MAX_TRANSFORMERS_VERSION` is an exclusive upper bound? I've just noticed it and in my opinion it is a bit confusing.

Is it common for transformers to do minor releases for older versions? If not, making the aforementioned condition a strict `>` would be more intuitive in my opinion. For this particular case we then could set `MAX_TRANSFORMERS_VERSION = "4.51.3"`.

tests/openvino/utils_tests.py
rkazants: It is a minimal size I managed to receive. `minicpmv` is about ~300MB and it is tested: https://huggingface.co/katuni4ka/tiny-random-minicpmv-2_6/tree/main

optimum/exporters/openvino/model_configs.py
rkazants: yes, this is exclusive upper bound. I checked and it works indeed for 4.51

tests/openvino/test_seq2seq.py
```diff
@@ -496,13 +496,15 @@ class OVModelForVisualCausalLMIntegrationTest(unittest.TestCase):
         SUPPORTED_ARCHITECTURES += ["gemma3", "smolvlm"]
     if is_transformers_version(">=", "4.51"):
         SUPPORTED_ARCHITECTURES += ["llama4"]
+    if is_transformers_version("<", "4.51"):
```
rkazants:
```suggestion
    if is_transformers_version("<", "4.52"):
```

tests/openvino/utils_tests.py
IlyasMoutawwakil: should be reduced as well

tests/openvino/test_exporters_cli.py
```diff
@@ -1020,7 +1043,7 @@ def test_exporters_cli_4bit(
             result = subprocess.run(
                 f"optimum-cli export openvino --model {MODEL_NAMES[model_type]} --task {task} --weight-format {option} {tmpdir}",
                 shell=True,
-                check=True,
+                #check=True,
```
rkazants:
```suggestion
                check=True,
```

tests/openvino/test_seq2seq.py
rkazants: Done

tests/openvino/utils_tests.py
rkazants: I reduced to 144MB. Minimal hidden_size is 128 for llm part: https://huggingface.co/rkazants/tiny-random-MiniCPM-o-2_6/blob/main/modeling_minicpmo.py#L209
 That also impacts apm and tts module size.

@IlyasMoutawwakil, @echarlaix, I propose to do further reduction in further PR(s) if any ideas. Now my other colleagues anticipate this PR merge, let us not block PR merge due to tiny model size. We know that the implemented logic are passing the tests in GHA.

tests/openvino/test_exporters_cli.py
```diff
@@ -733,6 +733,29 @@ class OVCLIExportTestCase(unittest.TestCase):
                 "vision_embeddings_model": {"int8": 16},
             },
         ),
+        (
+            "image-text-to-text",
+            "minicpmo",
+            "int4 --group-size 4 --ratio 0.8 --trust-remote-code",
+            {
+                "lm_model": {"int8": 16, "int4": 14},
```
rkazants:
```suggestion
                "lm_model": {"int8": 14, "int4": 14},
```

optimum/exporters/openvino/model_configs.py
nikita-savelyevv: Hi @echarlaix! This is a gentle reminder. Could you please take a look at the comment above?

tests/openvino/test_exporters_cli.py
```diff
@@ -733,6 +733,29 @@ class OVCLIExportTestCase(unittest.TestCase):
                 "vision_embeddings_model": {"int8": 16},
             },
         ),
+        (
+            "image-text-to-text",
+            "minicpmo",
+            "int4 --group-size 4 --ratio 0.8 --trust-remote-code",
+            {
+                "lm_model": {"int8": 14, "int4": 14},
```
rkazants:
```suggestion
                "lm_model": {"int8": 14, "int4": 12},
```

optimum/exporters/openvino/model_configs.py
rkazants: @nikita-savelyevv, @echarlaix, just clarify - this is a separate unrelated story, right?:) We can address this issue in separate PR and discuss offline.

IlyasMoutawwakil: we can use the `MAX_TRANSFORMERS_VERSION = "4.51.99"` notation

echarlaix: @nikita-savelyevv yes this was done to enable all patch releases by default when setting the `MAX_TRANSFORMERS_VERSION` to 4.x.0 (as we don't know the number of patch release that will happen until the next minor release), but I understand that it might be a bit confusing so no problem to change it in a following PR if that makes things clearer

nikita-savelyevv: I see, thanks for the explanation. Setting `MAX_TRANSFORMERS_VERSION = "4.51.99"` is ok with me. 

@rkazants Could you please update to 4.51.99?

tests/openvino/utils_tests.py
echarlaix: completely agree with @IlyasMoutawwakil comment, we should be super careful with our tiny random models size to not slow down the ci, could you extend on the different models parameters constraint @rkazants https://huggingface.co/rkazants/tiny-random-MiniCPM-o-2_6/blob/main/config.json#L20 for example I see `d_model` /  `decoder_ffn_dim` / `encoder_ffn_dim` respectively set to 1024, 1024 and 4096

echarlaix: Also if the PR really needs to be merged asap I'm ok with keeping this model but would like to have a following PR to change it to a smaller model or if that cannnot be done due to modeling constraint then would like to have more information on what are the constraints / why it cannot be done, would that sound reasonable @rkazants ?

optimum/exporters/openvino/model_configs.py
```diff
@@ -2682,6 +2682,12 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
         return super().patch_model_for_export(model, model_kwargs)
 
 
+@register_in_tasks_manager("minicpmo", *["image-text-to-text"], library_name="transformers")
+class MiniCPMOOpenVINOConfig(MiniCPMVOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.43.0"
+    MAX_TRANSFORMERS_VERSION = "4.52.0"
```
```suggestion
    MAX_TRANSFORMERS_VERSION = "4.51.99"
```

optimum/exporters/openvino/model_configs.py

tests/openvino/utils_tests.py
```diff
@@ -327,6 +328,12 @@
     "clip": {"model": 130},
     "mamba": {"model": 386},
     "falcon-mamba": {"model": 194},
+    "minicpmo": {
+        "lm_model": 30,
```
```suggestion
        "lm_model": 16,
```

tests/openvino/utils_tests.py
rkazants: Discussed offline with @echarlaix to proceed with the merge.
I will take this AR for further optimization. Indeed, there is a room for optimization such as `d_model`, `encoder_ffn_dim` but it will take some time because varying these parameters values needs to adjust several parameters from other modalities. It requires a bit deeper model understanding.
Thanks!

tests/openvino/test_exporters_cli.py
```diff
@@ -733,6 +733,29 @@ class OVCLIExportTestCase(unittest.TestCase):
...
+            "image-text-to-text",
+            "minicpmo",
+            "int4 --group-size 4 --ratio 0.8 --trust-remote-code",
+            {
+                "lm_model": {"int8": 12, "int4": 4},
+                "text_embeddings_model": {"int8": 1},
+                "vision_embeddings_model": {"int8": 8},
+                "resampler_model": {"int8": 6},
+            },
+        ),
```
nikita-savelyevv:
```suggestion
```

Data-aware case is enough.

docs/source/openvino/models.mdx
```diff
@@ -93,6 +93,7 @@ Here is the list of the supported architectures :
 - Marian
 - MiniCPM
 - MiniCPM3
+- MiniCPM-o
```
rkazants:
```suggestion
- MiniCPM-o 
```

docs/source/openvino/models.mdx
```diff
@@ -93,6 +93,7 @@ Here is the list of the supported architectures :
...
```
rkazants:
```suggestion
- MiniCPM-o
```

tests/openvino/test_quantization.py
```diff
@@ -1006,6 +1006,27 @@ class OVWeightCompressionTest(unittest.TestCase):
...
+                group_size=16,
+                dataset="contextual",
+                ratio=0.8,
+                sensitivity_metric="mean_activation_magnitude",
+                num_samples=1,
+                processor=MODEL_NAMES["minicpmo"],
+                trust_remote_code=True,
+            ),
+            {
+                "lm_model": {"int8": 4, "int4": 0},
```
nikita-savelyevv:
```suggestion
                group_size=4,
                dataset="contextual",
                ratio=0.8,
                sensitivity_metric="mean_activation_magnitude",
                num_samples=1,
                processor=MODEL_NAMES["minicpmo"],
                trust_remote_code=True,
            ),
            {
                "lm_model": {"int8": 6, "int4": 10},
```

rkazants: this is a problem of tests, not this PR. I propose to re-implement this test suite in separate PR and avoid test data duplication in separate files if you expect the same test data among these files. Now it is not obvious at all:)

nikita-savelyevv: That's a good idea. But until that I would propose to keep being aligned with the current approach of equal quantization parameters, the same as it is done for other models right now.

rkazants: ok, applied

nikita-savelyevv: Thanks!

docs/source/openvino/models.mdx
rkazants:
```suggestion
- MiniCPM-o 
```

docs/source/openvino/models.mdx
rkazants:
```suggestion
- MiniCPM-o
```

PR #1457, opened 2025-10-01
Fix editable mode and cleanup/adapt for optimum v2

IlyasMoutawwakil:
# What does this PR do?

Following https://github.com/huggingface/optimum/pull/2361

## Before submitting


optimum/exporters/openvino/model_patcher.py
```diff
@@ -2745,8 +2745,7 @@ def gptj_attn_forward(
 class GptJModelPatcher(OVDecoderModelPatcher):
     def __enter__(self):
         super().__enter__()
-
-        if is_transformers_version(">=", "4.49"):
+        if is_transformers_version(">=", "4.45.0"):
```
echarlaix: is this coming from bettertransformers being deprecated in which case shouldn't we also remove `patch_model_with_bettertransformer` https://github.com/huggingface/optimum-intel/blob/e95d8c781929205f795a40d2c9404992e664f9b2/optimum/exporters/openvino/model_patcher.py#L77

Also the min required transformers version is now v4.45 so we should also be able to remove `is_transformers_version(">=", "4.45.0")`

echarlaix (approved):
LGTM, thanks a lot for the great improvement @IlyasMoutawwakil

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: yes that's where it's coming from !

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4509,60 +4644,107 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+        self.patched_forward = patched_forward
 
     def __enter__(self):
         super().__enter__()
 
         if is_transformers_version(">=", "4.53.0"):
             # for OpenVINO, we use torch.finfo(torch.float16).min instead of torch.finfo(dtype).min
-            # Although I'm not sure this is the right way to handle this, we are basically pretending that -65,504 is -inf
+            # to aoid overflow issues on some hardware (e.g. Intel NPU)
```
echarlaix:
```suggestion
            # to avoid overflow issues on some hardware (e.g. Intel NPU)
```

.github/workflows/build_pr_documentation.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: "18"
+          node-version: "20"
```
echarlaix, line 29: thanks for the doc fix

optimum/exporters/openvino/model_configs.py
```diff
@@ -3854,20 +3845,9 @@ def flatten_past_key_values(self, flattened_output, name, idx, t):
 class WhisperOpenVINOConfig(WhisperOnnxConfig):
     def patch_model_for_export(
         self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
-    ) -> ModelPatcher:
+    ) -> OVModelPatcher:
         return OVSeq2SeqModelPatcher(self, model, model_kwargs=model_kwargs)
 
-    @property
-    def inputs(self):
```
echarlaix, line 3861: why can this be removed ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -65,53 +71,187 @@
...
+    PatchingSpec(torch, "rms_norm", onnx_compatible_rms_norm, torch.rms_norm),
+    PatchingSpec(torch.Tensor, "unfold", onnx_compatible_unfold, torch.Tensor.unfold),
+    PatchingSpec(torch.linalg, "norm", onnx_compatible_linalg_norm, torch.linalg.norm),
+    # TracerWarning: Using len to get tensor shape might cause the trace to be incorrect. Recommended usage would be tensor.shape[0]. Passing a tensor of different shape might lead to errors or silently give incorrect results.
+    PatchingSpec(torch.Tensor, "__len__", lambda x: x.shape[0], torch.Tensor.__len__),
+]
 
 
-BETTERTRANSFORMER_IGNORE = ["codegen", "gpt_neo"]
+class OVModelPatcher:
```
echarlaix: what's the difference with `ModelPatcher` https://github.com/huggingface/optimum-onnx/blob/671b84f78a244594dd21cb1a8a1f7abb8961ea60/optimum/exporters/onnx/model_patcher.py#L455

echarlaix (approved):
Thanks a lot @IlyasMoutawwakil

nikita-savelyevv (approved):
LGTM!

optimum/exporters/openvino/model_configs.py
IlyasMoutawwakil: because we are doing it correctly in optimum-onnx now ; seq2seq models used to make the decoder_input_ids dynamic only on the batch dim, supposing that the seq len dim will always be 1 when decoding, but it can be more (e.g. multi-turn chat). It will also be the case when we get rid of the merged decoders and have the encoder generate the cross attn.

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: tbh it was quick dirty fix until i can find a better way. 
basically I had to remove the part that fixes `check_model_inputs` method, because I was seeing a weird failure in ov seq2seq models (model has unreferenced parameter...) when the decoder was being converted to stateful, error was very cryptic so I removed that part from the model patcher and the error disappeared, it seems that the fix was also unwrapping this stateful-specific patch in the ov seq2seq patcher [(the one we do before even calling super().__init__ and apply directly to the mode's forward)](https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_patcher.py#L4519), I pushed a new solution that merges that patch with the other one in ov seq2seq model patcher, resolving the issue without rewriting the model patcher !

optimum/commands/export/openvino.py
```diff
@@ -21,15 +21,8 @@
 
 from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
 
-from ...exporters import TasksManager
-from ...intel.utils.import_utils import (
-    DIFFUSERS_IMPORT_ERROR,
-    is_diffusers_available,
-    is_nncf_available,
-)
-from ...intel.utils.modeling_utils import _infer_library_from_model_name_or_path
```
IlyasMoutawwakil, line 30: avoiding imports at the top level of commands help speedup the cli startup time

optimum/exporters/openvino/convert.py
```diff
@@ -227,63 +223,8 @@ def export(
...
-
-
-def export_tensorflow(
-    model: Union["PreTrainedModel", "ModelMixin"],
-    config: "OnnxConfig",
-    opset: int,
-    output: Path,
-    ov_config: Optional["OVConfig"] = None,
-    library_name: Optional[str] = None,
-):
```
IlyasMoutawwakil, line 254: exporting tf model was removed from optimum-onnx

IlyasMoutawwakil:
everything seems to be working ! I will remove a couple unnecessary patchers, for now:
- gpt bigcode: done in optimum-onnx
- vision-encode-decoder: it only sets a use_cache to true for trocr (can be done in seq2seq patcher)
pretty sure there are more patches we can get rid of now.

PR #1458, opened 2025-10-02
Remove automatic support of all supported ONNX models

echarlaix:
- Remove automatic support of all supported ONNX models, for untested models that are already supported (models from `optimum-onnx` >= v0.1.0) that are listed in `ONNX_SUPPORTED_ARCHITECTURES` we ensure support until v1.29
- Add tests test_find_untested_architectures to ensure all officially supported architectures are tested
- Enable loading of pix2struct models with `OVModelForVision2Seq` (previously only via `OVModelForPix2Struct`)

### DIFF
--- a/tests/openvino/test_modeling.py
+++ b/tests/openvino/test_modeling.py
@@ -49,9 +49,9 @@ from transformers import (
     AutoProcessor,
     AutoTokenizer,
     PretrainedConfig,
+    SamModel,
     pipeline,
     set_seed,
-    SamModel,
 )
 from transformers.onnx.utils import get_preprocessor
 from transformers.testing_utils import slow


optimum/exporters/openvino/model_configs.py
```diff
@@ -3969,6 +3969,7 @@ class M2M100OpenVINOConfig(BartOpenVINOConfig):
 )
 @register_in_tasks_manager("deepseek", *["text-generation", "text-generation-with-past"], library_name="transformers")
 class DeepseekOpenVINOConfig(MiniCPM3OpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.46.0"
```
echarlaix, line 3747: only tested for transformers >= v4.46
https://github.com/huggingface/optimum-intel/blob/v1.25.0/tests/openvino/test_modeling.py#L1163

tests/openvino/utils_tests.py
```diff
@@ -114,15 +115,17 @@
     "jais": "optimum-intel-internal-testing/tiny-random-jais",
     "levit": "optimum-intel-internal-testing/tiny-random-LevitModel",
     "lfm2": "optimum-intel-internal-testing/tiny-random-lfm2",
-    "longt5": "optimum-intel-internal-testing/tiny-random-longt5",
+    "longt5": "hf-internal-testing/tiny-random-LongT5Model",
     "llama": "optimum-intel-internal-testing/tiny-random-LlamaForCausalLM",
     "llama_awq": "optimum-intel-internal-testing/tiny-random-LlamaForCausalLM",
     "llama4": "optimum-intel-internal-testing/tiny-random-llama4",
+    "llama4_text": "trl-internal-testing/tiny-Llama4ForCausalLM",
```
echarlaix: will add all new tiny models to https://huggingface.co/optimum-intel-internal-testing before merging

tests/openvino/test_export.py
```diff
@@ -300,6 +304,30 @@ def test_export_fp16_model(self):
...
+        for supported_models_config in supported_model_types:
+            supported_models = getattr(TasksManager, supported_models_config)
+            for model_type, exporter_configs in supported_models.items():
+                if "onnx" in exporter_configs:
+                    onnx_architectures.add(model_type)
+                if "openvino" in exporter_configs:
+                    openvino_architectures.add(model_type)
+
+            # ensure no breaking changes (all models supported before this PR are still supported), can be removed
+            self.assertTrue(onnx_architectures.issubset(openvino_architectures))
```
echarlaix, line 325: only needs to be true before merging, can be removed 
```suggestion
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -3656,6 +3763,11 @@ def __init__(
         preprocessors: Optional[List[Any]] = None,
         **kwargs,
     ):
+        if task == "image-to-text":
```
echarlaix, line 3766: GOT-OCR2  should be exported using the `"image-text-to-text"` task and not `"image-to-text"`, if a user specify the task `"image-to-text"` the resulting model will still be equivalent https://github.com/huggingface/optimum-intel/blob/71c47d49715b09c9c6dc95da5331f3f0a6b36755/optimum/exporters/openvino/convert.py#L936

tests/openvino/test_modeling.py
```diff
@@ -1834,93 +1835,93 @@ def test_huggingface_pipeline_streaming(self, model_arch):
         gc.collect()
 
 
-class OVSamIntegrationTest(unittest.TestCase):
```
echarlaix, line 1837: removed to make sure all tests pass before merging

optimum/exporters/openvino/utils.py
```diff
@@ -307,6 +307,66 @@ def get_submodels(model):
...
+    "owlv2",
+    "owlvit",
+    "patchtsmixer",
+    "patchtst",
+    "pvt",
+    "regnet",
+    "rt_detr",
+    "rt_detr_v2",
+    "siglip-text",
+    "siglip-text-with-projection",
```
IlyasMoutawwakil: does this list contain only transformers models ? asking because i think these two are export model types that we define for diffusers

optimum/exporters/openvino/convert.py
```diff
@@ -561,6 +562,12 @@ def export_from_model(
     else:
         model_type = getattr(model.config, "model_type", None) or ""
 
+    if model_type in ONNX_SUPPORTED_ARCHITECTURES:
+        logger.warning(
+            f"The OpenVINO export of {model_type} models is not officially supported and will be deprecated in optimum-intel v1.29.\n"
+            f"If you want to support {model_type} please propose a PR or open up an issue."
+        )
```
IlyasMoutawwakil, line 568: do we wanna warn users about all these models ? my understanding is that they only lack testing, but supporting them should be as easy as imporitng their onnx config and wrapping it. a lot of them are probably not important but some are, like smollm3, dinov2, d_fine, modernbert..

tests/openvino/test_decoder.py
```diff
@@ -223,6 +244,59 @@ def mock_torch_compile(self, model_arch):
...
+            if tokenizer.eos_token is not None:
+                tokenizer.pad_token = tokenizer.eos_token
+            elif tokenizer.bos_token is not None:
+                tokenizer.pad_token = tokenizer.bos_token
+            else:
+                raise ValueError(
+                    f"Tokenizer for model {model_id} does not have a defined `pad_token`, `eos_token`, or `bos_token`."
+                )
+        tokenizer.padding_side = "left"
+        return tokenizer
```
IlyasMoutawwakil, line 255: thanks for adding this ! much needed

tests/openvino/test_modeling.py
IlyasMoutawwakil: maybe we can move it to a separate file to avoid it interrupting the modeling tests

echarlaix: agreed! moved in https://github.com/huggingface/optimum-intel/pull/1458/commits/fbc52109d425152441838ccedbacdc2b739631df (once fixed we can move other vision related tests there)

optimum/exporters/openvino/utils.py
echarlaix: thanks a lot for catching, model parts shouldn't be part of ONNX_SUPPORTED_ARCHITECTURES, fixed in https://github.com/huggingface/optimum-intel/pull/1458/commits/cfdadc7070c19dbdb725618ddf8ad3c62575c086

also added a comment in https://github.com/huggingface/optimum-intel/pull/1458/commits/b6d6c4c75e79c217747f2e7e7fd9f3ab67456858 to specify it can be any models from transformers, diffusers, timm or sentence transformers.

optimum/exporters/openvino/convert.py
echarlaix: not against removing, however from couple of testing looks like a lot of models have issues (like smollm3, helium, nemotron) like outputs mismatch when compared to the original model

tests/openvino/test_decoder.py
echarlaix: shamelessly stolen from https://github.com/huggingface/optimum-onnx/blob/682988bc2fd972f4594fb66725fa7e7c6c2b628d/tests/onnxruntime/test_decoder.py#L167

optimum/exporters/openvino/convert.py
echarlaix: updated in https://github.com/huggingface/optimum-intel/pull/1458/commits/f0b154d6dc2049f70888c1ea7726500db08a6e53

tests/openvino/test_modeling.py
echarlaix: Still getting the same issue for test_modeling.py : https://github.com/huggingface/optimum-intel/actions/runs/21249186795/job/61146035649 so reverting https://github.com/huggingface/optimum-intel/commit/e1fe54dea580633e745742a8fc5889b41d8b3ac8

PR #1461, opened 2025-10-07
[OV] Fix model inference not being affected by static quantization

nikita-savelyevv:
# What does this PR do?

**Reason for changes**
Executing multi-model pipelines after static quantization but before serialization leads to original non-quantized models being inferred instead. For example in case of vision embeddings model within a VLM pipeline, this is because the model is replaced at `OVModelForVisualCausalLM` level, but not at `OVVisionEmbedding` level.

**Changes**
- Introduced `OVModelHostMixin` to unify handling of  `openvino.Model` instances for all model classes containing them. This class implements:
  - `ov_models` property returns named instances of `openvino.Model` that the model contains (renamed from `ov_submodels`).
  - `components` property returns named instances of `OVModelHostMixin` that the model contains.
  - `replace_ov_model(current_model: openvino.Model, new_model: openvino.Model)` method is used to perform recursive replacement. OV models are matched by Python object id.
  - `compile()` method is self explanatory. Previously, for some classes it was called `_compile()` which is now renamed for consistency.
  - `clear_requests()`  method is self explanatory.
- `test_ovmodel_pipeline_quantization` -- now checks quantized model both before and after serialization.

## Before submitting


optimum/intel/openvino/modeling_base.py
```diff
@@ -210,17 +295,6 @@ def dtype(self) -> Optional[torch.dtype]:
 
         return None
 
-    @property
-    def ov_submodels(self) -> Dict[str, openvino.Model]:
-        return {submodel_name: getattr(self, submodel_name) for submodel_name in self._ov_submodel_names}
```
IlyasMoutawwakil, line 215: breaking change ?

nikita-savelyevv: Yes, but https://github.com/nikita-savelyevv/optimum-intel/blob/ns/fix-inplace-quantization/optimum/intel/openvino/modeling_base.py#L75-L79

nikita-savelyevv:
> Are you sure we need to refactorization to fix ? I understand the incentive of the refactorization but I would suggest fixing in one PR, and refactoring in another, because reviewing the refactorization will definitely take longer than the fix (which i assume is meant for the blog's benchmark)

Sure, I can create a fix for the VLM case

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, model: openvino.Model, parent_model: OVBaseModel) -> None:
         self._main_input = list(self.input_names.keys())[0]
 
     def forward(self, input_ids, **kwargs):
-        self._compile()
+        self.compile()
```
rkazants, line 61: Now `compile()` method gets a sort of external method but do we really need it? I am not sure that user API of optimum-intel should expose it. Compilation happens under the hood and belong to dev API only. So `_` still needed.

optimum/intel/openvino/modeling_open_clip.py
```diff
@@ -487,6 +489,18 @@ def __init__(
...
+    def _component_names(self) -> List[str]:
+        return ["text_model", "visual_model"]
+
+    @property
+    def _ov_model_names(self) -> List[str]:
+        return self._component_names
+
+    @property
+    def ov_models(self) -> Dict[str, openvino.Model]:
+        return {name: getattr(component, "model") for name, component in self.components.items()}
```
rkazants, line 502: A bit confusing what critiria to have method names w/o "_"

optimum/intel/openvino/modeling_text2speech.py
nikita-savelyevv: As of now, for some classes this method is named `compile()` and for some `_compile()`. I've decided it to align it by making all of them called `compile()`. According to documentation, `compile()` is a part of our public API https://huggingface.co/docs/optimum/en/intel/openvino/inference#compilation

optimum/intel/openvino/modeling_open_clip.py
nikita-savelyevv: This was discussed in one of the previous PRs where these methods were first introduced. As I remember, at that point we concluded that `*_names` properties don't need to be publicly exposed, possibly because calling `ov_models.components.keys()` will provide the names. In this PR I keep it the same way.

optimum/intel/openvino/modeling_base.py
```diff
@@ -57,12 +57,97 @@
...
+
+    @property
+    def ov_submodels(self) -> Dict[str, Union[openvino.Model, openvino.runtime.CompiledModel]]:
+        logger.warn(
+            "`ov_submodels` property is deprecated and will be removed in v1.27. Please use `ov_models` property instead."
+        )
+        return self.ov_models
+
+    @property
+    def _ov_model_names(self) -> List[str]:
```
echarlaix, line 82: from my understanding  `_ov_model_names` will differ from `_component_names` for encoder-only or decoder-only models? I'm wondering if we should modify both so that they have the same behavior as other models (having only one component) so that we can always iterate on the models "components" to compile / clear request / set device, also would remove the distinction between `_ov_model_names` and  `_component_names`  wdyt @nikita-savelyevv ?

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -830,7 +813,7 @@ def _prepare_cache_for_generation(self, *args, **kwargs):
         return
 
 
-class OVEncoder:
+class OVEncoder(OVModelHostMixin):
```
echarlaix: why is this needed? (my understanding that `OVModelHostMixin` was providing functionalities for OVModels and not for components themselves)

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -914,8 +897,15 @@ def _compile(self):
...
+    def clear_requests(self):
+        if self._compile_only:
+            raise ValueError(
+                "`clear_requests()` is not supported with `compile_only` mode, please initialize model without this option"
+            )
+        self.request = None
+
 
-class OVDecoder:
+class OVDecoder(OVModelHostMixin):
```
echarlaix: why not inherit from `OVModelPart` instead ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -697,23 +691,54 @@ def _export(
...
+        additional_components = [part for part in self.additional_parts if hasattr(self, part)]
         return base_components + additional_components
 
     @property
-    def components(self):
-        return {component_name: getattr(self, component_name) for component_name in self._component_names}
-
-    @property
-    def _ov_submodel_names(self):
+    def _ov_model_names(self):
```
echarlaix, line 700: why is it needed to make a distinction between `_component_names` and `_ov_model_names` here ?

optimum/intel/openvino/modeling_seq2seq.py
nikita-savelyevv: My idea was that every entity that has an `openvino.model` instance inside of it, should base from `OVModelHostMixin`. This includes components such as `OVEncoder` and `OVDecoder` too. In terms of this particular PR, this is needed for `OVModelForSeq2SeqLM` to call `replace_ov_model()` on its components which happen to be `OVEncoder` and `OVDecoder`.

optimum/intel/openvino/modeling_seq2seq.py
nikita-savelyevv: That's a good idea, I think it's possible 👍

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: > from my understanding _ov_model_names will differ from _component_names for encoder-only or decoder-only models?

Yes, that's right. For encoder-only or decoder-only models `_ov_model_names` is `["model"]` and `_component_names` is empty.

> I'm wondering if we should modify both so that they have the same behavior as other models (having only one component) so that we can always iterate on the models "components" to compile / clear request / set device, also would remove the distinction between _ov_model_names and _component_names wdyt @nikita-savelyevv ?

There is a distinction between "single-ov-model" models and "multi-ov-model" models. The first ones have a single `openvino.model` within them an others have many. The first type of models are for example `OVModelForCausalLM`, `OVPipelinePart`, `OVEncoder` and so on. The second type: `OVModelForSeq2SeqLM`, `OVDiffusionPipeline`, `OVModelForVisualCausalLM`, etc. 

The reason why I had to implement both `_ov_model_names` and `_component_names` is that because sometimes one "multi-ov-model" model can have another "multi-ov-model" model as a sub-model. Currently, there is only one example of that: `OVModelForVisualCausalLM` has a `OVModelWithEmbedForCausalLM` as a sub-model, and `OVModelWithEmbedForCausalLM` has two `openvino.Model` instances within it: a language model and a text embedding model.

I suppose the question here is: is that single example a bad design that should be refactored or a hint of the fact that in the future this could be more common? In my opinion is that in theory there could appear more examples of that kind. But if you have a suggestion on how to refactor `OVModelWithEmbedForCausalLM`, I will be also ok with that.

optimum/intel/openvino/modeling_visual_language.py
nikita-savelyevv: @echarlaix, that's a good question! I believe I've answered it in another comment above.

optimum/intel/openvino/modeling_seq2seq.py
nikita-savelyevv: @echarlaix Done

optimum/intel/openvino/modeling_base.py
echarlaix: Thanks for the clear explanation, my understanding was that `OVModelHostMixin` was providing functionalities for OVModels, and my point was that maybe all `OVModels` should actually be treated as “multi-ov-model" with “single-ov-model" having only one `OVModelPart` instead of currently directly dealing with the `openvino.model` directly: this would simplify a lot things for methods like `compile`, `clear_request` or when changing device (as we could just iterate on the `OVModel.ov_models`). But here I’m understanding that `OVModelHostMixin` was designed in mind for “every entity that has an `openvino.model` instance inside of it” as you explain it in https://github.com/huggingface/optimum-intel/pull/1461#discussion_r2439775813 and that this is needed for `OVModelWithEmbedForCausalLM` as well not just for the “single-ov-model" vs “multi-ov-model" behavior mismatch. For me that would make sense to have `OVModels` and `OVModelParts` more clearly separated (the first one handling a single or multiple `OVModelParts` and the later only having to deal with one `openvino.model`) but this is outside the scope of this PR. Happy to merge this PR however I'd prefer not to include in the following release as potentially what I'm suggesting above could be implemented in the meantime, what do you think @nikita-savelyevv

echarlaix (approved):

PR #1464, opened 2025-10-08
[OV] Fix VLM in-place static quantization

nikita-savelyevv:
# What does this PR do?

Executing VLMs after static quantization but before serialization leads to original non-quantized models being inferred instead. For example in case of vision embeddings model within a VLM pipeline, the model is replaced at OVModelForVisualCausalLM level, but on at OVVisionEmbedding level.

This is a fast fix for VLMs only, full fix is here #1461

## Before submitting


echarlaix (approved):
Thanks for the fix !

PR #1378, opened 2025-07-10
Add support for cb4_f8e4m3 quantization mode.

nikita-savelyevv:
# What does this PR do?

As in the title. The new `cb4` weight format option represents a codebook with 16 fixed fp8 values in E4M3 format. The `cb4_f8e4m3` quant mode adds on this the corresponding activations quantization in f8e4m3.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1465, opened 2025-10-09
[OpenVINO] Adopt new mxfp4 quantization logic

nikita-savelyevv:
# What does this PR do?

Changes:
- Now with NNCF develop version there is an explicit "mxfp4" compression mode. The group size value is fixed to 32 for it according to mxfp4 definition.
- Remove temporary cb4 pre-release logic introduced in #1378.

There are also other compression modes "mxfp8_e4m3", "fp4_e2m1" and "nvfp4" added to NNCF, but they are planned to be added to optimum-intel after the next NNCF release.

## Before submitting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,8 +67,8 @@ TESTS_REQUIRE = [
 QUALITY_REQUIRE = ["black~=23.1", "ruff==0.4.4"]
 
 EXTRAS_REQUIRE = {
-    "nncf": ["nncf>=2.16.0"],
-    "openvino": ["nncf>=2.16.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0"],
+    "nncf": ["nncf>=2.18.0"],
+    "openvino": ["nncf>=2.18.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0"],
     "neural-compressor": ["neural-compressor[pt]>=3.4.1", "accelerate", "transformers<4.46"],
     "ipex": ["intel-extension-for-pytorch>=2.6", "transformers>4.50,<4.53", "accelerate"],
     "diffusers": ["diffusers"],


optimum/intel/openvino/configuration.py
```diff
@@ -715,7 +715,7 @@ def __init__(
         )
         self.bits = bits
         self.sym = sym
-        self.group_size = group_size or (-1 if bits == 8 else 128)
```
nikita-savelyevv, line 718: Delegate group size value selection to NNCF. This is backward compatible in terms of the default value since NNCF also selects -1 for 8-bit types and 128 for 4-bit non-mxfp4 types by default.

optimum/intel/openvino/configuration.py
```diff
@@ -874,7 +869,7 @@ def to_nncf_dict(self) -> Dict[str, Any]:
         if mode in signed_bitness.values():
             mode += "_sym" if self.sym else "_asym"
         if mode == "mxfp4":
-            mode = "e2m1"
+            mode = "e2m1" if is_nncf_version("<=", "2.18") else "mxfp4"
```
rkazants, line 872: do we have any model and IR with mxffp4 that will work with mxfp4 weights?
will it work for gpt-oss? will it work with re-quantization from int4/int8?

tests/openvino/test_quantization.py
rkazants, line 1: any test with inference?

optimum/intel/openvino/configuration.py
nikita-savelyevv: @ljaljushkin could you please comment on that?

tests/openvino/test_quantization.py
nikita-savelyevv: Not for most of 4-bit types. We have some such tests only for int8 and int4 that check that the model infers without errors.

optimum/intel/openvino/configuration.py
ljaljushkin: @rkazants, gpt-oss and re-quantization is a separate story - it's in PoC stage (173837 ticket).
That PR is about bf16 -> mxfp4 quantization only that was before in optimum, but with incorrect compression type (e2m1).

tests/openvino/test_quantization.py
nikita-savelyevv: But that's a good point, we'll add such test on NNCF side including accuracy validation

rkazants: Why not here?:)
Let us add it here.

nikita-savelyevv: In my opinion this is a separate task not really relevant to changes in this PR. I've created a separate one with addition of inference checks https://github.com/huggingface/optimum-intel/pull/1470

rkazants: I meant tests for inference of IRs with mxfp4 weights. Now it looks to be absent.
What is an idea to produce IRs with mxfp4 but in reality this feature is not used for any model?

nikita-savelyevv: > I meant tests for inference of IRs with mxfp4 weights. Now it looks to be absent.

Yes, that's right. If you wish, we can merge #1470 first and then merge this one, because mxfp4 (with wrong group size) test is already present on the main branch https://github.com/huggingface/optimum-intel/blob/24d5834e3d38c96d53b91fca90c73b2695118f7b/tests/openvino/test_quantization.py#L659 and after #1470 is merged, mxfp4-quantized model will be inferred during testing.

> What is an idea to produce IRs with mxfp4 but in reality this feature is not used for any model?

It is used in tests, but the model is just not inferred. As I said, there are a limited number of tests that do infer int8 and int4 quantized models. That was enough as of now. I do agree that we can infer all quantized models just in case, but in general we have tests inside NNCF and OpenVINO repos and also OpenVINO LM validation for that.

optimum/intel/openvino/configuration.py
```diff
@@ -843,11 +843,6 @@ def post_init(self):
                 f"['int4', 'int8', 'mxfp4', 'nf4', 'cb4'], but found: {self.dtype}."
             )
         if self.dtype in ["mxfp4", "nf4", "cb4"]:
-            if self.dtype == "cb4" and is_nncf_version("<=", "2.17"):
```
echarlaix, line 846: this will be an issue still for users with nncf <=2.17 no ? if we want to remove, should we update the nncf version in https://github.com/huggingface/optimum-intel/blob/v1.25.2/setup.py#L69 ?

echarlaix (approved):

optimum/intel/openvino/configuration.py
nikita-savelyevv: Good point. Updated.

PR #1469, opened 2025-10-13
[OV] Recover cache_position for Whisper

rkazants:
# What does this PR do?

It recovers cache_position input in the exported whisper model. Regression happened in https://github.com/huggingface/optimum-intel/pull/1457

Fixes https://jira.devtools.intel.com/browse/CVS-174805

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 from transformers import AutoConfig, PretrainedConfig, PreTrainedModel
 
+from optimum.exporters.onnx.base import ConfigBehavior
 from optimum.exporters.onnx.config import OnnxConfig, TextDecoderOnnxConfig, TextDecoderWithPositionIdsOnnxConfig
 from optimum.exporters.onnx.model_configs import (
     BartOnnxConfig,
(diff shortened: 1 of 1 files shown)


optimum/exporters/openvino/model_configs.py
```diff
@@ -3586,6 +3587,14 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
 class WhisperOpenVINOConfig(WhisperOnnxConfig):
     _MODEL_PATCHER = OVSeq2SeqModelPatcher
 
+    @property
+    def inputs(self):
+        common_inputs = super().inputs
+        if self._behavior is not ConfigBehavior.ENCODER and self.use_past_in_inputs:
+            # since https://github.com/huggingface/transformers/pull/31166
+            common_inputs["cache_position"] = {0: "decoder_sequence_length"}
```
nikita-savelyevv, line 3595: Could you please add a condition somewhere in tests that would check that the exported OpenVINO decoder model has `cache_position` input? Also, are stateless whisper models also affected? If so, we should check for that case too.

optimum/exporters/openvino/model_configs.py
```diff
@@ -3586,6 +3587,14 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
 class WhisperOpenVINOConfig(WhisperOnnxConfig):
     _MODEL_PATCHER = OVSeq2SeqModelPatcher
 
+    @property
+    def inputs(self):
+        common_inputs = super().inputs
+        if self._behavior is not ConfigBehavior.ENCODER and self.use_past_in_inputs:
+            # since https://github.com/huggingface/transformers/pull/31166
```
IlyasMoutawwakil, line 3594: this comment only explains why it's needed from version 4.43 to 4.45, which is already covered in https://github.com/huggingface/optimum-onnx/blob/main/optimum/exporters/onnx/model_configs.py#L2009. it doesn't explain why it's **always** needed with openvino

PR #1470, opened 2025-10-13
[OpenVINO] Add model inference check to weight-only and pipeline quantization testing

nikita-savelyevv:
**Changes**

As in the title. Following up https://github.com/huggingface/optimum-intel/pull/1465#discussion_r2426157162 

## Before submitting

### DIFF
--- a/tests/openvino/test_quantization.py
+++ b/tests/openvino/test_quantization.py
@@ -584,7 +584,7 @@ class OVQuantizerTest(unittest.TestCase):
             ov_model = model_cls.from_pretrained(model_id, quantization_config=quantization_config)
             ov_model.save_pretrained(tmp_dir)
 
-            check_model_inference(ov_model, model_cls, model_id, trust_remote_code=False)
+            check_model_inference(ov_model, model_id, trust_remote_code=False)
 
             if model_cls in [OVModelForSpeechSeq2Seq, OVModelForSeq2SeqLM] and ov_model.decoder_with_past is None:
                 expected_fake_nodes_per_model.pop("decoder_with_past", None)
@@ -1290,7 +1290,7 @@ class OVWeightCompressionTest(unittest.TestCase):
 
             model.save_pretrained(tmp_dir)
             model = model_cls.from_pretrained(tmp_dir, trust_remote_code=trust_remote_code)
-            check_model_inference(model, model_cls, model_id, trust_remote_code)
+            check_model_inference(model, model_id, trust_remote_code)
 
             # At the moment the first model in the list is the only one we apply data-aware compression to
             wc_rt_info = next(iter(submodels.values())).get_rt_info()["nncf"]["weight_compression"]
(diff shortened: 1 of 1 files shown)


tests/openvino/test_quantization.py
```diff
@@ -2263,3 +2230,50 @@ def check_optimization_not_applicable_to_optimized_model(model, quantization_con
         match="Cannot apply optimization to the model because it was already optimized with the following config",
     ):
         quantizer.quantize(ov_config=OVConfig(quantization_config=quantization_config))
+
+
+def check_model_inference(ov_model, model_cls, model_id, trust_remote_code):
+    if model_cls in [OVModelForSpeechSeq2Seq, OVModelForSeq2SeqLM]:
```
rkazants: can you avoid this `if` and use `if` and `elif` below?

tests/openvino/test_quantization.py
```diff
@@ -2263,3 +2230,50 @@ def check_optimization_not_applicable_to_optimized_model(model, quantization_con
...
+        if model_cls == OVModelForSpeechSeq2Seq:
+            input_features = torch.randn((1, ov_model.config.num_mel_bins, 3000), dtype=torch.float32)
+            ov_model.generate(input_features)
```
rkazants: what is unclear here is that we don't explicitly specify generation config here. Let us set interation number and some other params.

rkazants: also, can we check inference results and compare with reference?

nikita-savelyevv: > what is unclear here is that we don't explicitly specify generation config here. Let us set interation number and some other params.

Copied generation config from seq2seq tests.

> also, can we check inference results and compare with reference?

In theory it is possible. What do you suggest to use as a reference?

tests/openvino/test_quantization.py
nikita-savelyevv: Thanks for the suggestion. These cases now share a single generation config, so it makes sense to keep this condition.

tests/openvino/test_quantization.py
```diff
@@ -2263,3 +2232,61 @@ def check_optimization_not_applicable_to_optimized_model(model, quantization_con
         match="Cannot apply optimization to the model because it was already optimized with the following config",
     ):
         quantizer.quantize(ov_config=OVConfig(quantization_config=quantization_config))
+
+
+def check_model_inference(ov_model, model_cls, model_id, trust_remote_code):
+    if model_cls in [OVModelForSpeechSeq2Seq, OVModelForSeq2SeqLM]:
```
echarlaix: why not ?

```suggestion
def check_model_inference(ov_model, model_id, trust_remote_code):
    if isinstance(ov_model, (OVModelForSpeechSeq2Seq, OVModelForSeq2SeqLM)):
```

nikita-savelyevv: Thanks for noticing, indeed!

PR #1473, opened 2025-10-14
CI secret with public token fallback

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/.github/workflows/test_openvino_nightly.yml
+++ b/.github/workflows/test_openvino_nightly.yml
@@ -9,10 +9,13 @@ on:
     branches:
       - v*-release
   pull_request:
+    branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/test_openvino_nightly.yml
```diff
@@ -9,10 +9,13 @@ on:
     branches:
       - v*-release
   pull_request:
+    branches:
+      - main
     types:
       - opened
       - labeled
       - reopened
+      - unlabeled
```
IlyasMoutawwakil, line 18: this allows stopping the label workflows when unlabeled

PR #1476, opened 2025-10-15
Add note about supported transformers version to export docs

helena-intel:
Add a sentence about supported transformers version to export docs. I didn't find a good place for it so added a Troubleshooting section but happy to move it elsewhere.


echarlaix (approved):
Thanks a lot for the addition @helena-intel

PR #1482, opened 2025-10-17
[OpenVINO] Add workaround logic for default int4 quantization of openai/gpt-oss-20b model

nikita-savelyevv:
# What does this PR do?

For correct 4-bit weight-only quantization of openai/gpt-oss-20b we currently have to run quantization twice. Unfortunately, we don't have such logic in optimum-intel yet, and its proper implementation will take a while. Because of this, this PR adds a workaround to enable this use case. The export command is a usual:
```
optimum-cli export openvino -m openai/gpt-oss-20b --weight-format int4 ./gpt-oss-20b
```

We hope that inference with bf16 weights will soon be supported in GPU plugin and then the second quantization step will not be required.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -1154,7 +1154,7 @@ class OVConfig(BaseConfig):
             dtypes = [OVConfig._get_dtype(config) for config in quantization_config.quantization_configs.values()]
             dtype = "_".join(dtypes)
         elif isinstance(quantization_config, _GPTOSSQuantizationConfig):
-            dtype = "int4_int8"
+            dtype = "int4"
         else:
             raise ValueError(f"Unsupported type of quantization config: {type(quantization_config)}")
         return dtype


optimum/exporters/openvino/__main__.py
```diff
@@ -520,9 +520,16 @@ class StoreAttr(object):
                     "Quantization of the weights requires nncf, please install it with `pip install nncf`"
                 )
 
+            from optimum.intel.openvino.configuration import _GPTOSSQuantizationConfig
             from optimum.intel.openvino.quantization import _weight_only_quantization
 
-            _weight_only_quantization(submodel, quantization_config)
+            if isinstance(quantization_config, _GPTOSSQuantizationConfig):
+                _weight_only_quantization(submodel, quantization_config.quantization_config1)
```
rkazants, line 527: please add a comment - why we set two configs specifically for GPT-OSS

optimum/intel/openvino/configuration.py
```diff
@@ -316,13 +316,17 @@ class OVQuantizationMethod(str, Enum):
...
-        "bits": 4,
-        "sym": True,
-        "group_size": 32,
-        "ignored_scope": {
-            "patterns": [".*self_attn.*", ".*router.*"],
+        "quantization_config1": {
+            "bits": 4,
+            "sym": True,
+            "group_size": 32,
+            "ignored_scope": {"patterns": [".*self_attn.*", ".*router.*"]},
```
rkazants, line 323: please add a commit why we ignore some weights. Or make a common comment for GPT-OSS - why we should use mixed precision.

optimum/intel/openvino/configuration.py
```diff
@@ -1149,6 +1153,8 @@ def _get_dtype(quantization_config):
         elif isinstance(quantization_config, OVPipelineQuantizationConfig):
             dtypes = [OVConfig._get_dtype(config) for config in quantization_config.quantization_configs.values()]
             dtype = "_".join(dtypes)
+        elif isinstance(quantization_config, _GPTOSSQuantizationConfig):
+            dtype = "int4_int8"
```
rkazants: how is it used because I see no usage?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -876,7 +876,13 @@ def _from_pretrained(
             init_cls = cls
 
         if isinstance(quantization_config, dict) and quantization_config == {"bits": 4}:
-            default_config = get_default_quantization_config(config.name_or_path, weight_format="int4")
+            if config.name_or_path == "openai/gpt-oss-20b":
+                raise NotImplementedError(
+                    "Quantization with the default 4-bit config is not supported through Python API for openai/gpt-oss-20b model. "
+                    "Please export the model via optimum-cli with `--weight-format int4` argument."
```
rkazants, line 882: Shall we mention that the result will be not pure 4-bit quantization?

optimum/exporters/openvino/__main__.py
nikita-savelyevv: Added

optimum/intel/openvino/configuration.py
nikita-savelyevv: Added

optimum/intel/openvino/configuration.py
nikita-savelyevv: This is used when `{"bits": 4}` is selected as the quantization config during loading via `from_pretrained()`:
```python
OVModelForCausalLM.from_pretrained("openai/gpt-oss-20b", quantization_config={"bits": 4})
```

optimum/intel/openvino/modeling_decoder.py
nikita-savelyevv: I believe it is fine, because we have default mixed int4/int8 configs for many of the current models and we don't mention that they will be not purely int4. Also, for this model only a small proportion (10%) of model parameters will actually be quantized to int8.

nikita-savelyevv:
> Thanks, Nikita! Could you please add the same config for `openai/gpt-oss-120b`?

Added

optimum/exporters/openvino/__main__.py
```diff
@@ -520,9 +520,16 @@ class StoreAttr(object):
...
+                _weight_only_quantization(
+                    submodel, quantization_config.quantization_config2, verify_not_optimized=False
+                )
```
echarlaix, line 530: will this ever be needed for other models or should this be deprecated soon?  also why not just make it a list and iterate on it ? (making it more general than `quantization_config1` / `quantization_config2`

optimum/intel/openvino/configuration.py
echarlaix: why do we need to first apply int4 quantization and then later int8 quantization ? also why do we need a new dtype `int4_int8` ? (if we have a mix of int4 and int8, isn't it already the case wuen applying int4 quantization with a ratio which is not 1

optimum/exporters/openvino/__main__.py
nikita-savelyevv: > will this ever be needed for other models or should this be deprecated soon?

At the moment we don't see any other models in the near future for which it will be needed and hope it will be deprecated soon.

> also why not just make it a list and iterate on it ? (making it more general than quantization_config1 / quantization_config2

There was a plan to add a proper `OVSequentialQuantizationConfig` which would contain a list of configs and be a more general solution, but it will require significant time to be implemented. That's why we've decided to implement a workaround for now. As for `_GPTOSSQuantizationConfig`, it can also hold a list instead of two configs, but I've decided to be more explicit to highlight that only two are required.

optimum/intel/openvino/configuration.py
nikita-savelyevv: > why do we need to first apply int4 quantization and then later int8 quantization ? 

There are some layers which if quantized to int4 ruin the model quality. At the same time, we can't keep them in original bf16 precision, because it is not supported by GPU plugin yet. That's why they have to quantized to int8 -- this is done in the second quantization pass.

> also why do we need a new dtype int4_int8 ? (if we have a mix of int4 and int8, isn't it already the case wuen applying int4 quantization with a ratio which is not 1

`OVConfig.get_dtype()` must return something for every type of quantization config. That's why I've made it to return an arbitrary value of `int4_int8`. To be honest, I'm not sure when this value is even needed. Seems like it's purely for logging.

I suppose we can update it to `int4` to be consistent with int4 case with ratio <1.

optimum/intel/openvino/configuration.py
```diff
@@ -1366,6 +1372,36 @@ def post_init(self):
             submodel_config.post_init()
 
 
+class _GPTOSSQuantizationConfig(QuantizationConfigMixin):
```
echarlaix, line 1375: before merging can you just make sure that  `_GPTOSSQuantizationConfig`  can be loaded correctly / saved ? (maybe this is already tested somewhere but dont see it)  I'm not sure that inhereting from `QuantizationConfigMixin` is enough to make it a valid config (that can be saved and re-loaded as expected)

echarlaix (approved):
LGTM, just left one comment, feel free to merge when resolved

optimum/intel/openvino/configuration.py
nikita-savelyevv: That's a good point! Added serialization/deserialization test cases.

PR #1484, opened 2025-10-21
Apply chat_template for MiniCPM-o-2_6

Wovchena:
# What does this PR do?

`MiniCPMOProcessor` doesn't have chat_template but `processor.tokenizer` does. That resulted in `preprocess_inputs()` skipping `apply_chat_template()`. `openbmb/MiniCPM-o-2_6`'s implementation applies chat_template: https://huggingface.co/openbmb/MiniCPM-o-2_6/blob/main/modeling_minicpmo.py#L959

The patch affects newly added `openbmb/MiniCPM-o-2_6` only to keep it safe . Although `openbmb/MiniCPM-V-2_6` has this problem as well.

```python
from PIL import Image
import requests

model_id="openbmb/MiniCPM-o-2_6"

prompt= "What is in the image?"

inputs = model.preprocess_inputs(prompt, image, processor)

result = model.generate(**inputs, max_new_tokens=20)
```

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -2085,11 +2085,6 @@ class _OVMiniCPMVForCausalLM(OVModelForVisualCausalLM):
         return vllm_embedding, attention_mask, position_ids
 
     @staticmethod
-    def image_tag_apply_chat_template(text, image, templater):
-        messages = [{"role": "user", "content": text if image is None else "(<image>./</image>)\n" + text}]
-        return templater.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
-
-    @staticmethod
     def preprocess_inputs(
         text: str,
         image: Optional["Image"] = None,
@@ -2106,9 +2101,8 @@ class _OVMiniCPMVForCausalLM(OVModelForVisualCausalLM):
         if audio is not None:
             raise ValueError("Audio input is not supported")
         if getattr(processor, "chat_template", None) is not None:
-            prompt = _OVMiniCPMVForCausalLM.image_tag_apply_chat_template(text, image, processor)
-        elif "MiniCPMOProcessor" == processor.__class__.__name__:
-            prompt = _OVMiniCPMVForCausalLM.image_tag_apply_chat_template(text, image, processor.tokenizer)
+            messages = [{"role": "user", "content": text if image is None else "(<image>./</image>)\n" + text}]
+            prompt = processor.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
         else:
             prompt = (
                 f"<|im_start|>user\n(<image>./</image>)\n{text}<|im_end|>\n<|im_start|>assistant\n"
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -2101,8 +2106,9 @@ def preprocess_inputs(
         if audio is not None:
             raise ValueError("Audio input is not supported")
         if getattr(processor, "chat_template", None) is not None:
-            messages = [{"role": "user", "content": text if image is None else "(<image>./</image>)\n" + text}]
-            prompt = processor.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
+            prompt = _prepend_minicpm_image_tag_apply_chat_template(text, image, processor)
```
rkazants: let us make it as method `image_tag_apply_chat_template` of class and just override this method for `_OVMiniCPMOForCausalLM` class below

Wovchena: Moved to the class. I didn't get the override part. What should be overridden with what and at what stage?

rkazants: Let us just override `preprocess_inputs` for `_OVMiniCPMOForCausalLM`. So it will be more clear. Some day we will return to `preprocess_inputs` for minicom-v when we will be sure that this is correct for minicpm-v as well after validation.
Sorry, I think new method is not needed

Wovchena: `_OVMiniCPMOForCausalLM` itself overrides `OVModelForVisualCausalLM`'s `preprocess_inputs()`. What approach should I take to override `_OVMiniCPMOForCausalLM.preprocess_inputs()`?

rkazants: just write own implementation of `preprocess_input` for `_OVMiniCPMOForCausalLM` and do not change one for MiniCPMV

Wovchena: Overridden

tests/openvino/test_seq2seq.py
```diff
@@ -625,9 +625,11 @@ def test_compare_to_transformers(self, model_arch):
         ov_model.config.eos_token_id = None
         transformers_model.config.eos_token_id = None
         ov_model.generation_config.do_sample = False
+        # minicpmo diverges after 20 tokens
+        tokens_to_generate = 20 if model_arch == "minicpmo" else 30
```
IlyasMoutawwakil, line 629: do we know why ?

Wovchena: Measurements on a tiny dataset (without my changes) show that there's a difference for sure between the full models. I don't know if it's caused by inference backends or the wrapping implementation. I tried forcing f32 kv cache for optimum but it didn't help

echarlaix: @Wovchena did you observe this with https://huggingface.co/rkazants/tiny-random-MiniCPM-o-2_6 or an other model such as https://huggingface.co/openbmb/MiniCPM-o-2_6 ?

IlyasMoutawwakil: does it diverge after exactly 20 ?

Wovchena: Dataset based test was run on `openbmb/MiniCPM-o-2_6`

The test fails if I set 21

rkazants: @Wovchena, does it diverge on the tiny model after iter. 20? Please clarify

Wovchena: Yes

rkazants: I believe that we have the same diverge for minicpm-v but it requires more iteration or applying this fix in chat template. This issue we should solve separately.

nikita-savelyevv (commented):
> The patch affects newly added openbmb/MiniCPM-o-2_6 only to keep it safe . Although openbmb/MiniCPM-V-2_6 has this problem as well.

Since we know that MiniCPM-V is also affected, why don't we fix both models at the same time? What risks does this have?

Wovchena:
That's @rkazants's request. openbmb/MiniCPM-o-2_6 isn't fixed because it has that problem for a long time so it's not so important to fix. He also mentioned that other MiniCPM models are being added and he'd prefer avoiding affecting that. If I introduce a new problem with this PR, it will leech resources which are already short.

IlyasMoutawwakil (approved):
LGTM ! thanks for the fix ! please investigate if the divergence hurt performance of the original model (not the tiny one).

PR #1485, opened 2025-10-22
Fix some minor bugs after upgrading transformers to 4.55 for IPEX

kaixuanliu:


optimum/intel/ipex/modeling_base.py
```diff
@@ -311,10 +320,15 @@ def __init__(
 
         self.generation_config = GenerationConfig.from_model_config(self.config)
         try:
-            self.model_cls = get_class_from_dynamic_module(
-                self.config.auto_map["AutoModelForCausalLM"], model_save_dir
-            )
-        except AttributeError:
+            # Use model_save_dir if available, otherwise use config's name_or_path
+            pretrained_model_name_or_path = model_save_dir or getattr(self.config, "_name_or_path", None)
```
echarlaix, line 324: instead of extracting from the config why not just get it from https://github.com/huggingface/optimum-intel/blob/fb8fd987b37776e9a8bb1954f62fedf0eb39bd80/optimum/intel/ipex/modeling_base.py#L185
by setting `model_save_dir=model_id`

kaixuanliu: Hi @echarlaix , sometimes we do not go into from_pretrained func defined here. For example:
```
generator = pipeline(
        "text-generation",
        model=model_id,
        tokenizer=tokenizer,
        model_kwargs=model_kwargs,
    )
generator.model = IPEXModelForCausalLM(generator.model, dtype=torch_dtype)
```
so it is safer to get from config.

echarlaix: imo it's expected to call `from_pretrained` when creating an instance of  `IPEXModel` so not sure to agree here.
For pipelines the recommended path should actually be to use optimum pipelines and just set the `accelerator="ipex"` as follows:

```python
from optimum.intel.pipelines import pipeline

pipe = pipeline("text-generation", model_id, accelerator="ipex")
 ```

echarlaix: in some cases the `_name_or_path` from the config is not correct, so I think you can apply the suggestion above https://github.com/huggingface/optimum-intel/pull/1485#discussion_r2455183093 and use the `model_id` when possible, and when it's not the case (for cases when `from_pretrained` wasn't called) you can fallback on `config._name_or_path`, what do you think ?

kaixuanliu: It makes sense. Have updated the code. Thx for your advice~

PR #1486, opened 2025-10-22
[OpenVINO][Docs] Update docs to mention how to apply default 4-bit config via Python API

nikita-savelyevv:
# What does this PR do?

As in the title.

## Before submitting

### DIFF
--- a/docs/source/openvino/optimization.mdx
+++ b/docs/source/openvino/optimization.mdx
@@ -894,7 +894,7 @@ from optimum.intel import OVModelForCausalLM
 model = OVModelForCausalLM.from_pretrained(model_id, quantization_config={"bits": 4})
 ```
 
-For some models, we provide preconfigured 4-bit weight-only quantization configurations ([link](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py)) that offer a good trade-off between quality and speed. This default 4-bit configuration is applied automatically when you specify `quantization_config={"bits": 4}`.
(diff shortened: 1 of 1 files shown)


docs/source/openvino/optimization.mdx
```diff
@@ -889,23 +889,26 @@ If quantization_config is not provided, model will be exported in 8 bits by defa
...
 ```python
-from optimum.intel import OVModelForCausalLM, OVWeightQuantizationConfig
+from optimum.intel import OVModelForCausalLM
 
-quantization_config = OVWeightQuantizationConfig(bits=4)
-model = OVModelForCausalLM.from_pretrained(model_id, quantization_config=quantization_config)
+model = OVModelForCausalLM.from_pretrained(model_id, quantization_config={"bits": 4})
 ```
 
+For some models, we provide preconfigured 4-bit weight-only quantization configurations ([link](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py)) that offer a good trade-off between quality and speed. This default 4-bit configuration is applied automatically when you specify `quantization_config={"bits": 4}`.
```
echarlaix: not very important but would have instead (feel free to close)

```suggestion
For some models, we provide preconfigured 4-bit weight-only quantization [configurations](https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py) that offer a good trade-off between quality and speed. This default 4-bit configuration is applied automatically when you specify `quantization_config={"bits": 4}`.
```

nikita-savelyevv: Sure!

PR #1489, opened 2025-10-23
[OpenVINO] Add transformers==4.52 constraint to vision_language_quantization notebook

nikita-savelyevv:
# What does this PR do?

To avoid possible issues on SPR CPUs. Should be replaced by "openvino>=2025.4" after its release.

## Before submitting


echarlaix (approved):
Thanks!

PR #1490, opened 2025-10-23
[OpenVINO] Add workaround logic for default int4 quantization of openai/gpt-oss-20b model

nikita-savelyevv:
# What does this PR do?

For correct 4-bit weight-only quantization of openai/gpt-oss-20b we currently have to run quantization twice. Unfortunately, we don't have such logic in optimum-intel yet, and its proper implementation will take a while. Because of this, this PR adds a workaround to enable this use case. The export command is a usual:
```
optimum-cli export openvino -m openai/gpt-oss-20b --weight-format int4 ./gpt-oss-20b
```

We hope that inference with bf16 weights will soon be supported in GPU plugin and then the second quantization step will not be required.

Duplicate of accidentally closed #1482

## Before submitting

### DIFF
--- a/tests/openvino/test_quantization.py
+++ b/tests/openvino/test_quantization.py
@@ -78,6 +78,7 @@ from optimum.intel.openvino.configuration import (
     _DEFAULT_4BIT_WQ_CONFIGS,
     _DEFAULT_4BIT_WQ_CONFIG,
     _quantization_config_from_dict,
+    _GPTOSSQuantizationConfig,
 )
 from optimum.intel.openvino.modeling_visual_language import _OVNanoLlavaForCausalLM
 from optimum.intel.openvino.utils import TemporaryDirectory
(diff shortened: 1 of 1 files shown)


nikita-savelyevv:
Merging as got the required approvals at #1482

optimum/intel/openvino/configuration.py
```diff
@@ -1149,6 +1170,8 @@ def _get_dtype(quantization_config):
         elif isinstance(quantization_config, OVPipelineQuantizationConfig):
             dtypes = [OVConfig._get_dtype(config) for config in quantization_config.quantization_configs.values()]
             dtype = "_".join(dtypes)
+        elif isinstance(quantization_config, _GPTOSSQuantizationConfig):
+            dtype = "int4_int8"
```
echarlaix, line 1174: not very important but in the other PR it was changed to int4 https://github.com/huggingface/optimum-intel/pull/1482/ (feel free to close if you think this is more adapted)

nikita-savelyevv: Thanks for noticing!

PR #1491, opened 2025-10-24
Support exaone4 model

zhaohb:
Just for EXAONE-4.0-1.2B

ticket : [175853](https://jira.devtools.intel.com/browse/CVS-175853)

# What does this PR do?

Conversion cmd-line for LGAI-EXAONE/EXAONE-4.0-1.2B:
```shell
optimum-cli export openvino -m  LGAI-EXAONE/EXAONE-4.0-1.2B EXAONE-4.0-1.2B
```
Inference of LGAI-EXAONE/EXAONE-4.0-1.2B using OpenVINO backend:

```python
import torch

model_dir="LGAI-EXAONE/EXAONE-4.0-1.2B"

messages = [
    {"role": "user", "content": prompt}
]
input_ids = tokenizer.apply_chat_template(
    messages,
    tokenize=True,
    add_generation_prompt=True,
    return_tensors="pt"
)

# Run inference
output_ids = model.generate(input_ids, max_new_tokens=10)

print(output_text)
```

## Before submitting
* [N/A] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
* [x]  Did you make sure to update the documentation with your changes?
* [x]  Did you write any new necessary tests?

### DIFF
--- a/docs/source/openvino/models.mdx
+++ b/docs/source/openvino/models.mdx
@@ -80,6 +80,7 @@ Here is the list of the supported architectures :
 - InternVL2
 - Jais
 - Levit
+- Lfm2
 - Llama
 - Llama4
 - Llava
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -104,6 +104,7 @@ from .model_patcher import (
     InternVL2ChatLangModelPatcher,
     InternVLChatImageEmbeddingModelPatcher,
     JaisModelPatcher,
+    Lfm2ModelPatcher,
     Llama4ImageEmbeddingsModelPatcher,
     Llama4TextModelPatcher,
     LlavaImageEmbeddingModelPatcher,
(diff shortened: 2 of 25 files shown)


optimum/exporters/openvino/model_configs.py
```diff
@@ -592,6 +592,20 @@ class GptOssOpenVINOConfig(LlamaOpenVINOConfig):
...
+    *[
+        "feature-extraction",
+        "feature-extraction-with-past",
+        "text-generation",
+        "text-generation-with-past",
+        "text-classification",
+    ],
+    library_name="transformers",
+)
+class Exaone4OpenVINOConfig(LlamaOpenVINOConfig):
```
rkazants, line 616: please add tests

zhaohb: Hi, how should I add tests, especially regarding the model storage path? 
Are there any PRs I can refer to?

zhaohb:
> Please provide tests and JIRA ticket.

Hi, I had add JITA ticket and tests. 
Thank you very much.

optimum/exporters/openvino/model_configs.py
zhaohb: Done

optimum/exporters/openvino/model_configs.py
```diff
@@ -592,6 +592,20 @@ class GptOssOpenVINOConfig(LlamaOpenVINOConfig):
...
     pass
 
+@register_in_tasks_manager(
+    "exaone4",
+    *[
+        "feature-extraction",
+        "feature-extraction-with-past",
+        "text-generation",
+        "text-generation-with-past",
+        "text-classification",
```
rkazants: do you validate tasks "feature-extraction", "text-classification" in the precommit?
Please check

docs/source/openvino/models.mdx
rkazants, line 1: since we have already supported previous "exaone" version. Please add tests for quantization. It should work as well.

optimum/exporters/openvino/model_configs.py
zhaohb: I tested the tasks "feature-extraction" and "text-classification" - the model conversion works correctly without quantization

docs/source/openvino/models.mdx
zhaohb: However, when applying quantization, feature-extraction works correctly, but text-classification throws an error:
```shell
nncf.errors.InvalidGroupSizeError: Compression of the '__module.model.rotary_emb/aten::matmul/MatMul' layer failed with the following error:
Channel size 1 should be divisible by size of group 128.
Ensure that the group size is divisible by the channel size, or include this node and others with similar issues in the ignored scope:
nncf.compress_weight(
        ...,
        ignored_scope=IgnoredScope(names=["__module.model.rotary_emb/aten::matmul/MatMul"]
        )
)
```

zhaohb: I see. It's likely that the shape of my dummy  model is incorrect. I may need to adjust the shape of the dummy model.

nikita-savelyevv: > However, when applying quantization, feature-extraction works correctly, but text-classification throws an error:
> 
> ```shell
> nncf.errors.InvalidGroupSizeError: Compression of the '__module.model.rotary_emb/aten::matmul/MatMul' layer failed with the following error:
> Channel size 1 should be divisible by size of group 128.
> Ensure that the group size is divisible by the channel size, or include this node and others with similar issues in the ignored scope:
> nncf.compress_weight(
>         ...,
>         ignored_scope=IgnoredScope(names=["__module.model.rotary_emb/aten::matmul/MatMul"]
>         )
> )
> ```

What version of nncf you are using?

zhaohb: @nikita-savelyevv I don't think this error is related to the NNCF. It's more likely an issue with my dummy model.
I will double check that. thank you.

nikita-savelyevv: @zhaohb It could be, because there was a fix addressing this issue. In any case I would suggest to update to NNCF v 2.18

zhaohb: @nikita-savelyevv
Hi, the issue disappeared after upgrading to version 2.18. It wasn't a problem with the dummy model after all. 
Thank you!

zhaohb: > since we have already supported previous "exaone" version. Please add tests for quantization. It should work as well.

@rkazants Hi, I was unable to find quantization tests for the "exaone" model.

optimum/exporters/openvino/model_configs.py
```diff
@@ -597,6 +597,21 @@ class ExaoneOpenVINOConfig(LlamaOpenVINOConfig):
...
 
 
+@register_in_tasks_manager(
+    "exaone4",
+    *[
+        "feature-extraction",
+        "feature-extraction-with-past",
+        "text-generation",
+        "text-generation-with-past",
+        "text-classification",
```
rkazants: `OVModelForSequenceClassification` is not validated. I see that this task is not validated in precommit.
The same problem is for `feature-extraction`.

optimum/exporters/openvino/model_configs.py
rkazants, line 1: please also add test in `test_export.py` to check that exporting works in Python.
Also, need quantization tests in precommit: weights compression, quantization.

optimum/exporters/openvino/model_configs.py
zhaohb:
```python
diff --git a/tests/openvino/test_exporters_cli.py b/tests/openvino/test_exporters_cli.py
index 61d13f6..05b43dd 100644
--- a/tests/openvino/test_exporters_cli.py
+++ b/tests/openvino/test_exporters_cli.py
@@ -116,6 +116,8 @@ class OVCLIExportTestCase(unittest.TestCase):
         SUPPORTED_ARCHITECTURES.extend(
             [
+                ("text-generation-with-past", "exaone4"),
+                ("feature-extraction", "exaone4"),
+                ("text-classification", "exaone4"),
             ]
         )
     EXPECTED_NUMBER_OF_TOKENIZER_MODELS = {
```

Do you mean to modify it like this?

tests/openvino/test_quantization.py
```diff
@@ -993,7 +1003,7 @@ def test_filtered_architectures(cls):
         elif is_transformers_version("<", "4.52"):
             expected = set()
         else:
-            expected = {"llava-qwen2", "phi3_v", "minicpmo"}
+            expected = {"llava-qwen2", "phi3_v", "minicpmo", "exaone4"}
```
nikita-savelyevv: This is more readable in my opinion:
```python
expected = set()
    expected.add("qwen2_5_vl")
if is_transformers_version("<", "4.51"):
    expected.add("llama4")
if is_transformers_version("<", "4.54"):
    expected.add("exaone4")
if is_transformers_version(">=", "4.54"):
    expected.update({"llava-qwen2", "phi3_v", "minicpmo"})
```

tests/openvino/test_quantization.py
```diff
@@ -979,14 +989,15 @@ class OVWeightCompressionTest(unittest.TestCase):
...
-            expected = set()
-        else:
-            expected = {"llava-qwen2", "phi3_v", "minicpmo"}
+            expected.add("qwen2_5_vl")
+        if is_transformers_version("<", "4.51"):
+            expected.add("llama4")
+        if is_transformers_version("<", "4.54"):
+            expected.add("exaone4")
+        if is_transformers_version(">=", "4.54"):
+            expected.update({"llava-qwen2", "phi3_v", "minicpmo"})
```
IlyasMoutawwakil, line 1000: should >= 4.52 no ?

nikita-savelyevv: For these models, the max transformers version is 4.53.3, e.g. https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_configs.py#L1889 That's why we should skip them when `transformers>=4.54`.

IlyasMoutawwakil: but this line defines when to add them, not when to skip them, or am i understanding this incorrectly ?

nikita-savelyevv: Yeah, this test can be a bit confusing. `expected` here is the expected set of models which will be skipped, i.e. later we compare it to `skipped` -- a set of models which were removed from `LOAD_IN_4_BITS_SCOPE`. So when we add a model to `expected`, we expect it to be skipped 🙂

IlyasMoutawwakil: thanks !

PR #1493, opened 2025-10-24
removed nf4_fp8 modes

ljaljushkin:
# What does this PR do?

removes `nf4_f8e4m3` and `nf4_f8e5m2` mixed precision modes. 
keep cb4_f8e4m3 instead.

# Reason
nf4_fp8 types don't make sense in terms of performance.
nf4 assumes fp16 values. even with fp8 activation, the compute will be done in fp16.
the alternative solution is using cb4 data type what can be casted to fp8 without accuracy loss. then with fp8 activation it would be possible to compute in fp8.

## Before submitting


echarlaix (approved):
LGTM, thanks for the PR!

PR #1497, opened 2025-10-27
Fix test_decoder for gpt_oss

rkazants:
# What does this PR do?

OpenVINO 2025.4 provides enhancements for GPT-OSS that leads to more optimized IR.
You can see results for nightly OpenVINO for which IR gets to contain two SDPAs for test model: https://github.com/huggingface/optimum-intel/actions/runs/18830164857/job/53720096823

Fixes # (issue)

## Before submitting


echarlaix (approved):
Thanks a lot for the fix !

PR #1499, opened 2025-10-29
[OpenVINO] Update SAM low precision references

nikita-savelyevv:
# What does this PR do?

Update int4/int8 reference numbers after an ignored pattern was added for positional embedding layer https://github.com/openvinotoolkit/nncf/pull/3700 .

## Before submitting

### DIFF
--- a/tests/openvino/test_exporters_cli.py
+++ b/tests/openvino/test_exporters_cli.py
@@ -458,7 +458,7 @@ class OVCLIExportTestCase(unittest.TestCase):
             "--dataset coco --num-samples 1",
             {
                 "vision_encoder": 75,
-                "prompt_encoder_mask_decoder": 61,
+                "prompt_encoder_mask_decoder": 61 if is_nncf_version("<=", "2.18") else 60,
             },
             {
                 "vision_encoder": {"int8": 75},


nikita-savelyevv:
Merging as failing tests are unrelated.

echarlaix (approved):

Not related to this PR but I'm seeing a new error for export in fp16 (for sam and gpt2 models) with openvino-nightly which previously wasn't present, any chances you have an idea on what could be the cause ? 
```
RuntimeError: Check 'node->constant_fold(outputs, node->input_values())' failed at src/core/src/xml_util/xml_serialize_util.cpp:59
```

nikita-savelyevv:
> LGTM, thanks @nikita-savelyevv !
> 
> Not related to this PR but I'm seeing a new error for export in fp16 (for sam and gpt2 models) with openvino-nightly which previously wasn't present, any chances you have an idea on what could be the cause ?
> 
> ```
> RuntimeError: Check 'node->constant_fold(outputs, node->input_values())' failed at src/core/src/xml_util/xml_serialize_util.cpp:59
> ```

Thanks @echarlaix! I don't have this understanding yet, we will investigate this internally.

cc @rkazants

Issue #1498, opened 2025-10-28
[BUG] Import optimum-intel broke export


AlexanderDokuchaev:
This issue occurs due to an incorrect task inference by `TasksManager._infer_task_from_model_or_model_class(model_class=cls.auto_model_class)`.
The expected task is `text-generation`, but the function incorrectly detects it as `image-text-to-text`.

Reproducer:

```sh
pip install optimum-intel[openvino]==1.26.0 onnxruntime==1.23.2
```

```python
from optimum.onnxruntime import ORTModelForCausalLM
from optimum.exporters.tasks import TasksManager

model_id = "tinyllama/tinyllama-1.1b-step-50k-105b"
model_hf = ORTModelForCausalLM.from_pretrained(model_id, export=True)  # Works


import optimum.exporters.openvino.model_configs
model_hf = ORTModelForCausalLM.from_pretrained(model_id, export=True)

# ValueError: Asked to export a llama model for the task image-text-to-text-with-past, but the Optimum ONNX exporter only supports the tasks feature-extraction, feature-extraction-with-past, text-generation, text-generation-with-past, text-classification for llama. Please use a supported task. Please open an issue at https://github.com/huggingface/optimum/issues if you would like the task image-text-to-text-with-past to be supported in the ONNX export for llama.
```

PR #1501, opened 2025-10-29
Fix TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS

echarlaix:
Fixes https://github.com/huggingface/optimum-intel/issues/1498
*  `TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS` shouldn't be updated as it'll cause issue when inferring task with `_infer_task_from_model_or_model_class` https://github.com/huggingface/optimum/blob/2e8e23285c87f74d0d56a4c0e676ec3ef8fe2583/optimum/exporters/tasks.py#L708 (this can results in issue for onnx export for example as image-text-to-text could be detected for onnx when it should actually be text-generation)
* deprecate `OVQuantizer` task attribute (not needed since nn.Modules are not compatible with `OVQuantizer` since v1.26)

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -1148,7 +1148,7 @@ class OVQuantizer(OptimumQuantizer):
         self.dataset_builder = OVCalibrationDatasetBuilder(model, seed)
         self._task = task
         if self._task is not None:
-            logger.warning("The `task` argument is ignored and will be remvoed in optimum-intel v1.27")
+            logger.warning("The `task` argument is ignored and will be removed in optimum-intel v1.27")
 
     @property
     def task(self) -> Dict[str, Union[openvino.Model, openvino.runtime.CompiledModel]]:


optimum/exporters/openvino/__main__.py
```diff
@@ -394,7 +394,7 @@ class StoreAttr(object):
             model = _OpenClipForZeroShotImageClassification.from_pretrained(model_name_or_path, cache_dir=cache_dir)
         else:
             model = TasksManager.get_model_from_task(
-                task,
+                "text-generation" if trust_remote_code and task == "image-text-to-text" else task,
```
echarlaix: temporary fix : we currently load "image-text-to-text" models with `AutoModelForCausalLM` and not `AutoModelForImageTextToText`, this seems to be needed for remote code models like [phi3_v](https://huggingface.co/microsoft/Phi-3.5-vision-instruct/blob/main/config.json#L9) internvl2, minicpmv, internvl2, nanollava, maira2

WIll open a PR to load remote code models using `config.auto_map` instead (for other models we can directly use `config.architectures`)

optimum/exporters/openvino/model_configs.py
```diff
@@ -198,12 +198,6 @@ def init_model_configs():
         "AutoModelForImageTextToText",
     )
 
-    TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS[
-        "image-text-to-text"
-    ] = TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS["text-generation"]
-
-    TasksManager._TRANSFORMERS_TASKS_TO_MODEL_LOADERS["video-text-to-text"] = "AutoModelForVision2Seq"
```
echarlaix, line 205: https://github.com/huggingface/transformers/blob/6efc1799c1d0665c41717c555f3551d974e4d61a/src/transformers/models/auto/modeling_auto.py#L2283

optimum/intel/openvino/quantization.py
```diff
@@ -1142,15 +1140,20 @@ def __init__(self, model: OVModel, task: Optional[str] = None, seed: int = 42, *
...
             seed (`int`, defaults to 42):
                 The random seed to use when shuffling the calibration dataset.
         """
         super().__init__()
         self.model = model
-        self.task = task
         self.dataset_builder = OVCalibrationDatasetBuilder(model, seed)
+        self._task = task
+        if self._task is not None:
+            logger.warning("The `task` argument is ignored and will be remvoed in optimum-intel v1.27")
```
nikita-savelyevv:
```suggestion
            logger.warning("The `task` argument is ignored and will be removed in optimum-intel v1.27")
```

nikita-savelyevv (approved):
Thanks @echarlaix!

optimum/intel/openvino/quantization.py
echarlaix: good catch thanks @nikita-savelyevv !

PR #1504, opened 2025-10-30
Fix Phi long context issue

helena-intel:
This is https://github.com/huggingface/optimum-intel/pull/1297 updated to latest main branch. 

Currently inference on Phi-3-mini and Phi-4-mini returns bad outputs (random characters) when context gets larger than about 2000 tokens. This PR, contributed by @eaidova , fixes that. This is not my code. The original PR is no longer being updated; I'm making this a new PR to make it easier to discuss and add updates.

I saw no negative impact on inference speed. I see slightly different outputs with shorter contexts on SPR (on inference with the model exported with the PR vs the model exported with main). Any suggestions to fix that would be much appreciated.

Draft PR for now, awaiting some feedback and testing, but I hope we can merge this soon.

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -1501,7 +1501,7 @@ def _phi3_longrope_forward(self, x, position_ids):
     the sequence length exceeds the original max position embeddings used during pretraining.
 
     Note: In transformers, the @dynamic_rope_update decorator replaces self.inv_freq before the forward pass.
-    Here we use torch.where to select between inv_freq and long_inv_freq and add the selection logic into the model graph.
+    Here we use torch.where to select between original_inv_freq and long_inv_freq and add the selection logic into the model graph.
     """
     seq_len = torch.max(position_ids) + 1
     original_max_position_embeddings = (
@@ -1515,7 +1515,7 @@ def _phi3_longrope_forward(self, x, position_ids):
     inv_freq = torch.where(
         seq_len > original_max_position_embeddings,
         self.long_inv_freq,
-        self.inv_freq,
+        self.original_inv_freq,
     )
 
     inv_freq_expanded = inv_freq[None, :, None].float().expand(position_ids.shape[0], -1, 1)
--- a/tests/openvino/test_decoder.py
+++ b/tests/openvino/test_decoder.py
@@ -56,7 +56,6 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "pegasus",
         "qwen",
         "phi",
-        "phi3-longrope",
         "internlm2",
         "orion",
         "falcon",
@@ -91,6 +90,7 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
 
     if is_transformers_version(">=", "4.49"):
         SUPPORTED_SSM_ARCHITECTURES += ("zamba2",)
+        SUPPORTED_ARCHITECTURES += ("phi3", "phi3-longrope")
 
     if is_transformers_version(">=", "4.54.0") and is_openvino_version(">=", "2025.4.0"):
         SUPPORTED_SSM_ARCHITECTURES += ("lfm2",)
@@ -110,9 +110,6 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         # autoawq install disabled for windows test environment
         if is_openvino_version(">=", "2024.6.0") and platform.system() != "Windows":
             SUPPORTED_ARCHITECTURES += ("mixtral_awq",)
-
-    if is_transformers_version(">=", "4.49.0"):
-        SUPPORTED_ARCHITECTURES += ("phi3",)
 
     if is_transformers_version(">", "4.49"):
         SUPPORTED_ARCHITECTURES += ("gemma3_text",)


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -358,10 +358,10 @@ def _export(
...
         )
 
-        if config.model_type == "phi3" and config.max_position_embeddings != getattr(
-            config, "original_max_position_embeddings", config.max_position_embeddings
-        ):
-            config.max_position_embeddings = config.original_max_position_embeddings
+        # if config.model_type == "phi3" and config.max_position_embeddings != getattr(
+        #     config, "original_max_position_embeddings", config.max_position_embeddings
+        # ):
+        #     config.max_position_embeddings = config.original_max_position_embeddings
```
helena-intel: This should be deleted but I left it here for now for context.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1493,15 +1493,54 @@ def _phi3_self_attn_sdpa_forward(
     return attn_output, None, past_key_value
 
 
+# @torch.jit.script
```
rkazants: I think it makes sense to add the test with long prompt. Is this issue reproduced on tiny-model?

nikita-savelyevv:
>  I see slightly different outputs with shorter contexts on SPR (on inference with the model exported with the PR vs the model exported with main). 

I believe minor differences are expected on SPR. But if possible, WWB similarity should be run to see if the difference is significant or not.

optimum/exporters/openvino/model_patcher.py
rkazants: please remove not needed comments and commented out code

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -950,6 +947,47 @@ def _from_pretrained(
         return causal_model
 
 
+class OVPhi3ForCausalLM(OVModelForCausalLM):
+    def prepare_inputs_for_generation(
```
echarlaix, line 914: would you mind adding a link to the original code 
https://github.com/huggingface/transformers/blob/v4.57.0/src/transformers/models/phi3/modeling_phi3.py#L493

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1493,16 +1493,51 @@ def _phi3_self_attn_sdpa_forward(
...
     def __enter__(self):
         super().__enter__()
 
-        # currently, long RoPE can not be traced for long context support, disable it for avoid potential accuracy issues
-        if self._model.config.max_position_embeddings != getattr(
-            self._model.config, "original_max_position_embeddings", self._model.config.max_position_embeddings
-        ):
-            self._model.config.max_position_embeddings = self._model.config.original_max_position_embeddings
-
+        # currently, long RoPE can not be traced for long context support, disable it to avoid potential accuracy issues
```
rkazants: now I think we don't need this comment. We solve the problem by this PR.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1493,16 +1493,51 @@ def _phi3_self_attn_sdpa_forward(
...
+        self.original_max_position_embeddings
+        if hasattr(self, "original_max_positional_embeddings")
+        else self.config.original_max_position_embeddings
+    )
+    max_position_embeddings = (
+        self.max_position_embeddings
+        if hasattr(self, "max_position_embeddings")
+        else self.config.max_position_embeddings
+    )
+    inv_freq = select_ext_factor(seq_len, original_max_position_embeddings, self.inv_freq, self.long_inv_freq)
```
rkazants: Let us add a comment:

`slow down all frequencies by scale factor for long prompts that makes attention more stable, i.e. preserve model accuracy`

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1529,6 +1564,23 @@ def __enter__(self):
...
+                torch.device("cpu"),
+                seq_len=self._model.config.original_max_position_embeddings + 1,
+            )
+            self._model.model.rotary_emb.long_inv_freq = long_inv_freq
+            self._model.model.rotary_emb._orig_forward = self._model.model.rotary_emb.forward
+            self._model.model.rotary_emb.forward = types.MethodType(long_rope, self._model.model.rotary_emb)
+        elif self._model.config.max_position_embeddings != getattr(
+            self._model.config, "original_max_position_embeddings", self._model.config.max_position_embeddings
+        ):
+            self._model.config.max_position_embeddings = self._model.config.original_max_position_embeddings
```
rkazants: shall we save original value for `max_position_embeddings` and recover it in `__exit__` method?

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -950,6 +947,47 @@ def _from_pretrained(
...
+        past_key_values=None,
+        attention_mask=None,
+        inputs_embeds=None,
+        cache_position=None,
+        position_ids=None,
+        use_cache=True,
+        logits_to_keep=None,
+        **kwargs,
+    ):
+        # Overwritten -- this model may need to switch between short and long rope, invalidating the cache in the
```
rkazants, line 926: Am I correct that we have a problem when we have short and long prompts in consecutive `generate` calls?
We can't re-initialize `inv_freqs` from `long_inv_freqs` to `short_inv_freqs` and vise-versa? How this problem is solved?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1590,13 +1644,47 @@ def _phi_moe_sparse_moe_block_forward(self, hidden_states: torch.Tensor) -> torc
...
-        super().__enter__()
+        # Call OVDecoderModelPatcher.__enter__() directly to skip Phi3ModelPatcher's longrope logic
+        # PhiMoE has a different rotary embedding structure, longrope is not yet supported
+        OVDecoderModelPatcher.__enter__(self)
+
+        if is_transformers_version("<", "4.48.0"):
+            self._model.model._orig_forward = self._model.model.forward
+            self._model.model.forward = types.MethodType(phi3_442_forward, self._model.model)
+
+        # init inv_freq for torchscript tracing for PhiMoE
```
rkazants: strange comment about torchscript tracing. Please revise.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1590,13 +1644,47 @@ def _phi_moe_sparse_moe_block_forward(self, hidden_states: torch.Tensor) -> torc
 
 class PhiMoEModelPatcher(Phi3ModelPatcher):
     def __enter__(self):
-        super().__enter__()
+        # Call OVDecoderModelPatcher.__enter__() directly to skip Phi3ModelPatcher's longrope logic
+        # PhiMoE has a different rotary embedding structure, longrope is not yet supported
```
echarlaix: why do we need to add all this modifications to `PhiMoEModelPatcher`? (if longrope is not yet supported then `self._model.model.rotary_emb` will never be set to "longrope")  If we want to make sure we can raise an error in case it's ever the case

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1493,16 +1493,51 @@ def _phi3_self_attn_sdpa_forward(
     return attn_output, None, past_key_value
 
 
+def select_ext_factor(
+    seq_len: torch.Tensor, max_pos_embeddings: torch.Tensor, short_factor: torch.Tensor, long_factor: torch.Tensor
+):
+    return torch.where(seq_len <= max_pos_embeddings, short_factor, long_factor)
+
+
+def long_rope(self, x, position_ids, seq_len=None):
```
echarlaix: would you mind adding a link to original code (https://github.com/huggingface/transformers/blob/v4.57.1/src/transformers/models/phi3/modeling_phi3.py#L324 ?)

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1493,16 +1493,51 @@ def _phi3_self_attn_sdpa_forward(
...
+    device_type = device_type if isinstance(device_type, str) and device_type != "mps" else "cpu"
+    freqs = (inv_freq_expanded.float() @ position_ids_expanded.float()).transpose(1, 2)
+    emb = torch.cat((freqs, freqs), dim=-1)
+
+    scale = max_position_embeddings / original_max_position_embeddings
+    if scale <= 1.0:
+        scaling_factor = 1.0
+    else:
+        scaling_factor = math.sqrt(1 + math.log(scale) / math.log(original_max_position_embeddings))
+    cos = emb.cos() * scaling_factor
```
echarlaix: here we can't use `self.attention_scaling` ? https://github.com/huggingface/transformers/blob/63fbd50fb4ff7b586ab1b59b67f7464e62f9df69/src/transformers/modeling_rope_utils.py#L519

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1493,16 +1493,51 @@ def _phi3_self_attn_sdpa_forward(
...
+    )
+    inv_freq = select_ext_factor(seq_len, original_max_position_embeddings, self.inv_freq, self.long_inv_freq)
+
+    inv_freq_expanded = inv_freq[None, :, None].float().expand(position_ids.shape[0], -1, 1)
+    position_ids_expanded = position_ids[:, None, :].float()
+
+    # Force float32 since bfloat16 loses precision on long contexts
+    # See https://github.com/huggingface/transformers/pull/29285
+    device_type = x.device.type
+    device_type = device_type if isinstance(device_type, str) and device_type != "mps" else "cpu"
```
echarlaix, line 1527: not used and should we ensure fp32 dtype also ?

echarlaix (commented):
Thanks a lot @helena-intel !!

optimum/exporters/openvino/model_patcher.py
helena-intel: Initially tests failed for phi_moe, see https://github.com/huggingface/optimum-intel/actions/runs/18952102871/job/54119192964 . We should have longrope support for the MoE model too but not in this PR. I would be happy with a simpler solution to not enable longrope for the MoE model (but still have it working as it is now).

optimum/intel/openvino/modeling_decoder.py
AlexanderKalistratov: Wouldn't this: https://github.com/huggingface/optimum-intel/pull/1504/files#r2518867960 solve the issue?

optimum/exporters/openvino/model_patcher.py
rkazants: Yes, it is a good point. @helena-intel, please use it.

optimum/exporters/openvino/model_patcher.py
rkazants: @helena-intel, you actually patch this function `https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_rope_utils.py#L442`
but I don't see that `short_factor` from model config is used in the patch. Please clarify it.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -950,6 +947,47 @@ def _from_pretrained(
...
+        # When the first time input length reached long and short factor switching point, enforce re-compute cache
+        # It will cause downside of slower at this single token position, however, better than current failure.
+        if (
+            past_key_values
+            and self.config.rope_scaling
+            and input_ids.shape[1] >= self.config.original_max_position_embeddings + 1
+        ):
+            past_length = cache_position[0]
+            if past_length <= self.config.original_max_position_embeddings:
+                past_key_values = None
```
rkazants, line 938: please add a link to  `https://github.com/huggingface/transformers/blob/main/src/transformers/models/phi3/modeling_phi3.py#L522` and comment that it is aligned with phi3 for long context.
And add a comment that we reset KV cache and it means that the next step will be prefill for extended (computed so far) tokens.

optimum/exporters/openvino/model_patcher.py
rkazants: @helena-intel, I think we need to re-write this patch more accurately to be aligned with `https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_rope_utils.py#L442` for `longrope`

optimum/exporters/openvino/model_patcher.py
helena-intel: I created a new tiny model with longrope support where we can reproduce the issue and will add a test. https://huggingface.co/optimum-intel-internal-testing/tiny-random-phi-4-mini-instruct

optimum/intel/openvino/modeling_decoder.py
helena-intel: As discussed offline: this is handled in optimum-intel by resetting the kv-cache when the number of input tokens is equal to the long rope boundary (e.g. 4096). This is done the same way in transformers code. Tested that this works as expected in chat context with https://gist.github.com/helena-intel/b55522cda91d9d61a644f153e71f0f98 .

tests/openvino/test_decoder.py
```diff
@@ -769,3 +769,40 @@ def test_load_with_different_dtype(self):
                 torch.allclose(torch.Tensor(ov_logits), ref_logits, atol=5e-3),
                 f"values are not close for {dtype if dtype is not None else 'None'}, max diff = {torch.abs(ov_logits - ref_logits).max()}",
             )
+
+    def test_phi3_longrope_support(self):
```
rkazants, line 786: no need in this new test. Just add your model into `SUPPORTED_ARCHITECTURES` above. All requited testing will be activated. You also need to add model id into `util_tests.py`

helena-intel:
> Why `phi4`? 

I wanted to use a model that is being used by people who reported this issue, and I figured it would be useful to have a phi-4 tiny model too. I can change it if needed. 

> Just name it as tiny-random-phi3

So it should replace the existing model? https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/utils_tests.py#L152

> no need in this new test. Just add your model into SUPPORTED_ARCHITECTURES above. All requited testing will be activated. 

I think it's useful to test both short and long context because it is also relevant to know if short context starts failing. And long context should be tested with prompts above the threshold value so if we rely on existing tests we should always remember that the generic model input needs to exceed the long context threshold. If someone changes the existing "same output as transformers" test, or the tiny model, the test may miss issues.

> Also, please initialize to some small `original_max_position_embeddings` to value I suggested. Just name it as `tiny-random-phi3`, make sure that `rope_type` is `longrope`.

I will look into that. Values probably need to be a bit higher, but can be lower than default. We can't just set the values to 10 and 20, the model is sensitive to parameters and it's easy to get collapsing outputs or differences between PyTorch and OpenVINO.

optimum/exporters/openvino/model_patcher.py
helena-intel: I will fix this in a better way.

optimum/exporters/openvino/model_patcher.py
helena-intel: `short_factor` is in the select_ext_factor function: `return torch.where(seq_len <= max_pos_embeddings, short_factor, long_factor)`

I agree it would be clearer to rewrite - but it is functionally working now. We see the same outputs as transformers, for both short and long context.

helena-intel: Done

optimum/exporters/openvino/model_patcher.py
helena-intel: Done

optimum/exporters/openvino/model_patcher.py
helena-intel: Thanks! Added the autocast line with enabled=False.

optimum/intel/openvino/modeling_decoder.py
helena-intel: Done

optimum/intel/openvino/modeling_decoder.py
helena-intel: Added the link a few lines above. The comment that was there was copied verbatim from the transformers code. I modified the second line a bit to make it clearer (transformers comment references "current failure" but it is not clear what that is).

optimum/exporters/openvino/model_patcher.py
helena-intel: @rkazants I refactored the function and added more comments. I think it is clearer now, please review.

optimum/exporters/openvino/model_patcher.py
helena-intel: Done

optimum/exporters/openvino/model_patcher.py
helena-intel: Done

optimum/exporters/openvino/model_patcher.py
helena-intel: I added a `_disable_longrope` property instead of the previous code.

optimum/exporters/openvino/model_patcher.py
helena-intel: Cleaned up phi3-moe code and deleted this section

tests/openvino/test_decoder.py
```diff
@@ -769,3 +769,40 @@ def test_load_with_different_dtype(self):
...
+            f"Model {model_id} does not have original_max_position_embeddings attribute required for LongRoPE",
+        )
+
+        set_seed(SEED)
+        ov_model = OVModelForCausalLM.from_pretrained(
+            model_id, export=True, ov_config=F32_CONFIG, device=OPENVINO_DEVICE
+        )
+
+        # Creating model inputs with more than original max position embeddings and enough variation for varied output tokens
+        tokens = torch.as_tensor(list(tokenizer.get_vocab().values())[: original_max_pos + 50]).unsqueeze(0)
```
IlyasMoutawwakil: the tokenizer is not really needed here, you can use torch.randint with model.config.vocab_size
also shouldn't we test staring with less than max position embeddings and generating enough to surpass it (to trigger cache re-computation)

tests/openvino/utils_tests.py
```diff
@@ -27,7 +27,8 @@
 
 SEED = 42
 
-F32_CONFIG = {"INFERENCE_PRECISION_HINT": "f32"}
+# With this config, inference runs in f32 and optimizations that may influence accuracy are disabled
+F32_CONFIG = {"EXECUTION_MODE_HINT": "ACCURACY"}
```
IlyasMoutawwakil: do we wanna change it for all models ?

tests/openvino/test_decoder.py
```diff
@@ -769,3 +769,40 @@ def test_load_with_different_dtype(self):
...
+            original_max_pos,
+            f"Model {model_id} does not have original_max_position_embeddings attribute required for LongRoPE",
+        )
+
+        set_seed(SEED)
+        ov_model = OVModelForCausalLM.from_pretrained(
+            model_id, export=True, ov_config=F32_CONFIG, device=OPENVINO_DEVICE
+        )
+
+        # Creating model inputs with more than original max position embeddings and enough variation for varied output tokens
```
rkazants, line 808: please test two cases when input_ids length exceeds threshold and when only max_new_tokens exceeds threshold

tests/openvino/utils_tests.py
rkazants: I think we should not change it globally. Let have it only for phi3 models.

tests/openvino/test_decoder.py
```diff
@@ -769,3 +769,40 @@ def test_load_with_different_dtype(self):
                 torch.allclose(torch.Tensor(ov_logits), ref_logits, atol=5e-3),
                 f"values are not close for {dtype if dtype is not None else 'None'}, max diff = {torch.abs(ov_logits - ref_logits).max()}",
             )
+
+    def test_phi3_longrope_support(self):
+        """Test LongRoPE support for Phi3 with inputs > 4096 tokens."""
+        set_seed(SEED)
+        model_id = "optimum-intel-internal-testing/tiny-random-phi-4-mini-instruct"
```
rkazants: please change model card id. Now it is quite confusing with phi-4 but this is not phi-4

tests/openvino/test_decoder.py
helena-intel: Done

tests/openvino/utils_tests.py
helena-intel: Reverted

tests/openvino/test_decoder.py
helena-intel: Done

tests/openvino/test_decoder.py
helena-intel: Changed test to use randint, and now we test both scenarios, where input tokens exceeds original_max_pos and where generation tokens exceeds it.

rkazants (commented):
@IlyasMoutawwakil, please let me review before merge. Thanks!

AlexanderKalistratov:
@helena-intel @rkazants is there any ETA for this PR?

tests/openvino/test_decoder.py
```diff
@@ -769,3 +771,54 @@ def test_load_with_different_dtype(self):
                 torch.allclose(torch.Tensor(ov_logits), ref_logits, atol=5e-3),
                 f"values are not close for {dtype if dtype is not None else 'None'}, max diff = {torch.abs(ov_logits - ref_logits).max()}",
             )
+
+    def test_phi3_longrope_support(self):
+        """Test LongRoPE support for Phi3 with inputs > 4096 tokens."""
+        if is_transformers_version("<", "4.49"):
+            self.skipTest("Incompatible transformers version: Phi3 longrope requires transformers>=4.49")
+        set_seed(SEED)
+        model_id = "optimum-intel-internal-testing/tiny-random-phi3-longrope"
```
rkazants, line 791: you have not-representable short factor in your tiny model:
{DCF9E543-D98D-404A-846A-F81F93BF542C}

Please regenerate model with different values in short factor. For example, real model has short factor with different values.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1493,41 +1493,69 @@ def _phi3_self_attn_sdpa_forward(
...
+# and https://github.com/huggingface/transformers/blob/v4.57.1/src/transformers/modeling_rope_utils.py#L30
+def _phi3_longrope_forward(self, x, position_ids):
+    """
+    LongRoPE uses different scaling factors (short_factor vs long_factor) depending on whether
+    the sequence length exceeds the original max position embeddings used during pretraining.
+
+    Note: In transformers, the @dynamic_rope_update decorator replaces self.inv_freq before the forward pass.
+    Here we use torch.where to select between inv_freq and long_inv_freq and add the selection logic into the model graph.
+    """
+    seq_len = torch.max(position_ids) + 1
```
rkazants, line 1506: I don't see that `short_factor` is extracted and used anyhow in this patch.
Please check reference impl.: https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_rope_utils.py#L454C5-L454C18

We need to be aligned with HF

helena-intel: short_factor is not used explicitly because inv_freq is used directly. inv_freq is set here https://github.com/huggingface/transformers/blob/v4.55.1/src/transformers/models/phi3/modeling_phi3.py#L313C1-L314C69 . It calls compute_longrope_parameters in https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_rope_utils.py . During model export we will not use inputs exceeding original_max_position_embeddings, so inv_freq will be set based on short factors.

This code with explicit short and long factors is needed for transformers because they use this for inference, but for model export inv_freq will be set with short factor correctly.

tests/openvino/test_decoder.py
helena-intel: Tiny model is based on phi-4-mini-instruct, which has only 1.0 for short factor: https://huggingface.co/microsoft/Phi-4-mini-instruct/blob/main/config.json#L85
I can add another model with different short factors, but I think it's useful to keep phi-4-mini-instruct too.

rkazants: please don't mix up tiny models for phi-3 and phi4. For phi4, it should be separate

optimum/exporters/openvino/model_patcher.py
rkazants: I see that short_factor is used in the code
{1CEF8B08-BF42-4CC4-95BF-FE8C199D1A0E}

Now we are not aligned with this code and can expect additional bugs in the future

helena-intel: We are using this code during model export, as part of model loading code. During model loading, compute_longrope_embeddings is actually called with seq_len=None, so self.inv_freq will be set with short_factor (see your screenshot: if seq_len is None,  ext_factors is set according to short_factor, which is defined earlier in the function from model config, and inv_freq is than computed based on this short_factor. So, self.inv_freq is always set with short factor, self.long_inv_freq with long factor, and then in the _phi3_longrope_forward inv_freq is set to self.long_inv_freq if seq is long and else to self.inv_freq. This is aligned with transformers.

tests/openvino/test_decoder.py
helena-intel: I initially named this model tiny-phi-4 but was asked to rename. phi-4-mini-instruct is phi3 architecture, so it is not that this tiny model is not representative for the phi3 model type. I will add a tiny model that is based on a phi-3- model.

helena-intel: I updated the tiny model and now used a base model that does not just use phi3 architecture as before but that also specifically has "phi-3-" in the model name.

optimum/exporters/openvino/model_patcher.py
rkazants: As I understand, it should not be executed during exporting of model. This code should be executed in run-time for each input. Otherwise, it is strange that this model is exported for some concrete seq_len.

helena-intel: The short_factor is used to compute inv_freq. We compute long_inv_freq in the patcher, but short/default inv_freq is computed correctly during model initialization, so self.inv_freq will already be set correctly. And then during inference,  we do "if seq len > max_pos: use long_inv_freq else: use default inv_freq". 

I replaced self.inv_freq with self.original_inv_freq in forward:
image
This is initialized here https://github.com/huggingface/transformers/blob/v4.55-release/src/transformers/models/phi3/modeling_phi3.py#L315 right after initializing self.inv_freq with short factors. We never update self.inv_freq (we override forward()) so self.original_inv_freq == self.inv_freq and self.original_inv_freq is clearer.

PR #1505, opened 2025-10-31
[OpenVINO] Update InferRequestWrapper to collect samples taking into account state of stateful models

nikita-savelyevv:
# What does this PR do?

### Changes
Add an extra value to input dict when wrapping infer request of a stateful model. This feature is not part of the latest NNCF v.2.19 release and depends on NNCF develop version.

### Reason for changes
Quantization quality of stateful Whisper models is poor because a state of a stateful model must be cleared with the same schedule as it is done during calibration input data collection.

Ticket CVS-172705

## Before submitting

### DIFF
--- a/tests/openvino/test_quantization.py
+++ b/tests/openvino/test_quantization.py
@@ -82,7 +82,7 @@ from optimum.intel.openvino.utils import TemporaryDirectory
 from copy import deepcopy
 
 from optimum.intel.openvino.quantization import InferRequestWrapper, OVCalibrationDatasetBuilder
-from optimum.intel.utils.import_utils import is_openvino_version, is_transformers_version, is_nncf_version
+from optimum.intel.utils.import_utils import is_transformers_version, is_nncf_version
 from utils_tests import (
     MODEL_NAMES,
     get_num_quantized_nodes,
@@ -1010,8 +1010,6 @@ class OVWeightCompressionTest(unittest.TestCase):
         (OVSanaPipeline, "sana", 19, 53),
     ]
 
-    IS_SUPPORT_STATEFUL = is_openvino_version(">=", "2023.3")
-
     DEFAULT_INT4_CONFIG = {"bits": 4, "sym": True, "group_size": 64, "all_layers": True}
 
     def test_filtered_architectures(cls):
@@ -1602,12 +1600,8 @@ class OVPipelineQuantizationTest(unittest.TestCase):
                                 lora_correction=True,
                                 ignored_scope={
                                     "patterns": [
-                                        "__module\\.model\\.layers\\.\\d+\\.(mlp\\.(gate_up_proj|down_proj)|self_attn\\.(qkv_proj|o_proj))"
-                                        + (
-                                            "/aten::mul/Multiply"
-                                            if is_openvino_version("<", "2025.2")
-                                            else "\\.lora_B\\.speech/aten::linear/MatMul"
-                                        ),
+                                        "__module\\.model\\.layers\\.\\d+\\.(mlp\\.(gate_up_proj|down_proj)|self_attn\\."
+                                        "(qkv_proj|o_proj))\\.lora_B\\.speech/aten::linear/MatMul",
                                     ],
                                 },
                             ),
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -31,7 +31,6 @@ from optimum.exporters.tasks import TasksManager
 from optimum.intel.utils.import_utils import (
     is_nncf_available,
     is_openvino_tokenizers_available,
-    is_openvino_version,
     is_transformers_version,
 )
 from optimum.intel.utils.modeling_utils import (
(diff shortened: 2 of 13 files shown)


optimum/intel/openvino/quantization.py
```diff
@@ -165,14 +165,29 @@ def __init__(
...
+        self.stateful = len(request.query_state()) > 0
+        self._reset_state_called = False
 
     def collect_inputs(self, inputs):
+        if self.stateful:
+            if isinstance(inputs, dict) and is_nncf_version(">", "2.19"):
+                from nncf.definitions import NNCF_DATASET_RESET_STATE_KEY
+
+                # To reflect the state resetting during NNCF calibration, we add a special key to the input dict
+                # Shallow copying is done on purpose shallow copy on purpose: we only need to add a key to the top-level dict
```
echarlaix:
```suggestion
                # Shallow copying is done on purpose: we only need to add a key to the top-level dict
```

optimum/intel/openvino/quantization.py
```diff
@@ -165,14 +165,29 @@ def __init__(
         self.apply_caching = apply_caching
         self.inference_result_mock = inference_result_mock
         self.tensor_cache = {}
+        self.stateful = len(request.query_state()) > 0
+        self._reset_state_called = False
 
     def collect_inputs(self, inputs):
+        if self.stateful:
+            if isinstance(inputs, dict) and is_nncf_version(">", "2.19"):
```
echarlaix: this would include any following patch release (if any) shouldn't we have instead ?
```suggestion
            if isinstance(inputs, dict) and is_nncf_version(">=", "2.20"):
```

tests/openvino/test_quantization.py
```diff
@@ -2158,13 +2161,28 @@ def test_calibration_data_uniqueness(self, model_name, apply_caching):
...
         data_hashes_per_key = defaultdict(list)
         data_id_per_key = defaultdict(set)
 
+        # Check that reset state flag is present and correctly set in collected inputs
+        if stateful and is_nncf_version(">", "2.19"):
+            from nncf.definitions import NNCF_DATASET_RESET_STATE_KEY
+
+            # All inputs should have reset state key
+            self.assertTrue(all(NNCF_DATASET_RESET_STATE_KEY in inputs_dict for inputs_dict in calibration_data))
+            # The number of times reset state flag is set to True should be equal to (2 * n_samples), because
```
echarlaix, line 2198: why is the state is reset twice for each generation ?

optimum/intel/openvino/quantization.py
```diff
@@ -165,14 +165,29 @@ def __init__(
...
+                from nncf.definitions import NNCF_DATASET_RESET_STATE_KEY
+
+                # To reflect the state resetting during NNCF calibration, we add a special key to the input dict
+                # Shallow copying is done on purpose shallow copy on purpose: we only need to add a key to the top-level dict
+                inputs = inputs.copy()
+                inputs[NNCF_DATASET_RESET_STATE_KEY] = self._reset_state_called
+            self._reset_state_called = False
```
echarlaix, line 181: would you mind adding a comment to state when `inputs[NNCF_DATASET_RESET_STATE_KEY]` should be set to True or False, I'm not sure to understand as I see that `self._reset_state_called` is set to True when `reset_state` is called + `reset_state` being called when `inputs[NNCF_DATASET_RESET_STATE_KEY] is True` https://github.com/openvinotoolkit/nncf/blob/c7afa57fa6910883d5595db66c5785059eb9cb11/src/nncf/openvino/engine.py#L52

optimum/intel/openvino/quantization.py
nikita-savelyevv: Thanks for noticing! Changed

optimum/intel/openvino/quantization.py
nikita-savelyevv: Good point, updated

tests/openvino/test_quantization.py
nikita-savelyevv: The state is first reset before the language-detection call of Whisper pipeline. Then it is reset before the start of token generation itself.

optimum/intel/openvino/quantization.py
nikita-savelyevv: I've added a comment. Please let me know if it does not clarify things enough.

echarlaix: thanks for adding a comment! my last question is why do we have 2 * n_samples and not n_samples ?

nikita-savelyevv: > thanks for adding a comment!

As always no problem!

> my last question is why do we have 2 * n_samples and not n_samples ?

During `.genereate()` call of Whisper pipeline, the first decoder call is actually a language detection call. Only then decoder starts generating actual text tokens. So for each sample, state is reset the first time before the language detection forward, and the second time before token generation phase.

echarlaix: thanks a lot for the explanation @nikita-savelyevv

PR #1507, opened 2025-10-31
[OpenVINO] Fix preprocess_inputs for Gemma3

rkazants:
# What does this PR do?

Now it provides exactly one bos token in preprocessed input ids. Before fix, it returns input ids with double bos token in the head. It leads to accuracy degradation.

Fixes https://jira.devtools.intel.com/browse/CVS-172722

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -3527,10 +3527,15 @@ class _OVGemma3ForCausalLM(OVModelForVisualCausalLM):
         text_prompt = processor.apply_chat_template(conversation, add_generation_prompt=True, tokenize=False)
 
         # switch off add_bos_token if chat template already includes it
+        orig_add_bos_token = processor.tokenizer.add_bos_token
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3526,6 +3526,10 @@ def preprocess_inputs(
 
         text_prompt = processor.apply_chat_template(conversation, add_generation_prompt=True, tokenize=False)
 
+        # switch off add_bos_token if chat template already includes it
+        if "bos_token" in processor.tokenizer.chat_template:
+            processor.tokenizer.add_bos_token = False
+
         inputs = processor(images=image, text=text_prompt, videos=video, return_tensors="pt")
```
nikita-savelyevv, line 3534: I think we should set `processor.tokenizer.add_bos_token` back to its original value after `processor()` call in order not to modify the provided processor instance. What do you think?

rkazants: done.

PR #1508, opened 2025-11-02
Fix typos 

ml0mbardi:
- Align kwargs mapping to `pixle_attention_mask` so the mask is propagated (previous `pixel_attention_mask` mismatch dropped it as None).

- Fix small typos in comments:
  bandwith (floa32) → bandwidth (float32)
  overide  → override
  appling  → applying 
  Recursivley → Recursively

### DIFF
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -916,7 +916,7 @@ class OVModelForVisualCausalLM(OVBaseModel, GenerationMixin):
                 "image_grid_thw": kwargs.get("image_grid_thw"),
                 "video_grid_thw": kwargs.get("video_grid_thw"),
                 "token_type_ids": kwargs.get("token_type_ids"),
-                "pixle_attention_mask": kwargs.get("pixle_attention_mask"),
+                "pixel_attention_mask": kwargs.get("pixel_attention_mask"),
                 "image_attention_mask": kwargs.get("image_attention_mask"),
                 "input_audio_embeds": kwargs.get("input_audio_embeds", kwargs.get("audio_input_features")),
                 "audio_embed_sizes": kwargs.get("audio_embed_sizes"),


optimum/intel/openvino/quantization.py
```diff
@@ -1449,6 +1449,121 @@ def _quantize_ovbasemodel(
...
             ov_config.save_pretrained(save_directory)
 
+    def _quantize_torchmodel(
+        self,
+        ov_config: OVConfig,
+        save_directory: Union[str, Path],
+        calibration_datasets: Optional[OVCalibrationDataset] = None,
+        file_name: Optional[str] = None,
+        **kwargs,
+    ):
```
nikita-savelyevv: I believe this is a merge artifact because this logic has been removed recently. Could you please go over the changes and check if they really belong to the PR?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -77,6 +77,78 @@ def patch_update_causal_mask(
             inner_model._update_causal_mask = types.MethodType(patch_fn, inner_model)
 
 
+# adopted from
```
rkazants: please remove this big addition. Check below as well

ml0mbardi: @ rkazants
Hi! All cleaned up! Just typo fixes now.

optimum/intel/openvino/quantization.py
ml0mbardi: @ nikita-savelyevv
Hi! Rebased and cleaned up already—merge artifacts are gone now. Thanks!

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -916,7 +916,7 @@ def prepare_inputs_for_generation(
                 "image_grid_thw": kwargs.get("image_grid_thw"),
                 "video_grid_thw": kwargs.get("video_grid_thw"),
                 "token_type_ids": kwargs.get("token_type_ids"),
-                "pixel_attetion_mask": kwargs.get("pixle_attetion_mask"),
+                "pixle_attention_mask": kwargs.get("pixle_attention_mask"),
```
nikita-savelyevv:
```suggestion
                "pixel_attention_mask": kwargs.get("pixel_attention_mask"),
```

PR #1467, opened 2025-10-11
IPEX transformers upgrade to 4.55

kaixuanliu:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1511, opened 2025-11-04
fix can_compile

jiqing-feng:
The `can_compile` function was broken in #1467. Removed `return True` by accident. Fixed in this PR.

### DIFF
--- a/optimum/intel/ipex/modeling_base.py
+++ b/optimum/intel/ipex/modeling_base.py
@@ -70,9 +70,9 @@ elif is_torch_version("<", "2.7"):
     _COMPILE_NOT_READY_MODEL_TYPES = ("llama", "falcon", "gpt2", "qwen2", "mistral")
 else:
     _COMPILE_NOT_READY_MODEL_TYPES = ("mistral",)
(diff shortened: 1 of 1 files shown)


jiqing-feng:
Hi @IlyasMoutawwakil . Please review this PR. Thanks!

optimum/intel/ipex/modeling_base.py
```diff
@@ -71,6 +71,9 @@
 else:
     _COMPILE_NOT_READY_MODEL_TYPES = ("mistral",)
 
+# TODO: Remove it in the next PR.
+_COMPILE_NOT_READY_MODEL_TYPES_WITHOUT_CACHE = ("gpt2", "gpt_bigcode")
```
IlyasMoutawwakil: please explain why these models can't be compiled.

jiqing-feng: It's a regression in torch/ipex 2.8. I have explain it in the code comments. We will fix it in the next PR. Before that, let's merge this PR to fix the general error.

PR #1512, opened 2025-11-06
[OpenVINO] Avoid use of deprecated openvino.runtime

rkazants:
# What does this PR do?

In OpenVINO 2026.0, openvino.runtime will be removed.
Now we are preparing PR with this removal in OpenVINO master targeted to 2026.0 release. To avoid massive failures on openvino-nightly, we also need to remove it here.

## Before submitting


echarlaix (approved):
Thanks @rkazants !

PR #1515, opened 2025-11-10
[OpenVINO] Support LFM2 model  

popovaan:
# What does this PR do?

Note: To convert this model `openvino>=2025.4.0` is required. You can install the nightly build with following command:
```sh
pip install --pre -U openvino --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
```

Conversion cmd-line for `LiquidAI/LFM2-350M`:
```sh
optimum-cli export openvino -m LiquidAI/LFM2-350M LFM2
```

Inference of `LiquidAI/LFM2-350M` using OpenVINO backend:

```python
import torch

model_dir="./LFM2"

# Prepare input

# Run inference

print(output_text)
```

Draft support of Lfm2 model.

## Before submitting
- [*] [N/A] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

### DIFF
--- a/tests/openvino/test_quantization.py
+++ b/tests/openvino/test_quantization.py
@@ -987,11 +987,11 @@ class OVWeightCompressionTest(unittest.TestCase):
 
     def test_filtered_architectures(cls):
         if is_transformers_version("<", "4.49"):
-            expected = {"llama4", "qwen2_5_vl", "lfm2"}
+            expected = {"llama4", "qwen2_5_vl"}
         elif is_transformers_version("<", "4.51"):
-            expected = {"llama4", "lfm2"}
+            expected = {"llama4"}
         elif is_transformers_version("<", "4.52"):
-            expected = {"lfm2"}
+            expected = set()
         else:
             expected = {"llava-qwen2", "phi3_v", "minicpmo"}
 


docs/source/openvino/models.mdx
```diff
@@ -83,6 +83,7 @@ Here is the list of the supported architectures :
 - Llava
 - Llava-Next
 - Llava-Next-Video
+- Lfm2
```
rkazants: need to move up a bit

popovaan: Done.

tests/openvino/test_exporters_cli.py
```diff
@@ -134,6 +136,7 @@ class OVCLIExportTestCase(unittest.TestCase):
         "stable-diffusion-3": 6 if is_tokenizers_version("<", "0.20") or is_openvino_version(">=", "2024.5") else 2,
         "flux": 4 if is_tokenizers_version("<", "0.20") or is_openvino_version(">=", "2024.5") else 0,
         "flux-fill": 4 if is_tokenizers_version("<", "0.20") or is_openvino_version(">=", "2024.5") else 0,
+        "lfm2": 0, # TODO: looks like tokenizers fail to convert
```
apaniukov: I didn't observe any issues with the direct tokenizer conversion and accuracy locally. Could you share the error logs?

popovaan: Here is the error log:
```
panas@IMU-NEX-OVINO-25-WS:~/git/optimum-intel$ convert_tokenizer "LiquidAI/LFM2-350M" --with_detokenizer
Loading Huggingface Tokenizer...
Converting Huggingface Tokenizer to OpenVINO...
  File "/home/panas/venv/new_py3_12/bin/convert_tokenizer", line 8, in <module>
    sys.exit(convert_hf_tokenizer())
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/openvino_tokenizers/cli.py", line 291, in convert_hf_tokenizer
    converted = convert_tokenizer(
                ^^^^^^^^^^^^^^^^^^
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/openvino_tokenizers/convert_tokenizer.py", line 54, in wrapper
    return func(args[0], params)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/openvino_tokenizers/convert_tokenizer.py", line 134, in convert_tokenizer
    ov_tokenizers = convert_fast_tokenizer(tokenizer_object, params)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/openvino_tokenizers/hf_parser.py", line 482, in convert_fast_tokenizer
    return tokenizer_model, pipeline.get_detokenizer_ov_subgraph()
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/openvino_tokenizers/tokenizer_pipeline.py", line 1620, in get_detokenizer_ov_subgraph
    outputs = self.create_decoding_pipeline([token_ids])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/openvino_tokenizers/tokenizer_pipeline.py", line 1607, in create_decoding_pipeline
    pipeline_step = step.get_ov_subgraph(input_nodes)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/openvino_tokenizers/tokenizer_pipeline.py", line 1294, in get_ov_subgraph
    return _get_factory().create("UTF8Validate", input_nodes, {"replace_mode": replace_mode}).outputs()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/openvino_tokenizers/__init__.py", line 128, in new_create
    return old_create(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/openvino/utils/node_factory.py", line 58, in create
    node = self.factory.create(op_type_name, arguments, attributes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
openvino._pyopenvino.GeneralFailure: Check 'node->get_input_element_type(input_index+2) == element::u8' failed at /home/jenkins/agent/workspace/private-ci/ie/build-linux-manylinux2014/b/repos/openvino_tokenizers/src/utils.cpp:33:
FrontEnd API failed with GeneralFailure:
Expected a u8 tensor as the third part of the decomposed string representation, got i32
```
openvino is from latest nightly:

```
openvino==2026.0.0.dev20251114
openvino-tokenizers==2026.0.0.0.dev20251114
```

helena-intel: > Yes, I observed this issue as well. It looks like a GPU plugin specific problem, as CPU works fine with the same inputs.

Inference also does not work well on SPR with BF16, or LNL CPU with FP16 so it's not GPU plugin specific. On CPU with fp16/bf16, model outputs are [0,0,0,...] so immediately collapsing to zero. On GPU the first couple of words in output are good. Could be different issues for both CPU and GPU plugin, but may be worth looking into if this can be addressed in model export?

rkazants: It is out of scope for this PR. CPU team will take a look. My assumption is that linear attention layer is new type of layer for CPU plugin that it is not properly supported. Here we check that FP32 is working that is sufficient to prove that export of this model is correct.
We will file a bug for them.

tests/openvino/test_decoder.py
```diff
@@ -337,7 +341,8 @@ def test_compare_to_transformers(self, model_arch):
         gen_config = GenerationConfig(
             max_new_tokens=30,
             min_new_tokens=30,
-            num_beams=1 if model_arch == "chatglm4" else 2,
+            # LFM2 fails with beam search, issue link: https://github.com/huggingface/transformers/issues/42257
+            num_beams=1 if model_arch == "chatglm4" or model_arch == "lfm2" else 2,
```
rkazants:
```suggestion
            num_beams=1 if model_arch in ["chatglm4", "lfm2"] else 2,
```

tests/openvino/test_export.py
```diff
@@ -90,6 +90,9 @@ class ExportModelTest(unittest.TestCase):
     if is_transformers_version(">=", "4.49"):
         SUPPORTED_ARCHITECTURES.update({"zamba2": OVModelForCausalLM})
 
+    if is_openvino_version(">=", "2025.4.0"):
```
rkazants: also need to check transformer version. 4.45 does not support it

popovaan: Added version check.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6954,3 +6954,220 @@ def __exit__(self, exc_type, exc_value, traceback):
             else:
                 continue
             mamba_layer.forward = mamba_layer._orig_forward
+
+
+def lfm2_short_conv_forward_patched(
```
IlyasMoutawwakil, line 6968: please add a comment here about what's being patched and why the original fails

optimum/exporters/openvino/model_patcher.py
```diff
@@ -6954,3 +6954,220 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+                attention_layer_idx = 0
+                for i in range(config.num_hidden_layers):
+                    if config.layer_types[i] == "full_attention":
+                        self.attention_layer_idx_mapping[i] = attention_layer_idx
+                        attention_layer_idx += 1
+                    elif config.layer_types[i] == "conv":
+                        self.conv_layer_idx_mapping[i] = conv_layer_idx
+                        conv_layer_idx += 1
+
+            def update(
```
IlyasMoutawwakil, line 7052: why redefined ? if i understand correctly the only think we want from this wrapper class to allow instantiation from tensor lists when they are given. please un-bloat this class as much as possible, we want to keep the patching code minimal.

rkazants: that's because we remove unneeded cache items that the source model contains. They don't care of it. In the meantime we don't want to have redundant cache items in IR so we create only required number. That's why we should have re-index cache items and update such method to work with cache items.

tests/openvino/test_exporters_cli.py
```diff
@@ -134,6 +136,7 @@ class OVCLIExportTestCase(unittest.TestCase):
         "stable-diffusion-3": 6 if is_tokenizers_version("<", "0.20") or is_openvino_version(">=", "2024.5") else 2,
         "flux": 4 if is_tokenizers_version("<", "0.20") or is_openvino_version(">=", "2024.5") else 0,
         "flux-fill": 4 if is_tokenizers_version("<", "0.20") or is_openvino_version(">=", "2024.5") else 0,
+        "lfm2": 0,  # Tokenizers fail to convert, TODO: ticket number
```
rkazants: please list CVS  ticket number for tokenizers component

optimum/exporters/openvino/model_patcher.py
rkazants: > that's because we remove unneeded cache items that the source model contains. They don't care of it. In the meantime we don't want to have redundant cache items in IR so we create only required number. That's why we should have re-index cache items and update such method to work with cache items.

@popovaan, please leave a comment about it.

optimum/exporters/openvino/model_patcher.py
popovaan: Done.

optimum/exporters/openvino/model_patcher.py
popovaan: Added comment about it.

tests/openvino/test_exporters_cli.py
popovaan: Done.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -1443,9 +1443,15 @@ def prepare_inputs_for_generation(
...
-                attention_mask = torch.ones_like(input_ids, dtype=torch.int64)
+
+                if self.config.model_type == "lfm2":
+                    # LFM2 requires the attention mask to be the length of the full context
+                    # so default mask from OVModelForCausalLM needs to be used
+                    pass
+                else:
+                    # other models like Mamba typically do not require an attention_mask
+                    # for the decoding step after the first token so use attention mask of ones
+                    attention_mask = torch.ones_like(input_ids, dtype=torch.int64)
```
nikita-savelyevv:
```suggestion
                if self.config.model_type != "lfm2":
                    # LFM2 requires the attention mask to be the length of the full context,
                    # so default mask from OVModelForCausalLM needs to be used.
                    # Other models like Mamba typically do not require an attention_mask
                    # for the decoding step after the first token so use attention mask of ones.
                    attention_mask = torch.ones_like(input_ids, dtype=torch.int64)
```

.github/workflows/test_openvino.yml
```diff
@@ -54,6 +54,8 @@ jobs:
         run: |
           pip install --upgrade pip uv
           uv pip install .[openvino,diffusers,tests]
+          # temporarily added to test on nightly
+          uv pip install --upgrade --pre openvino openvino-tokenizers --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
```
nikita-savelyevv: Please add this temporarily to `test_openvino_slow.yaml` as well

tests/openvino/utils_tests.py
```diff
@@ -339,6 +340,7 @@
         "resampler_model": 6,
     },
     "zamba2": {"model": 44},
+    "lfm2": {"model": 52},
```
nikita-savelyevv, line 343: Could you please also add a test case to TRANSFORMERS_4BIT_CONFIGURATIONS at https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_quantization.py#L558 ? Please use group_size=8 for tiny lfm2 model.

popovaan: Done.

.github/workflows/test_openvino.yml
popovaan: Done.

nikita-savelyevv (approved):
I'm ok to merge once all expected tests are passing

tests/openvino/test_exporters_cli.py
```diff
@@ -103,6 +103,8 @@ class OVCLIExportTestCase(unittest.TestCase):
         ("feature-extraction", "sam"),
         ("text-to-audio", "speecht5"),
         ("zero-shot-image-classification", "clip"),
+        ("text-generation", "lfm2"),
+        ("text-generation-with-past", "lfm2"),
```
rkazants: need to move under required `is_transformers_version`
https://github.com/huggingface/optimum-intel/actions/runs/19502143862/job/55818685084?pr=1515

for 4.45 transformers, the test should be skipped

popovaan: Done.

tests/openvino/test_quantization.py
```diff
@@ -938,6 +938,13 @@ class OVWeightCompressionTest(unittest.TestCase):
                 "resampler_model": {"int8": 6},
             },
         ),
+        (
+            OVModelForCausalLM,
+            "lfm2",
+            False,
+            dict(bits=4, backup_precision="none", group_size=8, ratio=0.5),
+            {"model": {"int4": 12}},
+        ),
```
rkazants: @nikita-savelyevv, propose to add this when we move to 2025.4 release because now PT FE fails with error:

```
While validating node 'util::ScatterNDBase ScatterNDBase_6758359 (opset1::Transpose Transpose_6758338[0]:f32[?,?,?], opset1::Unsqueeze Unsqueeze_6758347[0]:f32[?,1], opset1::ConvertLike ConvertLike_6758358[0]:f32[?,?,?]) -> (dynamic[...])' with friendly_name 'ScatterNDBase_6758359':
Indices element type must be i64 or i32
```

https://github.com/huggingface/optimum-intel/actions/runs/19509191839/job/55843987581?pr=1515

rkazants: now let us remove it

rkazants:
```suggestion
```

rkazants:
@IlyasMoutawwakil, @nikita-savelyevv, propose to merge. All new tests are passing on openvino-nighlty: [25af7a9](https://github.com/huggingface/optimum-intel/pull/1515/commits/25af7a947b5bfb966acdde58c397903212127637)
There are no new regressions.

Best regards,
Roman

PR #1517, opened 2025-11-11
[OpenVINO] Add `cache_position` input inside `prepare_inputs` method for Mamba

nikita-savelyevv:
# What does this PR do?

`cache_position` input is required to run inference. Needed for enabling data-aware quantization of Mamba model https://github.com/openvinotoolkit/nncf/pull/3725 .

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_decoder.py
+++ b/optimum/intel/openvino/modeling_decoder.py
@@ -1305,9 +1305,9 @@ class OVModelWithMambaForCausalLM(OVModelForCausalLM):
         **kwargs,
     ) -> Dict:
         if kwargs.get("past_key_values") is not None:
-            raise ValueError("`past_key_values` input is not supported for OVModelWithMambaForCausalLM")
+            raise ValueError("`past_key_values` input is not supported for `OVModelWithMambaForCausalLM`")
         if kwargs.get("position_ids") is not None:
-            raise ValueError("`position_ids` input is not supported for OVModelWithMambaForCausalLM")
+            raise ValueError("`position_ids` input is not supported for `OVModelWithMambaForCausalLM`")
 
         inputs = {"input_ids": input_ids}
         if "cache_position" in self.input_names:


optimum/intel/openvino/modeling_decoder.py
```diff
@@ -545,6 +545,9 @@ def prepare_inputs(
                 self.next_beam_idx if self.next_beam_idx is not None else np.arange(batch_size, dtype=int)
             )
 
+        if "cache_position" in self.input_names:
```
rkazants: For Mamba models we have concrete class `OVModelWithMambaForCausalLM` where we fill `cache_positions` for prefill stage: https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_decoder.py#L1437

I think you need to use the method of this class. So there is no need in these changes. Please check

nikita-savelyevv: `prepare_inputs_for_generation` you mention is called during `model.generate()` call, while `prepare_inputs()` is called during `model.forward()` call. We rely on the latter for quantization calibration data preparation:
https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/quantization.py#L675 , so it needs to be implemented properly. 

I've noticed there is some other input preparation logic inside `OVModelWithMambaForCausalLM.forward` so I've overriden `prepare_inputs` for it to stay consistent.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -1295,15 +1295,20 @@ def _has_cache_inputs(model: openvino.Model) -> bool:
...
         cache_params=None,
         use_cache: Optional[bool] = None,
         cache_position: Optional[torch.Tensor] = None,
         **kwargs,
-    ):
+    ) -> Dict:
+        if kwargs.get("past_key_values") is not None:
+            raise ValueError("`past_key_values` input is not supported for OVModelWithMambaForCausalLM")
+        if kwargs.get("position_ids") is not None:
+            raise ValueError("`position_ids` input is not supported for OVModelWithMambaForCausalLM")
```
rkazants:
```suggestion
            raise ValueError("`past_key_values` input is not supported for `OVModelWithMambaForCausalLM`")
        if kwargs.get("position_ids") is not None:
            raise ValueError("`position_ids` input is not supported for `OVModelWithMambaForCausalLM`")
```

optimum/intel/openvino/modeling_decoder.py
rkazants: Now it is better. I agree to override `prepare_inputs`.

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -1295,15 +1295,20 @@ def _has_cache_inputs(model: openvino.Model) -> bool:
...
         cache_params=None,
         use_cache: Optional[bool] = None,
         cache_position: Optional[torch.Tensor] = None,
         **kwargs,
-    ):
+    ) -> Dict:
+        if kwargs.get("past_key_values") is not None:
+            raise ValueError("`past_key_values` input is not supported for `OVModelWithMambaForCausalLM`")
+        if kwargs.get("position_ids") is not None:
+            raise ValueError("`position_ids` input is not supported for `OVModelWithMambaForCausalLM`")
```
IlyasMoutawwakil, line 1310: is this really needed ?

nikita-savelyevv: My motivation was to avoid possible confusion if anyone passes these arguments because parent's definition of `prepare_inputs()` has these arguments in its signature.

IlyasMoutawwakil: hmm i'm honestly not sure why this function is public, imo it should be private (it serves one purpose, an internal one)

nikita-savelyevv: It is used externally during quantization https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/quantization.py#L675

IlyasMoutawwakil: that's still internal though 😅.. what i mean is that it shouldn't  be used by a user
the downside of a public method is that we need to maintain its behaviour and only change it through deprecation process (over multiple versions).

Issue #1510, opened 2025-11-03
Potential remote code execution via untrusted quantization_config in OVQuantizer


Vancir:
### Description

The OpenVINO quantization flow in `optimum.intel` accepts a remote `openvino_config.json` that may contain loader kwargs (e.g. `trust_remote_code`) or references to other repos (e.g. a `tokenizer` path). These values are forwarded into `AutoTokenizer.from_pretrained` calls during `OVQuantizer.quantize(...)`. An attacker can publish a benign-looking model repo whose `openvino_config.json` (or `quantization_config`) contains `trust_remote_code: true` and a `tokenizer` value pointing to an attacker-controlled repo. When a victim calls the usual API to load the OpenVINO config and run quantization, the tool imports and executes code from the attacker's repository, resulting in remote code execution.

### Root Cause 

optimum-intel treats remote config fields as trusted and passes them into `AutoTokenizer.from_pretrained` calls without sanitization.

https://github.com/huggingface/optimum-intel/blob/8a6268a81811b14df7edf9bb4450efb80c803868/optimum/intel/openvino/quantization.py#L942-L951

### Impact

This issue turns remote configuration files into attack vectors. Any user who loads a malicious adapter config can have arbitrary code executed on their machine. The compromise is silent, requires no extra user action beyond `from_pretrained`, and is easy to weaponize by publishing a seemingly legitimate config that explicitly sets `trust_remote_code=True` and points to attacker code. Consequences include command execution, credential and data theft, file tampering, and worm infection if environment tokens or write permissions are present. This should be fixed urgently by treating config-supplied kwargs as untrusted: filter or reject sensitive parameters such as `trust_remote_code`. 

### Who can help?

@nikita-savelyevv 

### Reproduction

A malicious remote config can look like ([openvino_config.json](https://huggingface.co/XManFromXlab/optimum-quantization-rce/blob/main/openvino_config.json)):

```json
{
	"quantization_config": {
		"tokenizer": "XManFromXlab/optimum-quantization-rce",
		"trust_remote_code": true,
	}
}
```

When users are attracted to the repo and use peft to load the optimum-intel from remote repo

```python
from optimum.intel import OVConfig, OVModelForCausalLM, OVQuantizer

model_id = "XManFromXlab/optimum-quantization-rce"
ov_config = OVConfig.from_pretrained(model_id, trust_remote_code=False)
model = OVModelForCausalLM.from_pretrained(model_id, trust_remote_code=False)
[...]

PR #1520, opened 2025-11-13
[OpenVINO] Refactor from_pretrained quantization

nikita-savelyevv:
# What does this PR do?

Changes:
- Introduce `OVBaseModel._apply_quantization()` method -- a single end point where quantization is applied during `from_pretrained()` call. Deprecate quantization through `OVBaseModel.load_model()` as it is limited to data-free only quantization.
- Extend `OVBaseModel._prepare_quantization_config()` method to additionally load default quantization config given model id. Before, this logic was applied only for `OVBaseDecoderModel`.
- Deprecate providing `trust_remote_code` to `OVQuantizationConfigBase` instances, e.g. `OVWeightQuantizationConfig`. Instead provide it when initializing `OVQuantizer`. Ideally I would like to deprecate it until the next release due to https://github.com/huggingface/optimum-intel/issues/1510, but that's debatable.

Fixes https://github.com/huggingface/optimum-intel/issues/1510

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -465,6 +465,8 @@ class OVBaseModel(OptimizedModel, OVModelHostMixin):
                 Whether or not to only look at local files (i.e., do not try to download the model).
             load_in_8bit (`bool`, *optional*, defaults to `False`):
                 Whether or not to apply 8-bit weight quantization.
+            trust_remote_code (`bool`, *optional*, defaults to `False`):
+                Whether to trust remote code when loading model tokenizer/processor during quantization.
         """
         model_path = Path(model_id)
         default_file_name = ONNX_WEIGHTS_NAME if from_onnx else OV_XML_FILE_NAME
@@ -600,7 +602,7 @@ class OVBaseModel(OptimizedModel, OVModelHostMixin):
     ):
         if not isinstance(quantization_config, (dict, OVQuantizationConfigBase, type(None))):
             raise TypeError(
-                "Expected `quantization_config` to be either a dictionary, instance of OVQuantizationConfigBase object "
+                "Expected `quantization_config` to be either a dictionary, instance of `OVQuantizationConfigBase` object "
                 f"or None, got {type(quantization_config)}."
             )
 
@@ -644,6 +646,22 @@ class OVBaseModel(OptimizedModel, OVModelHostMixin):
         model_name_or_path: Optional[str] = None,
         trust_remote_code: Optional[bool] = False,
     ):
+        """
+        Apply quantization to the model.
+        Arguments:
+            model (`OVBaseModel`):
+                The model to quantize.
+            quantization_config (`OVQuantizationConfigBase`):
+                The quantization config to use.
+            compile_only (`bool`):
+                Whether the model was initialized with `compile_only=True`.
+            compile_model (`bool`):
+                Whether to compile the model after quantization.
+            model_name_or_path (`str`, *optional*):
+                The model name or path used to set tokenizer/processor in quantization config if they are not set.
+            trust_remote_code (`bool`, *optional*, defaults to `False`):
+                Whether to trust remote code when loading model tokenizer/processor during quantization.
+        """
         if not is_nncf_available():
             raise ImportError("Quantization of the weights requires nncf, please install it with `pip install nncf`")
 
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -507,8 +507,8 @@ class OVModelForVisualCausalLM(OVBaseModel, GenerationMixin):
                 openvino_decoder_with_past_model.xml, allowing to load the decoder model with a different name.
             local_files_only(`bool`, *optional*, defaults to `False`):
                 Whether or not to only look at local files (i.e., do not try to download the model).
-            trust_remote_code(`bool`, *optional*, defaults to `False`):
-                Whether or not to trust the remote code when loading a model from the Hub.
+            trust_remote_code (`bool`, *optional*, defaults to `False`):
+                Whether to trust remote code when loading model tokenizer/processor during quantization.
         """
         if use_auth_token is not None:
             warnings.warn(
--- a/optimum/intel/openvino/modeling_sam.py
+++ b/optimum/intel/openvino/modeling_sam.py
@@ -207,6 +207,8 @@ class OVSamModel(OVBaseModel):
                 Whether or not to apply 8-bit weight quantization.
             quantization_config(`Union[OVQuantizationConfigBase, Dict]`, *optional*, defaults to `None`):
                 Quantization configuration to apply to the model.
+            trust_remote_code (`bool`, *optional*, defaults to `False`):
+                Whether to trust remote code when loading model tokenizer/processor during quantization.
         """
         if use_auth_token is not None:
             warnings.warn(


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -555,40 +548,39 @@ def _from_pretrained(
...
+
+        if quantization_config:
+            if quantization_config.dataset is not None and ov_pipeline_class.export_feature != "text-to-image":
+                raise NotImplementedError(
+                    f"Data-aware quantization is not supported for {cls.__name__} with "
+                    f"{ov_pipeline_class.export_feature} task."
+                )
+
+            cls._apply_quantization(
+                ov_pipeline, quantization_config, compile_only, compile_model, trust_remote_code=trust_remote_code
```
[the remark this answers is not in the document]
nikita-savelyevv: This is an optional argument, not required.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -593,29 +594,26 @@ def _from_pretrained(
...
         )
 
-        if to_quantize:
-            from optimum.intel.openvino.quantization import OVQuantizer
-
+        if quantization_config:
             quantization_config_copy = copy.deepcopy(quantization_config)
-            quantization_config_copy.tokenizer = str(quantization_config.tokenizer or model_id)
             potential_processor_id = config.mm_vision_tower if isinstance(model, _OVNanoLlavaForCausalLM) else model_id
             quantization_config_copy.processor = str(quantization_config.processor or potential_processor_id)
```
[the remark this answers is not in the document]
nikita-savelyevv: This is fine, it is set inside `OVBaseModel._apply_quantization()`

optimum/intel/openvino/modeling_base.py
```diff
@@ -489,22 +493,21 @@ def _from_pretrained(
...
             )
 
+        quantization_config = cls._prepare_quantization_config(model_id, quantization_config, load_in_8bit)
+        compile_model = kwargs.pop("compile", True)
         model = cls(
             ov_model,
             config=config,
             model_save_dir=model_cache_path.parent,
             quantization_config=quantization_config,
+            compile=compile_model and not quantization_config,
```
IlyasMoutawwakil, line 505: why `and not quantization_config` ?

IlyasMoutawwakil: oh I see it's differed

nikita-savelyevv: This is to avoid one extra unnecessary model compilation

optimum/intel/openvino/configuration.py
```diff
@@ -537,7 +537,7 @@ def __init__(
         dataset: Optional[Union[str, List[str]]] = None,
         tokenizer: Optional[str] = None,
         processor: Optional[str] = None,
-        trust_remote_code: Optional[bool] = False,
+        trust_remote_code: Optional[bool] = None,
```
rkazants, line 540: why is it changed to `None`?

optimum/intel/openvino/modeling_base.py
```diff
@@ -489,22 +493,21 @@ def _from_pretrained(
...
-            from optimum.intel import OVQuantizer
-
-            quantizer = OVQuantizer(model)
-            quantization_config_copy = copy.deepcopy(quantization_config)
-            quantization_config_copy.tokenizer = str(quantization_config.tokenizer or model_id)
-            quantization_config_copy.processor = str(quantization_config.processor or model_id)
-            quantizer.quantize(ov_config=OVConfig(quantization_config=quantization_config_copy))
+        if quantization_config:
+            cls._apply_quantization(
+                model, quantization_config, compile_only, compile_model, model_id, trust_remote_code
```
rkazants, line 511: if the model is already compiled, how is it quantized?

optimum/intel/openvino/modeling_base.py
```diff
@@ -591,16 +594,78 @@ def from_pretrained(
...
     @staticmethod
     def _prepare_quantization_config(
-        quantization_config: Optional[Union[OVWeightQuantizationConfig, Dict]] = None, load_in_8bit: bool = False
+        model_name_or_path: str,
+        quantization_config: Optional[Union[OVQuantizationConfigBase, Dict]] = None,
+        load_in_8bit: bool = False,
     ):
-        # Give default quantization config if not provided and load_in_8bit=True
+        if not isinstance(quantization_config, (dict, OVQuantizationConfigBase, type(None))):
+            raise TypeError(
```
rkazants, line 604: I think this exception is not reachable because  we have `Optional[Union[OVQuantizationConfigBase, Dict]] = None,` in method signature

optimum/intel/openvino/modeling_base.py
```diff
@@ -591,16 +594,78 @@ def from_pretrained(
...
+                    "recommended quantization config will be used."
+                )
+            default_config = get_default_quantization_config(model_name_or_path, weight_format="int4")
+            quantization_config = _quantization_config_from_dict(default_config or _DEFAULT_4BIT_WQ_CONFIG)
+        else:
             quantization_config = _quantization_config_from_dict(quantization_config)
+            # Notify user if 4-bit quantization is requested and there is a recommended config for the model
+            if isinstance(quantization_config, OVWeightQuantizationConfig) and quantization_config.bits == 4:
+                default_config = get_default_quantization_config(model_name_or_path, weight_format="int4")
+                if default_config is not None:
```
rkazants, line 633: can it be `None`? If yes, shall we notify user that it is not possible to apply 4-bit quantization?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -471,6 +471,7 @@ def _from_pretrained(
         local_files_only: bool = False,
         load_in_8bit: bool = False,
         quantization_config: Union[OVWeightQuantizationConfig, Dict] = None,
+        trust_remote_code: bool = False,
```
rkazants, line 474: parameter is not documented below.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -593,29 +594,26 @@ def _from_pretrained(
...
-            from optimum.intel.openvino.quantization import OVQuantizer
-
+        if quantization_config:
             quantization_config_copy = copy.deepcopy(quantization_config)
-            quantization_config_copy.tokenizer = str(quantization_config.tokenizer or model_id)
             potential_processor_id = config.mm_vision_tower if isinstance(model, _OVNanoLlavaForCausalLM) else model_id
             quantization_config_copy.processor = str(quantization_config.processor or potential_processor_id)
-            OVQuantizer(model).quantize(ov_config=OVConfig(quantization_config=quantization_config_copy))
+            cls._apply_quantization(
+                model, quantization_config_copy, compile_only, compile_model, model_id, trust_remote_code
```
rkazants, line 617: please clarify why do we need `trust_remote_code` for quantization. A model is already loaded and converted by this time.

optimum/intel/openvino/configuration.py
nikita-savelyevv: This is because `trust_remote_code` argument is being deprecated now. Please see a warning inside `OVQuantizationConfigBase`'s constructor. For the reasons for deprecation please take a look at https://github.com/huggingface/optimum-intel/issues/1510.

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: If quantization is to be applied, at this point the model will not be compiled because above we set `compile=compile_model and not quantization_config`.

Nevertheless, it still can be compiled, because compilation does not contradict quantization. We can quantize an already compiled model without errors, it is just that it will require and additional re-compilation after the quantization. That's why to avoid an unnecessary compilation of the original non-quantized model, `compile=compile_model and not quantization_config` was added.

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: Ideally, yes. But type hints are purely annotation-only and will not result in an exception if a different type of object is provided. This exception is just to be safer from an unintended usage, mostly by developers.

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: > can it be None?

Yes, it can. `get_default_quantization_config` can return None, for example if we don't have a default config for the given model_id/quantization type. 

> If yes, shall we notify user that it is not possible to apply 4-bit quantization?

It is possible to apply, but the default int4 config (`_DEFAULT_4BIT_WQ_CONFIG`) will be applied since we don't have a custom one for this model. Please note this line above:
```python
quantization_config = _quantization_config_from_dict(default_config or _DEFAULT_4BIT_WQ_CONFIG)
```

optimum/intel/openvino/modeling_visual_language.py
nikita-savelyevv: Thanks, added.

optimum/intel/openvino/modeling_visual_language.py
nikita-savelyevv: We need it in some cases to prepare quantization calibration data. Namely, to initialize model tokenizer or processor to prepare inputs for the model to run on. For example please take a look at https://github.com/huggingface/optimum-intel/blob/499e6e1fcbcb004fcb65d173826cad98301ed940/optimum/intel/openvino/quantization.py#L697

optimum/intel/openvino/configuration.py
```diff
@@ -553,16 +553,21 @@ def __init__(
...
-                Allows to use custom code for the modeling hosted in the model repository. This option should only be
-                set for repositories you trust and in which you have read the code, as it will execute on your local
-                machine arbitrary code present in the model repository.
+            trust_remote_code (`bool`, defaults to `None`):
+                This is a deprecated argument and will be removed in optimum-intel v1.27.0. Please provide
+                `trust_remote_code` parameter to `PreTrainedModel.from_pretrained` or `OVQuantizer.__init__` methods instead.
         """
+        if trust_remote_code is not None:
+            logger.warning(
+                "`trust_remote_code` parameter in OVQuantizationConfigBase is deprecated and will be removed in "
```
rkazants:
```suggestion
                "`trust_remote_code` parameter in `OVQuantizationConfigBase` is deprecated and will be removed in "
```

optimum/intel/openvino/modeling_base.py
```diff
@@ -310,6 +312,12 @@ def load_model(
                 Quantization config to apply after model is loaded.
         """
 
+        if quantization_config is not None:
+            logger.warning(
+                "Providing `quantization_config` to `load_model` is deprecated and will be removed in v1.28.0. "
```
rkazants, line 317: will it be removed in 1.27 or 1.28? Please check. I may be wrong here.

nikita-savelyevv: My plan was to deprecate `trust_remote_code` from `OVQuantizationConfigBase` in 1.27 and `quantization_config` from `load_model` in 1.28, because the first one is urgent.

optimum/intel/openvino/configuration.py
nikita-savelyevv: Updated

optimum/intel/openvino/modeling_base.py
```diff
@@ -591,16 +594,78 @@ def from_pretrained(
...
     def _prepare_quantization_config(
-        quantization_config: Optional[Union[OVWeightQuantizationConfig, Dict]] = None, load_in_8bit: bool = False
+        model_name_or_path: str,
+        quantization_config: Optional[Union[OVQuantizationConfigBase, Dict]] = None,
+        load_in_8bit: bool = False,
     ):
-        # Give default quantization config if not provided and load_in_8bit=True
+        if not isinstance(quantization_config, (dict, OVQuantizationConfigBase, type(None))):
+            raise TypeError(
+                "Expected `quantization_config` to be either a dictionary, instance of OVQuantizationConfigBase object "
```
rkazants:
```suggestion
                "Expected `quantization_config` to be either a dictionary, instance of `OVQuantizationConfigBase` object "
```

optimum/intel/openvino/modeling_base.py
```diff
@@ -591,16 +594,78 @@ def from_pretrained(
...
+            return OVWeightQuantizationConfig(bits=8)
+
+        # Nothing to do if quantization_config is already an OVQuantizationConfigBase instance or None
+        if isinstance(quantization_config, (OVQuantizationConfigBase, type(None))):
+            return quantization_config
+
+        # Quantization config is a dictionary
+        if quantization_config == {"bits": 4}:
+            # If config is given as {"bits": 4}, use the default 4-bit quantization config
+            if model_name_or_path in ["openai/gpt-oss-20b", "openai/gpt-oss-120b"]:
```
rkazants, line 620: may be better to use `model_type` instead of model name?

optimum/intel/openvino/modeling_base.py
```diff
@@ -427,6 +435,7 @@ def _from_pretrained(
         local_files_only: bool = False,
         load_in_8bit: bool = False,
         quantization_config: Union[OVWeightQuantizationConfig, Dict] = None,
+        trust_remote_code: bool = False,
```
rkazants, line 438: We need documentation for this parameter below . Please apply this to other places where you add `trust_remote_code`

optimum/intel/openvino/quantization.py
```diff
@@ -1135,17 +1159,21 @@ class OVQuantizer(OptimumQuantizer):
...
         """
         Args:
-            model (`OVModel`):
-                The [OVModel](https://huggingface.co/docs/optimum-intel/en/openvino/reference) to quantize.
+            model (`OVBaseModel`):
+                The [OVBaseModel](https://huggingface.co/docs/optimum-intel/en/openvino/reference) to quantize.
             seed (`int`, defaults to 42):
                 The random seed to use when shuffling the calibration dataset.
+            trust_remote_code (`bool`, *optional*, defaults to `False`):
+                Whether to trust remote code when loading model tokenizer/processor.
```
rkazants, line 1172:
```suggestion
                Whether or not to trust the remote code when loading a model from the Hub.
```
Shall we align different descriptions to the single one?

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: This is done the way it is because for these specific model ids we have a workaround int4 configs set up https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py#L318

optimum/intel/openvino/quantization.py
nikita-savelyevv: In the case of the OVQuantizer we won't actually use the `trust_remote_code` flag to load a model from the Hub, so such docstring will be misleading in my opinion.

nikita-savelyevv: I'll update the description to mention loading of model tokenizer/processor when this flag will actually be used for this.

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: Added

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: Thanks. I've now added a docstring for this argument wherever there is a docstring for the corresponding method at all. Adding complete docstrings where there are missing I believe should be done separately in order not to spawn extra diff.

PR #1521, opened 2025-11-13
[OpenVINO] Fix jais-13b int4 config

nikita-savelyevv:
# What does this PR do?

Avoid failures during import when nncf is not installed.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -354,16 +354,18 @@ _DEFAULT_4BIT_WQ_CONFIGS = {
 
 if is_nncf_available():
     # TODO: Remove after update to NNCF 2.19 because `group_size_fallback` argument will be added to OVWeightQuantizationConfig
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/configuration.py
```diff
@@ -350,17 +350,24 @@ class OVQuantizationMethod(str, Enum):
...
-        "sym": False,
-        "group_size": 128,
-        "ratio": 1.0,
-        "advanced_parameters": nncf.AdvancedCompressionParameters(
-            group_size_fallback_mode=nncf.GroupSizeFallbackMode.ADJUST,
-        ),
-    },
 }
 
+if is_nncf_available():
```
rkazants, line 355: shall we check NNCF version? When nncf.GroupSizeFallbackMode.ADJUST is introduced?

nikita-savelyevv: It was introduced in NNCF 2.18, this is also the minimal requirement for NNCF https://github.com/huggingface/optimum-intel/blob/main/setup.py#L67

PR #1522, opened 2025-11-14
[OpenVINO] NNCF 2.19 update

nikita-savelyevv:
# What does this PR do?

- Update minimal required NNCF version to 2.19.
- Add `group_size_fallback` argument for weight only quantization. It defines the behavior when the specified group size is not compatible with the weight shape. Possible values:
  - "error": raises an error if the group size is not compatible with the weight shape (default);
  - "ignore": skips quantization for the layers where the group size is not compatible with the weight shape;
  - "adjust": automatically adjusts the group size to the maximum compatible value for each weight tensor, if there is no valid value greater or equal to 32, then the node is quantized to the backup precision which is int8_asym by default.
- Remove some conditions on the previous NNCF version.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/__init__.py
+++ b/optimum/intel/openvino/__init__.py
@@ -18,6 +18,7 @@ import warnings
 from ..utils.import_utils import (
     is_diffusers_available,
     is_nncf_available,
+    is_nncf_version,
     is_sentence_transformers_available,
 )
 from .utils import (
@@ -46,6 +47,12 @@ from .configuration import (
 if is_nncf_available():
     logging.disable(logging.INFO)
     import nncf
+
+    if is_nncf_version("<", "2.19"):
+        raise ImportError(
+            "NNCF version 2.19 or higher is required to use NNCF-based quantization. "
+            f"Please upgrade your NNCF installation. The current version of NNCF is {nncf.__version__}."
+        )
 
     logging.disable(logging.NOTSET)
 


docs/source/openvino/export.mdx
```diff
@@ -101,6 +101,14 @@ Optional arguments:
   --group-size GROUP_SIZE
                         The group size to use for quantization. Recommended value is 128 and -1 uses per-column
                         quantization.
+  --group-size-fallback {error,ignore,adjust}
+                        Specifies how to handle operations that do not support the given group size. Possible values are:
+                        `error`: raise an error if the given group size is not supported by a node, this is the default
+                        behavior;
+                        `ignore`: skip nodes that cannot be compressed with the given group size;
+                        `adjust`: adjust the group size to the maximum supported value for each problematic node, if
+                        there is no valid value greater or equal to 32, then the node is quantized to the backup
```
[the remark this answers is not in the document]
nikita-savelyevv: Done

optimum/commands/export/openvino.py
```diff
@@ -141,6 +141,19 @@ def parse_args_openvino(parser: "ArgumentParser"):
...
+        "--group-size-fallback",
+        type=str,
+        choices=["error", "ignore", "adjust"],
+        default=None,
+        help=(
+            "Specifies how to handle operations that do not support the given group size. Possible values are: "
+            "`error`: raise an error if the given group size is not supported by a node, this is the default behavior; "
+            "`ignore`: skip nodes that cannot be compressed with the given group size; "
+            "`adjust`: adjust the group size to the maximum supported value for each problematic node, if there is no "
+            "valid value greater or equal to 32, then the node is quantized to the backup precision which is int8_asym by default. "
```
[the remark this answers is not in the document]

optimum/intel/openvino/configuration.py
```diff
@@ -914,6 +910,13 @@ def post_init(self):
         if self.gptq and self.lora_correction:
             raise ValueError("The GPTQ and LoRA Correction algorithms can't be applied simultaneously")
 
+        valid_group_size_fallback_values = [e.value for e in nncf.GroupSizeFallbackMode]
```
echarlaix, line 913: wondering if we should add a check somewhere (like in https://github.com/huggingface/optimum-intel/blob/75d6b7d3bc9544487e2111a610b59f8d62e0ef89/optimum/intel/openvino/__init__.py#L46) to ensure that the nncf version is compatible with our integration (in our case >= v2.19 I imagine ?)

tests/openvino/test_quantization.py
```diff
@@ -938,6 +935,35 @@ class OVWeightCompressionTest(unittest.TestCase):
...
+            dict(
+                bits=4,
+                sym=True,
+                group_size_fallback="adjust",
+            ),
+            {"model": {"int8": 28, "int4": 2}},
+        ),
+        (
+            OVModelForCausalLM,
+            "llama_awq",
```
echarlaix: why not instead 
```suggestion
            "llama",
```

optimum/intel/openvino/configuration.py
nikita-savelyevv: That's a good point! Added

tests/openvino/test_quantization.py
nikita-savelyevv: No particular reason, changed

optimum/intel/openvino/configuration.py
echarlaix: thanks!

PR #1100, opened 2025-01-07
Fp8 implementation

KodiaqQ:
# What does this PR do?

- Introduce new options for the `--quant-mode` parameter: `f8e4m3`, `f8e5m2`.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

PR #1524, opened 2025-11-18
[NNCF] FP8/FP4 support

daniil-lyakhov:
# What does this PR do?

[NNCF] FP8/FP4 support

## Before submitting

### DIFF
--- a/docs/source/openvino/export.mdx
+++ b/docs/source/openvino/export.mdx
@@ -31,7 +31,7 @@ Check out the help for more options:
 
 ```text
 usage: optimum-cli export openvino [-h] -m MODEL [--task TASK] [--framework {pt}] [--trust-remote-code]
-                                   [--weight-format {fp32,fp16,int8,int4,mxfp4,nf4,cb4}]
+                                   [--weight-format {fp32,fp16,int8,int4,mxfp4,nf4,fp4,fp8_e4m3,cb4}]
                                    [--quant-mode {int8,f8e4m3,f8e5m2,cb4_f8e4m3,int4_f8e4m3,int4_f8e5m2}]
                                    [--library {transformers,diffusers,timm,sentence_transformers,open_clip}]
                                    [--cache_dir CACHE_DIR] [--pad-token-id PAD_TOKEN_ID] [--ratio RATIO] [--sym]
@@ -66,7 +66,7 @@ Optional arguments:
   --trust-remote-code   Allows to use custom code for the modeling hosted in the model repository. This option should
                         only be set for repositories you trust and in which you have read the code, as it will execute
                         on your local machine arbitrary code present in the model repository.
-  --weight-format {fp32,fp16,int8,int4,mxfp4,nf4,cb4}
+  --weight-format {fp32,fp16,int8,int4,mxfp4,fp4,fp8_e4m3,nf4,cb4}
                         The weight format of the exported model. Option 'cb4' represents a codebook with 16
                         fixed fp8 values in E4M3 format.
   --quant-mode {int8,f8e4m3,f8e5m2,cb4_f8e4m3,int4_f8e4m3,int4_f8e5m2}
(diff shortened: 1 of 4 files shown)


docs/source/openvino/export.mdx
```diff
@@ -31,7 +31,7 @@ Check out the help for more options:
 
 ```text
 usage: optimum-cli export openvino [-h] -m MODEL [--task TASK] [--framework {pt}] [--trust-remote-code]
-                                   [--weight-format {fp32,fp16,int8,int4,mxfp4,nf4,cb4}]
+                                   [--weight-format {fp32,fp16,int8,int4,mxfp4,nf4,fp4,fp8_e4m3,cb4}]
```
nikita-savelyevv, line 34: In my opinion we should rename `fp8_e4m3` to `f8e4m3` to keep being aligned with `f8e4m3` option for `--quant-mode`.

cc @ljaljushkin

tests/openvino/test_exporters_cli.py
```diff
@@ -494,6 +494,18 @@ class OVCLIExportTestCase(unittest.TestCase):
             "mxfp4",
             {"model": {"int8": 4, "f4e2m1": 72, "f8e8m0": 72}},
         ),
+        (
+            "text-generation-with-past",
+            "opt125m",
```
nikita-savelyevv, line 531: Could you please use some other model, e.g. `llama`? `opt125m` is too large and in the future we'd like to replace it with a  different one. Perhaps, group size will need to reduced.

docs/source/openvino/export.mdx
ljaljushkin: to be honest, don't like that optimum and nncf names are different and not aligned. 
not clear why quant-mode in optimum is f8e4m3, but nncf has `QuantizationMode.FP8_E4M3`

MXFP4, MXFP8 and NVFP4 are established names and don't follow `mxf4_e2m1` and `nvf4_e2m1` format convention.
fp32 and fp16 are also not f32 and f16. not sure it was a good choice for `--quant-mode`. Can we reconsider it in optimum? will it affect someone?

nikita-savelyevv: Unfortunately, I don't remember the exact reasons why these data types were introduced with the names different from the ones in NNCF, because it was done by Nikita M. (https://github.com/huggingface/optimum-intel/pull/1100, ticket 160144). It was approved by Alexander K. back then so I believe there was at least some reasoning from his side regarding this.

I personally don't see much difference between the optimum-intel and NNCF names as long they are consistent within a single repo. The only thing is that if we decide to rename these data types in optimum-intel, the transition will be a bit painful, because we will have to keep both names for a couple of releases in order to properly deprecate the old ones.

nikita-savelyevv: @helena-intel could you please provide your feedback on this? The proposal is to rename some low precision data type names in optimum-intel to be consistent with their corresponding names in NNCF, for example `f8e4m3` -> `fp8_e4m3`.

helena-intel: I strongly prefer having the same names, and using standard established names. One reason why this is useful: people read NNCF documentation to learn more about different quantization options, with the notion that the examples use NNCF, but the concepts apply to optimum-intel too. Having different names is then unexpected, also if you then go search for examples/source code for what you just read. Not terrible, but not great either. I think now is a better time to change too, since quant-mode is not yet very widely used, but it will probably become more important.

ljaljushkin: at the same time we already have not matching types like int8/int4 and --sym in optimum, and int4_asym, int4_sym,... in nncf. 
since int4 is reserved for default parameters: https://huggingface.co/docs/optimum-intel/en/openvino/export#default-quantization-configs

@nikita-savelyevv  propose keep as is and unblock long awaited fp4 option for NPU.

nikita-savelyevv: @ljaljushkin just to clarify, by "as is" you mean `fp8_e4m3`?

ljaljushkin: yes

tests/openvino/test_exporters_cli.py
daniil-lyakhov: Done

ljaljushkin (approved):

PR #1525, opened 2025-11-19
[OpenVINO] Refactor CLI quantization

nikita-savelyevv:
# What does this PR do?

### Changes

This PR refactors the way how quantization is applied during model export via CLI.
1. Quantization application logic is moved from `optimum/commands/export/openvino.py` to `optimum/exporters/openvino/__main__.py` represented by `_main_quantize` function. When quantization config is explicitly provided (i.e. `--weight-format` or `--quant-mode` options) the export flow now looks like this:
    1. Original floating point model is exported to a temporary directory. This is to avoid confusion in the case when quantization fails and an intermediate floating point model ends up at the target location.
    2. A model is loaded from the temp. dir. using `OVModelFor*` class
    3. Quantization is run using `model._apply_quantization()` -- the entry point introduced in #1520
    4. Quantized model is saved to another temporary directory. This is to avoid loading and saving to the same location.
    5. The loaded model object is unloaded and quantized model's files are moved to the floating point model's temporary directory.
    6. Final model's files are moved to the target `output` directory.
3. There are also additional changes to `OVBaseModel`. `OVBaseModel._prepare_quantization_config()` is decoupled into 
    1. `_resolve_default_quantization_config()` --  matches default int4 config based on model id
    2. `_quantization_config_from_dict()` -- called inside `_apply_quantization()`
    3.  `_preprocess_quantization_config()` -- applies model-specific updates to the config, such as setting tokenizer/processor.

### Reason for changes

1. This logic is required for implementing default ignored scope (ticket 175336) and default dynamic quantization group size (ticket 176390) matching based on model id. 
Since such matching is needed to be done through export via both CLI and API, for both data-free and data-aware quantization and for every low-precision data type, some unification was needed. The problem with the current CLI approach is that data-free and data-aware quantization paths are different: the data-free one is applied at `__main__.py::main_export()` while data-aware quantization is applied directly at `openvino.py::run()` through `from_pretrained` call. Because of this, it is complicated to match default ignored scope for data-free case.
For an example of how default ignored scope matching will be implemented based on changes in this PR please take a look [here](https://github.com/nikita-savelyevv/optimum-intel/blob/3a1c6be3e04f5689453fd6e88ca036bf5e7d1d6d/optimum/intel/openvino/modeling_base.py#L692).
3. There is an ask for encapsulating OpenVINO export path through a single `main_export()` call similar to optimum-onnx to allow programmatic usage by third-party components like Olive (ticket 176928). With the changes in this PR, this procedure is clearly separated into `main_export` for exporting a model and `main_quantize` for quantizing the exported model.

### Potential drawbacks
1. It is possible, that peak RAM will increase during model export with explicit data-free weight-only quantization of multimodal pipelines (except VLMs). This is because, before, most of these pipelines were quantized per-model [here](https://github.com/huggingface/optimum-intel/blob/a7011b5ca88a4e8a59eee3a6f3eadbcc7d35f43e/optimum/exporters/openvino/__main__.py#L562), i.e. each quantized model was written to the disk right away. Now in this case all quantized models will need to be gathered in memory in order to be written to the disk. That said, most commonly peak RAM is reached during model export, not quantization. See for example a figure for `stabilityai/stable-diffusion-3.5-large` exported with `--weight-format int8` below.

| Before    | After |
| -------- | ------- |
| sd_3 5_large_main |  sd_3 5_large_branch |

Single-model pipelines are not affected since there is no "accumulation" effect. Among multi-model pipelines, VLMs are not affected because they are quantized the "new" way already at this moment. 

In the future I believe it should be possible to improve this by introducing immediate serialization logic to `OVQuantizer.quantize()` when `save_directory` is provided. This will require however an API for mapping submodel names to their paths similar to `OVBaseModel.ov_models`.

2. The same behavior affects the peak free disk size needed during export. Because of the fact that we save to a temporary directory first the max disk size footprint is `original_model_size + quantized_model_size`, while before for data-free weight-only quantization it was `original_model_size + largest_quantized_submodel_size`.

The immediate serialization logic suggested above will help with this too, but again it requires some additional work. In general, I don't see this as a major blocker considering the fact that both drawbacks already hold for data-aware quantization as of this moment.

#### Update
I was actually able to implement the aforementioned immediate serialization logic and it demonstrates the same memory profile as "Before" case on the figure above. We can proceed with implementation of this approach after this PR is merged. Diff: https://github.com/nikita-savelyevv/optimum-intel/commit/9c260c9f8a7047c8078ffec8c60c0f7b0f38d28f.

### Related tickets

175336, 176390, 176928

## Before submitting

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -340,7 +340,7 @@ class OVExportCommand(BaseOptimumCLICommand):
         return parse_args_openvino(parser)
 
     def run(self):
-        from ...exporters.openvino.__main__ import _merge_move, main_export, main_quantize
+        from ...exporters.openvino.__main__ import _main_quantize, _merge_move, main_export
         from ...intel.openvino.configuration import (
             _DEFAULT_4BIT_WQ_CONFIG,
             OVConfig,
@@ -483,7 +483,7 @@ class OVExportCommand(BaseOptimumCLICommand):
                 # **input_shapes,
             )
             if apply_main_quantize:
-                main_quantize(
+                _main_quantize(
                     model_name_or_path=self.args.model,
                     task=self.args.task,
                     library_name=library_name,
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -554,7 +554,7 @@ def main_export(
                 AutoBitLinear.load_hook = orig_load_hook
 
 
-def main_quantize(
+def _main_quantize(
     model_name_or_path: str,
     task: str,
     library_name: str,


optimum/exporters/openvino/__main__.py
```diff
@@ -200,6 +201,9 @@ def main_export(
     """
     from optimum.exporters.openvino.convert import export_from_model
 
+    from ...intel.openvino.configuration import _GPTOSSQuantizationConfig
+    from ...intel.openvino.utils import TemporaryDirectory
```
echarlaix: if possible would replace with absolute imports  
```suggestion
    from optimum.intel.openvino.configuration import _GPTOSSQuantizationConfig
    from optimum.intel.openvino.utils import TemporaryDirectory
```

optimum/exporters/openvino/__main__.py
```diff
@@ -621,3 +599,187 @@ def maybe_convert_tokenizers(library_name: str, output: Path, model=None, prepro
...
+            del submodel
+            gc.collect()
+            continue
+
+        if not is_nncf_available():
+            raise ImportError("Quantization of the weights requires nncf, please install it with `pip install nncf`")
+
+        from optimum.intel.openvino.quantization import _weight_only_quantization
+
+        from ...intel.openvino.configuration import _GPTOSSQuantizationConfig
```
echarlaix: why not ?
```suggestion
        from optimum.intel.openvino.configuration import _GPTOSSQuantizationConfig
```

optimum/exporters/openvino/__main__.py
```diff
@@ -621,3 +599,187 @@ def maybe_convert_tokenizers(library_name: str, output: Path, model=None, prepro
...
+        class_name = diffusers_config.get("_class_name", None)
+        ov_class_name = f"OV{class_name}"
+        try:
+            model_cls = getattr(__import__("optimum.intel", fromlist=[ov_class_name]), ov_class_name)
+        except (AttributeError, ImportError) as e:
+            raise RuntimeError(f"Wasn't able to locate OpenVINO class for {class_name} diffusion model.") from e
+    else:
+        try:
+            model_cls_name = _HEAD_TO_AUTOMODELS[task.replace("-with-past", "")]
+            if model_cls_name == "OVModelForFeatureExtraction" and library_name == "sentence_transformers":
```
echarlaix: wouldn't 
```suggestion
            if library_name == "sentence_transformers":
```
be equivalent ?

optimum/exporters/openvino/__main__.py
```diff
@@ -621,3 +599,187 @@ def maybe_convert_tokenizers(library_name: str, output: Path, model=None, prepro
...
+        compile=False,
+        trust_remote_code=trust_remote_code,
+        cache_dir=cache_dir,
+        **((model_kwargs or {}) | additional_model_kwargs),
+    )
+
+    # Step 3. Apply quantization
+    with TemporaryDirectory() as tmpdir:
+        # Save quantized model to a temporary directory to avoid conflicts when reading and writing from the same directory
+        model._apply_quantization(
```
echarlaix, line 666: why not give the `quantization_config` when calling `from_pretrained` above instead ?

optimum/exporters/openvino/__main__.py
```diff
@@ -409,6 +413,19 @@ def bitnet_load_hook(self, state_dict, prefix, *args, **kwargs):
         if loading_kwargs.get("torch_dtype") == "auto":
             loading_kwargs["torch_dtype"] = dtype
 
+    temporary_directory = None
+    original_output = None
+    # TODO: Remove GPT-OSS workaround when possible
+    quantization_config = None if ov_config is None else ov_config.quantization_config
+    is_generic_quantization = quantization_config and not isinstance(quantization_config, _GPTOSSQuantizationConfig)
+    if is_generic_quantization:
```
echarlaix: here the difference is that quantization will be applied after loading the model with an OVModel class is this correct ? Can you add a comment to specify it ? Current workflow is a bit confusing to me still, especially for case where we use `OVModelForXxx` to export and quantize a model, does it mean that we will create a new instance of `OVModelForXxx` in `main_export`? https://github.com/huggingface/optimum-intel/blob/33b78760d7347acb985baec5f7f404c4f6518ea5/optimum/intel/openvino/modeling_decoder.py#L340

optimum/intel/openvino/modeling_base.py
```diff
@@ -637,27 +635,40 @@ def _prepare_quantization_config(
 
         return quantization_config
 
-    @staticmethod
-    def _apply_quantization(
-        model: "OVBaseModel",
+    def _preprocess_quantization_config(
```
echarlaix, line 638: wondering if we should replace with more explicit name like `_set_quantization_config_processor` (or do you think that other steps/modifications not related to this will be added in the future?)

nikita-savelyevv: There is already a case when it does something different for diffusion models: https://github.com/huggingface/optimum-intel/blob/fd728a1b80ea698213e5057260aa4fde8d0ea12e/optimum/intel/openvino/modeling_diffusion.py#L713

optimum/exporters/openvino/__main__.py
nikita-savelyevv: If we provide `quantization_config` to `from_pretrained` call, then there will be an extra unnecessary step of attempting to match default quantization config with `_resolve_default_quantization_config`: https://github.com/huggingface/optimum-intel/blob/fd728a1b80ea698213e5057260aa4fde8d0ea12e/optimum/intel/openvino/modeling_base.py#L509-L511  

When quantizing through CLI, this step was already performed during creation of `quantization_config` from CLI args.

In theory, we could run the second matching step that will do nothing, because the ` quantization_config == {"bits": 4}`  condition in any case will not hold, but there could be a second info message in such case https://github.com/huggingface/optimum-intel/blob/fd728a1b80ea698213e5057260aa4fde8d0ea12e/optimum/intel/openvino/modeling_base.py#L632-L634 which is not ideal.

Another reason is that in the future we'll probably need to pass custom args to `apply_quantization` which will only be added during CLI quantization flow https://github.com/nikita-savelyevv/optimum-intel/blob/9c260c9f8a7047c8078ffec8c60c0f7b0f38d28f/optimum/exporters/openvino/__main__.py#L662 , this is the immediate quantized model serialization approach mentioned in the PR description.

optimum/exporters/openvino/__main__.py
nikita-savelyevv: > here the difference is that quantization will be applied after loading the model with an OVModel class is this correct ? Can you add a comment to specify it ?

Yes, that's right. I've added such comment right above `_apply_generic_quantization` call if that's ok.

> Current workflow is a bit confusing to me still, especially for case where we use OVModelForXxx to export and quantize a model, does it mean that we will create a new instance of OVModelForXxx in main_export?

That's a good question! In case a model is exported through `OVModelForX` class, the generic quantization part will not be executed because we never pass a `quantization_config` through `OVConfig` in such case, e.g. for the base class: https://github.com/huggingface/optimum-intel/blob/fd728a1b80ea698213e5057260aa4fde8d0ea12e/optimum/intel/openvino/modeling_base.py#L795-L798 

In this case quantization will be applied at `_from_pretrained` call, e.g. https://github.com/huggingface/optimum-intel/blob/fd728a1b80ea698213e5057260aa4fde8d0ea12e/optimum/intel/openvino/modeling_base.py#L509-L511

optimum/exporters/openvino/__main__.py
nikita-savelyevv: It should be, changed

optimum/exporters/openvino/__main__.py
nikita-savelyevv: My IDE automatically inserts relative imports for some reason, changed.

optimum/exporters/openvino/__main__.py
nikita-savelyevv: Thanks! Changed.

optimum/exporters/openvino/__main__.py
echarlaix: what do you think of having `main_export` only reserved for the conversion step instead (keeping it called in https://github.com/huggingface/optimum-intel/blob/33b78760d7347acb985baec5f7f404c4f6518ea5/optimum/intel/openvino/modeling_decoder.py#L340), and have everything related to quantization in an `OVQuantizer` (or something similar) which would apply quantization on a model loaded with `OVModelForXxx` (so potentially a combination of multiple components),  that could be used when applying quantization when loading a model directly with `.from_pretrained()`.  The workflow could be simplified this way and we would have the same resulting model no matter what we path we used to obtain it (via the CLI or `.from_pretrained()`)

echarlaix: > There is an ask for encapsulating OpenVINO export path through a single main_export()

I think we need to have a function only for export and can create one to combine export + quantization instead of having only `main_export`, wdyt?

nikita-savelyevv: Thanks for the ideas! They seem very reasonable. I've implemented some of the changes you propose. I hope other points could be addressed in upcoming PRs.

.github/workflows/test_openvino_slow.yml
```diff
@@ -89,6 +89,12 @@ jobs:
...
           pytest tests/openvino/test_modeling_basic.py
 
-      - name: Test with Pytest (slow)
+      - if: ${{ matrix.os == 'windows-2022' }}  # Run all tests on windows
+        name: Test with Pytest
         run: |
-          pytest tests/openvino -m "run_slow" --durations=0
+          pytest tests/openvino --durations=0
+
+      - if: ${{ matrix.os != 'windows-2022' }}
```
echarlaix: the slow tests could also be triggered for windows no ?

optimum/exporters/openvino/__main__.py
```diff
@@ -574,6 +519,111 @@ def bitnet_load_hook(self, state_dict, prefix, *args, **kwargs):
...
+            model repository.
+        model_kwargs (`Optional[Dict[str, Any]]`, defaults to `None`):
+            Experimental usage: keyword arguments to pass to the model during
+            the export. This argument should be used along the `custom_export_configs` argument
+            in case, for example, the model inputs/outputs are changed (for example, if
+            `model_kwargs={"output_attentions": True}` is passed).
+
+    """
+    from ...intel.openvino.utils import _HEAD_TO_AUTOMODELS, TemporaryDirectory
+    from ...intel.utils.import_utils import DIFFUSERS_IMPORT_ERROR, is_diffusers_available
```
echarlaix: if possible would replace with absolute import and move to file top

optimum/commands/export/openvino.py
```diff
@@ -444,138 +447,28 @@ def run(self):
...
+            original_output = Path(self.args.output)
+            temporary_directory = TemporaryDirectory()
+            output = Path(temporary_directory.name)
         else:
+            output = Path(self.args.output)
+
+        try:
             # TODO : add input shapes
-            main_export(
+            library_name, task = main_export(
```
echarlaix: instead of returning library_name / task which would be a breaking change, why not inferring the `library_name` and `task` before calling `main_export` ?

optimum/exporters/openvino/__main__.py
```diff
@@ -574,6 +519,111 @@ def bitnet_load_hook(self, state_dict, prefix, *args, **kwargs):
...
+            if library_name == "sentence_transformers":
+                model_cls_name = "OVSentenceTransformer"
+            model_cls = getattr(__import__("optimum.intel", fromlist=[model_cls_name]), model_cls_name)
+        except (AttributeError, ImportError, KeyError) as e:
+            raise RuntimeError(f"Wasn't able to locate OpenVINO class for task {original_task} ({task}).") from e
+
+    # Step 2. Load the exported model
+    additional_model_kwargs = (
+        {"use_cache": task.endswith("with-past")}
+        if "generation" in task or task.startswith("automatic-speech-recognition")
```
echarlaix: shouldn't be the same for automatic-speech-recognition ? (task being automatic-speech-recognition-with-past)

echarlaix (approved):
Looks great, thanks a lot for the PR @nikita-savelyevv !

.github/workflows/test_openvino_slow.yml
nikita-savelyevv: Yes, but with this I'd like to verify that with introduction of `_merge_move` export works fine on Windows. The slow test as it is right now does not trigger any tests other than the ones marked with "run_slow" tag.

These temporary changes run all the tests on windows. I'll revert them once I verify that the failures we see right now are not caused by changes in the PR.

optimum/exporters/openvino/__main__.py
nikita-savelyevv: Could you please rephrase your question?

With this we check if the task ends with "with-past". This check only makes sense if either there is a "generation" word in task, e.g. text2text-generation, or if the task name starts with "automatic-speech-recognition".

echarlaix: ha yes my bad! then checking if any task ends with "with-past" should also work and would also take into account any task with the "-with-past" suffix like feature-extraction-with-past https://github.com/huggingface/optimum-intel/blob/46557e9a0f780a89bd4f880f4ebb0c584719a0f5/optimum/exporters/openvino/model_configs.py#L568
what do you think? (not sure this is needed so feel free to close)

nikita-savelyevv: Ah, that's a good point and would indeed simplify things. I remember I had some issues when setting `use_cache=False` explicitly, but let me check once more.

.github/workflows/test_openvino_slow.yml
nikita-savelyevv: There was indeed an issue with export of Zamba2/LFM2 models which I fixed. Now the set of failed Windows exporters_cli tests is the same as on the main branch ([main](https://github.com/huggingface/optimum-intel/actions/runs/20108257270?pr=734) vs [pr](https://github.com/huggingface/optimum-intel/actions/runs/20136514796?pr=1525)). We should address the failed Windows tests separately. Also would be nice to run all test scope on Windows on a regular basis.

optimum/exporters/openvino/__main__.py
nikita-savelyevv: Moved import_utils imports to the top and changed utils import to an absolute one.

optimum/commands/export/openvino.py
nikita-savelyevv: I agree, done

optimum/exporters/openvino/__main__.py
nikita-savelyevv: Seems like there are no issues with this approach, changed to a simple `use_cache=task.endswith("with-past")`

echarlaix: thanks!

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -1240,16 +1261,10 @@ def __init__(
                     self.ssm_cache_names.append(state.name)
                 elif "cache_params.present.conv" in state.name:
                     self.conv_cache_names.append(state.name)
-
-        self.key_cache_names = sorted(self.key_cache_names)
```
echarlaix, line 1244: why is this modification needed ?  (cache being created during compilation instead)

nikita-savelyevv: I briefly touched on it in https://github.com/huggingface/optimum-intel/pull/1525#discussion_r2611121868. But let me provide more details.

I had some issues on Windows when I was not able to delete model files after a model was loaded with `.from_preptrained()` and then deleted (`del model; gc.collect()`). Among the tested cases, this happened only for Zamba2 and LFM2. After some investigation, I noticed that for this types of models, an openvino model is compiled every time during initialization despite `compile=False` is provided. This was done to build the variables like `self.key_cache_names` and others. Because model was compiled, it interfered with OS thinking that model files are still in use and prevented them from being deleted.

To solve this, I moved initialization of these variable to happen during the first time model is compiled. This is ok because they are needed only during `forward` call. I believe this is how it should have been implemented in the first place, because the previous logic contradicts with `compile` argument behavior we expect.

echarlaix: thanks a lot for the explanation and fix!

optimum/exporters/openvino/__main__.py
nikita-savelyevv: @echarlaix  Found it:
```
optimum-cli export openvino -m TinyLlama/TinyLlama-1.1B-Chat-v1.0 ./tmp
...
File "/home/nsavel/workspace/optimum-intel/optimum/intel/openvino/modeling_decoder.py", line 195, in raise_error
    raise ValueError(
ValueError: `use_cache` was set to `False` but the loaded model only supports `use_cache=True`. Please load your current model with `use_cache=True` or export the original model once again with `use_cache=False` when calling the `from_pretrained` method. To export your model, simply set `export=True`.
```

PR #1526, opened 2025-11-20
fix bug for attention_mask when model is not patched

kaixuanliu:


kaixuanliu:
@echarlaix @IlyasMoutawwakil  This PR solves the performance drop problem  when using optimum-intel with not patched model, pls help review, thx.

IlyasMoutawwakil (approved):
Thanks

optimum/intel/ipex/modeling_base.py
```diff
@@ -448,7 +448,8 @@ def _reorder_cache(self, past_key_values, beam_idx):
 
     def prepare_inputs_for_generation(self, *args, **kwargs):
         input_kwargs = self.model.prepare_inputs_for_generation(*args, **kwargs)
-        input_kwargs["attention_mask"] = kwargs.get("attention_mask", None)
+        if self.add_patch:
+            input_kwargs["attention_mask"] = kwargs.get("attention_mask", None)
```
IlyasMoutawwakil, line 453: can you please add a comment explaining why this is needed

IlyasMoutawwakil: @kaixuanliu

kaixuanliu: Sure. For models wrapped by `IPEXModelForCausalLM` but not patched, they should use vanilla transformers modeing, where attention_mask is processed by functions like `sdpa_mask_recent_torch` in [L248](https://github.com/huggingface/transformers/blob/v4.55.0/src/transformers/masking_utils.py#L248), and this will have performance benefit. However, here I forgot to add `if self.add_patch` judgement, and this will let attention_mask processed by `sdpa_mask_recent_torch` be changed back to original one. And as a result, the performance will be worse than that of using vanilla transformers modeling.

kaixuanliu: comment added.

PR #1220, opened 2025-04-01
[POC] Use torch.export for converting

mvafin:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1531, opened 2025-11-27
[OpenVINO] Env var for switching to torch.export()

popovaan:
# What does this PR do?

Added env var **OPENVINO_DYNAMO_EXPORT** that switches exporting from torch.script() to torch.export().

Continuation of https://github.com/huggingface/optimum-intel/pull/1220.

## Before submitting

### DIFF
--- a/.github/workflows/test_openvino_nightly.yml
+++ b/.github/workflows/test_openvino_nightly.yml
@@ -69,7 +69,19 @@ jobs:
           - nncf-version: "nncf-stable"
             openvino-version: "openvino-stable"
             export-mode: "torchdynamo"
-            test-pattern: "*{genai,decoder,seq2seq,modeling}*"
+            test-pattern: "*modeling*"
+          - nncf-version: "nncf-stable"
+            openvino-version: "openvino-stable"
+            export-mode: "torchdynamo"
+            test-pattern: "*decoder*"
+          - nncf-version: "nncf-stable"
+            openvino-version: "openvino-stable"
+            export-mode: "torchdynamo"
+            test-pattern: "*seq2seq*"
+          - nncf-version: "nncf-stable"
+            openvino-version: "openvino-stable"
+            export-mode: "torchdynamo"
+            test-pattern: "*diffusion*"
 
     runs-on: ubuntu-22.04
 


optimum/commands/export/openvino.py
```diff
@@ -293,6 +293,12 @@ def parse_args_openvino(parser: "ArgumentParser"):
         type=json.loads,
         help=("Any kwargs passed to the model forward, or used to customize the export for a given model."),
     )
+    optional_group.add_argument(
+        "--use-torch-export",
```
nikita-savelyevv: Have you considered the name `--torch-export` for this argument? I think it is more concise without any drawbacks.

tests/openvino/excluded_tests_with_torch_export.txt
```diff
@@ -0,0 +1,504 @@
+OVModelBasicIntegrationTest::test_pipeline_2_hf_internal_testing_tiny_random_mbart
```
nikita-savelyevv: Unfortunately this approach is not very robust. For example, if some test is added or removed, the test order may change and this will affect the naming. 

I also notice that for each type of test the list of failing architectures is similar, e.g. there are t5, gpt2, bart, etc. Have you considered introducing a variable like `ARCHITECTURES_FAILING_WITH_TORCH_EXPORT` containing a list of all model architectures which fail to convert with torch.export and skipping tests based on that?

With this approach we will gradually remove models from this list until hopefully there will be no models left there.

optimum/commands/export/openvino.py
nikita-savelyevv: Please add this argument to `docs/source/openvino/export.mdx`

.github/workflows/test_openvino.yml
```diff
@@ -38,6 +38,11 @@ jobs:
             "*quantization*",
           ]
         transformers-version: ["4.45.0", "latest"]
+        use_torch_export: ["True", "False"]
```
nikita-savelyevv: Could you please rework this in such a way that workflows will be renamed this way? 
- `build (*export*, 4.45.0, False)` -> `build (*export*, 4.45.0, torch_script)`
- `build (*export*, 4.45.0, True)` -> `build (*export*, 4.45.0, torch_export)`

This will improve readability in my opinion.

tests/openvino/excluded_tests_with_torch_export.txt
popovaan: Not all test names have model architecture, for example:
```
OVCLIExportTestCase::test_export_25_text_generation_with_past
OVCLIExportTestCase::test_exporters_cli_24_text_generation
OVCLIExportTestCase::test_exporters_cli_25_text_generation_with_past
OVCLIExportTestCase::test_exporters_cli_fp16_28_text_generation_with_past
OVCLIExportTestCase::test_exporters_cli_full_quantization_00_automatic_speech_recognition
```
But I agree that some tests, which have architecture names, can be skipped based on that. So we can probably combine these two approaches.

nikita-savelyevv: It seems to me that it should be possible to do the parsing not by the test name, but somehow through a decorator. Because from what I see, for (all?) tests one of the arguments is the architecture name. Then for all tests where you added the USE_TORCH_EXPORT argument, in theory you could also add a decorator that skips the test if USE_TORCH_EXPORT env. var. is "True" and the arguments contain an unsupported architecture name.

nikita-savelyevv: However, there are tests where model_id is hardcoded:
https://github.com/popovaan/optimum-intel/blob/torch_exprt/tests/openvino/test_modeling.py#L633 

Maybe in that case we could extract this into an argument and add a `parameterized` with a single element.

optimum/commands/export/openvino.py
popovaan: Agree, renamed.

.github/workflows/test_openvino.yml
popovaan: Yes, done.

tests/openvino/excluded_tests_with_torch_export.txt
rkazants: I agree with @nikita-savelyevv. I missed a fact that test case id can be changed in case of removing or adding another test case.
Is it possible to slightly refactor tests to have arch name in test case name instead of their ids. If yes, we can proceed with "excluded tests" approach. It will be more convenient way because we avoid solid changes in tests and will work only with this excluded tests.
If not possible, let us proceed with Nikita's approach. For me, it is not principal. Now it is more important to start transition to torch.export().

optimum/exporters/openvino/convert.py
```diff
@@ -407,18 +417,38 @@ def ts_patched_forward(*args, **kwargs):
...
+            if torch_export:
+                if hasattr(torch.ops, "_prepare_4d_causal_attention_mask_for_sdpa"):
+                    # patch_everywhere breaks torch.ops namespace
+                    del torch.ops._prepare_4d_causal_attention_mask_for_sdpa
+                dynamic_shapes = _get_dynamic_shapes_info(model, config, dummy_inputs)
+                _export_kwargs = {"args": (), "kwargs": _normalize_dummy_inputs(dummy_inputs, get_model_dtype(model))}
+                _export_kwargs["dynamic_shapes"] = dynamic_shapes
+
+                try:
+                    from nncf.torch.dynamic_graph.patch_pytorch import disable_patching
```
nikita-savelyevv: After https://github.com/huggingface/optimum-intel/pull/1557 we don't have `nncf.torch` import anywhere. So I believe this is no longer needed.

tests/openvino/utils_tests.py
```diff
@@ -500,6 +505,105 @@ def get_num_sdpa(model):
...
+            if param in ["model_arch", "model_id", "model_type", "model_name"]:
+                model_id_idx = idx
+                break
+        if model_id_idx != -1:
+            model_id = args[model_id_idx]
+            if USE_TORCH_EXPORT and model_id in _ARCHITECTURES_NOT_SUPPORTED_WITH_TORCH_EXPORT:
+                pytest.skip("{} is unsupported with torch.export.".format(model_id))
+        return fn(*args, **kwargs)
+
+    return run_test
```
nikita-savelyevv:
```suggestion
def skip_architectures_unsupported_with_torch_export(fn):
    @functools.wraps(fn)
    def run_test(*args, **kwargs):
        if not USE_TORCH_EXPORT:
            return fn(*args, **kwargs)
        test_signature = signature(fn)
        model_id_idx = -1
        for idx, param in enumerate(test_signature.parameters):
            if param in ["model_arch", "model_id", "model_type", "model_name"]:
                if model_id_idx != -1:
                    raise RuntimeError("Detected ambiguous model id arguments.")
                model_id_idx = idx
        if model_id_idx == -1:
            raise ValueError("Model id argument not found.")
        model_id = args[model_id_idx]
        if model_id in _ARCHITECTURES_NOT_SUPPORTED_WITH_TORCH_EXPORT:
            pytest.skip(f"{model_id} is unsupported with torch.export.")
        return fn(*args, **kwargs)

    return run_test
```

Looks great! I have some minor suggestions to make it more robust.

optimum/commands/export/openvino.py
popovaan: Done.

tests/openvino/excluded_tests_with_torch_export.txt
popovaan: Used decorator to skip tests by architecture name according Nikita's suggestion.

optimum/exporters/openvino/convert.py
popovaan: Removed.

docs/source/openvino/export.mdx
```diff
@@ -40,7 +40,7 @@ usage: optimum-cli export openvino [-h] -m MODEL [--task TASK] [--framework {pt}
                                    [--lora-correction] [--sensitivity-metric SENSITIVITY_METRIC]
                                    [--quantization-statistics-path QUANTIZATION_STATISTICS_PATH]
                                    [--num-samples NUM_SAMPLES] [--disable-stateful] [--disable-convert-tokenizer]
-                                   [--smooth-quant-alpha SMOOTH_QUANT_ALPHA]
+                                   [--smooth-quant-alpha SMOOTH_QUANT_ALPHA] [--use-torch-export]
```
nikita-savelyevv:
```suggestion
                                   [--smooth-quant-alpha SMOOTH_QUANT_ALPHA] [--torch-export]
```

docs/source/openvino/export.mdx
```diff
@@ -165,6 +165,7 @@ Optional arguments:
   --smooth-quant-alpha SMOOTH_QUANT_ALPHA
                         SmoothQuant alpha parameter that improves the distribution of activations before MatMul layers
                         and reduces quantization error. Valid only when activations quantization is enabled.
+  --use-torch-export    Use torch.export() for scripting of PyTorch models.
```
nikita-savelyevv:
```suggestion
  --torch-export    Use torch.export() for scripting of PyTorch models.
```

optimum/exporters/openvino/utils.py
```diff
@@ -120,6 +121,72 @@ def _get_input_info(
...
+    new_dummy = {}
+    for name, value in dummy_inputs.items():
+        new_dummy[name] = _normalize_element(value, dtype)
+    return new_dummy
+
+
+def get_model_dtype(model):
+    for param in model.parameters():
+        return param.dtype
+    return getattr(model, "dtype", torch.float32)
```
IlyasMoutawwakil: this doesn't work with diffusion pipelines

.github/workflows/test_openvino.yml
```diff
@@ -39,6 +39,15 @@ jobs:
...
         transformers-version: ["4.45.0", "latest"]
+        export_mode: ["torch_script", "torch_export"]
+        include:
+          - export_mode: "torch_export"
+            use_torch_export_bool: "True"
+          - export_mode: "torch_script"
+            use_torch_export_bool: "False"
+        exclude:
+          - transformers-version: "4.45.0"
+            export_mode: "torch_export"
```
IlyasMoutawwakil: we can't double the workload of our ci for a parameter that fails most of the time 😅

IlyasMoutawwakil (changes requested):
I don't think this is the right way to integrate dynamo export, consider the following scenario, a user sees this argument in the docs, wants to use this feature, how do they know if their model can be exported with torch.dynamo ? should they simply *try it out* and see for themselves if it's working with some models and failing with others, with no clear explanation, just a long torch.export traceback 😅
in my opinion, taking the approach of "dynamo export as a feature" doubles the complexity and testing time of our integration, all while still failing with half the models we support.
the alternative I suggest is a "migration" approach, which makes more sense given that we need to migrate from torch script to torch dynamo, not support both.
Rather than supporting both paths and having to test both of them for each model, we start migrating models one by one into using dynamo export by default (or based on a condition like torch_version >= 2.5). for example in the openvino config, adding a property "DYNAMO"_EXPORT=True (or torch_version>=2.5" for models that support it, this way when we add this "DYNAMO"_EXPORT=True" to a model we are 100% sure and guaranteeing that it works with this export backend because now all the tests will be using the dynamo exported ov model instead of the torchscript one. 
We don't really need to support both export paths, if one path is clearly better, let's go the better route.
btw @popovaan thanks for the tremendous work, tell me if the approach i'm suggesting is clear and if it makes sense for you.

.github/workflows/test_openvino.yml
popovaan: Models will be migrated from torch.script() to torch.export() once they pass full validation, therefore this is a temporary measure.

optimum/exporters/openvino/utils.py
popovaan: Yes, among the diffusion group, only the “latent-consistency” architecture works at the moment. But the intent of this PR was not to support all models, but to introduce a convenient way to switch between torch.script() and torch.export().

IlyasMoutawwakil: totally get you, it's just when things are introduced it's usually hard to get get rid of, especially when not private, please add an "_" to mark it as private/internal util

popovaan: Done.

.github/workflows/test_openvino_nightly.yml
```diff
@@ -51,6 +51,7 @@ jobs:
             "*diffusion*",
             "*quantization*",
           ]
+        export_mode: [ "torch_script", "dynamo_export" ]
```
IlyasMoutawwakil: export_mode doubles the ci matrix without ever being used

IlyasMoutawwakil:
```suggestion
        openvino_dynamo_export: ["False"]
```

.github/workflows/test_openvino_nightly.yml
```diff
@@ -61,6 +62,10 @@ jobs:
           - nncf-version: "nncf-develop"
             openvino-version: "openvino-nightly"
             test-pattern: "*{quantization,export}*"
+          - export_mode: "dynamo_export"
+            dynamo_export_bool: "True"
+          - export_mode: "torch_script"
+            dynamo_export_bool: "False"
```
IlyasMoutawwakil:
```suggestion
          - nncf-version: "nncf-stable"
            openvino-version: "openvino-stable"
            openvino_dynamo_export: "True"
            test-pattern: "*"
```

optimum/exporters/openvino/convert.py
```diff
@@ -407,18 +410,32 @@ def ts_patched_forward(*args, **kwargs):
...
-                __make_16bit_traceable(model)
             check_dummy_inputs_are_allowed(model, dummy_inputs)
             input_info = _get_input_info(model, config, dummy_inputs)
-            ts_decoder = TorchScriptPythonDecoder(model, example_input=dummy_inputs, **ts_decoder_kwargs)
-            ov_model = convert_model(
-                ts_decoder,
-                example_input=dummy_inputs,
-                input=[(item.shape, item.type) for item in input_info],
-            )
+            torch_export = os.getenv("OPENVINO_DYNAMO_EXPORT", "False") == "True"
```
IlyasMoutawwakil:
```suggestion
            torch_export = os.getenv("OPENVINO_DYNAMO_EXPORT", "false").lower() == "true"
```

.github/workflows/test_openvino_nightly.yml
```diff
@@ -51,6 +51,7 @@ jobs:
             "*diffusion*",
             "*quantization*",
           ]
+        openvino_dynamo_export: ["False"]
```
IlyasMoutawwakil:
```suggestion
        export_mode: ["torchscript"]
```

.github/workflows/test_openvino_nightly.yml
```diff
@@ -61,6 +62,10 @@ jobs:
           - nncf-version: "nncf-develop"
             openvino-version: "openvino-nightly"
             test-pattern: "*{quantization,export}*"
+          - nncf-version: "nncf-stable"
+            openvino-version: "openvino-stable"
+            openvino_dynamo_export: "True"
```
IlyasMoutawwakil:
```suggestion
            export_mode: "torchdynamo"
```

.github/workflows/test_openvino_nightly.yml
```diff
@@ -103,5 +108,7 @@ jobs:
           pip freeze
 
       - name: Test with Pytest
+        env:
+          OPENVINO_DYNAMO_EXPORT: ${{ matrix.dynamo_export_bool }}
```
IlyasMoutawwakil:
```suggestion
          OPENVINO_DYNAMO_EXPORT: ${{ matrix.export_mode == "torchdynamo" }}
```

nikita-savelyevv:
I see that the dynamo test job is canceled before finishing: https://github.com/huggingface/optimum-intel/actions/runs/20429776830/job/58792970015?pr=1531. I have a concern that if this is how it always behaves, then we won't see and actual test log.

popovaan:
I split the torchdynamo nightly run into two groups, now they are not cancelled, and we can see the full logs.
There are some failures that sporadically appear in torchscript tests, but they do not seem to be related to this PR, as I see similar failures in other PRs as well, for example: https://github.com/huggingface/optimum-intel/actions/runs/20594462767/job/59146091357?pr=1575

@IlyasMoutawwakil can we merge the PR?

.github/workflows/test_openvino_nightly.yml
```diff
@@ -61,6 +62,14 @@ jobs:
...
             openvino-version: "openvino-nightly"
             test-pattern: "*{quantization,export}*"
+          - nncf-version: "nncf-stable"
+            openvino-version: "openvino-stable"
+            export-mode: "torchdynamo"
+            test-pattern: "*{quantization,export,diffusion}*"
+          - nncf-version: "nncf-stable"
+            openvino-version: "openvino-stable"
+            export-mode: "torchdynamo"
+            test-pattern: "*{genai,decoder,seq2seq,modeling}*"
```
IlyasMoutawwakil: this one is too big, maybe spread it out

.github/workflows/test_openvino_nightly.yml
```diff
@@ -61,6 +62,14 @@ jobs:
           - nncf-version: "nncf-develop"
             openvino-version: "openvino-nightly"
             test-pattern: "*{quantization,export}*"
+          - nncf-version: "nncf-stable"
+            openvino-version: "openvino-stable"
+            export-mode: "torchdynamo"
+            test-pattern: "*{quantization,export,diffusion}*"
```
IlyasMoutawwakil:
```suggestion
            test-pattern: "*{quantization,export}*"
```

.github/workflows/test_openvino_nightly.yml
IlyasMoutawwakil:
```suggestion
          - nncf-version: "nncf-stable"
            openvino-version: "openvino-stable"
            export-mode: "torchdynamo"
            test-pattern: "*modeling*"
          - nncf-version: "nncf-stable"
            openvino-version: "openvino-stable"
            export-mode: "torchdynamo"
            test-pattern: "*decoder*"
          - nncf-version: "nncf-stable"
            openvino-version: "openvino-stable"
            export-mode: "torchdynamo"
            test-pattern: "*seq2seq*"
          - nncf-version: "nncf-stable"
            openvino-version: "openvino-stable"
            export-mode: "torchdynamo"
            test-pattern: "*diffusion*"
```

IlyasMoutawwakil: we split them because the runners disk memory can't handle storing all the related models.

PR #1533, opened 2025-11-28
[OpenVINO] Remove deprecated openvino.runtime module

almilosz:
`openvino.runtime` is deprecated in favor of flat `openvino` and is set to be removed in the upcoming 26.0 OpenVINO release.

- https://github.com/openvinotoolkit/openvino/pull/32698

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -1179,7 +1179,7 @@ class OVQuantizer(OptimumQuantizer):
             logger.warning("The `task` argument is ignored and will be removed in optimum-intel v1.27")
 
     @property
-    def task(self) -> Dict[str, Union[openvino.Model, openvino.CompiledModel]]:
+    def task(self) -> str:
         logger.warning("The `task` attribute is deprecated and will be removed in v1.27.")
         return self._task
 


almilosz:
@echarlaix @nikita-savelyevv Please take a look

optimum/intel/openvino/quantization.py
```diff
@@ -1179,7 +1179,7 @@ def __init__(
             logger.warning("The `task` argument is ignored and will be removed in optimum-intel v1.27")
 
     @property
-    def task(self) -> Dict[str, Union[openvino.Model, openvino.runtime.CompiledModel]]:
+    def task(self) -> Dict[str, Union[openvino.Model, openvino.CompiledModel]]:
```
nikita-savelyevv:
```suggestion
    def task(self) -> str:
```

That's weird that the type hint for `task` is `Dict[str, Union[openvino.Model, openvino.runtime.CompiledModel]]`. It should be `str`.

almilosz: I applied your suggestion.

rkazants:
@nikita-savelyevv, @echarlaix, @IlyasMoutawwakil, could you please merge it?

Thanks,
Roman

Issue #1528, opened 2025-11-24
Errors during inference of a whisper model exported with --task automatic-speech-recognition


nikita-savelyevv:
When whisper model is exported with explicitly provided `--task automatic-speech-recognition`, inferencing it fails with the error:
```
Traceback (most recent call last):
  File "/home/nsavel/workspace/optimum-intel/tmp.py", line 27, in <module>
    transcription = processor.batch_decode(model.generate(input_features), skip_special_tokens=True)[0]
  File "/home/nsavel/venvs/optimum-intel/lib/python3.10/site-packages/transformers/models/whisper/generation_whisper.py", line 866, in generate
    ) = self.generate_with_fallback(
  File "/home/nsavel/venvs/optimum-intel/lib/python3.10/site-packages/transformers/models/whisper/generation_whisper.py", line 1038, in generate_with_fallback
    seek_outputs = super().generate(
  File "/home/nsavel/venvs/optimum-intel/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
  File "/home/nsavel/venvs/optimum-intel/lib/python3.10/site-packages/transformers/generation/utils.py", line 2617, in generate
    result = self._sample(
  File "/home/nsavel/venvs/optimum-intel/lib/python3.10/site-packages/transformers/generation/utils.py", line 3591, in _sample
    model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
  File "/home/nsavel/workspace/optimum-intel/optimum/intel/openvino/modeling_seq2seq.py", line 1346, in prepare_inputs_for_generation
    past_length = self.decoder._get_past_length(past_key_values)
  File "/home/nsavel/workspace/optimum-intel/optimum/intel/openvino/modeling_seq2seq.py", line 992, in _get_past_length
    return past_key_values[0][0].shape[-2]
IndexError: tuple index out of range
```
Lately, whisper models are exported with `--task automatic-speech-recognition{*}-with-past{*}`. Nevertheless, the legacy behavior should work correctly.

Reproducer:
```
optimum-cli export openvino --task automatic-speech-recognition --model optimum-intel-internal-testing/tiny-random-whisper --weight-format fp16 --trust-remote-code ./whisper
python reproducer.py
```

reproducer.py:
```python
import numpy as np
from datasets import load_dataset
from transformers import AutoProcessor

from optimum.intel import OVModelForSpeechSeq2Seq

[...]

PR #1534, opened 2025-12-01
Fix whisper inference for models exported without pkv 

echarlaix:
Fixes https://github.com/huggingface/optimum-intel/issues/1528

Fix inference for models exported with `use_cache=False` for which past_key_values will be an empty tuple 

Alsi add a corresponding test


tests/openvino/test_seq2seq.py
```diff
@@ -312,21 +283,35 @@ def test_compare_to_transformers(self, model_arch):
...
-            **pt_features, generation_config=gen_config, **generate_kwrgs
-        )
-
-        self.assertTrue(torch.equal(generated_tokens, ov_generated_tokens))
-        self.assertTrue(torch.equal(generated_tokens, ov_stateless_generated_tokens))
-
         del transformers_model
-        del ov_model
+
+        for use_cache, stateful in ((True, True), (True, False), (False, False)):
```
IlyasMoutawwakil: can we have this in the test parameters ? it makes debugging easier imo

echarlaix: makes sense!  added in https://github.com/huggingface/optimum-intel/pull/1534/commits/de41bf67e7684ccd3aa33f87b1504adb07cbe322

PR #1538, opened 2025-12-02
[OpenVINO] Optimize IR for Mamba models

rkazants:
# What does this PR do?

Simplify IR for Mamba models. Now it contains `ReduceSum` operation, instead of convolution with a kernel of ones. Also, now it has the common flag `is_prefill` used in two places.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_seq2seq.py
+++ b/optimum/intel/openvino/modeling_seq2seq.py
@@ -958,7 +958,6 @@ class OVDecoder(OVModelPart):
         self.request.wait()
         logits = torch.from_numpy(self.request.get_tensor("logits").data).clone().to(self.device)
         self._past_length += input_ids.shape[1]
-
         out_past_key_values = ((),)
 
         if not self.stateful:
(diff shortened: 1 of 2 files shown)


tests/openvino/utils_tests.py
```diff
@@ -338,8 +338,8 @@
         "vocoder": 80,
     },
     "clip": {"model": 130},
-    "mamba": {"model": 386},
-    "falcon-mamba": {"model": 194},
+    "mamba": {"model": 322},
+    "falcon-mamba": {"model": 162},
```
rkazants, line 342:
```suggestion
    "falcon-mamba": {"model": 162 },
```

tests/openvino/utils_tests.py
```diff
@@ -338,8 +338,8 @@
         "vocoder": 80,
     },
     "clip": {"model": 130},
-    "mamba": {"model": 386},
-    "falcon-mamba": {"model": 194},
+    "mamba": {"model": 322},
+    "falcon-mamba": {"model": 162 },
```
```suggestion
    "falcon-mamba": {"model": 162},
```

PR #1546, opened 2025-12-04
[OpenVINO] Take into account that `pillow` may be not installed

nikita-savelyevv:
# What does this PR do?

`pillow` is no longer installed with NNCF (develop version) and needs to be installed manually for data-aware quantization methods to work.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -729,7 +729,7 @@ class OVCalibrationDatasetBuilder:
         Currently, collects data only for a language model component.
         """
         if not is_pillow_available():
-            raise ValueError(
+            raise ImportError(
                 PILLOW_IMPORT_ERROR.format("OVCalibrationDatasetBuilder._prepare_visual_causal_lm_calibration_data")
             )
 
@@ -1073,7 +1073,7 @@ class OVCalibrationDatasetBuilder:
         seq_len: int = 128,
     ) -> OVCalibrationDataset:
         if not is_pillow_available():
-            raise ValueError(
+            raise ImportError(
                 PILLOW_IMPORT_ERROR.format(
                     "OVCalibrationDatasetBuilder._prepare_text_image_encoder_model_calibration_data"
                 )


setup.py
```diff
@@ -64,8 +64,8 @@
 QUALITY_REQUIRE = ["black~=23.1", "ruff==0.4.4"]
 
 EXTRAS_REQUIRE = {
-    "nncf": ["nncf>=2.19.0"],
-    "openvino": ["nncf>=2.19.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0"],
+    "nncf": ["nncf>=2.19.0", "pillow"],
+    "openvino": ["nncf>=2.19.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0", "pillow"],
```
rkazants: why does openvino need pillow?

rkazants:
```suggestion
    "openvino": ["nncf>=2.19.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0"],
```

setup.py
```diff
@@ -64,8 +64,8 @@
 QUALITY_REQUIRE = ["black~=23.1", "ruff==0.4.4"]
 
 EXTRAS_REQUIRE = {
-    "nncf": ["nncf>=2.19.0"],
-    "openvino": ["nncf>=2.19.0", "openvino>=2025.1.0", "openvino-tokenizers>=2025.1.0"],
+    "nncf": ["nncf>=2.19.0", "pillow"],
```
rkazants: I think you need to adjust GHA job for nightly and manually install `pillow` there. I don't want user to install it in a production

setup.py
nikita-savelyevv: OpenVINO lists nncf as dependency and some data-aware quantization features require pillow. Previously, pillow was installed together with nncf, but now it is not the case.

setup.py
nikita-savelyevv: Removed pillow from setup.py

optimum/intel/openvino/quantization.py
```diff
@@ -727,6 +728,13 @@ def _prepare_visual_causal_lm_calibration_data(
         Prepares calibration data for VLM pipelines.
         Currently, collects data only for a language model component.
         """
+        if not is_pillow_available():
+            raise ValueError(
```
echarlaix: would replace with
```suggestion
            raise ImportError(
```

optimum/intel/openvino/quantization.py
```diff
@@ -1064,6 +1072,15 @@ def _prepare_text_image_encoder_model_calibration_data(
         dataset: "Dataset",
         seq_len: int = 128,
     ) -> OVCalibrationDataset:
+        if not is_pillow_available():
+            raise ValueError(
```
echarlaix:
```suggestion
            raise ImportError(
```

optimum/intel/openvino/quantization.py
nikita-savelyevv: I agree, replaced

optimum/intel/openvino/quantization.py
nikita-savelyevv: I agree, replaced

PR #1547, opened 2025-12-05
[OpenVINO] Add gsm8k as a dataset option for CausalLM quantization

nikita-savelyevv:
# What does this PR do?

In this PR "gsm8k" dataset option is added to apply data-aware quantization to causal language modes. Calibrating on "gsm8k" in some cases provides better results compared to "wikitext2":

| Model                   | Precision         | Calibration Dataset | seqlen | gsm8k_cot_llama (strict-match) | gsm8k_cot_llama (flexible-extract) |
|-------------------------|-------------------|----------------------|-------------|-------------------------------|-------------------------------------|
| Llama-3.1-8B-Instruct   | nf4 per-channel   | wikitext2            | 32          | 78.70%                       | 78.77%                              |
| Llama-3.1-8B-Instruct   | nf4 per-channel   | gsm8k                | 64          | 80.29%                       | 80.82%                              |
| Llama-3.1-8B-Instruct   | nf4 per-channel   | gsm8k                | 128         | 79.30%                       | 79.53%                              |
| Llama-3.1-8B-Instruct   | nf4 per-channel   | gsm8k                | 256         | **81.20%**                     | **81.50%**                          |

## Before submitting

### DIFF
--- a/tests/openvino/test_exporters_cli.py
+++ b/tests/openvino/test_exporters_cli.py
@@ -318,7 +318,7 @@ class OVCLIExportTestCase(unittest.TestCase):
             "text-generation",
             "llama",
             "int4_f8e5m2",
-            "--dataset wikitext2 --num-samples 1 --group-size 16 --trust-remote-code",
+            "--dataset gsm8k --num-samples 1 --group-size 16 --trust-remote-code",
             {
                 "model": 15,
             },
@@ -531,7 +531,7 @@ class OVCLIExportTestCase(unittest.TestCase):
         (
             "text-generation-with-past",
             "llama_awq",
-            "int4 --ratio 1.0 --sym --group-size 16 --awq --dataset wikitext2 --num-samples 100 "
+            "int4 --ratio 1.0 --sym --group-size 16 --awq --dataset gsm8k --num-samples 100 "
             "--sensitivity-metric max_activation_variance",
             {"model": {"int8": 4, "int4": 14}},
         ),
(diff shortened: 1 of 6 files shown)


tests/openvino/test_exporters_cli.py
```diff
@@ -319,7 +318,7 @@ class OVCLIExportTestCase(unittest.TestCase):
             "text-generation",
             "llama",
             "int4_f8e5m2",
-            "--dataset wikitext2 --num-samples 1 --group-size 16 --trust-remote-code",
+            "--dataset gsm8k --num-samples 1 --group-size 16 --trust-remote-code",
```
IlyasMoutawwakil, line 321: shouldn't we just add a test instead of removing wikitext2  here ?

tests/openvino/test_exporters_cli.py
```diff
@@ -529,16 +528,10 @@ class OVCLIExportTestCase(unittest.TestCase):
...
-            "text-generation-with-past",
-            "gpt2",
-            "int4 --sym --group-size-fallback adjust",
-            {"model": {"int8": 4, "int4": 20}},
-        ),
         (
             "text-generation-with-past",
             "llama_awq",
-            "int4 --ratio 1.0 --sym --group-size 16 --awq --dataset wikitext2 --num-samples 100 "
+            "int4 --ratio 1.0 --sym --group-size 16 --awq --dataset gsm8k --num-samples 100 "
```
IlyasMoutawwakil, line 534: same question

IlyasMoutawwakil (approved):
LGTM ! isn't the perf increase to be expected since you are calibrating on on test data/distribution ?

tests/openvino/test_exporters_cli.py
nikita-savelyevv: My motivation was not to spawn additional tests. There are plenty of existing ones with `wikitext2` dataset that should cover that part of functionality.

rkazants (approved):
Is it possible and reasonable to specify several datasets (e.g., wikitext2 and gsm8k) simultaneously for calibration? Can it provide higher accuracy?

I think it is reasonable to update documentation and mention in what cases to use gsm8k dataset.

PR #1548, opened 2025-12-10
Revert "[Docs] Export whisper as stateless during quantization (#1435)"

nikita-savelyevv:
# What does this PR do?

This PR reverts #1435. After fixes were merged in https://github.com/huggingface/optimum-intel/pull/1505, using stateless models for quantization is no longer required as long as the latest NNCF version is used.

Ticket CVS-172705

## Before submitting

### DIFF
--- a/docs/source/openvino/optimization.mdx
+++ b/docs/source/openvino/optimization.mdx
@@ -1014,7 +1014,7 @@ ov_model = OVModelForSpeechSeq2Seq.from_pretrained(
 )
 ```
 
-With this, encoder, decoder and decoder-with-past models of the Whisper pipeline will be fully quantized, including activations.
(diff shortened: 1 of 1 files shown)


docs/source/openvino/optimization.mdx
```diff
@@ -1005,7 +1005,6 @@ The speech-to-text Whisper model can be quantized without the need for preparing
 model_id = "openai/whisper-tiny"
 ov_model = OVModelForSpeechSeq2Seq.from_pretrained(
     model_id,
-    stateful=False,
```
echarlaix, line 1008: now that the model is stateful, we should also modify the line below to remove the decoder-with-past mention

nikita-savelyevv: Good catch! Thanks

PR #1550, opened 2025-12-11
[OpenVINO] Accept nncf dev 2.19.0 builds

wkobielx:
# What does this PR do?
Accept custom installed from branch/sha versions - this should allow to use at least 2.19.0 dev0 builds

### DIFF
--- a/optimum/intel/openvino/__init__.py
+++ b/optimum/intel/openvino/__init__.py
@@ -58,7 +58,7 @@ if is_nncf_available():
 
     if not is_nncf_version(">=", "2.19.0.dev0"):
         raise ImportError(
-            "NNCF version 2.19.0.dev0 or higher is required to use NNCF-based quantization. "
+            "NNCF version 2.19 or higher is required to use NNCF-based quantization. "
             f"Please upgrade your NNCF installation. The current version of NNCF is {nncf.__version__}."
         )
 


optimum/intel/openvino/__init__.py
```diff
@@ -56,9 +56,9 @@
     logging.disable(logging.INFO)
     import nncf
 
-    if is_nncf_version("<", "2.19"):
+    if not is_nncf_version(">=", "2.19.0.dev0"):
         raise ImportError(
-            "NNCF version 2.19 or higher is required to use NNCF-based quantization. "
+            "NNCF version 2.19.0.dev0 or higher is required to use NNCF-based quantization. "
```
nikita-savelyevv:
```suggestion
            "NNCF version 2.19 or higher is required to use NNCF-based quantization. "
```

I think keeping 2.19 is ok in order not to confuse most of the users with "dev0" part.

wkobielx: Updated exception message

nikita-savelyevv (changes requested):
On the second look, using `is_nncf_version(">=", "2.19.0.dev0")` would allow any custom commit between NNCF 2.18 and 2.19 which is not ideal. This won't lead to any failures, but still if someone has such "in-between" version installed, they should be prompted to upgrade to the latest NNCF version. 

I wonder if instead validation can use custom commit of optimum-intel and not the latest version?

optimum/intel/openvino/__init__.py
```diff
@@ -56,7 +56,7 @@
     logging.disable(logging.INFO)
     import nncf
 
-    if is_nncf_version("<", "2.19"):
+    if not is_nncf_version(">=", "2.19.0.dev0"):
```
rkazants, line 59: I think we can just cut off the current version value to have in format xx.yy.zz and after that we can do comparation like is_nncf_version("<", "2.19"). It should fix the problem

nikita-savelyevv: Thanks for the suggestion! Unfortunately that won't work because in NNCF we increase dev version to the next one right after the release, i.e. the version was updated to 2.19.0.dev0 back in August https://github.com/openvinotoolkit/nncf/commit/5ba2f29db2271c67f39b423ce6f7f63f1b02b6f6. That's why we'll have the same problem as described in https://github.com/huggingface/optimum-intel/pull/1550#pullrequestreview-3566354385.

PR #1452, opened 2025-09-27
[OpenVINO][Draft]Add Qwen3VL and Qwen3VL-MOE

openvino-dev-samples:

PR #1551, opened 2025-12-11
[OpenVINO] Support Qwen3VL model

popovaan:
# What does this PR do?

Conversion cmd-line for `Qwen/Qwen3-VL-2B-Instruct`:
```sh
optimum-cli export openvino -m Qwen/Qwen3-VL-2B-Instruct ./Qwen3-VL
```

Inference of `Qwen/Qwen3-VL-2B-Instruct` using OpenVINO backend:

```python
from transformers.video_utils import load_video

model_dir = "./Qwen3-VL/"

# Prepare video input
video_path = hf_hub_download(
                repo_id="raushan-testing-hf/videos-test",
                filename="sample_demo_1.mp4",
                repo_type="dataset",
            )
input_video, _ = load_video(video_path, num_frames=10, backend="opencv")
question = "Why is this video funny?"
inputs = model.preprocess_inputs(processor=processor, text=question, video=input_video)

# Run inference

print(output_text)
```

Continuation of https://github.com/huggingface/optimum-intel/pull/1452

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -3437,6 +3437,13 @@ else:
         pass
 
 
+# The inheritance from Qwen3VLModel is needed to get access to methods:
+# get_placeholder_mask(): https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L1066
+# get_rope_index(): https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L916
+# get_video_features(): https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L1035
+#
+# and inheritance from Qwen3VLVisionModel is needed for accessing the following method:
+# rot_pos_emb(): https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L603
 class _OVQwen3VLForCausalLM(OVModelForVisualCausalLM, Qwen3VLModel, Qwen3VLVisionModel):
     additional_parts = ["vision_embeddings_merger", "vision_embeddings_pos"]
 


optimum/exporters/openvino/model_configs.py
```diff
@@ -3489,6 +3641,180 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
...
+    TEXT_EMBEDDINGS = "text_embeddings"
+    VISION_EMBEDDINGS_POS = "vision_embeddings_pos"
+
+
+@register_in_tasks_manager(
+    "qwen3_vl",
+    *["image-text-to-text"],
+    library_name="transformers",
+)
+class Qwen3_VLOpenVINOConfig(BaseVLMOpenVINOConfig):
```
```suggestion
class Qwen3VLOpenVINOConfig(BaseVLMOpenVINOConfig):
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -3489,6 +3641,180 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
...
+            self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
+            self._normalized_config.use_embed_dim = False
+        if self._behavior == Qwen3VLConfigBehavior.VISION_EMBEDDINGS_MERGER and hasattr(config, "vision_config"):
+            self._config = config.vision_config
+            self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
+            self._normalized_config.use_embed_dim = True
+        if self._behavior == Qwen3VLConfigBehavior.VISION_EMBEDDINGS_POS and hasattr(config, "vision_config"):
+            self._config = config.vision_config
+            self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
+            self._normalized_config.use_embed_dim = True
```
rkazants: looks the same. You can combine these two if-branches

optimum/exporters/openvino/model_configs.py
```diff
@@ -3489,6 +3641,180 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
...
+            )
+        if behavior == Qwen3VLConfigBehavior.VISION_EMBEDDINGS_POS:
+            return self.__class__(
+                self._orig_config,
+                task=self.task,
+                int_dtype=self.int_dtype,
+                float_dtype=self.float_dtype,
+                behavior=behavior,
+                preprocessors=self._preprocessors,
+            )
```
rkazants: please combine it into the single if condition

optimum/exporters/openvino/model_configs.py
```diff
@@ -3489,6 +3641,180 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
...
+
+    @property
+    def outputs(self) -> Dict[str, Dict[int, str]]:
+        if self._behavior == Qwen3VLConfigBehavior.VISION_EMBEDDINGS:
+            return {"last_hidden_state": {0: "seq_len"}}
+        if self._behavior == Qwen3VLConfigBehavior.VISION_EMBEDDINGS_MERGER:
+            return {"last_hidden_state": {0: "seq_len"}, "deepstack_feature_lists": {0: "seq_len"}}
+        if self._behavior == Qwen3VLConfigBehavior.VISION_EMBEDDINGS_POS:
+            return {"last_hidden_state": {0: "seq_len", 1: "seq_len"}}
+        return {}
```
rkazants: please define outputs for other behaviors. They are not empty, right?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4062,6 +4062,51 @@ def __exit__(self, exc_type, exc_value, traceback):
         self._model.forward = self._model.__orig_forward
 
 
+class Qwen3VLLanguageModelPatcher(OVDecoderModelPatcher):
```
rkazants, line 4065: please try to re-use `Qwen2VLLanguageModelPatcher`.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -188,11 +190,27 @@ def prepare_inputs(
             if past_len:
                 position_ids = position_ids[:, -inputs_embeds.shape[1] :]
 
-            if self.config.model_type == "qwen2_vl" and position_ids.ndim != 3:
+            if (
+                self.config.model_type == "qwen2_vl" or self.config.model_type == "qwen3_vl"
```
rkazants:
```suggestion
                self.config.model_type in ["qwen2_vl", "qwen3_vl"]
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -307,6 +313,93 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+        self,
+        input_name: str,
+        framework: str = "pt",
+        int_dtype: str = "int64",
+        float_dtype: str = "fp32",
+        bool_dtype: str = "bool",
+    ):
+        if input_name == "deepstack_visual_embeds":
+            return self.random_float_tensor(
+                [3, 2 * self.sequence_length, self.embed_dim], framework=framework, dtype=float_dtype
```
rkazants: let us explicitly retrieve 3 as:
num_layers = len(self.config.vision_config.deepstack_visual_indexes)

optimum/exporters/openvino/model_configs.py
popovaan: Done.

optimum/exporters/openvino/model_configs.py
popovaan: Done.

optimum/exporters/openvino/model_configs.py
popovaan: These fields are not used, as for text model and text embeddings outputs specification is taken from export model config in class `LMInputEmbedsConfigHelper`, but I can define them here as well to make this class complete.

popovaan: Done.

optimum/exporters/openvino/model_configs.py
popovaan: Done.

optimum/exporters/openvino/model_patcher.py
popovaan: I tried to reuse it, but the signature of the forward method doesn’t match `Qwen2VL`, so the patcher can’t be reused as is. I also tried to derive `Qwen3VL` patcher from `Qwen2VL`, but the initialization of the `Qwen2VL` patcher breaks the rest of pipeline, because it sets an incorrect patched forward method. And we can't avoid this initialization unless we rewrite `Qwen2VL` patcher somehow and move setting of patched method outside of init.

Generally the matching code in these patchers is only the wrapping of `past_key_values` into `DynamicCache`, but we can’t avoid a separate patcher either way because the forward signatures differ and the patched methods are in different parts of models (`model.forward` vs `model.language_model.forward`). So since implementing a common `DynamicCache` wrapper doesn’t seem trivial, I suggest leaving this patcher as it is unless there are examples of similar logic or you have a suggestion on how to handle this.

rkazants (approved):

optimum/exporters/openvino/model_configs.py
```diff
@@ -3328,6 +3422,65 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+
+    def __init__(
+        self,
+        task: str,
+        normalized_config: NormalizedVisionConfig,
+        batch_size: int = 1,
+        num_channels: int = DEFAULT_DUMMY_SHAPES["num_channels"],
+        width: int = 420,
+        height: int = 420,
+        **kwargs,
```
IlyasMoutawwakil: why so big ? doesn't this impact export speed ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3420,6 +3507,572 @@ def _update_model_kwargs_for_generation(
...
+                "pixel_values": pixel_values,
+                "pixel_values_videos": pixel_values_videos,
+                "image_grid_thw": image_grid_thw,
+                "video_grid_thw": video_grid_thw,
+                "cache_position": cache_position,
+            }
+        )
+        return model_inputs
+
+    # Copied from https://github.com/huggingface/transformers/blob/8ac2b916b042b1f78b75c9eb941c0f5d2cdd8e10/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L916
```
IlyasMoutawwakil: this comment is for all the copied methods, can't we take them directly from transformers the same way you did above with the rotary embedding class ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3400,6 +3437,56 @@ def preprocess_inputs(
...
+            quantization_config=quantization_config,
+            **kwargs,
+        )
+        self.rope_deltas = None  # cache rope_deltas here
+
+        if is_transformers_version(">=", "4.57.0"):
+            from transformers.models.qwen3_vl.modeling_qwen3_vl import Qwen3VLVisionRotaryEmbedding
+            from transformers.models.qwen3_vl.modeling_qwen3_vl import (
+                Qwen3VLVisionRotaryEmbedding as VisionRotaryEmbedding,
+            )
```
IlyasMoutawwakil: same class, two names, why ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3400,6 +3437,56 @@ def preprocess_inputs(
...
+            self.num_grid_per_side = int(config.vision_config.num_position_embeddings**0.5)
+            self.spatial_merge_size = config.vision_config.spatial_merge_size
+            head_dim = config.vision_config.hidden_size // config.vision_config.num_heads
+            self.rotary_pos_emb = Qwen3VLVisionRotaryEmbedding(head_dim // 2)
+
+        else:
+            raise ValueError(
+                f"Initialization model for {self.config.model_type} required at least transformers >= 4.57"
+            )
```
IlyasMoutawwakil: instead of 

```python
if ..:
     do
else: 
     raise
```

let's rather

```python
if .. :
    raise

do
```

optimum/intel/openvino/modeling_visual_language.py
IlyasMoutawwakil: anther way to get them is through inheritance, see whisper in https://github.com/huggingface/optimum-intel/blob/ea1f56de55714453dfabe1596a55db3b2bbbc32f/optimum/intel/openvino/modeling_seq2seq.py#L1301

popovaan: I thought about taking these methoud directly from transformers, but I saw that for qwen2vl, get_rope_index() method was copied here https://github.com/huggingface/optimum-intel/pull/1319/files in the PR with support of a particular transformers version.
So as I understood it was needed to keep a fixed implementation in our code as in newer transformers  may change something. But if you think that the better strategy is to import them, I surely will do it.

IlyasMoutawwakil: better use the inherited one because in our testing we compare a model that exported with a transformers version to the openvino one running on the same version, these functions also get fixed sometimes so i'm more inclined to using the latest one unless broken.

optimum/exporters/openvino/model_configs.py
popovaan: Yes, it indeed impacts the speed, fixed.

optimum/intel/openvino/modeling_visual_language.py
popovaan: No particular reason, seems to be an artifact from earlier versions. Removed it.

optimum/exporters/openvino/model_configs.py
```diff
@@ -171,6 +173,10 @@ def init_model_configs():
         "transformers",
         "AutoModelForImageTextToText",
     )
+    TasksManager._CUSTOM_CLASSES[("pt", "qwen3_vl", "image-text-to-text")] = (
+        "transformers",
+        "AutoModelForImageTextToText",
+    )
```
```suggestion
```
`AutoModelForImageTextToText` should already be default for the "image-text-to-text" task no ? https://github.com/huggingface/optimum/blob/114647f51404f9747410afc287a8079f1f16389b/optimum/exporters/tasks.py#L157

optimum/exporters/openvino/model_configs.py
```diff
@@ -307,6 +313,94 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+    def generate(
+        self,
+        input_name: str,
+        framework: str = "pt",
+        int_dtype: str = "int64",
+        float_dtype: str = "fp32",
+        bool_dtype: str = "bool",
+    ):
+        if input_name == "deepstack_visual_embeds":
+            num_layers = len(self.normalized_config.deepstack_visual_indexes)
```
echarlaix: why not ? is it because `deepstack_visual_embeds`  is not always needed as an input and `deepstack_visual_indexes` not always defined in the config ? 
```suggestion
        self.embed_dim = normalized_config.hidden_size
        self.num_layers = len(self.normalized_config.deepstack_visual_indexes)

    def generate(
        self,
        input_name: str,
        framework: str = "pt",
        int_dtype: str = "int64",
        float_dtype: str = "fp32",
        bool_dtype: str = "bool",
    ):
        if input_name == "deepstack_visual_embeds":
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -307,6 +313,94 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        common_inputs = super().inputs
+        common_inputs["visual_pos_masks"] = {0: "batch_size", 1: "sequence_length"}
+        common_inputs["deepstack_visual_embeds"] = {0: "num_layers", 1: "visual_seqlen"}
+        return common_inputs
+
+    def patch_model_for_export(
+        self, model: Union["PreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ) -> "ModelPatcher":
+        return OVDecoderModelPatcher(self, model, model_kwargs=model_kwargs)
```
```suggestion
    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
    _MODEL_PATCHER = OVDecoderModelPatcher

    @property
    def inputs(self) -> Dict[str, Dict[int, str]]:
        common_inputs = super().inputs
        common_inputs["visual_pos_masks"] = {0: "batch_size", 1: "sequence_length"}
        common_inputs["deepstack_visual_embeds"] = {0: "num_layers", 1: "visual_seqlen"}
        return common_inputs
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -3328,6 +3422,65 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+    )
+
+    def __init__(
+        self,
+        task: str,
+        normalized_config: NormalizedVisionConfig,
+        batch_size: int = 1,
+        num_channels: int = DEFAULT_DUMMY_SHAPES["num_channels"],
+        width: int = DEFAULT_DUMMY_SHAPES["width"],
+        height: int = DEFAULT_DUMMY_SHAPES["height"],
```
echarlaix: why not inherit from https://github.com/huggingface/optimum-intel/blob/ea1f56de55714453dfabe1596a55db3b2bbbc32f/optimum/exporters/openvino/model_configs.py#L3266 directly ? (and keep the given default values for `width` / `height`)

optimum/exporters/openvino/model_configs.py
```diff
@@ -3328,6 +3422,65 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+            return self.random_mask_tensor(
+                [1, grid_t * grid_h * grid_w, grid_t * grid_h * grid_w], framework=framework, dtype=float_dtype
+            )
+
+        if input_name == "rotary_pos_emb":
+            dim = self.embed_dim // self.num_heads // 2
+            return self.random_float_tensor([grid_h * grid_t * grid_w, dim], framework=framework, dtype=float_dtype)
+
+        if input_name == "input":
+            return self.constant_tensor([4, 2520], framework=framework, value=0, dtype=DTYPE_MAPPER.pt(int_dtype))
```
echarlaix: where does 4 and 2520 come from ? any chance these values could change depending on the model ? (in which case we should use the config to determine both)

optimum/exporters/openvino/model_configs.py
```diff
@@ -3489,6 +3642,175 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
...
+    TEXT_EMBEDDINGS = "text_embeddings"
+    VISION_EMBEDDINGS_POS = "vision_embeddings_pos"
+
+
+@register_in_tasks_manager(
+    "qwen3_vl",
+    *["image-text-to-text"],
+    library_name="transformers",
+)
+class Qwen3VLOpenVINOConfig(BaseVLMOpenVINOConfig):
```
echarlaix: why not inherit from `Qwen2VLOpenVINOConfig` ? https://github.com/huggingface/optimum-intel/blob/ea1f56de55714453dfabe1596a55db3b2bbbc32f/optimum/exporters/openvino/model_configs.py#L3340

optimum/intel/openvino/modeling_visual_language.py
popovaan: Done.

optimum/intel/openvino/modeling_visual_language.py
popovaan: Removed this line, as the whole class is now defined only if transformers are 4.57 or higher.

optimum/exporters/openvino/model_configs.py
popovaan: `4` is a constant that represents `floor(h), ceil(h), floor(w), ceil(w)`, where `h,w` are interpolated height and width of feature shapes for each image or video frame. A list of these values is passed to the "vision_embeddings_pos" model, which adds embeddings to them.
So `4` value remains constant as it is a part of hw-interpolation: https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L686-L687

Regarding `2520`, it is the length of this list of visual features. This dimension should be dynamic, as it depends on size of video or number of images. It is defined as dynamic in 3795. 
So the value `2520` in this dummy input should not affect the resulting IR, since the corresponding dimension becomes dynamic. But I think, I should better set a smaller value to reduce memory consumption during the conversion here.

popovaan: Changed `2025` to `DEFAULT_DUMMY_SHAPES["sequence_length"]`

optimum/exporters/openvino/model_configs.py
popovaan: No, it can be done this way, I applied the suggestion.

optimum/exporters/openvino/model_configs.py
popovaan: Agree, done.

optimum/exporters/openvino/model_configs.py
popovaan: Agree. I also reused `QwenVLConfigBehavior` and made this class common. It allowed to reuse a bit more functionality from `Qwen2VLOpenVINOConfig`.

tests/openvino/test_seq2seq.py
```diff
@@ -596,6 +599,10 @@ def get_transformer_model_class(self, model_arch):
             from transformers import Qwen2VLForConditionalGeneration
 
             return Qwen2VLForConditionalGeneration
+        if model_arch == "qwen3_vl":
+            from transformers import Qwen3VLForConditionalGeneration
+
+            return Qwen3VLForConditionalGeneration
```
echarlaix: we should be able to load with `AutoModelForImageTextToText` directly no ?

tests/openvino/test_seq2seq.py
```diff
@@ -691,7 +698,7 @@ def test_compare_to_transformers(self, model_arch):
         self._check_device_and_request(ov_model, test_device, False)
 
         # pytorch minicpmv and internvl_chat are not designed to be used via forward
-        if model_arch not in ["minicpmv", "minicpmo", "internvl_chat"]:
+        if model_arch not in ["minicpmv", "minicpmo", "internvl_chat", "qwen3_vl"]:
```
echarlaix: why removed here ?

tests/openvino/test_quantization.py
```diff
@@ -1205,11 +1257,11 @@ class OVWeightCompressionTest(unittest.TestCase):
...
         expected = set()
         if is_transformers_version("<", "4.49"):
-            expected.add("qwen2_5_vl")
+            expected.update({"qwen3_vl", "qwen2_5_vl"})
         if is_transformers_version("<", "4.51"):
-            expected.add("llama4")
+            expected.update({"qwen3_vl", "llama4"})
         if is_transformers_version("<", "4.54"):
-            expected.add("exaone4")
+            expected.update({"qwen3_vl", "exaone4"})
```
echarlaix: should just be added if `is_transformers_version("<", "4.57")`  no ?

optimum/exporters/openvino/model_configs.py
```diff
@@ -3575,11 +3697,143 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+            self._config = config.vision_config
+            self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
+            self._normalized_config.use_embed_dim = False
+        if self._behavior in (
+            QwenVLConfigBehavior.VISION_EMBEDDINGS_MERGER,
+            QwenVLConfigBehavior.VISION_EMBEDDINGS_POS,
+        ) and hasattr(config, "vision_config"):
+            self._config = config.vision_config
+            self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
+            self._normalized_config.use_embed_dim = True
```
echarlaix: not equivalent to ?
```suggestion
        if self._behavior == QwenVLConfigBehavior.VISION_EMBEDDINGS_POS and hasattr(config, "vision_config"):
            self._config = config.vision_config
            self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
            self._normalized_config.use_embed_dim = True
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -388,6 +390,90 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+                shape=[self.batch_size, self.sequence_length],
+                framework=framework,
+                value=1,
+                dtype=DTYPE_MAPPER.pt(bool_dtype),
+            )
+        return super().generate(input_name, framework, int_dtype, float_dtype)
+
+
+@register_in_tasks_manager(
+    "qwen3_vl_text",
```
echarlaix, line 454: could you extend on why we need this config (instead of adding any modifications to the qwen3_vl config, depending on `_behavior`) ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4062,6 +4062,51 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    def __init__(
+        self,
+        config: "OnnxConfig",
+        model: Union["PreTrainedModel"],
+        model_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        # Adopted from https://github.com/huggingface/transformers/blob/v4.51.3/src/transformers/models/phi4_multimodal/modeling_phi4_multimodal.py#L2156-L2178
+        # moved audio and vision features processing outside model
```
echarlaix, line 4073: would you mind adding a link to the original model as well ? 
https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L1344-L1362

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4276,6 +4321,47 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+class Qwen3VLVisionEmbMergerPatcher(ModelPatcher):
+    def __init__(
+        self,
+        config: "OnnxConfig",
+        model: Union["PreTrainedModel"],
+        model_kwargs: Dict[str, Any] = None,
+    ):
+        model.__orig_forward = model.forward
+
+        # Modified from https://github.com/huggingface/transformers/blob/v4.45.2/src/transformers/models/qwen2_vl/modeling_qwen2_vl.py#L1118
```
echarlaix, line 4334: would you mind adding a link to the original modeling code ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3384,25 +3421,415 @@ def preprocess_inputs(
...
+                prefix_allowed_tokens_fn,
+                synced_gpus,
+                assistant_model,
+                streamer,
+                negative_prompt_ids,
+                negative_prompt_attention_mask,
+                use_model_defaults,
+                custom_generate,
+                **kwargs,
+            )
```
echarlaix: would instead have 



```suggestion
        def generate(self, *args, **kwargs):
            # Clear cached rope delta from previous generations
            self.rope_deltas = None

            return super().generate((*args, **kwargs)
```

to avoid any potential issue if transformers generate signature changes between two versions 
like https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/generation/utils.py#L2234
and https://github.com/huggingface/transformers/blob/v5.0.0/src/transformers/generation/utils.py#L2271

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3384,25 +3421,415 @@ def preprocess_inputs(
...
+            assistant_model=None,
+            streamer: Optional = None,
+            negative_prompt_ids: Optional[torch.Tensor] = None,
+            negative_prompt_attention_mask: Optional[torch.Tensor] = None,
+            use_model_defaults: Optional[bool] = None,
+            custom_generate=None,
+            **kwargs,
+        ):
+            # Clear cached rope delta from previous generations
+            self.rope_deltas = None
```
echarlaix: not directly related to this PR but what if the models forward is called multiple times (not through `generate`) ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -188,11 +190,25 @@ def prepare_inputs(
...
+            else:
+                inputs["visual_pos_masks"] = torch.zeros(1, 1, dtype=torch.bool)
+
+        if "deepstack_visual_embeds" in self.input_names:
+            num_layers = len(self.config.vision_config.deepstack_visual_indexes)
+            emd_dim = self.config.text_config.hidden_size
+            if deepstack_visual_embeds is not None:
+                inputs["deepstack_visual_embeds"] = torch.Tensor(deepstack_visual_embeds)
+            else:
+                inputs["deepstack_visual_embeds"] = torch.zeros((num_layers, 1, emd_dim), dtype=torch.float32)
```
echarlaix: would replace with 
```suggestion
            if deepstack_visual_embeds is not None:
                inputs["deepstack_visual_embeds"] = torch.Tensor(deepstack_visual_embeds)
            else:
                num_layers = len(self.config.vision_config.deepstack_visual_indexes)
                emd_dim = self.config.text_config.hidden_size
                inputs["deepstack_visual_embeds"] = torch.zeros((num_layers, 1, emd_dim), dtype=torch.float32)
```

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3384,25 +3421,415 @@ def preprocess_inputs(
...
+                pos_embed = (
+                    pos_embed.view(t, h // merge_size, merge_size, w // merge_size, merge_size, -1)
+                    .permute(0, 1, 3, 2, 4, 5)
+                    .flatten(0, 4)
+                )
+                patch_pos_embeds_permute.append(pos_embed)
+            patch_pos_embeds = torch.cat(patch_pos_embeds_permute)
+            return patch_pos_embeds
+
+        def get_vision_embeddings(self, pixel_values, grid_thw, **kwargs):
```
echarlaix: would you mind adding link to the original code https://github.com/huggingface/transformers/blob/v5.0.0/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L771 (for all methods)

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3384,25 +3421,415 @@ def preprocess_inputs(
...
-        if getattr(outputs, "rope_deltas", None) is not None:
-            model_kwargs["rope_deltas"] = outputs.rope_deltas
+if is_transformers_version(">=", "4.57.0"):
+    from transformers.models.qwen3_vl.modeling_qwen3_vl import (
+        Qwen3VLModel,
+        Qwen3VLVisionModel,
+        Qwen3VLVisionRotaryEmbedding,
+    )
+
+    class _OVQwen3VLForCausalLM(OVModelForVisualCausalLM, Qwen3VLModel, Qwen3VLVisionModel):
```
echarlaix: would instead
```suggestion
if is_transformers_version(">=", "4.57.0"):
    from transformers.models.qwen3_vl.modeling_qwen3_vl import Qwen3VLModel, Qwen3VLVisionModel
else:
    Qwen3VLModel = object
    Qwen3VLVisionModel = object

class _OVQwen3VLForCausalLM(OVModelForVisualCausalLM, Qwen3VLModel, Qwen3VLVisionModel):
```

and raise an exception if an instance of `_OVQwen3VLForCausalLM` is created and transformers < 4.57

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4405,3 +4832,6 @@ def preprocess_inputs(
     "llama4": _OVLlama4ForCausalLM,
```
echarlaix, line 4821:
```suggestion
    "llama4": _OVLlama4ForCausalLM,
    "qwen3_vl": _OVQwen3VLForCausalLM,
```

tests/openvino/test_seq2seq.py
popovaan: Looks like an artifact from older versions, removed this change, it's not needed.

tests/openvino/test_seq2seq.py
popovaan: Yes, it works. Used `AutoModelForImageTextToText` here.

tests/openvino/test_quantization.py
popovaan: Agree, done.

optimum/exporters/openvino/model_patcher.py
popovaan: Yes, done.

optimum/exporters/openvino/model_patcher.py
popovaan: Done.

optimum/intel/openvino/modeling_visual_language.py
popovaan: Done.

optimum/exporters/openvino/model_configs.py
popovaan: Yes, applied.

optimum/intel/openvino/modeling_visual_language.py
popovaan: Applied.

optimum/intel/openvino/modeling_visual_language.py
popovaan: Applied.

optimum/intel/openvino/modeling_visual_language.py
popovaan: Applied.

optimum/intel/openvino/modeling_visual_language.py
popovaan: Applied, but replaced "`Qwen3VLModel = object`" and "`Qwen3VLVisionModel = object`" with creating of dummy empty classes, as usage of two `object`'s as parent classes triggers error:
`E TypeError: duplicate base class object`

optimum/exporters/openvino/model_configs.py
popovaan: If we want to avoid a separate text config, the qwen3_vl config should support handling of past key values. Simple inheritance from `TextDecoderWithPositionIdsOnnxConfig `does not solve this issue, as the config still behaves like a visual-language config and the past key values functionality is not triggered for some reason.

I tried several approaches to work around this, but none of them work properly. If avoiding a separate config is crucial, I suggest making this in a separate PR, as this model is awaited by the customer and it seems to be not trivial.

optimum/intel/openvino/modeling_visual_language.py
popovaan: I suppose accuracy will be low, as `rope_deltas` need to be cleared before each forward.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3384,25 +3421,397 @@ def preprocess_inputs(
...
+else:
+
+    class Qwen3VLModel:
+        pass
+
+    class Qwen3VLVisionModel:
+        pass
+
+
+class _OVQwen3VLForCausalLM(OVModelForVisualCausalLM, Qwen3VLModel, Qwen3VLVisionModel):
```
echarlaix, line 3447: would you mind adding a comment to explain why we need to inherit from `Qwen3VLModel` and `Qwen3VLVisionModel` respectively (+ link to original methods like https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L603)

popovaan: Done.

PR #1554, opened 2025-12-12
[OpenVINO] Revert NNCF version check

nikita-savelyevv:
# What does this PR do?

Revert NNCF version check to be a warning until a proper solution is found. 

The current problem is that when NNCF is manually installed from the release branch, i.e. `pip install git+https://github.com/openvinotoolkit/nncf.git@release_v2190`, the version becomes `2.19.0.dev0+8a9d5871e` and it does not pass the condition. We also can't truncate the version because it will allow installing from any intermediate commit between NNCF 2.18 and 2.19 (in NNCF we raise version on the develop branch to the next one right after the release).

One solution is to do `if is_nncf_version("<", "2.19") and not is_nncf_version("==", "2.19.0.dev0+8a9d5871e"):`, but it looks hacky. Perhaps, we'll make changes on NNCF side to improve on this.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/__init__.py
+++ b/optimum/intel/openvino/__init__.py
@@ -52,9 +52,17 @@ from .configuration import (
 )
 
 
+logger = logging.getLogger(__name__)
+
 if is_nncf_available():
     logging.disable(logging.INFO)
     import nncf
+
+    if is_nncf_version("<", "2.19"):
+        logger.warning(
+            "NNCF version 2.19 or higher is required to use NNCF-based quantization. "
+            f"Please upgrade your NNCF installation. The current version of NNCF is {nncf.__version__}."
+        )
 
     logging.disable(logging.NOTSET)
 


rkazants (approved):

optimum/intel/openvino/__init__.py
```diff
@@ -56,12 +56,6 @@
     logging.disable(logging.INFO)
     import nncf
```
echarlaix, line 60: why not replace with    
```suggestion
    if not is_nncf_version(">", "2.18"):
```

nikita-savelyevv: This would allow NNCF versions like `2.19.0.dev0...` which is not desired. As described at the title, "We also can't truncate the version because it will allow installing from any intermediate commit between NNCF 2.18 and 2.19 (in NNCF we raise version on the develop branch to the next one right after the release).". The argument is the same here.

echarlaix: ha sorry didn't see this in the description, however I think it's better to accept any commit after nncf v2.18 that no check at all (and accepting incompatible nncf versions which might result in errors) no ? In any case I think we can still keep a warning

nikita-savelyevv: Let's keep it as warning

PR #1556, opened 2025-12-15
[OpenVINO] Add model-specific quantization ignored scopes

nikita-savelyevv:
# What does this PR do?

- Added a configuration dict `_DEFAULT_IGNORED_SCOPE_CONFIGS` for storing default quantization ignored scope per model id. Such ignored scope will be applied irrespective of quantization method.
- Reworked `test_quantization.py::OVWeightCompressionTest.test_ovmodel_4bit_auto_compression` test to test on more model types besides `OVModelForCausalLM`.
- Added `test_quantization.py::OVWeightCompressionTest.test_ovmodel_default_ignored_scope` and `test_exporters_cli.py::OVCLIExportTestCase.test_exporters_cli_with_default_ignored_scope` to test the new default ignored scope matching logic.

Tickets CVS-175336, CVS-166099.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_diffusion.py
+++ b/optimum/intel/openvino/modeling_diffusion.py
@@ -570,6 +570,7 @@ class OVDiffusionPipeline(OVBaseModel, DiffusionPipeline):
                     f"Data-aware quantization is not supported for {cls.__name__} with "
                     f"{ov_pipeline_class.export_feature} task."
                 )
+            # export_model_id is needed because _name_or_path is not necessarily present in the model config
             model_id = export_model_id or name_or_path
             quantization_config = cls._resolve_default_quantization_config(model_id, quantization_config)
             ov_pipeline._apply_quantization(
@@ -648,7 +649,7 @@ class OVDiffusionPipeline(OVBaseModel, DiffusionPipeline):
             quantization_config=quantization_config,
             load_in_8bit=load_in_8bit,
             compile_only=compile_only,
-            export_model_id=model_id,
+            export_model_id=model_id,  # needed to resolve default quantization config during export
             **kwargs,
         )
 


optimum/intel/openvino/configuration.py
```diff
@@ -537,6 +554,68 @@ def get_default_quantization_config(
...
+    quantization_config_copy = None
+    model_id_candidates = _get_model_id_candidates(model_id_or_path)
+    for model_id in model_id_candidates:
+        default_ignored_scopes_per_model = _DEFAULT_IGNORED_SCOPE_CONFIGS.get(model_id)
+        if not default_ignored_scopes_per_model:
+            continue
+        quantization_config_copy = quantization_config.clone()
+        for ov_model_name, default_ignored_scope in default_ignored_scopes_per_model.items():
+            q_config = quantization_config_copy.quantization_configs.get(
+                ov_model_name, quantization_config_copy.default_config
```
echarlaix: we could add a check to make sure `quantization_config` is in instance of `OVPipelineQuantizationConfig` (to avoid any potential issue here)

optimum/intel/openvino/modeling_base.py
```diff
@@ -507,10 +511,9 @@ def _from_pretrained(
         )
 
         if quantization_config:
-            quantization_config = cls._resolve_default_quantization_config(str(model_id), quantization_config)
-            model._apply_quantization(
-                quantization_config, compile_only, compile_model, str(model_id), trust_remote_code
-            )
+            model_id = export_model_id or getattr(config, "name_or_path", model_id)
```
echarlaix: why use `export_model_id` instead of directly using the config ? (also here looks like  `export_model_id` will be used later on to load the config anyway to get `config.name_or_path` to add as a candidate)
```suggestion
            model_id = getattr(config, "name_or_path", model_id)
```

likely related to,  

> This is needed for more robust matching of default quantization configuration / ignored scope when exporting a model via OVModelForX.from_pretrained(model_id, export=True)

would you mind developing ?

optimum/intel/openvino/configuration.py
```diff
@@ -537,6 +554,68 @@ def get_default_quantization_config(
...
+    """
+    Applies default ignored scope configuration to the given quantization configuration based on the model ID or path.
+    Args:
+        model_id_or_path (`str`):
+            id of the model or path to it.
+        quantization_config (`OVPipelineQuantizationConfig`):
+            The quantization configuration to which the default ignored scope will be applied.
+    Returns:
+        Updated quantization configuration with the default ignored scope applied.
+    """
```
echarlaix, line 593: we could add a check to make sure `quantization_config` is in instance of `OVPipelineQuantizationConfig`

tests/openvino/test_exporters_cli.py
```diff
@@ -1261,6 +1271,81 @@ def test_exporters_cli_with_default_config(
...
+        auto_model_cls,
+        ov_model_cls,
+        options,
+    ):
+        with TemporaryDirectory() as tmpdir:
+            tmpdir = "tmp"
+            # 1. Create a local copy of the model so that we can override _name_or_path
+            pt_model = auto_model_cls.from_pretrained(MODEL_NAMES[test_model_id])
+            pt_model.save_pretrained(tmpdir)
+            try:
```
echarlaix: here you could use instead `maybe_save_preprocessors`  https://github.com/huggingface/optimum/blob/0227a1ce9652b1b02da5a510bf513c585608f8c2/optimum/utils/save_utils.py#L65

optimum/intel/openvino/configuration.py
nikita-savelyevv: Added

tests/openvino/test_exporters_cli.py
nikita-savelyevv: Good idea! Added

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: Thanks for paying attention! I had some issues with VLMs and diffusers because they did not contain proper model id inside `config.name_or_path`. My understanding was that there is no guarantee that this field is correctly defined. But from my experiments I now see that for transformers model it is.

The issue I had with VLMs came from the fact that config was re-initialized after export: https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_visual_language.py#L681 
That's why `config.name_or_path` contained `/tmp/X`-like string. I do not see any reason to re-initialize config, so I removed this line. **Update:** config reinitialization seems to be needed after all. I now set `name_or_path` manually instead.

Regarding diffusers models, it seems like configs there don't necessarily contain `name_or_path` (or `_name_or_path`) property, at least this is the case for `stabilityai/stable-diffusion-3.5-medium`. So I think we have to keep `export_model_id` argument for diffusers.

optimum/intel/openvino/configuration.py
echarlaix: thanks!

optimum/intel/openvino/modeling_base.py
echarlaix: thanks for the explanation, works for me! would you mind adding a comment just to mention that `export_model_id` is needed for diffusers models as their config don't necessarily have `_name_or_path`?

nikita-savelyevv: Added

PR #1559, opened 2025-12-15
[OpenVINO] Save OV models immediately after quantization via CLI

nikita-savelyevv:
# What does this PR do?

Changes:
- Introduced `OVModelHostMixin._all_ov_model_paths` -- a class-level property containing a dictionary of all OpenVINO model file paths that could be present in the pipeline and `OVModelHostMixin._ov_model_paths` -- a similar property, but it is instance-level and contains only the entries which are actually present in the model. Such distinction is needed because there could be differences between these two dicts for some pipelines. For example for diffusion models, `_all_ov_model_paths` will contain both `unet` and `transformer` keys, but `_ov_model_paths` will contain only the one which is actually present for that particular model instance.
- Added `immediate_save` argument to `OVQuantizer.quantize()` method. When it is `True`, after each submodel is quantized, it is immediately saved to `save_directory`. The whole model won't be saved afterwards, so it is assumed that full precision model is already present at the target location. This logic relies on the introduced `_ov_model_paths` property. To be able to do this, each openvino model has to be deleted from memory beforehand, so the quantized model won't be usable after such quantization. This is ok though for the export + save CLI usage path.
- Removed a step of saving quantized to a temporary directory inside `_main_quantize` function because with immediate saving we can save directly to the full precision model location.

These changes allow to save on RAM and disk usage. For RAM, please see the figure below for `optimum-cli export openvino -m stabilityai/stable-diffusion-3.5-large --weight-format int8 save_dir`. Disk usage is saved because there is no moment when we store a full copy of full precision model and a full copy of quantized model as it is the case with current implementation.

| Before    | After* |
| -------- | ------- |
| sd_3 5_large_main  | sd_3 5_large_main |

\* technically this was also the case for data-free quantization before #1525.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -434,8 +434,10 @@ class OVBaseModel(OptimizedModel, OVModelHostMixin):
             raise ValueError(
                 "`save_pretrained()` is not supported with `compile_only=True` mode, to save your model please initialize your model with compile_only=False"
             )
-        dst_path = os.path.join(save_directory, self._ov_model_paths["model"])
-        openvino.save_model(self.model, dst_path, compress_to_fp16=False)
+        for name, model in self.ov_models.items():
+            dst_path = os.path.join(save_directory, self._ov_model_paths[name])
+            openvino.save_model(model, dst_path, compress_to_fp16=False)
+
         generation_config = getattr(self, "generation_config", None)
         if generation_config is not None:
             try:
--- a/optimum/intel/openvino/modeling_text2speech.py
+++ b/optimum/intel/openvino/modeling_text2speech.py
@@ -274,28 +274,6 @@ class _OVModelForSpeechT5ForTextToSpeech(OVModelForTextToSpeechSeq2Seq):
     def ov_models(self) -> Dict[str, openvino.Model]:
         return {name: getattr(component, "model") for name, component in self.components.items()}
 
-    def _save_pretrained(self, save_directory: Union[str, Path]):
-        """
-        Saves the model to the OpenVINO IR format so that it can be re-loaded using the
-        [`~optimum.intel.openvino.modeling.OVModel.from_pretrained`] class method.
-
-        Arguments:
-            save_directory (`str` or `Path`):
-                The directory where to save the model files.
-        """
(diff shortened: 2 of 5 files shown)


optimum/intel/openvino/modeling_base.py
```diff
@@ -65,6 +66,19 @@ class OVModelHostMixin:
     Mixin class for models that contain OpenVINO models as submodels.
     """
 
+    @classproperty
+    def all_ov_model_paths(cls) -> Dict[str, str]:
```
echarlaix: for internal use so would add an underscore as prefix so that we can add any modification in the future (+ same comment for `unload_ov_model` or even `ov_model_paths`)
```suggestion
    def _all_ov_model_paths(cls) -> Dict[str, str]:
```

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -277,17 +283,9 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
...
-            self.OV_ENCODER_MODEL_NAME,
-            self.OV_DECODER_MODEL_NAME,
-            self.OV_POSTNET_MODEL_NAME,
-            self.OV_VOCODER_MODEL_NAME,
-        ]
-
-        for src_model, dst_file_name in zip(src_models, dst_file_names):
-            dst_path = os.path.join(save_directory, dst_file_name)
-            openvino.save_model(src_model, dst_path, compress_to_fp16=False)
+        for ov_model_name, ov_model_path in self.ov_model_paths.items():
```
echarlaix: do we need to overwrite `_save_pretrained` for every modeling now that we have `self.ov_model_paths` ?

optimum/intel/openvino/modeling_base.py
nikita-savelyevv: Renamed

optimum/intel/openvino/modeling_text2speech.py
nikita-savelyevv: Thanks for the suggestion! It is indeed now possible to remove most of `_save_pretrained` definitions. I've kept a couple of custom ones for decoder and diffusion models because they contain some custom logic.

echarlaix: perfect thanks!

PR #1561, opened 2025-12-16
Fix NNCF versions comparison

skuros:
Fixes:
```
ImportError: NNCF version 2.19 or higher is required to use NNCF-based quantization. Please upgrade your NNCF installation. The current version of NNCF is 2.19.0.
```


nikita-savelyevv:
Hi @skuros! Thanks for the PR. We have already replaced the ImportError with a warning: https://github.com/huggingface/optimum-intel/pull/1554 . Does it solve the problem for you?

PR #1564, opened 2025-12-17
[OpenVINO] Introduce extended quantization dataset options like 'wikitext2:seq_len=128'

nikita-savelyevv:
# What does this PR do?

Sequence length during quantization calibration data collection can now be provided as a string like `"wikitext2:seq_len=128"`. The motivation is that sometimes it makes sense to adjust default sequence length value, and this way it will be possible to configure it in those particular cases. For example for configs inside `_DEFAULT_4BIT_WQ_CONFIGS`.

Based on PR by Copilot: https://github.com/nikita-savelyevv/optimum-intel/pull/4 

Ticket: CVS-176623

## Before submitting

### DIFF
--- a/tests/openvino/test_quantization.py
+++ b/tests/openvino/test_quantization.py
@@ -2544,8 +2544,8 @@ class TestDatasetParsing(unittest.TestCase):
         """Test that non-integer seq_len values raise ValueError."""
         with pytest.raises(ValueError) as exc_info:
             OVQuantizationConfigBase(dataset="wikitext:seq_len=abc")
-        assert "Invalid value 'abc' for seq_len" in str(exc_info.value)
-        assert "Expected an integer" in str(exc_info.value)
+        self.assertIn("Invalid value 'abc' for seq_len", str(exc_info.value))
+        self.assertIn("Expected an integer", str(exc_info.value))
 
     def test_dataset_empty_string_option(self):
         """Test that empty seq_len value raises ValueError."""


optimum/intel/openvino/configuration.py
```diff
@@ -646,14 +646,55 @@ def __init__(
             num_samples (`int`, *optional*):
                 The maximum number of samples composing the calibration dataset.
             dataset (`str or List[str]`, *optional*):
-                The dataset used for data-aware optimization with NNCF.
+                The dataset used for data-aware optimization with NNCF. Can be a dataset name (e.g., 'wikitext')
+                or a string with options (e.g., 'wikitext2:seq_len=128'). Currently supported option: seq_len.
```
echarlaix: Wouldn't that be redundant with `num_sample` ? Also what if both are provided ?

nikita-savelyevv: There is a distinction between those: `num_samples` is a number of sequences in the dataset, while `seq_len` is a length of each sequence. Let me update the description to clarify what `seq_len` stands for.

optimum/intel/openvino/configuration.py
```diff
@@ -646,14 +646,56 @@ def __init__(
             num_samples (`int`, *optional*):
                 The maximum number of samples composing the calibration dataset.
             dataset (`str or List[str]`, *optional*):
-                The dataset used for data-aware optimization with NNCF.
+                The dataset used for data-aware optimization with NNCF. Can be a dataset name (e.g., 'wikitext2')
+                or a string with options (e.g., 'wikitext2:seq_len=128'). The only currently supported option is `seq_len`
```
ljaljushkin, line 650: There are some assumptions for seqlen=32 in gptq utils:
https://github.com/huggingface/optimum/blob/main/optimum/gptq/data.py#L128-L129
please check whether this code will be valid for other seq_len.

probably, it makes sense to copy paste this helper code to optimum-intel and extend for other seq_len
```python
    limit = num_samples * seqlen // 4  # ~1k for 128 samples with seqlen=32 to be aligned with optimum
    text = "".join([" \n" if s == "" else s for s in traindata["text"][:limit]])
```

nikita-savelyevv: Good catch! Added

ljaljushkin (approved):

optimum/intel/openvino/configuration.py
echarlaix: very clear, thanks @nikita-savelyevv !

tests/openvino/test_quantization.py
```diff
@@ -2456,3 +2470,97 @@ def check_model_inference(ov_model, model_id, trust_remote_code):
...
+        config = OVQuantizationConfigBase(dataset=dataset_list)
+        self.assertEqual(config.dataset, dataset_list)
+        self.assertEqual(config._dataset_kwargs, {})
+
+    def test_dataset_unsupported_option(self):
+        """Test that unsupported options raise ValueError."""
+        with pytest.raises(ValueError) as exc_info:
+            OVQuantizationConfigBase(dataset="wikitext:foo=bar")
+        assert "Unsupported dataset option 'foo'" in str(exc_info.value)
+        assert "Only 'seq_len' is supported" in str(exc_info.value)
```
echarlaix:
```suggestion
        self.assertIn("Unsupported dataset option 'foo'", str(exc_info.value))
        self.assertIn("Only 'seq_len' is supported", str(exc_info.value))
```

tests/openvino/test_quantization.py
```diff
@@ -2456,3 +2470,97 @@ def check_model_inference(ov_model, model_id, trust_remote_code):
...
+
+    def test_dataset_malformed_option_no_equals(self):
+        """Test that options without '=' raise ValueError."""
+        with pytest.raises(ValueError) as exc_info:
+            OVQuantizationConfigBase(dataset="wikitext:seq_len")
+        assert "Malformed dataset option" in str(exc_info.value)
+        assert "Expected format: 'key=value'" in str(exc_info.value)
```
echarlaix:
```suggestion
        self.assertIn("Malformed dataset option", str(exc_info.value))
        self.assertIn("Expected format: 'key=value'", str(exc_info.value))
```

tests/openvino/test_quantization.py
```diff
@@ -2456,3 +2470,97 @@ def check_model_inference(ov_model, model_id, trust_remote_code):
...
+
+    def test_dataset_invalid_seq_len_value(self):
+        """Test that non-integer seq_len values raise ValueError."""
+        with pytest.raises(ValueError) as exc_info:
+            OVQuantizationConfigBase(dataset="wikitext:seq_len=abc")
+        assert "Invalid value 'abc' for seq_len" in str(exc_info.value)
+        assert "Expected an integer" in str(exc_info.value)
```
echarlaix:
```suggestion
        self.assertIn("Invalid value 'abc' for seq_len", str(exc_info.value))
        self.assertIn("Expected an integer", str(exc_info.value))
```

tests/openvino/test_quantization.py
```diff
@@ -2456,3 +2470,97 @@ def check_model_inference(ov_model, model_id, trust_remote_code):
...
+
+    def test_dataset_empty_string_option(self):
+        """Test that empty seq_len value raises ValueError."""
+        with pytest.raises(ValueError) as exc_info:
+            OVQuantizationConfigBase(dataset="wikitext:seq_len=")
+        assert "Invalid value '' for seq_len" in str(exc_info.value)
```
echarlaix:
```suggestion
        self.assertIn("Invalid value '' for seq_len", str(exc_info.value))
```

PR #1566, opened 2025-12-17
Remove IPEX and INC from index and installations documentation sections 

echarlaix:
Following discussion that INC and IPEX integrations in optimum-intel will get deprecated

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -4,14 +4,6 @@
   - local: installation
     title: Installation
   - sections:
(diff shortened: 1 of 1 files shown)


rkazants (commented):
Shall we remove Gaudi from README.md?
What is the purpose of it in optimum-intel? OpenVINO does not support Gaudi

echarlaix:
@rkazants it was added to redirect gaudi users to `optimum-habana`. Now that `optimum-habana` is in maintenance mode it can make sense to remove though cc @regisss

PR #1568, opened 2025-12-18
Add depreciation warning for INC and IPEX 

echarlaix:
Following discussion that INC and IPEX integrations in optimum-intel will get deprecated as discussed on tuesday  @yao-matrix @hshen14

### DIFF
--- a/optimum/intel/ipex/__init__.py
+++ b/optimum/intel/ipex/__init__.py
@@ -34,7 +34,7 @@ if is_sentence_transformers_available():
 
 
 warnings.warn(
-    "`optimum.intel.ipex` is deprecated and will be removed in the next major release of optimum-intel. ",
+    "`optimum.intel.ipex` is deprecated and will be removed in the next major release of `optimum-intel`.",
     FutureWarning,
     stacklevel=2,
 )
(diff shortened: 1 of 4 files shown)


optimum/intel/__init__.py
```diff
@@ -27,10 +27,9 @@
 from .version import __version__
 
 
+# Base objects, independent of any specific backend
+# TODO: add openvino once ipex and inc officially deprecated
 _import_structure = {
-    "ipex": [],
-    "neural_compressor": [],
-    "openvino": [],
```
echarlaix, line 33: removing to avoid that `from optimum.intel import *` raising a `ModuleNotFoundError` if ipex, openvino and neural_compressor are not all installed

optimum/intel/ipex/__init__.py
```diff
@@ -28,3 +31,10 @@
 
 if is_sentence_transformers_available():
     from .modeling_sentence_transformers import IPEXSentenceTransformer
+
+
+warnings.warn(
+    "`optimum.intel.ipex` is deprecated and will be removed in the next major release of `optimum-intel`.",
```
echarlaix, line 37: cc @IlyasMoutawwakil  let me know if that works for you

IlyasMoutawwakil (approved):
LGTM ! thanks ella !

PR #1569, opened 2025-12-18
[OpenVINO] Support Arcee Trinity (aka Afmoe) models collection

rkazants:
# What does this PR do?

**Note:** Currently, the model is a remote-code model for transformers "4.x.y" version.

Example of conversion cmd-line for `arcee-ai/Trinity-Nano-Preview`:

```sh
optimum-cli export openvino -m arcee-ai/Trinity-Nano-Preview Trinity-Nano-Preview --trust-remote-code
```

Example of inference for `arcee-ai/Trinity-Nano-Preview` using OpenVINO backend:

```python

model_path = "arcee-ai/Trinity-Nano-Preview"

tokenizer = AutoTokenizer.from_pretrained(model_path)
model = OVModelForCausalLM.from_pretrained(model_path, trust_remote_code=True)

# change input text as desired
input_text = "The capital of France is"
# tokenize the text
input_tokens = tokenizer(input_text, return_tensors="pt")
# generate output tokens
output = model.generate(**input_tokens, max_length=10)
# decode output tokens into text
output = tokenizer.batch_decode(output)
print(output[0])
```

## Before submitting
- [X] Did you make sure to update the documentation with your changes?
- [X] Did you write any new necessary tests?

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -7484,3 +7484,4 @@ class AfmoeModelPatcher(ModelPatcher):
             if layer.moe_enabled:
                 afmoe_moe = layer.mlp
                 afmoe_moe.forward = afmoe_moe._orig_forward
+                del afmoe_moe.down_projs, afmoe_moe.gate_projs, afmoe_moe.up_projs
--- a/tests/openvino/test_decoder.py
+++ b/tests/openvino/test_decoder.py
@@ -735,8 +735,9 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
                 f"generation config : {gen_config}, transformers output {transformers_outputs}, ov_model_stateful output {ov_stateful_outputs}",
             )
 
-            # beam search does not work for stateless afmoe
-            # won't be fixed since it is not used in OpenVINO GenAI
+            # beam search does not work for stateless afmoe:
+            # generation results for the second padded prompt in the batch is incorrect
+            # won't be fixed since stateless model is not used for OpenVINO GenAI
             if model_arch in ["afmoe"]:
                 continue
 


optimum/exporters/openvino/model_configs.py
```diff
@@ -4558,3 +4559,16 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+@register_in_tasks_manager(
+    "afmoe",
+    *[
+        "text-generation",
+        "text-generation-with-past",
+    ],
+    library_name="transformers",
+)
+class AfmoeOpenVINOConfig(LlamaOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.55.4"
```
echarlaix: Since this architecture is already supported in transformers https://github.com/huggingface/transformers/blob/v5.0.0rc0/src/transformers/models/afmoe/modeling_afmoe.py I think we shouldn't add support to the remote code version. We would also need to add support to the latest transformers version (will open a PR for this soon)

rkazants: Hi @echarlaix, we have urgent request from customer for this model. optimum-intel doesn't support v5 transformers yet. I propose to proceed with remote-code support for v4 now.

rkazants:
> thanks for the addition @rkazants there are still some failures in _export_, can you check those please.

Added some export tests. Now they should be passing.
Unfortunately. export cli tests infrastructure is not supporting remote-code model. I tried to add the test but `main_export` requires `custom_export_configs`. It is better to add such tests in a separate PR for all models.

optimum/exporters/openvino/model_configs.py
echarlaix: @rkazants what if a user export a model using transformers directly (trust_remote_code=False) once v5 is supported ?  I thinks we need to also set MAX_TRANSFORMERS_VERSION = "4.57.3" to ensure this doesn't happen

echarlaix: also why setting v4.55.4 and not v4.55.0 for the `MIN_TRANSFORMERS_VERSION` ?

rkazants:
> I observe that the export and inference takes a very long time. Could you please add the same warning as the one we've added recently for Zamba2?
> 
> Another issue is that weight quantization is very slow because there are 20k weight constants to quantize. I believe we had a similar issue for Qwen/Qwen3-30B-A3B until MoE merging was added to OpenVINO. Do you know, is it expected for the same approach to work for this model?
> 
> Also, please take a look at the failing tests.

I re-implemented the patch for MoE in a vectorized form. Conversion gets speed-up more by 4x, total conversion time is about 24 minutes. Inference time was also improved.

optimum/exporters/openvino/model_configs.py
rkazants: Done, thanks!

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7409,3 +7409,74 @@ def unpatch_sparse_moe(sparse_moe_layer):
...
+    # combine weights from all experts to get the common gate, up and down weights
+    down_projs = torch.concat(
+        tuple(self.experts[i].down_proj.weight.float().unsqueeze(0) for i in range(num_experts)), dim=0
+    ).transpose(1, 2)
+    gate_projs = torch.concat(
+        tuple(self.experts[i].gate_proj.weight.float().unsqueeze(0) for i in range(num_experts)), dim=0
+    ).transpose(1, 2)
+    up_projs = torch.concat(
+        tuple(self.experts[i].up_proj.weight.float().unsqueeze(0) for i in range(num_experts)), dim=0
+    ).transpose(1, 2)
```
IlyasMoutawwakil: you don't need to combine them at runtime, because that will result in this being traced in the IR as part of the graph. You can do it in the patching loop, attribute them to the class and use them in the forward pass directly.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7409,3 +7409,74 @@ def unpatch_sparse_moe(sparse_moe_layer):
...
+
+
+class AfmoeModelPatcher(ModelPatcher):
+    def __enter__(self):
+        super().__enter__()
+        for idx, layer in enumerate(self._model.model.layers):
+            if layer.moe_enabled:
+                afmoe_moe = layer.mlp
+                afmoe_moe._orig_forward = afmoe_moe.forward
+                afmoe_moe.forward = types.MethodType(afmoe_moe_forward_patched, afmoe_moe)
```
IlyasMoutawwakil, line 7465:
```suggestion
                afmoe_moe.forward = types.MethodType(afmoe_moe_forward_patched, afmoe_moe)
                afmoe_moe.down_proj = torch.concat(...
                afmoe_moe.gate_proj = ...
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7409,3 +7409,74 @@ def unpatch_sparse_moe(sparse_moe_layer):
...
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+        for idx, layer in enumerate(self._model.model.layers):
+            if layer.moe_enabled:
+                afmoe_moe = layer.mlp
+                afmoe_moe.forward = afmoe_moe._orig_forward
```
IlyasMoutawwakil, line 7486:
```suggestion
                afmoe_moe.forward = afmoe_moe._orig_forward
                del afmoe_moe.down_proj, afmoe_moe.gate_proj, ...
```

IlyasMoutawwakil:
@rkazants looks good for me, I left one suggestion to cache the concatenated expert weights.

optimum/exporters/openvino/model_patcher.py
rkazants: totally agree, I will do it.

rkazants: Done

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7409,3 +7409,78 @@ def unpatch_sparse_moe(sparse_moe_layer):
...
+                    tuple(afmoe_moe.experts[i].down_proj.weight.float().unsqueeze(0) for i in range(num_experts)),
+                    dim=0,
+                ).transpose(1, 2)
+                afmoe_moe.gate_projs = torch.concat(
+                    tuple(afmoe_moe.experts[i].gate_proj.weight.float().unsqueeze(0) for i in range(num_experts)),
+                    dim=0,
+                ).transpose(1, 2)
+                afmoe_moe.up_projs = torch.concat(
+                    tuple(afmoe_moe.experts[i].up_proj.weight.float().unsqueeze(0) for i in range(num_experts)), dim=0
+                ).transpose(1, 2)
```
IlyasMoutawwakil, line 7479:
```suggestion
                afmoe_moe.down_projs = torch.concat(
                    tuple(afmoe_moe.experts[i].down_proj.weight.unsqueeze(0) for i in range(num_experts)),
                    dim=0,
                ).transpose(1, 2)
                afmoe_moe.gate_projs = torch.concat(
                    tuple(afmoe_moe.experts[i].gate_proj.weight.unsqueeze(0) for i in range(num_experts)),
                    dim=0,
                ).transpose(1, 2)
                afmoe_moe.up_projs = torch.concat(
                    tuple(afmoe_moe.experts[i].up_proj.weight.unsqueeze(0) for i in range(num_experts)), dim=0
                ).transpose(1, 2)
```

docs/source/openvino/models.mdx
```diff
@@ -14,6 +14,7 @@ Here is the list of the supported architectures :
 
 ## [Transformers](https://huggingface.co/docs/transformers/index)
 
+- Afmoe (aka Arcee Trinity)
```
rkazants:
```suggestion
- Afmoe (aka Arcee  Trinity)
```

docs/source/openvino/models.mdx
```diff
@@ -14,6 +14,7 @@ Here is the list of the supported architectures :
 
 ## [Transformers](https://huggingface.co/docs/transformers/index)
 
+- Afmoe (aka Arcee  Trinity)
```
rkazants:
```suggestion
- Afmoe (aka Arcee Trinity)
```

docs/source/openvino/models.mdx
```suggestion
- AFMoE (aka Arcee Trinity)
```

tests/openvino/test_decoder.py
```diff
@@ -752,6 +735,11 @@ def test_beam_search(self, model_arch):
                 f"generation config : {gen_config}, transformers output {transformers_outputs}, ov_model_stateful output {ov_stateful_outputs}",
             )
 
+            # beam search does not work for stateless afmoe
+            # won't be fixed since it is not used in OpenVINO GenAI
```
echarlaix: can you extend on what the issue is ? also should we disable the export of stateless AFMoE or is it needed for npus ?

optimum/exporters/openvino/model_patcher.py
rkazants: Fixed.

tests/openvino/test_decoder.py
rkazants: The comment was extended. Let us not disable export for stateless because we anyway plan to remove it. For greedy it looks working.

PR #1570, opened 2025-12-18
[OpenVINO] Add default 8-bit quantization configs with configurable dynamic quantization group size

nikita-savelyevv:
# What does this PR do?

Changes:
- Deprecate `OVDynamicQuantizationConfig` in favor of extending `OVWeightQuantizationConfig` with `dq_group_size` parameter.
- If `OVWeightQuantizationConfig.dq_group_size` is set, a corresponding runtime info flag is added to a model during weight only quantization. Adding `DYNAMIC_QUANTIZATION_GROUP_SIZE` to `ov_config` is no longer needed as the value is written directly to the model IR.
- Added `_DEFAULT_8BIT_WQ_CONFIGS` similar to the current `_DEFAULT_4BIT_WQ_CONFIGS`. Model configs here may contain an additional DQ group size value.

Ticket CVS-176390

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_base.py
+++ b/optimum/intel/openvino/modeling_base.py
@@ -650,7 +650,10 @@ class OVBaseModel(OptimizedModel, OVModelHostMixin):
         """
         if quantization_config == {"bits": 4} or quantization_config == {"bits": 8}:
             # If config is given as {"bits": N}, use the default N-bit quantization config
-            if quantization_config == {"bits": 4} and model_name_or_path in ["openai/gpt-oss-20b", "openai/gpt-oss-120b"]:
+            if quantization_config == {"bits": 4} and model_name_or_path in [
+                "openai/gpt-oss-20b",
+                "openai/gpt-oss-120b",
+            ]:
                 raise NotImplementedError(
                     "Quantization with the default 4-bit config is not supported through Python API for openai/gpt-oss-20b model. "
                     "Please export the model via optimum-cli with `--weight-format int4` argument. This way the "


optimum/intel/openvino/configuration.py
```diff
@@ -493,6 +494,11 @@ class OVQuantizationMethod(str, Enum):
 }
 
 
+# Default dynamic quantization group size configs. For each model id it is a dict of `{ov_model_name: group_size}`.
+# For possible values of `ov_model_name` please take a look at `_ov_model_names` property of corresponding OVModel class.
+_DEFAULT_DYNAMIC_QUANTIZATION_GROUP_SIZE_CONFIGS = {"Qwen/Qwen2.5-Coder-3B-Instruct": {"model": 128}}
```
ljaljushkin: I'd explicitly wrote this to int4/int8 default weight compression configurations to keep all defaults in one place

nikita-savelyevv: Thanks for the suggestion! I agree, done.

optimum/intel/openvino/configuration.py
```diff
@@ -366,6 +367,9 @@ class OVQuantizationMethod(str, Enum):
     },
 }
 
+_DEFAULT_8BIT_WQ_CONFIGS = {
+    "Qwen/Qwen2.5-Coder-3B-Instruct": {"bits": 8, "sym": False, "dq_group_size": 128},
```
ljaljushkin, line 371: seems like forgot to add for the list of 4bit configs

nikita-savelyevv: There is already a 4bit config for this model, I've added `"dq_group_size": 128` for it.

ljaljushkin (approved):

optimum/intel/openvino/modeling_base.py
```diff
@@ -649,26 +648,31 @@ def _resolve_default_quantization_config(
             quantization_config (`OVQuantizationConfigBase` or `Dict`):
                 The quantization config to resolve.
         """
-        if quantization_config == {"bits": 4}:
-            # If config is given as {"bits": 4}, use the default 4-bit quantization config
+        if quantization_config == {"bits": 4} or quantization_config == {"bits": 8}:
+            # If config is given as {"bits": N}, use the default N-bit quantization config
             if model_name_or_path in ["openai/gpt-oss-20b", "openai/gpt-oss-120b"]:
```
echarlaix:
```suggestion
            if quantization_config == {"bits": 4} and model_name_or_path in ["openai/gpt-oss-20b", "openai/gpt-oss-120b"]:
```

optimum/intel/openvino/configuration.py
```diff
@@ -903,6 +917,8 @@ def post_init(self):
             raise ValueError("`ratio` must between 0 and 1.")
         if self.group_size is not None and self.group_size != -1 and self.group_size <= 0:
             raise ValueError("`group_size` must be greater than 0 or equal to -1")
+        if self.dq_group_size is not None and self.dq_group_size < 0:
+            raise ValueError("`dq_group_size` must not be less than 0")
```
echarlaix:
```suggestion
            raise ValueError(f"`dq_group_size` must not be less than 0, but found {self.dq_group_size}")
```

optimum/intel/openvino/configuration.py
```diff
@@ -892,6 +905,7 @@ def __init__(
         self.dtype = dtype
         self.statistics_path = statistics_path
         self.group_size_fallback = group_size_fallback
+        self.dq_group_size = dq_group_size
```
echarlaix, line 908: maybe `act_group_size` could be clearer here, any preference @IlyasMoutawwakil @andrey-churkin ?

PR #1578, opened 2026-01-04
Update OpenVINO tests to use tiny-random-minicpm-o-2_6 model

ahmezk11:
Summary
This PR updates the OpenVINO test suite to use the tiny-random-minicpm-o-2_6 model
instead of the optimum-intel-internal-testing model for faster testing.

Changes
- Updated `tests/openvino/utils_tests.py` line 135 to reference the new tiny model

 Model Details
- Model ID: `ahmedzk11/tiny-random-minicpm-o-2_6`
- Size: 37.7 MB (74% reduction)
- Architecture**: MiniCPMO
- Task: image-text-to-text

Related
- PR 1454: Support openbmb/MiniCPM-o-2_6 for image-text-to-text task

### DIFF
--- a/tests/openvino/utils_tests.py
+++ b/tests/openvino/utils_tests.py
@@ -132,7 +132,7 @@ MODEL_NAMES = {
     "minicpm": "optimum-intel-internal-testing/tiny-random-minicpm",
     "minicpm3": "optimum-intel-internal-testing/tiny-random-minicpm3",
     "minicpmv": "optimum-intel-internal-testing/tiny-random-minicpmv-2_6",
-    "minicpmo": "optimum-intel-internal-testing/tiny-random-MiniCPM-o-2_6",
+    "minicpmo": "ahmedzk11/tiny-random-minicpm-o-2_6",
     "mistral": "optimum-intel-internal-testing/tiny-random-mistral",
     "mistral-nemo": "optimum-intel-internal-testing/tiny-random-mistral-nemo",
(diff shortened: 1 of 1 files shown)


tests/openvino/utils_tests.py
```diff
@@ -132,7 +132,7 @@
     "minicpm": "optimum-intel-internal-testing/tiny-random-minicpm",
     "minicpm3": "optimum-intel-internal-testing/tiny-random-minicpm3",
     "minicpmv": "optimum-intel-internal-testing/tiny-random-minicpmv-2_6",
-    "minicpmo": "optimum-intel-internal-testing/tiny-random-MiniCPM-o-2_6",
+    "minicpmo": "ahmedzk11/tiny-random-minicpm-o-2_6",
```
echarlaix, line 135: thanks a lot for the PR @ahmezk11 !! Unfortunately minicpmo is a remote code models, which means that it needs to be validated as well as hosted directly under one of our organization (so that we keep full control over it as we will need to set `trust_remote_code=True`)

ahmezk11: Hi @echarlaix,

Thanks for reviewing! I understand the concern about `trust_remote_code=True`. Since the model needs to be in an organization you control for security, I have a couple of options:

1. I can transfer the model to the optimum-intel organization if you'd like to take ownership
2. Or I'm happy to re-upload it if you provide the target location/organization

What would work best for your workflow? I want to make sure this integrates smoothly with your testing infrastructure.

PR #1582, opened 2026-01-12
Support Paraformer ASR model exporting to OpenVINO.

10vliu13:
This PR would like to support Paraformer ASR model exporting to OpenVINO. The command to export to OpenVINO is showed in below:

cmd:  optimum-cli export openvino --trust-remote-code --model funasr/paraformer-zh paraformer-zh

Please have a review and comment. Thanks

### DIFF
--- a/optimum/exporters/openvino/modeling_paraformer.py
+++ b/optimum/exporters/openvino/modeling_paraformer.py
@@ -1,0 +1,2105 @@
+import torch
+from torch import nn
+import logging
+from typing import Dict, Optional, List, Tuple
+import numpy as np
+import types
+import math
+import os
+import json
+import copy
+from omegaconf import OmegaConf, DictConfig, ListConfig
+
+# Copied from https://github.com/modelscope/FunASR/blob/main/funasr/models/transformer/utils/repeat.py#L14 (Apache 2.0)
+class MultiSequential(torch.nn.Sequential):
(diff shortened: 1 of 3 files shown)


optimum/exporters/openvino/modeling_paraformer.py
```diff
@@ -0,0 +1,2105 @@
...
+        dim (int): Dimension to be normalized.
+
+    """
+
+    def __init__(self, nout, dim=-1):
+        """Construct an LayerNorm object."""
+        super(LayerNorm, self).__init__(nout, eps=1e-12)
+        self.dim = dim
+
+class BaseTransformerDecoder(nn.Module):
```
echarlaix, line 192: @10vliu13 we shouldn't add a modeling code in optimum-intel, we can instead add export support via `export_from_model` https://github.com/huggingface/optimum-intel/blob/132f70dd17d15834902f4a422541982b60ca3772/optimum/exporters/openvino/convert.py#L545
when the model is already loaded using funasr for example. If we want to show users how to export their model, adding a script should be enough (the number of downloads for this architecture seems too low to add model loading via `TasksManager.get_model_from_task`)

10vliu13: It seems SpeechT5 TTS model also a example to uses export_from_model. I will refers to SpeechT5 TTS model code flow and reimplement my patch. Thanks for the comment.

PR #1583, opened 2026-01-12
support input shapes export

danielliuce:
# What does this PR do?

Fixes Issue # 1581

Tested with command:
`optimum-cli export openvino -m microsoft/resnet-50  --batch_size 1 --num_channels 3 --height 224 --width 224 static_shapes` and looked at the OpenVINO IR to see that the input shapes were the same as specified

Tested with
`optimum-cli export openvino -m microsoft/resnet-50 testing` and looked at the OpenINO IR to see that the input shapes were -1 indicating that they were dynamic

Inspected the OpenVINO XML IR and saw that the input shapes were static and specified

## Before submitting

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -560,7 +560,7 @@ class OVExportCommand(BaseOptimumCLICommand):
                 library_name=library_name,
                 variant=self.args.variant,
                 model_kwargs=self.args.model_kwargs,
-                **input_shapes
+                **input_shapes,
             )
             if apply_main_quantize:
                 _main_quantize(


rkazants (changes requested):
This is not needed. You can always use `reshape` method for ov.Model to set static shapes, for example.

echarlaix:
Makes sense to me to leave the possibility for users to specify the input shapes to use during model conversion via the CLI, like what we have with ONNX https://github.com/huggingface/optimum-onnx/blob/v0.1.0/optimum/commands/export/onnx.py#L259 
Can be useful for users who only uses the CLI + for cases where the inputs shapes can have an impact on the resulting model (for example for [LongT5](https://github.com/huggingface/transformers/blob/v5.0.0rc2/src/transformers/models/longt5/modeling_longt5.py#L188) conversion depending on [sequence_length](https://github.com/huggingface/optimum/issues/285))

```bash
optimum-cli export openvino --model microsoft/resnet-50 --batch_size 2 --num_channels 3 --height 224 --width 224 output_dir
```

@rkazants @danielliuce what do you think ?

danielliuce:
> > For the sake of usability and maintainability, I believe we should ensure consistency across the optimum-cli utility. OpenVINO should use argument names and conventions similar to other backends like ONNX.
> > Treating OpenVINO differently introduces confusion for users, complicates documentation and automation, and makes the CLI harder to maintain. Unless there’s a strong technical reason, I recommend aligning OpenVINO’s interface with the established patterns used by other tools.
> 
> optimum-onnx is not a reference for us. `input_shapes` can be set through inference APIs of OpenVINO GenAI using `reshape` method when you know which input is expected. I think that `optimum-cli` should output generic IRs for the input model. I vote for minimal set of parameters for conversion because it mitigates problems with using IRs for incorrect input shape data, for example, and we need to establish additional regular validation for this option. Now we have a clear way to set static shape and it is validated well and is quite enough.

Thanks for the response.
Just to make sure I understand,
To summarize:
* You are proposing that users use `optimum-cli export openvino` to download a model from huggingface into OpenVINO IR with dynamic shapes, and if they know the input shapes, to write an additional Python/C++ OpenVINO snippet that converts this IR into IR with static input shapes. and then finally able to use this IR, as opposed to letting them specify the input shapes directly.
* And the reasoning for this is because there would be too much validation / testing needed for users to optionally specify the static input shapes through the CLI?

Is my understanding correct?

PR #1588, opened 2026-01-19
[OpenVINO] Eagle3 support

rkazants:
# What does this PR do?

This feature allows to export eagle3 models prepared with AngleSlim framework. The exported model then can be used for speculative decoding in OpenVINO GenAI.

**Note:** Install OpenVINO 2026.0 or higher:

```sh
pip install --pre -U openvino openvino-genai openvino-tokenizers --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
```

1. Export the draft model:

```sh
optimum-cli export openvino -m AngelSlim/Qwen3-1.7B_eagle3 Qwen3-1.7B_eagle3 --task=text-generation-with-past --trust-remote-code
```

2. Export the main model:

```sh
optimum-cli export openvino -m Qwen/Qwen3-1.7B Qwen3-1.7B
```

3. Run the speculative decoding using OpenVINO GenAI:

```python
import openvino_genai

draft_model_path = "./Qwen3-1.7B_eagle3"
main_model_path = "./Qwen3-1.7B"

ov_draft_model = openvino_genai.draft_model(draft_model_path, "CPU")
ov_eagle3_pipe = openvino_genai.LLMPipeline(main_model_path, "CPU", draft_model=ov_draft_model)

prompt = "Paris is the capital of"
genai_eagle3_output = ov_eagle3_pipe.generate(prompt, echo=True, apply_chat_template=False, ignore_eos=True, max_new_tokens=10)

print(genai_eagle3_output)
```

Performance comparison script, I got ~2x speedup for 1000 new tokens generation, for example, on 12th Gen Intel(R) Core(TM) i9-12900K:

```python
import time
import openvino_genai

# Paths
draft_model_path = "./Qwen3-1.7B_eagle3"
main_model_path = "./Qwen3-1.7B"

# Initialize models
ov_draft_model = openvino_genai.draft_model(draft_model_path, "CPU")
ov_eagle3_pipe = openvino_genai.LLMPipeline(main_model_path, "CPU", draft_model=ov_draft_model)
ov_pipe = openvino_genai.LLMPipeline(main_model_path, "CPU")

prompt = "Paris is the capital of"

# --- Eagle-3 pipeline timing ---
start_time = time.time()
genai_eagle3_output = ov_eagle3_pipe.generate(
    prompt,
    echo=True,
    apply_chat_template=False,
    ignore_eos=True,
    max_new_tokens=1000
)
eagle3_time = time.time() - start_time
#print("Eagle-3 pipeline output:", genai_eagle3_output)
print(f"Eagle-3 generation time: {eagle3_time:.4f} seconds")

# --- Regular pipeline timing ---
start_time = time.time()
genai_regular_output = ov_pipe.generate(
    prompt,
    echo=True,
    apply_chat_template=False,
    ignore_eos=True,
    max_new_tokens=1000
)
regular_time = time.time() - start_time
#print("Regular pipeline output:", genai_regular_output)
print(f"Regular generation time: {regular_time:.4f} seconds")
```

## Before submitting

### DIFF
--- a/tests/openvino/test_decoder.py
+++ b/tests/openvino/test_decoder.py
@@ -840,7 +840,6 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
     @parameterized.expand(EAGLE3_MODELS.items())
     @pytest.mark.skipif(is_transformers_version("<", "4.54"), reason="Eagle3 requires transformers >= 4.54")
     def test_load_and_infer_with_eagle3_model(self, model_arch, model_pair):
-
         draft_model_id, target_model_id = model_pair
 
         ov_model = OVModelForCausalLM.from_pretrained(draft_model_id, export=True, trust_remote_code=True)


optimum/intel/openvino/configuration.py
```diff
@@ -691,13 +691,11 @@ def __init__(
                         self._dataset_kwargs[key] = int(value)
                     except ValueError:
                         raise ValueError(
-                            f"Invalid value '{value}' for seq_len in dataset spec '{dataset}'. "
-                            f"Expected an integer."
+                            f"Invalid value '{value}' for seq_len in dataset spec '{dataset}'. Expected an integer."
```
rkazants:
```suggestion
                            f"Invalid value '{value}' for seq_len in dataset spec '{dataset}'. "
                            f"Expected an integer."
```

optimum/intel/openvino/configuration.py
```diff
@@ -691,13 +691,11 @@ def __init__(
...
                         )
                 else:
                     raise ValueError(
-                        f"Unsupported dataset option '{key}' in dataset spec '{dataset}'. "
-                        f"Only 'seq_len' is supported."
+                        f"Unsupported dataset option '{key}' in dataset spec '{dataset}'. Only 'seq_len' is supported."
```
rkazants:
```suggestion
                        f"Unsupported dataset option '{key}' in dataset spec '{dataset}'. "
                        f"Only 'seq_len' is supported."
```

optimum/intel/openvino/configuration.py
```diff
@@ -700,13 +700,11 @@ def __init__(
                         self._dataset_kwargs[key] = int(value)
                     except ValueError:
                         raise ValueError(
-                            f"Invalid value '{value}' for seq_len in dataset spec '{dataset}'. "
-                            f"Expected an integer."
+                            f"Invalid value '{value}' for seq_len in dataset spec '{dataset}'. Expected an integer."
```
rkazants:
```suggestion
                            f"Invalid value '{value}' for seq_len in dataset spec '{dataset}'. "
                            f"Expected an integer."
```

optimum/intel/openvino/configuration.py
```diff
@@ -700,13 +700,11 @@ def __init__(
...
                         )
                 else:
                     raise ValueError(
-                        f"Unsupported dataset option '{key}' in dataset spec '{dataset}'. "
-                        f"Only 'seq_len' is supported."
+                        f"Unsupported dataset option '{key}' in dataset spec '{dataset}'. Only 'seq_len' is supported."
```
rkazants:
```suggestion
                        f"Unsupported dataset option '{key}' in dataset spec '{dataset}'. "
                        f"Only 'seq_len' is supported."
```

tests/openvino/utils_tests.py
```diff
@@ -355,6 +357,7 @@
     "zamba2": {"model": 44},
     "exaone4": {"model": 16},
     "lfm2": {"model": 52},
+    "qwen3_eagle3": {"model": 52},
```
```suggestion
    "qwen3_eagle3": {"model": 20},
```

optimum/exporters/openvino/convert.py
```diff
@@ -367,6 +371,7 @@ def export_pytorch(
 
         # Check that inputs match, and order them properly
         dummy_inputs = config.generate_dummy_inputs(framework="pt", **input_shapes)
+
```
rkazants:
```suggestion
```

optimum/exporters/openvino/convert.py
```diff
@@ -847,6 +851,18 @@ def _add_runtime_options_to_rt_info(model: Model, options: Dict):
     return model
 
 
+def _add_eagle3_mode_to_rt_info(model: Model):
+    """
+    Add eagle3 mode
+    """
+    try:
+        model.set_rt_info("True", ["eagle3_mode"])
+    except Exception:
```
andrey-churkin, line 864: Is it possible to catch a more specific exception here?

optimum/exporters/openvino/model_configs.py
```diff
@@ -653,6 +685,48 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        common_inputs = super().inputs
+        # Eagle3 model has additional conditional input
+        if self.eagle3:
+            common_inputs["hidden_states"] = {0: "batch_size", 1: "sequence_length", 2: "hidden_size"}
+        return common_inputs
+
+    @property
+    def outputs(self) -> dict[str, dict[int, str]]:
```
andrey-churkin: Just for consistency, I would use either `Dict[str, Dict[int, str]]` or `dict[str, dict[int, str]]`, not both

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7519,3 +7543,264 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        self,
+        input_emb: torch.Tensor,
+        hidden_states: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
+        past_key_value: Cache = None,
+        position_embeddings: tuple[torch.Tensor, torch.Tensor] = None,
+        output_attentions: Optional[bool] = False,
+        use_cache: Optional[bool] = False,
+    ) -> Tuple[torch.FloatTensor, Optional[Tuple[torch.FloatTensor, torch.FloatTensor]]]:
```
andrey-churkin, line 7690: Same comment as above, but for `Tuple` and `tuple`.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7519,3 +7543,264 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    def forward(
+        self,
+        input_emb: torch.Tensor,
+        hidden_states: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
+        past_key_value: Cache = None,
+        position_embeddings: tuple[torch.Tensor, torch.Tensor] = None,
+        output_attentions: Optional[bool] = False,
+        use_cache: Optional[bool] = False,
```
andrey-churkin, line 7689: https://docs.python.org/3/library/typing.html#typing.Optional

> An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional.

I guess that `None` is not desired value for these parameters.

```suggestion
        output_attentions: bool = False,
        use_cache: bool = False,
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7519,3 +7543,264 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        self.input_layernorm = LlamaRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        self.post_attention_layernorm = LlamaRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+
+    def forward(
+        self,
+        input_emb: torch.Tensor,
+        hidden_states: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
+        past_key_value: Cache = None,
```
andrey-churkin:
```suggestion
        past_key_value: Cache | None = None,
```

andrey-churkin (approved):
LGTM, just several non-critical comments.

optimum/exporters/openvino/convert.py
rkazants: we have the same approach in other similar functions to set rt info. I think no reason to differentiate exceptions. Let us see if we can see any reason in future, If yes, we will add it.

optimum/exporters/openvino/model_patcher.py
rkazants: we use `-> Tuple[torch.FloatTensor, Optional[Tuple[torch.FloatTensor, torch.FloatTensor]]]`

optimum/exporters/openvino/model_patcher.py
rkazants: We have the same declaration in transformers library: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L312
 I tend to use the same approach because we call transformers function here inside.

optimum/exporters/openvino/model_patcher.py
rkazants: Done

optimum/exporters/openvino/model_configs.py

tests/openvino/utils_tests.py
```diff
@@ -223,8 +223,10 @@
     "sana-sprint": "optimum-intel-internal-testing/tiny-random-sana-sprint",
     "ltx-video": "optimum-intel-internal-testing/tiny-random-ltx-video",
     "zamba2": "optimum-intel-internal-testing/tiny-random-zamba2",
+    "qwen3_eagle3": "AngelSlim/Qwen3-1.7B_eagle3",
```
echarlaix, line 227: can you replace it with a tiny model and have it under https://huggingface.co/optimum-intel-internal-testing ?

optimum/exporters/openvino/model_configs.py
```diff
@@ -639,6 +639,38 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+    ):
+        self.batch_size = batch_size
+        self.sequence_length = sequence_length
+        self.hidden_size = normalized_config.hidden_size
+
+    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+        shape = (
+            self.batch_size,
+            self.sequence_length,
+            self.hidden_size * 3,
```
echarlaix, line 756: please add a comment to explain why self.hidden_size * 3

optimum/exporters/openvino/__main__.py
```diff
@@ -301,6 +316,11 @@ def main_export(
         quantization_config = getattr(config, "quantization_config", None)
         quant_method = quantization_config.get("quant_method", None) if quantization_config else None
 
+        # update config to load eagle3 models
+        archs = getattr(config, "architectures", None)
+        if isinstance(archs, list) and len(archs) > 0 and "eagle3" in archs[0].lower():
```
echarlaix: shouldn't it be  
```suggestion
        if isinstance(archs, list) and len(archs) > 0 and archs[0] == "LlamaForCausalLMEagle3":
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7519,3 +7543,264 @@ def __exit__(self, exc_type, exc_value, traceback):
                 afmoe_moe = layer.mlp
                 afmoe_moe.forward = afmoe_moe._orig_forward
                 del afmoe_moe.down_projs, afmoe_moe.gate_projs, afmoe_moe.up_projs
+
+
+class LlamaEagle3Attention(LlamaAttention):
```
echarlaix, line 7638: please add a link to original code https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/llama/modeling_llama.py#L197 (same comment for LlamaMLP, LlamaPreTrainedModel, LlamaDecoderLayer)

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7519,3 +7543,264 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    two separate inputs:
+        - `hidden_states`: hidden states produced by the main (target) model
+        - `input_emb`: input token embeddings corresponding to the same positions
+
+    This layer is used exclusively within the Eagle-3 draft model and is not
+    compatible with token-only decoding pipelines.
+    """
+
+    def __init__(self, config, last=True):
+        from transformers.models.llama.modeling_llama import LlamaMLP
```
echarlaix: would you mind moving to top file?

tests/openvino/test_decoder.py
```diff
@@ -826,3 +827,29 @@ def test_load_with_different_dtype(self):
                 torch.allclose(torch.Tensor(ov_logits), ref_logits, atol=5e-3),
                 f"values are not close for {dtype if dtype is not None else 'None'}, max diff = {torch.abs(ov_logits - ref_logits).max()}",
             )
+
+    @parameterized.expand(EAGLE3_MODELS.items())
+    def test_load_and_infer_with_eagle3_model(self, model_arch, model_pair):
+        if is_transformers_version("<", "4.54"):
+            self.skipTest("Eagle3 requires transformers >= 4.54")
```
echarlaix: would replace
```suggestion
    @parameterized.expand(EAGLE3_MODELS.items())
    @pytest.mark.skipif(is_transformers_version("<", "4.54"), reason="Eagle3 requires transformers >= 4.54")
    def test_load_and_infer_with_eagle3_model(self, model_arch, model_pair):
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7519,3 +7543,264 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    with the main model vocabulary during speculative decoding.
+    """
+
+    logits: Optional[torch.FloatTensor] = None
+    past_key_values: Optional[Cache] = None
+    hidden_states: Optional[tuple[torch.FloatTensor, ...]] = None
+    d2t: Optional[torch.LongTensor] = None
+
+
+class LlamaEagle3ForCausalLM(LlamaPreTrainedModel, GenerationMixin):
```
echarlaix, line 7850: I'm not in favor of adding modeling code to `optimum-intel`, in my opinion we should only target models that are already supported by compatible libraries (transformers, diffusers,...) + remote code models when needed (limited support). We can make an exception here, but it should also be specified that we will not commit for any compatibility with future transformers versions

rkazants: We should discuss it very soon:)
May be we should rather consider new extras for Eagle3 models from AngelSlim framework. Thanks for the comment.

optimum/exporters/openvino/model_patcher.py
rkazants: Done

optimum/exporters/openvino/model_configs.py
rkazants: Added.

optimum/exporters/openvino/model_patcher.py
rkazants: Added

tests/openvino/utils_tests.py
rkazants: Let me please address this in a separate PR. That is because I also need to generate compatible tiny main model. It takes some time but tomorrow is our final pre-release check.

tests/openvino/utils_tests.py
```diff
@@ -224,8 +224,10 @@
...
 }
 
+EAGLE3_MODELS = {"qwen3_eagle3": ("AngelSlim/Qwen3-1.7B_eagle3", "Qwen/Qwen3-1.7B")}
```
echarlaix, line 230: these models will be changed in a following PR cc @rkazants

PR #1590, opened 2026-01-20
[OpenVINO] Register attention mask functions for LFM2 model

andrew-k-park:
# What does this PR do?
Register attention mask functions for LFM2 models in transformers >= 4.53.0 to enable proper model conversion

## Before submitting


optimum/exporters/openvino/model_patcher.py
```diff
@@ -7153,6 +7153,16 @@ def __enter__(self):
         super().__enter__()
         setattr(self._model, self.orig_forward_name, self.patched_forward)
 
+        if is_transformers_version(">=", "4.53.0"):
```
rkazants: instead of this change, let us try to inherit this patcher class from `OVDecoderModelPatcher` that includes this registrations

andrew-k-park: applied it as you suggested

rkazants (approved):

PR #1592, opened 2026-01-21
[OpenVINO] Use attention_mask_without_vmap for Arcee Trinity model

rkazants:
# What does this PR do?

Use attention_mask_without_vmap for Arcee Trinity model. It will fix nan values for non-stateful models on CPU

## Before submitting


rkazants:
Looks the problem is fixed. We are ready for merging

PR #1597, opened 2026-01-26
Add a condition based on the NNCF version to rename the mode from cb4 to cb4_f8e4m3

andrey-churkin:
# What does this PR do?

The [PR#3827](https://github.com/openvinotoolkit/nncf/pull/3827) was recently merged into NNCF. This PR renames the `nncf.CompressWeightsMode.CB4_F8E4M3` parameter to `nncf.CompressWeightsMode.CB4`.

This PR reflects these changes and makes the required updates in optimum-intel.

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -27,7 +27,10 @@ from transformers.utils.quantization_config import QuantizationConfigMixin
 
 from optimum.configuration_utils import BaseConfig
 
-from ..utils.import_utils import is_nncf_available
+from ..utils.import_utils import (
+    is_nncf_available,
+    is_nncf_version,
+)
 from .utils import (
     PREDEFINED_CAUSAL_LANGUAGE_DATASETS,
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/configuration.py
```diff
@@ -1060,7 +1063,8 @@ def to_nncf_dict(self) -> Dict[str, Any]:
         mode = self.dtype if self.dtype else signed_bitness[self.bits]
         if mode in signed_bitness.values():
             mode += "_sym" if self.sym else "_asym"
-        if mode == "cb4":
+
+        if mode == "cb4" and is_nncf_version("<=", "2.19"):
```
echarlaix, line 1067: should we include any potential future minor release ? 
```suggestion
        if mode == "cb4" and is_nncf_version("<", "2.20"):
```

andrey-churkin: As far as I know, we don’t plan any minor releases. The next release will be 3.0.

PR #1599, opened 2026-01-27
[OpenVINO] Fix mismatch type in MoE for Trinity (Afmoe)

rkazants:
# What does this PR do?

Fixes error during Trinity model conversion:

```
Exception:
expected m1 and m2 to have the same dtype, but got: float != c10::Half
```

## Before submitting

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -297,40 +297,39 @@ def parse_args_openvino(parser: "ArgumentParser"):
 
 
 def no_compression_parameter_provided(args):
-    from optimum.exporters.openvino.__main__ import _no_compression_parameter_provided
-
-    logger.warning(
-        "Calling `no_compression_parameter_provided` from `openvino.py` is deprecated and will be removed in future versions."
-        "Please call it from `exporters.openvino.__main__` as `_no_compression_parameter_provided` instead."
-    )
-    return _no_compression_parameter_provided(
-        args.ratio,
-        args.group_size,
-        args.sym,
-        args.all_layers,
-        args.dataset,
-        args.num_samples,
-        args.awq,
-        args.scale_estimation,
-        args.gptq,
-        args.lora_correction,
-        args.sensitivity_metric,
-        args.backup_precision,
+    # Except statistics path
+    return all(
+        (
+            it is None
+            for it in (
+                args.ratio,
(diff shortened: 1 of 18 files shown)


optimum/exporters/openvino/model_patcher.py
```diff
@@ -7484,17 +7484,29 @@ def __enter__(self):
...
-                afmoe_moe.up_projs = torch.concat(
-                    tuple(afmoe_moe.experts[i].up_proj.weight.unsqueeze(0) for i in range(num_experts)), dim=0
-                ).transpose(1, 2)
+                afmoe_moe.down_projs = (
+                    torch.concat(
+                        tuple(afmoe_moe.experts[i].down_proj.weight.unsqueeze(0) for i in range(num_experts)),
+                        dim=0,
+                    )
+                    .transpose(1, 2)
+                    .float()
```
echarlaix, line 7497: why is it needed and should it always be float ?

rkazants: This is needed for model itself because the weights are of bloat16 type and we need to align it with `hidden_state` of float32. Otherwise, we face an issue:

```
Exception:
expected m1 and m2 to have the same dtype, but got: float != c10::Half
```

IlyasMoutawwakil: not sure if upcasting model's weights is the right solution here, given that openvino supports half precision to keep memory usage comparable between torch and OV, why is the hidden state in fp32 at this point ?

rkazants: Hi @IlyasMoutawwakil, great comment. I totally agree with you. After our offline discussion, I did investigation why this mismatch happens: hidden_states - fp32 and weights - bf16 in torch.bmm. As it turned out, hidden_states has correct expected type fp32 that just properly applied bf16 patching (https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/convert.py#L424). It correctly patched above layer (nn.Linear for embedding) and it returns fp32, OV transformations will just properly handle it and preserve bf16 weights.
As for torch.bmm with bf16 weights, there are no correct patching and tracing happens on original type for weights.
Now I provide this temporal WA and created a high-exposure ticket for PT FE for further addressing by @mvafin.

PR #1602, opened 2026-01-29
Deprecate extras, install OpenVINO+NNCF by default and update README

rkazants:
# What does this PR do?

We deprecate extras (`openvino`, `nncf`, `neural-compressor`, `ipex`). Now OpenVINO backend will be installed by default with `pip install optimum-intel`.
Also, update README.md to concentrate more on logical pipeline for model running (from installation step to inference step). Recommend to use `OpenVINO GenAI`.

## Before submitting

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,11 +29,13 @@ python -m pip install "optimum-intel"@git+https://github.com/huggingface/optimum
 ## Export:
 
 To export your model to [OpenVINO IR](https://docs.openvino.ai/2025/documentation/openvino-ir-format.html) format, use the optimum-cli tool.
-Below is an example of exporting `TinyLlama/TinyLlama_v1.1` model:
+Below is an example of exporting [TinyLlama/TinyLlama_v1.1](https://huggingface.co/TinyLlama/TinyLlama_v1.1) model:
 
 ```sh
 optimum-cli export openvino --model TinyLlama/TinyLlama_v1.1 ov_TinyLlama_v1_1
 ```
+
+Additional information on exporting models is available in the [documentation](https://huggingface.co/docs/optimum-intel/en/openvino/export).
 
 ## Inference:
 
@@ -49,6 +51,8 @@ tokenizer = AutoTokenizer.from_pretrained(model_id)
 pipe = pipeline("text-generation", model=model, tokenizer=tokenizer)
 results = pipe("Hey, how are you doing today?", max_new_tokens=100)
 ```
+
+For more details on Optimum Intel inference, refer to the [documentation](https://huggingface.co/docs/optimum-intel/en/openvino/inference).
 
 **Note:** Alternatively, an exported model can also be inferred using [OpenVINO GenAI](https://github.com/openvinotoolkit/openvino.genai) framework,
 that provides optimized execution methods for highly performant Generative AI.


README.md
```diff
@@ -14,69 +14,46 @@
 To install the latest release of 🤗 Optimum Intel with the corresponding required dependencies, you can use `pip` as follows:
 
 ```bash
-python -m pip install --upgrade-strategy eager "optimum-intel[openvino]"
+python -m pip install -U "optimum-intel[openvino]"
```
echarlaix, line 17:
```suggestion
python -m pip install -U "optimum-intel"
```

README.md
```diff
@@ -14,69 +14,46 @@
...
-
-## OpenVINO
-
-Below are examples of how to use OpenVINO and its [NNCF](https://docs.openvino.ai/2025/openvino-workflow/model-optimization.html) framework to accelerate inference.
 
-#### Export:
+## Export:
 
-It is also possible to export your model to the [OpenVINO IR](https://docs.openvino.ai/2025/documentation/openvino-ir-format.html) format with the CLI :
+To export your model to [OpenVINO IR](https://docs.openvino.ai/2025/documentation/openvino-ir-format.html) format, use the optimum-cli tool.
```
echarlaix, line 31: would add a link to the documentation as well https://huggingface.co/docs/optimum-intel/en/openvino/export

README.md
```diff
@@ -14,69 +14,46 @@
...
+Below is an example of exporting `TinyLlama/TinyLlama_v1.1` model:
```
echarlaix: would add a link to the model as well https://huggingface.co/TinyLlama/TinyLlama_v1.1

README.md
```diff
@@ -14,69 +14,46 @@
...
 
-You can also apply 8-bit weight-only quantization when exporting your model : the model linear, embedding and convolution weights will be quantized to INT8, the activations will be kept in floating point precision.
+## Inference:
 
-```plain
-optimum-cli export openvino --model meta-llama/Meta-Llama-3-8B --weight-format int8 ov_llama_int8/
-```
-
-Quantization in hybrid mode can be applied to Stable Diffusion pipeline during model export. This involves applying hybrid post-training quantization to the UNet model and weight-only quantization for the rest of the pipeline components. In the hybrid mode, weights in MatMul and Embedding layers are quantized, as well as activations of other layers.
+To load an exported model and run inference using Optimum Intel, use the corresponding `OVModelForXxx` class instead of `AutoModelForXxx`:
```
echarlaix, line 42: same could make sense to add a link to https://huggingface.co/docs/optimum-intel/en/openvino/inference

README.md
```diff
@@ -14,69 +14,46 @@
...
 
-```plain
-optimum-cli export openvino --model stabilityai/stable-diffusion-2-1 --dataset conceptual_captions --weight-format int8 ov_model_sd/
+```python
+from optimum.intel import OVModelForSeq2SeqLM, OVModelForCausalLM
```
echarlaix:
```suggestion
from optimum.intel import OVModelForCausalLM
```

setup.py
```diff
@@ -31,6 +31,9 @@
     "optimum-onnx@git+https://github.com/huggingface/optimum-onnx.git@main",
     "transformers>=4.45,<4.58",
     "setuptools",
+    "nncf>=2.19.0",
+    "openvino>=2025.4.0",
+    "openvino-tokenizers>=2025.4.0",
```
echarlaix, line 36:
```suggestion
    "openvino>=2025.3.0",
    "openvino-tokenizers>=2025.3.0",
```

not related to this PR but would be great to enable v2025.3 as well given that [since 2025.4 we are facing issues](https://github.com/huggingface/optimum-intel/actions/runs/21479267196/job/61871097972)

We would also need to update https://github.com/huggingface/optimum-intel/blob/ddc3919ec46aeff94dabc955d0ccb85124547480/optimum/intel/openvino/__init__.py#L39

README.md
rkazants: It is intentionally not updated since it does not work for latest release. We will update this cmd-line for new release

setup.py
rkazants: We should leave 2025.4 version here since we have models (Zamba2, some MoE models) that required 2025.4 release

README.md
echarlaix: makes sense, then shouldn't we keep the extra as well when installing it from source ?

README.md
```diff
@@ -14,69 +14,46 @@
...
-
-We recommend creating a [virtual environment](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment) and upgrading
-pip with `python -m pip install --upgrade pip`.
-
-Optimum Intel is a fast-moving project, and you may want to install from source with the following command:
+Optimum Intel is a fast-moving project with regular additions of new model support, so you may want to install from source with the following command:
 
 ```bash
-python -m pip install "optimum-intel[openvino]"@git+https://github.com/huggingface/optimum-intel.git
+python -m pip install "optimum-intel"@git+https://github.com/huggingface/optimum-intel.git
```
echarlaix, line 23:
```suggestion
python -m pip install "optimum-intel[openvino]"@git+https://github.com/huggingface/optimum-intel.git
```

README.md
echarlaix: https://github.com/huggingface/optimum-intel/pull/1602/changes#r2746752583

setup.py
echarlaix: @rkazants then would you mind investigating https://github.com/huggingface/optimum-intel/pull/1602/changes#r2746752583 (only happens with openvino >= v2025.4)

README.md
echarlaix: @rkazants feel free to ignore in which case I'll add it in a following PR

README.md
rkazants: Once we merge this PR, it is not needed for installation from source due to this change: https://github.com/huggingface/optimum-intel/pull/1602/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R34

README.md
rkazants: not needed, it will work once we merge this PR:)

README.md
echarlaix: you're right, resolving !

echarlaix (approved):
Thanks for the addition ! Feel free to merge

setup.py
rkazants: @echarlaix, sure. I will do.

README.md
rkazants: Added

README.md
rkazants: Added, thanks

README.md
rkazants: I added, thanks:)

PR #1603, opened 2026-01-30
4-bit compression config for Trinity model

MaximProshin:
# What does this PR do?

Adds the compression configuration file which improves the Similarity of int4 model from 0.019552 to 0.920634 (CPU). This model is MoE and requires the ignored scope similar to GPT-OSS.

Fixes # 179269

## Before submitting

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -388,7 +388,7 @@ _DEFAULT_4BIT_WQ_CONFIGS = {
             "sym": False,
             "weight_only": True,
         },
-    },    
+    },
 }
 
 _DEFAULT_8BIT_WQ_CONFIGS = {


MaximProshin:
@andrey-churkin @rkazants FYI

rkazants (approved):

optimum/intel/openvino/configuration.py
```diff
@@ -374,6 +374,21 @@ class OVQuantizationMethod(str, Enum):
...
+    "arcee-ai/Trinity-Nano-Preview": {
+        "quantization_config1": {
+            "bits": 4,
+            "sym": False,
+            "group_size": 64,
+            # With ignored scope below we keep some weights in their original precision during the first quantization
+            # run and then quantize them to int8 in the second run.
+            "ignored_scope": {"patterns": [".*self_attn.*", ".*router.*"]},
+        },
+        "quantization_config2": {
```
echarlaix, line 386: Would also make sense to also rename `_GPTOSSQuantizationConfig` if having multiple quantization configs is something that's not only reserved for gpt-oss models https://github.com/huggingface/optimum-intel/blob/ddc3919ec46aeff94dabc955d0ccb85124547480/optimum/intel/openvino/configuration.py#L1538

echarlaix:
@bot /style

optimum/intel/openvino/configuration.py
MaximProshin: I agree, but let's address it in a separate PR. @andrey-churkin FYI

PR #1609, opened 2026-02-10
[OpenVINO] LFM2-MoE support for transformers v4.

popovaan:
# What does this PR do?

Added LFM2-MoE support for transformers v4.

## Before submitting
- [N/A as the model is private.] Did you write any new necessary tests?

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -4594,9 +4594,6 @@ class LFM2MoeOpenVINOConfig(LFM2OpenVINOConfig):
     MAX_TRANSFORMERS_VERSION = "4.57.99"
     _MODEL_PATCHER = Lfm2MoeModelPatcher
 
-    DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, Lfm2DummyPastKeyValuesGenerator)
-    DUMMY_PKV_GENERATOR_CLASS = Lfm2DummyPastKeyValuesGenerator
-
 
 @register_in_tasks_manager(
     "granitemoehybrid", *["text-generation", "text-generation-with-past"], library_name="transformers"
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -4597,37 +4597,6 @@ class LFM2OpenVINOConfig(LFM2OpenVINOConfig):
     DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, Lfm2DummyPastKeyValuesGenerator)
     DUMMY_PKV_GENERATOR_CLASS = Lfm2DummyPastKeyValuesGenerator
 
-    # def add_past_key_values(self, inputs_or_outputs: Dict[str, Dict[int, str]], direction: str):
-    #     if direction not in ["inputs", "outputs"]:
-    #         raise ValueError(f'direction must either be "inputs" or "outputs", but {direction} was given')
-    #
-    #     if direction == "inputs":
-    #         decoder_sequence_name = "past_sequence_length"
-    #         cache_name_prefix = "cache_params.past"
-    #     else:
-    #         decoder_sequence_name = "past_sequence_length + sequence_length"
-    #         cache_name_prefix = "cache_params.present"
-    #
-    #     self.num_conv_layers = self._normalized_config.layer_types.count("conv")
-    #     self.num_atten_layers = self._normalized_config.layer_types.count("full_attention")
-    #
-    #     for i in range(self.num_conv_layers):
-    #         inputs_or_outputs[f"{cache_name_prefix}.conv.{i}"] = {0: "batch_size"}
-    #
-    #     for i in range(self.num_atten_layers):
-    #         inputs_or_outputs[f"{cache_name_prefix}.key.{i}"] = {0: "batch_size", 2: decoder_sequence_name}
-    #         inputs_or_outputs[f"{cache_name_prefix}.value.{i}"] = {0: "batch_size", 2: decoder_sequence_name}
-    #
-    # @property
-    # def inputs(self) -> Dict[str, Dict[int, str]]:
-    #     common_inputs = {
-    #         "input_ids": {0: "batch_size", 1: "sequence_length"},
-    #         "attention_mask": {0: "batch_size", 1: "sequence_length"},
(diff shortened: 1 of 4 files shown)


optimum/exporters/openvino/model_configs.py
```diff
@@ -4902,6 +4582,22 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
 
+@register_in_tasks_manager(
+    "lfm2_moe",
+    *[
+        "text-generation",
+        "text-generation-with-past",
+    ],
+    library_name="transformers",
+)
+class LFM2OpenVINOConfig(LFM2OpenVINOConfig):
```
```suggestion
class LFM2MoeOpenVINOConfig(LFM2OpenVINOConfig):
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -4902,6 +4582,22 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+    MIN_TRANSFORMERS_VERSION = "4.54.0"
```
rkazants: let us put `MAX_TRANSFORMERS_VERSION = 4.57.99`

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7280,12 +7167,101 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        num_experts,
+        device=hidden_states.device,
+        dtype=hidden_states.dtype,
+    )
+    dense_routing_weights.scatter_(dim=1, index=selected_experts, src=routing_weights)
+    hidden_states_expanded = hidden_states.repeat(num_experts, 1) # (num_experts * num_tokens, hidden_dim)
+    hidden_states_expanded = hidden_states_expanded.view(num_experts, -1, hidden_dim) # (num_experts, num_tokens, hidden_dim)
+
+    # Stack expert parameters
+    w1_stacked = torch.stack([e.w1.weight.T for e in self.experts])
```
rkazants: you also transpose weight, right? May be not needed

popovaan: No, without it shapes don't match. 
I see that in Afmoe these weights are transposed as well, but concatenated beforehand. I can implement it the same way https://github.com/huggingface/optimum-intel/blob/2c48d6430c265ac259c1b264f3e2c4025cdd7b76/optimum/exporters/openvino/model_patcher.py#L7604C16-L7611C18

optimum/exporters/openvino/model_configs.py
popovaan: Done.

optimum/exporters/openvino/model_configs.py
```diff
@@ -163,6 +163,7 @@
     InternVLChatImageEmbeddingModelPatcher,
     JaisModelPatcher,
```
rkazants, line 164: please update https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/utils.py#L308 with `lfm2_moe`. It can help with an inference

optimum/exporters/openvino/model_patcher.py
rkazants: Got it

rkazants: @popovaan, btw, can we re-use existing MoE patching for this model?

optimum/exporters/openvino/model_configs.py
popovaan: Yes, it helped, now the model can be infered and text looks adequate, but it doesn't match original model text.

optimum/exporters/openvino/model_patcher.py
popovaan: The closest patching to this version of MoE is Qwen3 MoE block, but it has a slightly diffrent preprocessing of routing_weights (sofmax is applied instead of dividing by sum), so I suppose we can't reuse it as is.

optimum/exporters/openvino/model_configs.py
popovaan: Changed the patching in this commit, now the output text seems fine: https://github.com/huggingface/optimum-intel/pull/1609/commits/e7ad8fb036818dd9994e9a65e425148dcb399b15

popovaan: Returned torch.bmm patching as agreed. Accuracy is fine with it.

optimum/exporters/openvino/model_configs.py
```diff
@@ -4902,6 +4582,22 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
+        "text-generation-with-past",
+    ],
+    library_name="transformers",
+)
+class LFM2MoeOpenVINOConfig(LFM2OpenVINOConfig):
+    MAX_TRANSFORMERS_VERSION = "4.57.99"
+    _MODEL_PATCHER = Lfm2MoeModelPatcher
+
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, Lfm2DummyPastKeyValuesGenerator)
+    DUMMY_PKV_GENERATOR_CLASS = Lfm2DummyPastKeyValuesGenerator
```
```suggestion
class LFM2MoeOpenVINOConfig(LFM2OpenVINOConfig):
    MAX_TRANSFORMERS_VERSION = "4.57.99"
    _MODEL_PATCHER = Lfm2MoeModelPatcher
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7275,17 +7162,110 @@ def __enter__(self):
...
+                    tuple(sparse_moe_block.experts[i].w1.weight.unsqueeze(0) for i in range(num_experts)),
+                    dim=0,
+                ).transpose(1, 2)
+                sparse_moe_block.w2_stacked = torch.concat(
+                    tuple(sparse_moe_block.experts[i].w2.weight.unsqueeze(0) for i in range(num_experts)),
+                    dim=0,
+                ).transpose(1, 2)
+                sparse_moe_block.w3_stacked = torch.concat(
+                    tuple(sparse_moe_block.experts[i].w3.weight.unsqueeze(0) for i in range(num_experts)), dim=0
+                ).transpose(1, 2)
```
rkazants: please remove weight constants in `exit`

popovaan: Done.

rkazants:
Already merged another PR for official LFM2-MoE

PR #1610, opened 2026-02-11
Fix incorrect task inference for the Phi-4-multimodal-instruct model

andrey-churkin:
# What does this PR do?

Fixed the following error, which occurred on commit 2c48d6430c265ac259c1b264f3e2c4025cdd7b76:

```
RuntimeError: Exception from src/inference/src/cpp/core.cpp:97:
Check 'util::directory_exists(path) || util::file_exists(path)' failed at src/frontends/common/src/frontend.cpp:113:
FrontEnd API failed with GeneralFailure:
ir: Could not open the file: "/tmp/tmp2qkkght3/openvino_encoder_model.xml"
```

This error occurred when attempting to execute the command:

```
optimum-cli export openvino -m microsoft/Phi-4-multimodal-instruct tmp --trust-remote-code --weight-format int4 --group-size -1
```

**Root Cause:** The inferred task for this model.

https://github.com/huggingface/optimum-intel/blob/2c48d6430c265ac259c1b264f3e2c4025cdd7b76/optimum/exporters/openvino/__main__.py#L621

is `'automatic-speech-recognition'`. So the corresponding class

https://github.com/huggingface/optimum-intel/blob/2c48d6430c265ac259c1b264f3e2c4025cdd7b76/optimum/exporters/openvino/__main__.py#L656

is the `OVModelForSpeechSeq2Seq`. We can't load the model that was exported during `main_export()` via `OVModelForSpeechSeq2Seq.from_pretrained()`. The provided error occurred.

This fix is based on https://github.com/huggingface/optimum-intel/pull/1293

Versions:
```
transformers==4.51.3
peft==0.17
```

Fixes CVS-179936

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -643,7 +643,7 @@ def _main_quantize(
     # explicitly. Because of this, we get an error.
     if original_task == "auto" and library_name == "transformers":
         config = AutoConfig.from_pretrained(
-            output,
+            model_name_or_path,
             subfolder=subfolder,
             revision=revision,
             cache_dir=cache_dir,


optimum/exporters/openvino/__main__.py
```diff
@@ -125,6 +125,10 @@ def infer_task(
         else:
             model_type = config.model_type
         custom_architecture = model_type not in TasksManager._SUPPORTED_MODEL_TYPE
+
+        if original_task == "auto" and model_type in MULTI_MODAL_TEXT_GENERATION_MODELS:
+            task = "image-text-to-text"
```
rkazants: I think the problem is connected with this line: https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_configs.py#L273

As I can see, it can be exported for two tasks (speech recognition or vlm), `_main_quantize` should select the right one, may be based on IR directory content or we should store task value in some config file during export and use it.

optimum/exporters/openvino/__main__.py
```diff
@@ -125,6 +125,10 @@ def infer_task(
...
+
```
rkazants: please add test that reproduces this issue on the corresponding tiny-model. Just add optimum-cli cmd line test with `group_size` parameter.

andrey-churkin: done

optimum/exporters/openvino/__main__.py
andrey-churkin: Let me explain my point of view on this issue.

First of all, let’s understand how it worked **before this PR**. We execute the following command
```
optimum-cli export openvino -m microsoft/Phi-4-multimodal-instruct tmp --trust-remote-code --weight-format int4 --group-size -1
```
The `--task` was not specified, so it will be auto-inferred. Inside the `main_export()` method, the task is inferred

https://github.com/huggingface/optimum-intel/blob/2c48d6430c265ac259c1b264f3e2c4025cdd7b76/optimum/exporters/openvino/__main__.py#L284

and it equals `automatic-speech-recognition`. So far, so good. The rest of the code in `main_export()` works with this task and successfully exports the model. After that, inside `_main_quantize()` we infer the task again (using the same logic as in `main_export()`), and it equals `automatic-speech-recognition`. We get the `OVModelForSpeechSeq2Seq` model class using the inferred task `automatic-speech-recognition`. It perfectly aligns with the [documentation](https://huggingface.co/docs/optimum-intel/openvino/inference). After that, we try to load the model using the `OVModelForSpeechSeq2Seq.from_pretrained()` method, but it fails with an error. However, the exported model is loaded using the `OVModelForVisualCausalLM` class without any errors.

I can't say for sure why [here](https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_configs.py#L273) we have `automatic-speech-recognition`. It is probably because `TasksManager._infer_task_from_model_name_or_path()` returns it during task inference and it was necessary to add the task there because of this.

According to the PR https://github.com/huggingface/optimum-intel/pull/1293 models from `MULTI_MODAL_TEXT_GENERATION_MODELS` should be loaded using `OVModelForVisualCausalLM` class.

To me, it seems inconsistent that we can export a model for some task but can't load the exported model using the class corresponding to that task. This is the main reason why I changed the `infer_task()` method.

optimum/exporters/openvino/__main__.py
```diff
@@ -637,6 +637,26 @@ def _main_quantize(
...
+            subfolder=subfolder,
+            revision=revision,
+            cache_dir=cache_dir,
+            token=token,
+            trust_remote_code=trust_remote_code,
+        )
+        if hasattr(config, "export_model_type"):
+            model_type = config.export_model_type
+        else:
+            model_type = config.model_type
```
echarlaix: hasattr(config, "export_model_type") should always be false no ?
```suggestion
        model_type = config.model_type
```

tests/openvino/test_exporters_cli.py
```diff
@@ -788,6 +788,22 @@ class OVCLIExportTestCase(unittest.TestCase):
                 "resampler_model": {"int8": 6},
             },
         ),
+        (
+            "auto",
+            "phi4mm",
```
echarlaix: I don't think this test will ever be triggered

optimum/exporters/openvino/__main__.py
andrey-churkin: Thanks, fixed.

tests/openvino/test_exporters_cli.py
andrey-churkin: Yes, I can see that it wasn’t triggered. Do you have any idea why?

echarlaix: phi4mm only supported for couple of transformers version which are currently not tested in our CI
https://github.com/huggingface/optimum-intel/blob/2c48d6430c265ac259c1b264f3e2c4025cdd7b76/optimum/exporters/openvino/model_configs.py#L3290-L3291

echarlaix: I think it makes sense to remove this test for this reason

andrey-churkin: Done

optimum/exporters/openvino/__main__.py
```diff
@@ -637,6 +637,23 @@ def _main_quantize(
             token=token,
         )
 
+    # NOTE: The Phi-4-multimodal-instruct model card contains a pipeline_tag set to automatic-speech-recognition,
+    # which is returned as the inferred task. As a result, we try to load the exported model using the
+    # OVModelForSpeechSeq2Seq class instead of the OVModelForVisualCausalLM class when the task is not specified
+    # explicitly. Because of this, we get an error.
+    if original_task == "auto" and library_name == "transformers":
+        config = AutoConfig.from_pretrained(
+            output,
```
rkazants: why `output` is used? I think it should be `model_name_or_path`

andrey-churkin: thanks, fixed

Issue #1608, opened 2026-02-08
GigaChat3-10B-A1.8B Support ❤️


savvadesogle:
Hi. Please add support for this model 🙏. It supports many CIS languages.

https://huggingface.co/ai-sage/GigaChat3-10B-A1.8B-bf16/blob/main/config.json

An error occurs during the current conversion.

```
(openarc) c:\llm\openarc\201>optimum-cli export openvino --model "T:\models\ai-sage\GigaChat3-10B-A1.8B-bf16" --task text-generation --weight-format int4 "C:\llm\models\ov\GigaChat3-10B-A1.8B-ov-int4"
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 11/11 [00:23<00:00,  2.14s/it]
[...]

PR #1626, opened 2026-02-28
[OpenVINO] support ai-sage/GigaChat3-10B-A1.8B-bf16

Mohamed-Ashraf273:
# What does this PR do?

Conversion cmd-line for ai-sage/GigaChat3-10B-A1.8B-bf16:
```bash
optimum-cli export openvino -m ai-sage/GigaChat3-10B-A1.8B-bf16 ./output_dir --task text-generation-with-past --weight-format fp16
```

Inference of `ai-sage/GigaChat3-10B-A1.8B-bf16` using OpenVINO backend:
 
```python

model_dir = "output_dir"

    model_dir,
    ov_config={"INFERENCE_PRECISION_HINT": "f32"},
)

inputs.pop("token_type_ids", None)

output_text = tokenizer.decode(output_ids[0], skip_special_tokens=True)

print(output_text)
```
 Solving Issue: #1608 
## Before submitting
* [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
* [x]  Did you make sure to update the documentation with your changes?
* [x]  Did you write any new necessary tests?

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -4110,7 +4110,8 @@ class M2M100OpenVINOConfig(BartOpenVINOConfig):
 @register_in_tasks_manager("deepseek", *["text-generation", "text-generation-with-past"], library_name="transformers")
 class DeepseekOpenVINOConfig(MiniCPM3OpenVINOConfig):
     MIN_TRANSFORMERS_VERSION = "4.53.0"
-    MAX_TRANSFORMERS_VERSION = None
+    # TODO: add v5 support
+    MAX_TRANSFORMERS_VERSION = "4.57.6"
     _MODEL_PATCHER = DeepseekPatcher
 
 
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -4109,8 +4109,8 @@ class M2M100OpenVINOConfig(BartOpenVINOConfig):
 )
 @register_in_tasks_manager("deepseek", *["text-generation", "text-generation-with-past"], library_name="transformers")
 class DeepseekOpenVINOConfig(MiniCPM3OpenVINOConfig):
-    MIN_TRANSFORMERS_VERSION = "4.46.0"
-    MAX_TRANSFORMERS_VERSION = "4.53.3"
+    MIN_TRANSFORMERS_VERSION = "4.53.0"
+    MAX_TRANSFORMERS_VERSION = None
     _MODEL_PATCHER = DeepseekPatcher
 
 
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -4251,7 +4251,7 @@ class M2M100OpenVINOConfig(BartOpenVINOConfig):
 class DeepseekOpenVINOConfig(MiniCPM3OpenVINOConfig):
     MIN_TRANSFORMERS_VERSION = "4.53.0"
     # TODO: add v5 support
-    MAX_TRANSFORMERS_VERSION = "4.57.6"
+    MAX_TRANSFORMERS_VERSION = "4.57.99"
     _MODEL_PATCHER = DeepseekPatcher
 
 
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -3896,25 +3896,96 @@ class DeepseekPatcher(OVDecoderModelPatcher):
                 block.self_attn._orig_forward = block.self_attn.forward
                 block.self_attn.forward = types.MethodType(self_attn_fwd, block.self_attn)
             if hasattr(block.mlp, "moe_infer"):
-                block.mlp._org_moe_infer = block.mlp.moe_infer
+                # old interface (transformers < 4.57): moe_infer(self, x, topk_ids, topk_weight)
+                block.mlp._orig_moe_infer = block.mlp.moe_infer
+                block.mlp._orig_moe = None
+                block.mlp.ep_rank = getattr(block.mlp, "ep_rank", 0)
+                block.mlp.experts_per_rank = getattr(block.mlp, "experts_per_rank", len(block.mlp.experts))
+                block.mlp.moe_infer = types.MethodType(deepseek_moe_infer, block.mlp)
+            elif hasattr(block.mlp, "moe") and hasattr(block.mlp, "experts"):
+                # new interface (transformers >= 4.57): moe(self, hidden_states, topk_indices, topk_weights)
+                block.mlp._orig_moe = block.mlp.moe
+                block.mlp._orig_moe_infer = None
+                num_experts = len(block.mlp.experts)
+
+                # Concatenate expert weights
+                gate_projs = torch.concat(
+                    tuple(block.mlp.experts[i].gate_proj.weight.unsqueeze(0) for i in range(num_experts)),
+                    dim=0,
+                )
+                up_projs = torch.concat(
+                    tuple(block.mlp.experts[i].up_proj.weight.unsqueeze(0) for i in range(num_experts)),
+                    dim=0,
+                )
+                down_projs = torch.concat(
+                    tuple(block.mlp.experts[i].down_proj.weight.unsqueeze(0) for i in range(num_experts)),
+                    dim=0,
+                )
+
+                if is_openvino_version("<", "2026.1.0"):
+                    logger.warning(
+                        "This model works best with OpenVINO 2026.1 or later. "
+                        "Earlier versions require float() conversion for MoE weights, "
+                        "which may affect performance. "
+                        "OpenVINO 2026.1 includes a fix for torch.bmm dtype handling."
+                    )
+                    block.mlp.gate_projs = gate_projs.float()
+                    block.mlp.up_projs = up_projs.float()
+                    block.mlp.down_projs = down_projs.float()
+                else:
+                    block.mlp.gate_projs = gate_projs
+                    block.mlp.up_projs = up_projs
+                    block.mlp.down_projs = down_projs
+
+                block.mlp.moe = types.MethodType(deepseek_moe, block.mlp)
+            elif hasattr(block.mlp, "experts"):
+                # fallback: patch by injecting moe_infer with required attributes
+                block.mlp._orig_moe_infer = None
+                block.mlp._orig_moe = None
+                block.mlp.ep_rank = 0
+                block.mlp.experts_per_rank = len(block.mlp.experts)
                 block.mlp.moe_infer = types.MethodType(deepseek_moe_infer, block.mlp)
 
     def __exit__(self, exc_type, exc_value, traceback):
         super().__exit__(exc_type, exc_value, traceback)
         for block in self._model.model.layers:
-            block.self_attn.forward = block.self_attn._orig_forward
+            if hasattr(block.self_attn, "_orig_forward"):
+                block.self_attn.forward = block.self_attn._orig_forward
+            if hasattr(block.mlp, "_orig_moe"):
+                if block.mlp._orig_moe is not None:
+                    block.mlp.moe = block.mlp._orig_moe
+                if hasattr(block.mlp, "gate_projs"):
+                    del block.mlp.gate_projs
+                if hasattr(block.mlp, "up_projs"):
+                    del block.mlp.up_projs
+                if hasattr(block.mlp, "down_projs"):
+                    del block.mlp.down_projs
+                delattr(block.mlp, "_orig_moe")
             if hasattr(block.mlp, "_orig_moe_infer"):
-                block.mlp.moe_infer = block.mlp._orig_moe_infer
+                if block.mlp._orig_moe_infer is not None:
+                    block.mlp.moe_infer = block.mlp._orig_moe_infer
+                else:
+                    if hasattr(block.mlp, "moe_infer"):
+                        delattr(block.mlp, "moe_infer")
+                    if hasattr(block.mlp, "ep_rank"):
+                        delattr(block.mlp, "ep_rank")
+                    if hasattr(block.mlp, "experts_per_rank"):
+                        delattr(block.mlp, "experts_per_rank")
+                delattr(block.mlp, "_orig_moe_infer")
 
 
 def deepseek_v3_attn_forward(
     self,
     hidden_states: torch.Tensor,
+    position_embeddings=None,
     attention_mask: Optional[torch.Tensor] = None,
     position_ids: Optional[torch.LongTensor] = None,
     past_key_value=None,
+    past_key_values=None,
+    cache_position: Optional[torch.LongTensor] = None,
     output_attentions: bool = False,
     use_cache: bool = False,
+    **kwargs,
 ) -> Tuple[torch.Tensor, Optional[torch.Tensor], Optional[Tuple[torch.Tensor]]]:
     # modified from https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/modeling_deepseek.py#L751
     def rotate_half(x):
@@ -3938,9 +4009,13 @@ def deepseek_v3_attn_forward(
             hidden_states=hidden_states,
             attention_mask=attention_mask,
             position_ids=position_ids,
+            position_embeddings=position_embeddings,
             past_key_value=past_key_value,
+            past_key_values=past_key_values,
             output_attentions=output_attentions,
             use_cache=use_cache,
+            cache_position=cache_position,
+            kwargs=kwargs,
         )
 
     bsz, q_len, _ = hidden_states.size()
@@ -3949,50 +4024,89 @@ def deepseek_v3_attn_forward(
         q = self.q_proj(hidden_states)
     else:
         q = self.q_b_proj(self.q_a_layernorm(self.q_a_proj(hidden_states)))
-    q = q.view(bsz, q_len, self.num_heads, self.q_head_dim).transpose(1, 2)
+    q = q.view(bsz, q_len, self.num_heads, self.qk_head_dim).transpose(1, 2)
     q_nope, q_pe = torch.split(q, [self.qk_nope_head_dim, self.qk_rope_head_dim], dim=-1)
 
     compressed_kv = self.kv_a_proj_with_mqa(hidden_states)
-    compressed_kv, k_pe = torch.split(compressed_kv, [self.kv_lora_rank, self.qk_rope_head_dim], dim=-1)
-    k_pe = k_pe.view(bsz, q_len, 1, self.qk_rope_head_dim).transpose(1, 2)
-    kv = (
-        self.kv_b_proj(self.kv_a_layernorm(compressed_kv))
-        .view(bsz, q_len, self.num_heads, self.qk_nope_head_dim + self.v_head_dim)
-        .transpose(1, 2)
+
+    k_pass, k_rot = torch.split(
+        compressed_kv,
+        [self.kv_lora_rank, self.qk_rope_head_dim],
+        dim=-1,
     )
 
-    k_nope, value_states = torch.split(kv, [self.qk_nope_head_dim, self.v_head_dim], dim=-1)
-    kv_seq_len = value_states.shape[-2]
-    if past_key_value is not None:
-        if self.layer_idx is None:
-            raise ValueError(
-                f"The cache structure has changed since version v4.36. If you are using {self.__class__.__name__} "
-                "for auto-regressive decoding with k/v caching, please make sure to initialize the attention class "
-                "with a layer index."
-            )
-        kv_seq_len += past_key_value.get_usable_length(kv_seq_len, self.layer_idx)
-    cos, sin = self.rotary_emb(value_states, seq_len=kv_seq_len)
+    k_pass = self.kv_b_proj(self.kv_a_layernorm(k_pass))
+    k_pass = k_pass.view(bsz, q_len, self.num_heads, self.qk_nope_head_dim + self.v_head_dim).transpose(1, 2)
 
-    q_pe, k_pe = apply_rotary_pos_emb(q_pe, k_pe, cos, sin, position_ids)
+    k_pass, value_states = torch.split(
+        k_pass,
+        [self.qk_nope_head_dim, self.v_head_dim],
+        dim=-1,
+    )
 
-    # Difference with original code, k_pe.new_empty create constant tensor in torchscript
-    query_states = torch.concat([q_nope, q_pe], dim=-1)
-    # query_states = k_pe.new_empty(bsz, self.num_heads, q_len, self.q_head_dim)
-    # query_states[:, :, :, : self.qk_nope_head_dim] = q_nope
-    # query_states[:, :, :, self.qk_nope_head_dim :] = q_pe
-    key_states = torch.concat([k_nope, k_pe.expand(-1, self.num_heads, -1, -1)], dim=-1)
-    # key_states = k_pe.new_empty(bsz, self.num_heads, q_len, self.q_head_dim)
-    # key_states[:, :, :, : self.qk_nope_head_dim] = k_nope
-    # key_states[:, :, :, self.qk_nope_head_dim :] = k_pe
-    if past_key_value is not None:
-        cache_kwargs = {"sin": sin, "cos": cos}  # Specific to RoPE models
-        key_states, value_states = past_key_value.update(key_states, value_states, self.layer_idx, cache_kwargs)
+    k_rot = k_rot.view(bsz, 1, q_len, self.qk_rope_head_dim)
 
-    if attention_mask is not None:
-        if attention_mask.size() != (bsz, 1, q_len, kv_seq_len):
-            raise ValueError(
-                f"Attention mask should be of size {(bsz, 1, q_len, kv_seq_len)}, but is {attention_mask.size()}"
-            )
+    new_interface = position_embeddings is not None and not hasattr(self, "rotary_emb")
+
+    if new_interface:
+        from transformers.models.deepseek_v3.modeling_deepseek_v3 import (
+            apply_rotary_pos_emb as deepseek_v3_apply_rotary_pos_emb,
+        )
+        from transformers.models.deepseek_v3.modeling_deepseek_v3 import (
+            apply_rotary_pos_emb_interleave as deepseek_v3_apply_rotary_pos_emb_interleave,
+        )
+
+        cos, sin = position_embeddings
+
+        if getattr(self.config, "rope_interleave", False):
+            try:
+                q_pe, k_rot = deepseek_v3_apply_rotary_pos_emb_interleave(q_pe, k_rot, cos, sin)
+            except Exception as e:
+                raise RuntimeError(
+                    "Failed to apply interleaved rotary position embeddings, "
+                    f"may due to incompatible transformers version, try to `pip install transformers>=4.57.1`: {e}"
+                )
+        else:
+            q_pe, k_rot = deepseek_v3_apply_rotary_pos_emb(q_pe, k_rot, cos, sin)
+
+        kv_cache = past_key_values if past_key_values is not None else past_key_value
+
+    else:
+        kv_seq_len = value_states.shape[-2]
+        if past_key_value is not None:
+            if self.layer_idx is None:
+                raise ValueError(
+                    f"The cache structure has changed since version v4.36. If you are using {self.__class__.__name__} "
+                    "for auto-regressive decoding with k/v caching, please make sure to initialize the attention class "
+                    "with a layer index."
+                )
+            kv_seq_len += past_key_value.get_usable_length(kv_seq_len, self.layer_idx)
+        cos, sin = self.rotary_emb(value_states, seq_len=kv_seq_len)
+        q_pe, k_rot = apply_rotary_pos_emb(q_pe, k_rot, cos, sin, position_ids)
+
+        if attention_mask is not None:
+            if attention_mask.size() != (bsz, 1, q_len, kv_seq_len):
+                raise ValueError(
+                    f"Attention mask should be of size {(bsz, 1, q_len, kv_seq_len)}, but is {attention_mask.size()}"
+                )
+
+        kv_cache = past_key_value
+
+    k_rot = k_rot.expand(*k_pass.shape[:-1], -1)
+    query_states = torch.cat((q_nope, q_pe), dim=-1)
+    key_states = torch.cat((k_pass, k_rot), dim=-1)
+
+    if kv_cache is not None:
+        if new_interface:
+            cache_kwargs = {"sin": sin, "cos": cos, "cache_position": cache_position}
+            key_states, value_states = kv_cache.update(key_states, value_states, self.layer_idx, cache_kwargs)
+
+            if attention_mask is not None:
+                attention_mask = attention_mask[:, :, :, : key_states.shape[-2]]
+
+        else:
+            cache_kwargs = {"sin": sin, "cos": cos}
+            key_states, value_states = kv_cache.update(key_states, value_states, self.layer_idx, cache_kwargs)
 
     # SDPA with memory-efficient backend is currently (torch==2.1.2) bugged with non-contiguous inputs with custom attn_mask,
     # Reference: https://github.com/pytorch/pytorch/issues/112577.
@@ -4009,6 +4123,7 @@ def deepseek_v3_attn_forward(
         dropout_p=self.attention_dropout if self.training else 0.0,
         # The q_len > 1 is necessary to match with AttentionMaskConverter.to_causal_4d that does not create a causal mask in case q_len == 1.
         is_causal=self.is_causal and attention_mask is None and q_len > 1,
+        scale=None if not new_interface else self.scaling,
     )
 
     attn_output = attn_output.transpose(1, 2).contiguous()
@@ -4016,6 +4131,9 @@ def deepseek_v3_attn_forward(
     attn_output = attn_output.reshape(bsz, q_len, self.num_heads * self.v_head_dim)
 
     attn_output = self.o_proj(attn_output)
+
+    if new_interface:
+        return attn_output, None
 
     return attn_output, None, past_key_value
 
@@ -4174,6 +4292,28 @@ def deepseek_moe_infer(self, x, topk_ids, topk_weight):
         .to(new_x.dtype)
     )
     return final_out
+
+
+def deepseek_moe(self, hidden_states: torch.Tensor, topk_indices: torch.Tensor, topk_weights: torch.Tensor):
+    """
+    Vectorized MoE that matches original behavior.
+    """
+    orig_dtype = hidden_states.dtype
+    num_experts = len(self.experts)
+    batch_tokens, _ = hidden_states.shape
+    compute_dtype = torch.promote_types(hidden_states.dtype, self.gate_projs.dtype)
+    routing = torch.zeros(batch_tokens, num_experts, dtype=compute_dtype, device=hidden_states.device)
+    routing.scatter_(1, topk_indices, topk_weights.to(dtype=compute_dtype))
+    expanded = hidden_states.to(dtype=compute_dtype).unsqueeze(0).expand(num_experts, -1, -1)
+    act_fn = self.experts[0].act_fn
+    gate = torch.bmm(expanded, self.gate_projs.to(dtype=compute_dtype).transpose(1, 2))
+    up = torch.bmm(expanded, self.up_projs.to(dtype=compute_dtype).transpose(1, 2))
+    gate_up = act_fn(gate) * up
+    next_states = torch.bmm(gate_up, self.down_projs.to(dtype=compute_dtype).transpose(1, 2))
+    routing = routing.transpose(0, 1).unsqueeze(-1)
+    next_states = next_states * routing
+    next_states = next_states.sum(dim=0)
+    return next_states.to(orig_dtype)
 
 
 class Qwen2VLLanguageModelPatcher(OVDecoderModelPatcher):
--- a/tests/openvino/test_decoder.py
+++ b/tests/openvino/test_decoder.py
@@ -106,12 +106,12 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
     if is_transformers_version(">=", "4.46.0"):
         SUPPORTED_ARCHITECTURES += ("glm", "mistral-nemo", "phimoe")
 
-        if is_transformers_version("<", "4.54.0"):
-            SUPPORTED_ARCHITECTURES += ("deepseek",)
-
         # gptq and awq install disabled for windows test environment
         if platform.system() != "Windows" and is_transformers_version("<", "4.56.0"):
             SUPPORTED_ARCHITECTURES += ("opt_gptq", "mixtral_awq")
+
+    if is_transformers_version(">=", "4.53.0"):
+        SUPPORTED_ARCHITECTURES += ("deepseek", "gigachat3")
 
     if is_transformers_version(">", "4.47"):
         SUPPORTED_ARCHITECTURES += ("olmo2",)
@@ -239,6 +239,7 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
         "minicpm3": 6,
         "phimoe": 2,
         "deepseek": 2,
+        "gigachat3": 2,
         "opt_gptq": 12,
         "mixtral_awq": 2,
         "gemma3_text": 2,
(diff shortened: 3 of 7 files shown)


tests/openvino/test_decoder.py
```diff
@@ -383,7 +383,7 @@ def test_compare_to_transformers(self, model_arch):
                 transformers_outputs = transformers_model(**tokens)
 
         # Compare tensor outputs
-        atol = 3e-3 if model_arch in ["minicpm", "qwen2-moe"] else 1e-4
+        atol = 3e-2 if model_arch in ["deepseek"] else 3e-3 if model_arch in ["minicpm", "qwen2-moe"] else 1e-4
```
Mohamed-Ashraf273: ## Numerical Difference Investigation

The numerical difference observed during export is **NOT caused by the patcher implementation**.

### Key Findings:
1. **Patcher code is correct** - Implementation is taken directly from the native transformers deepseek_v3 architecture: 
   https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v3/modeling_deepseek_v3.py

2. **Differences arise from the export process itself**, not from the patcher logic

3. That's why we use a tolerance of `3e-2` for deepseek models in tests

### Test Evidence:
```python
# Simple test showing the inherent export difference
from transformers import AutoModelForCausalLM, AutoTokenizer
import torch
import tempfile

model_id = 'optimum-intel-internal-testing/tiny-random-deepseek-v3'
tokens = tokenizer('Hello world', return_tensors='pt')

tf_model = AutoModelForCausalLM.from_pretrained(model_id)
with torch.no_grad():
    tf_out = tf_model(**tokens)

with tempfile.TemporaryDirectory() as tmp:
    ov_model = OVModelForCausalLM.from_pretrained(model_id, export=True, compile=False, save_directory=tmp)
    with torch.no_grad():
        ov_out = ov_model(**tokens)

diff = torch.abs(ov_out.logits - tf_out.logits)
print(f'Max diff: {diff.max():.6f}')  # ~0.03

**Review by rkazants — COMMENTED** [review]
please also add export tests. The same test set that you have added for the previuos model.
Update documentation.

File: tests/openvino/utils_tests.py [hunk 2]
```diff
@@ -70,6 +70,7 @@
     "deberta-v2": "optimum-intel-internal-testing/tiny-random-DebertaV2Model",
     "decilm": "optimum-intel-internal-testing/tiny-random-decilm",
     "deepseek": "optimum-intel-internal-testing/tiny-random-deepseek-v3",
+    "gigachat3": "mohamedahraf273/tiny-random-gigachat3",
```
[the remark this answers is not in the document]
Mohamed-Ashraf273: Model not published on `optimum-intel-internal-testing` yet.

Mohamed-Ashraf273:
> Hi. Can I help test the model?
> You are so great, thank you so much!♥️🔥😊

Hi!
That would be great, thank you so much!
Please feel free to test it and let me know if you encounter any issues or unexpected behavior.

tests/openvino/test_decoder.py
```diff
@@ -410,7 +417,7 @@ def test_compare_to_transformers(self, model_arch):
         ov_outputs = ov_model.generate(**tokens, generation_config=gen_config)
 
         # TODO: add back once https://huggingface.co/katuni4ka/tiny-random-minicpm3/discussions/1 merged (for all models) as current modeling incompatible with transformers >= v4.49
-        if model_arch in {"deepseek"} and is_transformers_version(">=", "4.49"):
+        if model_arch in {"deepseek", "gigachat3"} and is_transformers_version(">=", "4.49"):
```
IlyasMoutawwakil: i see that you remove max transformers version above but still skipping this part ?

tests/openvino/utils_tests.py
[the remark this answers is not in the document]
IlyasMoutawwakil: i have sent you an invite to the org, don't hesitate to move it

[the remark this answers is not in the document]
Mohamed-Ashraf273: Thanks! I’ve received the invite. I’ll move the model to the organization.

rkazants:
@Mohamed-Ashraf273 , please check tests locally on your machine:
{520710FC-2874-4953-9C49-56D57B8B8064}

tests/openvino/utils_tests.py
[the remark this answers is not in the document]
Mohamed-Ashraf273: Done! published.

tests/openvino/test_decoder.py
Mohamed-Ashraf273: Thanks for the feedback!
I've removed it.

SearchSavior:
@Mohamed-Ashraf273 great work on this PR! 

were you able to verify that @savvadesogle issue above were solved in your export changes?

PS We are very excited to deploy this model in OpenArc with a qwen ASR/TTS system I am working on, thanks for tackling the issue so quickly.

Mohamed-Ashraf273:
Hi @SearchSavior 
Thank you so much!

Yes, I verified that the issue mentioned by @savvadesogle is resolved. I’ve attached an image below showing the successful result.
Screenshot from 2026-03-06 07-02-05

optimum/exporters/openvino/model_configs.py
```diff
@@ -4080,8 +4080,8 @@ class M2M100OpenVINOConfig(BartOpenVINOConfig):
 )
 @register_in_tasks_manager("deepseek", *["text-generation", "text-generation-with-past"], library_name="transformers")
 class DeepseekOpenVINOConfig(MiniCPM3OpenVINOConfig):
-    MIN_TRANSFORMERS_VERSION = "4.46.0"
-    MAX_TRANSFORMERS_VERSION = "4.53.3"
+    MIN_TRANSFORMERS_VERSION = "4.53.0"
+    MAX_TRANSFORMERS_VERSION = None
```
IlyasMoutawwakil: why explicitly setting it to None ?

Mohamed-Ashraf273: `MiniCPM3OpenVINOConfig` (the parent class) sets `MAX_TRANSFORMERS_VERSION = "4.53.3"` because it was originally written when support for `transformers >= 4.54` was uncertain.

`DeepseekOpenVINOConfig` explicitly overrides this value to None to indicate that there is no upper bound. The `Deepseek/GigaChat3` export has been validated with transformers up to `4.57.x`, and it is expected to remain compatible with future versions.

Without this override, `DeepseekOpenVINOConfig` would inherit the `4.53.3` limit from `MiniCPM3OpenVINOConfig`, which would incorrectly block export on newer transformers versions. Setting it to None explicitly removes that inherited ceiling and ensures the correct behavior.

Mohamed-Ashraf273: Hi @IlyasMoutawwakil,
I’d really appreciate it if you could review my latest updates.
Thanks in advance!

popovaan (approved):

optimum/exporters/openvino/convert.py
```diff
@@ -721,7 +721,27 @@ def export_from_model(
...
+                    from pathlib import Path as _Path
+
+                    gen_cfg_path = _Path(output) / "generation_config.json"
+                    if gen_cfg_path.exists():
+                        with open(gen_cfg_path, "r", encoding="utf-8") as _f:
+                            _cfg = _json.load(_f)
+                        if _cfg.get("transformers_version") != orig_transformers_version:
+                            _cfg["transformers_version"] = orig_transformers_version
+                            with open(gen_cfg_path, "w", encoding="utf-8") as _f:
+                                _json.dump(_cfg, _f, indent=2)
```
popovaan: Can we avoid this change? This modifies the common code for all models, which is undesirable.

Mohamed-Ashraf273: Thanks for your feedback!

I reverted the changes and set `gen_config.do_sample = False` specifically for `deepseek` in `test_decoder.py`.

Could you please take another look and let me know if anything else should be adjusted?

Thanks!

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4051,6 +3968,29 @@ def deepseek_moe_infer(self, x, topk_ids, topk_weight):
...
+    """
+    Replacement for DeepseekV3MoE.moe (transformers >= 4.57).
+    The original skips experts with no tokens (data-dependent control flow that breaks tracing).
+    This version unconditionally runs all experts to produce a traceable static graph.
+    """
+    final_hidden_states = torch.zeros_like(hidden_states, dtype=topk_weights.dtype)
+    expert_mask = torch.nn.functional.one_hot(topk_indices, num_classes=len(self.experts))
+    expert_mask = expert_mask.permute(2, 0, 1)
+
+    for expert_idx in range(len(self.experts)):
```
IlyasMoutawwakil: that's kinda inefficient, especially during decoding. do we replace this with some optimized MoE operator in openvino later ?

IlyasMoutawwakil (approved):
lgtm! approved to merge if all tests are passing, however we still don't solve the real perf issue in exported MoEs which is having an efficient implementation of either:
- torch.grouped_mm operator
- entire MoE operator

optimum/exporters/openvino/model_patcher.py
Mohamed-Ashraf273: Correct, this is inefficient for decoding. The current implementation intentionally runs all experts to avoid the data-dependent control flow in the original MoE (skipping experts with no tokens), which breaks `torch.jit.trace` required for OpenVINO export. So this change mainly serves as a temporary tracing workaround to produce a static, exportable graph.

Yes, the plan is to replace this with a custom OpenVINO MoE operator (similar to `convert_recurrent_attention_cell` used in `Qwen3Next`) so we can restore sparse execution and avoid loading all expert weights during decoding. This PR just unblocks model export for now, and the optimized operator is intended as a follow-up improvement.

Mohamed-Ashraf273:
Hi @popovaan , @rkazants 
Could you please take a look?
Please let me know if anything else is needed.

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: then maybe minimizing and standardizing the experts forward at least, in some models we use the batching trick to make experts exportable with minimal graph layout (the loop results in very long graphs), see http://github.com/huggingface/optimum-intel/blob/439b6319368c1667f3119ef508812ef167b0fef5/optimum/exporters/openvino/model_patcher.py#L7562 @rkazants

Mohamed-Ashraf273: Thanks for the suggestion!
I refactored the `DeepSeek MoE` patch to follow the same batching pattern used in the `AFMoE` implementation. The expert projections are now pre-stacked in the patcher, and the forward pass uses vectorized `bmm` operations instead of looping over experts, which helps keep the exported graph compact.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3771,250 +3771,201 @@ def __enter__(self):
...
+                # new interface (transformers >= 4.57): moe(self, hidden_states, topk_indices, topk_weights)
+                block.mlp._orig_moe = block.mlp.moe
+                block.mlp._orig_moe_infer = None
+                num_experts = len(block.mlp.experts)
+                block.mlp.gate_projs = (
+                    torch.concat(
+                        tuple(block.mlp.experts[i].gate_proj.weight.unsqueeze(0) for i in range(num_experts)),
+                        dim=0,
+                    )
+                    .transpose(1, 2)
```
popovaan: Please move the transposes of MoE expert weights to the patched `deepseek_moe` method. This is related to the specifics of OpenVINO transformations, which apply MoE optimizations only if the model follows a certain pattern and we need these transposes to match that pattern.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3771,250 +3771,201 @@ def __enter__(self):
...
+                    .float()
```
popovaan: Please apply the conversion to `float()` only for OpenVINO versions 2026.0 or earlier and add a warning that this model requires OpenVINO 2026.1 or later for better optimization. A fix in 2026.1 resolves the type issue related to `torch.bmm`, so this workaround will no longer be needed.

optimum/exporters/openvino/model_patcher.py
Mohamed-Ashraf273: Thanks for the feedback!
Moved transposes from patcher to `deepseek_moe` method as requested.

optimum/exporters/openvino/model_patcher.py
Mohamed-Ashraf273: Added OpenVINO version check with regex parsing; `float()` conversion and warning only for versions `<= 2026.0`.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3771,250 +3771,219 @@ def __enter__(self):
...
+
+                from packaging import version
+
+                import openvino as ov
+
+                ov_version_str = ov.__version__
+                version_match = re.match(r'(\d+\.\d+\.\d+)', ov_version_str)
+                if version_match:
+                    ov_version = version.parse(version_match.group(1))
+                    if ov_version <= version.parse("2026.0.0"):
```
popovaan: Please, use `is_openvino_version()` method here
https://github.com/huggingface/optimum-intel/blob/abe58d751b4e2cf88e05e87eaf9f01ee23dfded8/optimum/intel/utils/import_utils.py#L395

Mohamed-Ashraf273: Thanks for the suggestion!
I’ve updated it to use `is_openvino_version()`. Could you please take a look?
Thanks!

Mohamed-Ashraf273:
Hey @popovaan,

I’ve completed all the requested revisions. Could you please take a look?
Let me know if there’s anything else I should update or improve.
Thanks!

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3771,250 +3776,216 @@ def __enter__(self):
...
+
+                # Handle OpenVINO version check
+                import re
+                import warnings
+
+                import openvino as ov
+
+                ov_version_str = ov.__version__
+                version_match = re.match(r'(\d+\.\d+\.\d+)', ov_version_str)
+                if version_match:
```
popovaan: This check no longer seems necessary.
Please keep only the `is_openvino_version()` check below.

Mohamed-Ashraf273: Thanks for the feedback!
You are right, I've removed the unnecessary check.
Let me know if there’s anything else I should update or improve.
Thanks!

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3760,8 +3765,8 @@ class DeepseekPatcher(OVDecoderModelPatcher):
     def __enter__(self):
         super().__enter__()
         self_attn = {
-            "deepseek_v3": deepseek_v3_attn_forward,
-            "deepseek_v2": deepseek_v2_attn_forward,
```
rkazants, line 3913: please avoid this refactoring. use existing patch + use new patch for MoE

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3771,250 +3776,207 @@ def __enter__(self):
...
-    attention_mask: Optional[torch.Tensor] = None,
-    position_ids: Optional[torch.LongTensor] = None,
-    past_key_value=None,
-    output_attentions: bool = False,
-    use_cache: bool = False,
-) -> Tuple[torch.Tensor, Optional[torch.Tensor], Optional[Tuple[torch.Tensor]]]:
-    # modified from https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/modeling_deepseek.py#L751
-    def rotate_half(x):
-        """Rotates half the hidden dims of the input."""
-        x1 = x[..., : x.shape[-1] // 2]
```
rkazants, line 3946: avoid this refactoring and below

optimum/exporters/openvino/model_patcher.py
Mohamed-Ashraf273: Thanks for your review!
I have reverted all the refactoring steps and added the necessary changes to support `gigachat3`

optimum/exporters/openvino/model_patcher.py
Mohamed-Ashraf273: Done!

Mohamed-Ashraf273:
@popovaan , @rkazants 
Also, here is the running test:
```python
from transformers import AutoModelForCausalLM, AutoTokenizer, BitsAndBytesConfig
import whowhatbench

model_id = "ai-sage/GigaChat3-10B-A1.8B-bf16"
base_model = AutoModelForCausalLM.from_pretrained(model_id)
quant_config = None
optimized_model = AutoModelForCausalLM.from_pretrained(model_id, quantization_config=quant_config, device_map="cpu")

evaluator = whowhatbench.TextEvaluator(base_model=base_model, tokenizer=tokenizer)
metrics_per_prompt, metrics = evaluator.score(optimized_model)

metric_of_interest = "similarity"
print(metric_of_interest, ": ", metrics["similarity"][0])

worst_examples = evaluator.worst_examples(top_k=5, metric=metric_of_interest)
print("Metric: ", metric_of_interest)
for e in worst_examples:
    print("\t=========================")
    print("\tPrompt: ", e["prompt"])
    print("\tBaseline Model:\n ", "\t" + e["source_model"])
    print("\tOptimized Model:\n ", "\t" + e["optimized_model"])

```
image

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3767,29 +3772,92 @@ def __enter__(self):
...
 def deepseek_v3_attn_forward(
     self,
     hidden_states: torch.Tensor,
     attention_mask: Optional[torch.Tensor] = None,
+    position_embeddings: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
     position_ids: Optional[torch.LongTensor] = None,
     past_key_value=None,
     output_attentions: bool = False,
     use_cache: bool = False,
+    cache_position: Optional[torch.LongTensor] = None,  # ← ADD THIS
```
```suggestion
    cache_position: Optional[torch.LongTensor] = None,
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3892,6 +3976,9 @@ def apply_rotary_pos_emb(q, k, cos, sin, position_ids, unsqueeze_dim=1):
 
     attn_output = self.o_proj(attn_output)
 
+    if new_interface:
+        return attn_output, None
```
rkazants, line 4160: can you please a comment why updated past_key_values is not returned?

rkazants (approved):

rkazants (changes requested):
none

rkazants (approved):

Mohamed-Ashraf273:
Hi @rkazants , @popovaan

I’ve completed all the requested revisions. Could you please take a look?
Let me know if there’s anything else I should update or improve.
Thanks!

popovaan (approved):
Looks good to me.

rkazants:
@Mohamed-Ashraf273, please rebase your PR

Mohamed-Ashraf273:
> @Mohamed-Ashraf273, please rebase your PR

Done!

popovaan:
There are some issues with transformers 5.0, @Mohamed-Ashraf273 could you please take a look?
https://github.com/huggingface/optimum-intel/actions/runs/24568820116/job/71837598487?pr=1626

```
FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_44_deepseek - TypeError: object of type 'DeepseekV3NaiveMoe' has no len()
FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_45_gigachat3 - TypeError: object of type 'DeepseekV3NaiveMoe' has no len()
```

Mohamed-Ashraf273:
> There are some issues with transformers 5.0, @Mohamed-Ashraf273 could you please take a look? https://github.com/huggingface/optimum-intel/actions/runs/24568820116/job/71837598487?pr=1626
> 
> ```
> FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_44_deepseek - TypeError: object of type 'DeepseekV3NaiveMoe' has no len()
> FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_45_gigachat3 - TypeError: object of type 'DeepseekV3NaiveMoe' has no len()
> ```

Hi @popovaan 
There are a few compatibility breaks with `transformers 5.x` for these models at the moment, so for now I limited support to `transformers < 5` to keep the current behavior stable.

A proper `transformers 5.x` migration looks bigger than a small fix. It touches model patching, cache handling/imports, masking behavior, and some config-loading paths, so I’d prefer to handle that in a separate PR focused on the full refactor rather than mixing it into this one.

Mohamed-Ashraf273:
> There are some issues with transformers 5.0, @Mohamed-Ashraf273 could you please take a look?
> 
> https://github.com/huggingface/optimum-intel/actions/runs/24568820116/job/71837598487?pr=1626
> 
> 
> 
> ```
> 
> FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_44_deepseek - TypeError: object of type 'DeepseekV3NaiveMoe' has no len()
> 
> FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_45_gigachat3 - TypeError: object of type 'DeepseekV3NaiveMoe' has no len()
> 
> ```

Okay Thanks,
I'm working on it!

Mohamed-Ashraf273:
Hi @popovaan ,
There are a few compatibility breaks with `transformers 5.x` for these models at the moment, so for now I limited support to `transformers < 5` to keep the current behavior stable.

A proper `transformers 5.x` migration looks bigger than a small fix.  I’d prefer to handle that in a separate PR focused on the full refactor rather than mixing it into this one.

Mohamed-Ashraf273:
Hi @popovaan 
Tests for `transformers==4.57.6` are passing:
```text
(env) mohamed-ashraf@mohamed-ashraf-LOQ-15IRX9:~/Desktop/projects/GSoC26/optimum-intel$ python -m pytest tests/openvino/*.py -xvs -k "deepseek or gigachat" 2>&1
============================================================================== test session starts ==============================================================================
platform linux -- Python 3.12.3, pytest-7.4.4, pluggy-1.6.0 -- /home/mohamed-ashraf/Desktop/projects/env/bin/python
cachedir: .pytest_cache
rootdir: /home/mohamed-ashraf/Desktop/projects/GSoC26/optimum-intel
configfile: pyproject.toml
plugins: anyio-4.12.1, mock-3.15.1, langsmith-0.7.10
collecting ... Multiple distributions found for package optimum. Picked distribution: optimum-intel
collecting 696 items                                                                                                                                                            The provided dataset won't have any effect on the resulting compressed model because no data-aware quantization algorithm is selected and compression ratio is 1.0.
The provided dataset won't have any effect on the resulting compressed model because no data-aware quantization algorithm is selected and compression ratio is 1.0.
collected 1281 items / 1270 deselected / 11 selected                                                                                                                            

tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_beam_search_46_deepseek SKIPPED (test is slow)
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_beam_search_47_gigachat3 SKIPPED (test is slow)
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_46_deepseek `torch_dtype` is deprecated! Use `dtype` instead!
This model works best with OpenVINO 2026.1 or later. Earlier versions require float() conversion for MoE weights, which may affect performance. OpenVINO 2026.1 includes a fix for torch.bmm dtype handling.
`loss_type=None` was set in the config but it is unrecognized. Using the default loss: `ForCausalLMLoss`.
`generation_config` default values have been modified to match model-specific defaults: {'temperature': 0.3, 'top_p': 0.95, 'bos_token_id': 100000}. If this is not desired, please set these values explicitly.
The following generation flags are not valid and may be ignored: ['temperature', 'top_p']. Set `TRANSFORMERS_VERBOSITY=info` for more details.
PASSED
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_47_gigachat3 This model works best with OpenVINO 2026.1 or later. Earlier versions require float() conversion for MoE weights, which may affect performance. OpenVINO 2026.1 includes a fix for torch.bmm dtype handling.
`generation_config` default values have been modified to match model-specific defaults: {'bos_token_id': 1}. If this is not desired, please set these values explicitly.
PASSED
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_pipeline_46_deepseek SKIPPED (test is slow)
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_pipeline_47_gigachat3 SKIPPED (test is slow)
tests/openvino/test_export.py::ExportModelTest::test_export_27_deepseek This model works best with OpenVINO 2026.1 or later. Earlier versions require float() conversion for MoE weights, which may affect performance. OpenVINO 2026.1 includes a fix for torch.bmm dtype handling.
This model works best with OpenVINO 2026.1 or later. Earlier versions require float() conversion for MoE weights, which may affect performance. OpenVINO 2026.1 includes a fix for torch.bmm dtype handling.
PASSED
tests/openvino/test_export.py::ExportModelTest::test_export_28_gigachat3 '(ReadTimeoutError("HTTPSConnectionPool(host='huggingface.co', port=443): Read timed out. (read timeout=10)"), '(Request ID: 54be03ae-c0b2-42cd-9ed8-c690fc638d5b)')' thrown while requesting HEAD https://huggingface.co/optimum-intel-internal-testing/tiny-random-gigachat3/resolve/main/tokenizer_config.json
Retrying in 1s [Retry 1/5].
This model works best with OpenVINO 2026.1 or later. Earlier versions require float() conversion for MoE weights, which may affect performance. OpenVINO 2026.1 includes a fix for torch.bmm dtype handling.
This model works best with OpenVINO 2026.1 or later. Earlier versions require float() conversion for MoE weights, which may affect performance. OpenVINO 2026.1 includes a fix for torch.bmm dtype handling.
PASSED
tests/openvino/test_quantization.py::OVQuantizationConfigTest::test_named_default_configurations_49_deepseek_ai_DeepSeek_R1_Distill_Qwen_1_5B PASSED
tests/openvino/test_quantization.py::OVQuantizationConfigTest::test_named_default_configurations_50_deepseek_ai_DeepSeek_R1_Distill_Qwen_7B PASSED
tests/openvino/test_quantization.py::OVQuantizationConfigTest::test_named_default_configurations_51_deepseek_ai_DeepSeek_R1_Distill_Llama_8B PASSED

=============================================================================== warnings summary ================================================================================
../../../../../../usr/lib/python3.12/multiprocessing/popen_fork.py:66: 1 warning
tests/openvino/test_decoder.py: 118 warnings
tests/openvino/test_export.py: 223 warnings
  /usr/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=14952) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()

<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute

<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: builtin type SwigPyObject has no __module__ attribute

../../env/lib/python3.12/site-packages/torch/jit/_script.py:1480
  /home/mohamed-ashraf/Desktop/projects/env/lib/python3.12/site-packages/torch/jit/_script.py:1480: DeprecationWarning: `torch.jit.script` is deprecated. Please switch to `torch.compile` or `torch.export`.
    warnings.warn(

tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_46_deepseek
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_47_gigachat3
tests/openvino/test_export.py::ExportModelTest::test_export_27_deepseek
tests/openvino/test_export.py::ExportModelTest::test_export_27_deepseek
tests/openvino/test_export.py::ExportModelTest::test_export_28_gigachat3
tests/openvino/test_export.py::ExportModelTest::test_export_28_gigachat3
  /home/mohamed-ashraf/Desktop/projects/env/lib/python3.12/site-packages/torch/jit/_trace.py:1000: DeprecationWarning: `torch.jit.trace` is deprecated. Please switch to `torch.compile` or `torch.export`.
    warnings.warn(

tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_46_deepseek
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_47_gigachat3
tests/openvino/test_export.py::ExportModelTest::test_export_27_deepseek
tests/openvino/test_export.py::ExportModelTest::test_export_27_deepseek
tests/openvino/test_export.py::ExportModelTest::test_export_28_gigachat3
tests/openvino/test_export.py::ExportModelTest::test_export_28_gigachat3
  /home/mohamed-ashraf/Desktop/projects/env/lib/python3.12/site-packages/torch/jit/_trace.py:1139: DeprecationWarning: `torch.jit.trace_method` is deprecated. Please switch to `torch.compile` or `torch.export`.
    warnings.warn(

tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_46_deepseek
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_47_gigachat3
tests/openvino/test_export.py::ExportModelTest::test_export_27_deepseek
tests/openvino/test_export.py::ExportModelTest::test_export_28_gigachat3
  /home/mohamed-ashraf/Desktop/projects/env/lib/python3.12/site-packages/transformers/cache_utils.py:132: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    if not self.is_initialized or self.keys.numel() == 0:

tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_46_deepseek
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_47_gigachat3
tests/openvino/test_export.py::ExportModelTest::test_export_27_deepseek
tests/openvino/test_export.py::ExportModelTest::test_export_28_gigachat3
  /home/mohamed-ashraf/Desktop/projects/env/lib/python3.12/site-packages/transformers/masking_utils.py:207: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    if (padding_length := kv_length + kv_offset - attention_mask.shape[-1]) > 0:

tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_46_deepseek
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_47_gigachat3
tests/openvino/test_export.py::ExportModelTest::test_export_27_deepseek
tests/openvino/test_export.py::ExportModelTest::test_export_28_gigachat3
  /home/mohamed-ashraf/Desktop/projects/GSoC26/optimum-intel/optimum/exporters/openvino/model_patcher.py:264: TracerWarning: torch.tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
    torch.tensor(0.0, device=mask.device, dtype=dtype),

tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_46_deepseek
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_compare_to_transformers_47_gigachat3
tests/openvino/test_export.py::ExportModelTest::test_export_27_deepseek
tests/openvino/test_export.py::ExportModelTest::test_export_28_gigachat3
  /home/mohamed-ashraf/Desktop/projects/GSoC26/optimum-intel/optimum/exporters/openvino/model_patcher.py:265: TracerWarning: torch.tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
    torch.tensor(torch.finfo(torch.float16).min, device=mask.device, dtype=dtype),

tests/openvino/test_decoder.py: 58 warnings
  /home/mohamed-ashraf/Desktop/projects/GSoC26/optimum-intel/optimum/intel/openvino/modeling_decoder.py:820: DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments. To learn more, see the migration guide https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword
    np.array(beam_idx) if not self._second_iter_beam_search else self.next_beam_idx

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
==================================================== 7 passed, 4 skipped, 1270 deselected, 431 warnings in 113.07s (0:01:53) ====================================================
```
Formatting test:
```text
(env) mohamed-ashraf@mohamed-ashraf-LOQ-15IRX9:~/Desktop/projects/GSoC26/optimum-intel$ ruff check optimum/ --config pyproject.toml
warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `pyproject.toml`:
  - 'ignore' -> 'lint.ignore'
  - 'select' -> 'lint.select'
  - 'isort' -> 'lint.isort'
  - 'per-file-ignores' -> 'lint.per-file-ignores'
All checks passed!
```
and I disabled tests for `transformers > 5.x` so `CI` tests should be passing as well
Let me know if there’s anything else I should update or improve.
Thanks!

Mohamed-Ashraf273:
@popovaan

openvino-agent (dismissed):
no comments

rkazants (changes requested):
@Mohamed-Ashraf273, OV output for the real model (using script from your PR description) does not match the reference.

Please check that IR for language model contains states for KV-cache.

Mohamed-Ashraf273:
@rkazants , @popovaan 
I updated the PR description and retested the real model.

The exported IR does contain KV-cache states:
- ReadValue: 52
- Assign: 52
- query_state(): 52 states

The earlier OV/HF generation mismatch was not caused by missing KV-cache. It came from the comparison setup: the default export path for this large model applied weight compression, and the script compared a BF16 HF reference against OV FP16/INT8-compressed numerics. Since several greedy-decoding candidates have very close logits, that precision/compression difference can flip the selected token.

I updated the PR description to export with `--weight-format fp16`. I also verified separately against an FP16 HF reference; with that setup, generation matches exactly for the tested prompt, and the per-step logits are close. For example, at step 4 the max diff is ~0.06 and both models choose `Which`.

Example output:
Python script:
```python
import torch
from transformers import AutoTokenizer, AutoModelForCausalLM

model_id = "ai-sage/GigaChat3-10B-A1.8B-bf16"
model_dir = "output_dir"

ov_model = OVModelForCausalLM.from_pretrained(
    model_dir,
    ov_config={"INFERENCE_PRECISION_HINT": "f32"},
)

hf_model = AutoModelForCausalLM.from_pretrained(
    model_id,
    torch_dtype=torch.float16,
)

inputs.pop("token_type_ids", None)

ov_output_ids = ov_model.generate(**inputs, max_new_tokens=10)
ov_output_text = tokenizer.decode(ov_output_ids[0], skip_special_tokens=True)

hf_output_ids = hf_model.generate(**inputs, max_new_tokens=10)
hf_output_text = tokenizer.decode(hf_output_ids[0], skip_special_tokens=True)

print("=" * 50)
print("HF (REFERENCE):")
print(hf_output_text)

print("\nOV (OPENVINO):")
print(ov_output_text)

print("\nSame ids:", hf_output_ids.tolist() == ov_output_ids.cpu().tolist())
print("=" * 50)
```
Output:
```text
==================================================
HF (REFERENCE):
What is the capital of France?  
Paris

Which country has the largest population in

OV (OPENVINO):
What is the capital of France?  
Paris

Which country has the largest population in

Same ids: True
==================================================
```

Mohamed-Ashraf273:
> @rkazants , @popovaan I updated the PR description and retested the real model.
> 
> The exported IR does contain KV-cache states:
> 
> * ReadValue: 52
> * Assign: 52
> * query_state(): 52 states
> 
> The earlier OV/HF generation mismatch was not caused by missing KV-cache. It came from the comparison setup: the default export path for this large model applied weight compression, and the script compared a BF16 HF reference against OV FP16/INT8-compressed numerics. Since several greedy-decoding candidates have very close logits, that precision/compression difference can flip the selected token.
> 
> I updated the PR description to export with `--weight-format fp16`. I also verified separately against an FP16 HF reference; with that setup, generation matches exactly for the tested prompt, and the per-step logits are close. For example, at step 4 the max diff is ~0.06 and both models choose `Which`.
> 
> Example output: Python script:
> 
> ```python
> import torch
> from transformers import AutoTokenizer, AutoModelForCausalLM
> from optimum.intel.openvino import OVModelForCausalLM
> 
> model_id = "ai-sage/GigaChat3-10B-A1.8B-bf16"
> model_dir = "output_dir"
> 
> tokenizer = AutoTokenizer.from_pretrained(model_dir)
> 
> ov_model = OVModelForCausalLM.from_pretrained(
>     model_dir,
>     ov_config={"INFERENCE_PRECISION_HINT": "f32"},
> )
> 
> hf_model = AutoModelForCausalLM.from_pretrained(
>     model_id,
>     torch_dtype=torch.float16,
> )
> 
> prompt = "What is the capital of France?"
> inputs = tokenizer(prompt, return_tensors="pt")
> inputs.pop("token_type_ids", None)
> 
> ov_output_ids = ov_model.generate(**inputs, max_new_tokens=10)
> ov_output_text = tokenizer.decode(ov_output_ids[0], skip_special_tokens=True)
> 
> hf_output_ids = hf_model.generate(**inputs, max_new_tokens=10)
> hf_output_text = tokenizer.decode(hf_output_ids[0], skip_special_tokens=True)
> 
> print("=" * 50)
> print("HF (REFERENCE):")
> print(hf_output_text)
> 
> print("\nOV (OPENVINO):")
> print(ov_output_text)
> 
> print("\nSame ids:", hf_output_ids.tolist() == ov_output_ids.cpu().tolist())
> print("=" * 50)
> ```
> 
> Output:
> 
> ```
> ==================================================
> HF (REFERENCE):
> What is the capital of France?  
> Paris
> 
> Which country has the largest population in
> 
> OV (OPENVINO):
> What is the capital of France?  
> Paris
> 
> Which country has the largest population in
> 
> Same ids: True
> ==================================================
> ```

@rkazants , @popovaan 
Could you please take a look? I’d really appreciate it.

Thanks!

Mohamed-Ashraf273:
@rkazants , @popovaan
Could you please take a look? I’d really appreciate it.

Thanks!

popovaan (approved):

popovaan:
@rkazants please review this PR.

optimum/exporters/openvino/model_configs.py
```diff
@@ -4249,8 +4249,9 @@ class M2M100OpenVINOConfig(BartOpenVINOConfig):
 )
 @register_in_tasks_manager("deepseek", *["text-generation", "text-generation-with-past"], library_name="transformers")
 class DeepseekOpenVINOConfig(MiniCPM3OpenVINOConfig):
-    MIN_TRANSFORMERS_VERSION = "4.46.0"
-    MAX_TRANSFORMERS_VERSION = "4.53.3"
+    MIN_TRANSFORMERS_VERSION = "4.53.0"
+    # TODO: add v5 support
+    MAX_TRANSFORMERS_VERSION = "4.57.6"
```
```suggestion
    MAX_TRANSFORMERS_VERSION = "4.57.99"
```

optimum/exporters/openvino/model_patcher.py
rkazants: @Mohamed-Ashraf273, I think this question is still relevant. KV-cache should be returned with a new pair of KVs. Please check

optimum/exporters/openvino/model_configs.py
Mohamed-Ashraf273: Done!

optimum/exporters/openvino/model_patcher.py
Mohamed-Ashraf273: I checked the exported OpenVINO IR. The KV cache is present in the exported model as stateful OpenVINO variables. The XML contains 52 `ReadValue` and 52 `Assign` ops, matching 26 layers times key/value tensors. For example, layer 0 has `ReadValue`/`Assign` with variable id `past_key_values.0.keypresent.0.key`. The only external output is `logits` because after stateful conversion the `present.*` cache outputs are hidden as internal state.

Also, in the new Transformers DeepSeek-V3 interface, the attention cache is updated in-place here:
```python
key_states, value_states = past_key_values.update(
    key_states, value_states, self.layer_idx, cache_kwargs
)
```
Our patched code does the same in the new-interface branch:
```python
key_states, value_states = kv_cache.update(
    key_states, value_states, self.layer_idx, cache_kwargs
)
```
So returning only `(attn_output, None)` from attention seems correct for the new interface: the cache object is mutated in-place, the model-level forward still carries the updated cache, and the OpenVINO export captures it as state.

Here is the exported model link for your reference:
https://drive.google.com/file/d/1xakTYEGE0JRtTsuOZS86gGYABK_aeZSs/view?usp=drive_link

popovaan:
@rkazants can we merge the PR?

PR #1627, opened 2026-03-02
[OpenVINO] Add support for Mistral3

### LINKED ISSUE #1338: Please add support for  mistral3 models for openvino export
Need support for mistral3 model: https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503 

Currently it gives below error:

Image

kyoui-dev:
# What does this PR do?
Conversion cmd-line for `mistralai/Mistral-Small-3.1-24B-Instruct-2503`:
```bash
optimum-cli export openvino -m mistralai/Mistral-Small-3.1-24B-Instruct-2503 ./Mistral-Small-3.1-24B-Instruct-2503 --task image-text-to-text
```

Inference of `mistralai/Mistral-Small-3.1-24B-Instruct-2503` using OpenVINO backend:
```python
from transformers.image_utils import load_image

model_dir = "./Mistral-Small-3.1-24B-Instruct-2503"

# Prepare image input
image_path = hf_hub_download(
                repo_id="raushan-testing-hf/images_test",
                filename="australia.jpg",
                repo_type="dataset",
        )
image_input = load_image(image_path)
prompt = "Describe this image."
inputs = model.preprocess_inputs(processor=processor, text=prompt, image=image_input)

# Run inference

print(output_text)
```

Fixes #1338

## Before submitting
- N/A This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).


rkazants (commented):
please add tests and create tiny model for it

optimum/exporters/openvino/model_configs.py
```diff
@@ -257,6 +259,10 @@ def init_model_configs():
         "transformers",
         "AutoModelForVision2Seq",
     )
+    TasksManager._CUSTOM_CLASSES[("pt", "mistral3", "image-text-to-text")] = (
+        "transformers",
+        "AutoModelForImageTextToText",
+    )
```
echarlaix: can be removed as image-text-to-text should use `AutoModelForImageTextToText` in any case no ?
```suggestion
```

docs/source/openvino/models.mdx
```diff
@@ -105,6 +105,7 @@ Here is the list of the supported architectures :
 - MiniCPM-o
 - MiniCPMV
 - Mistral
+- Mistral3
```
```suggestion
- Mistral 3
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -2053,6 +2059,138 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
...
+    def generate(
+        self,
+        input_name: str,
+        framework: str = "pt",
+        int_dtype: str = "int64",
+        float_dtype: str = "fp32",
+    ):
+        input_dim = self.hidden_size * self.spatial_merge_size ** 2
+        shape = [self.num_merged_patches, input_dim]
+        return self.random_float_tensor(shape, framework=framework, dtype=float_dtype)
```
echarlaix: why not inherit fro `DummyLLavaMultiModalProjectorInputGenerator` and just add the `spatial_merge_size` attribute ? `num_merged_patches` can be inferred from `self.num_patches` no ? Also here the resulting input shape will not vary depending on `batch_size` could you extend on why ?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3307,6 +3307,66 @@ def __exit__(self, exc_type, exc_value, traceback):
         self._model.forward = self._model.__orig_forward
 
 
+# Adopted from https://github.com/huggingface/transformers/blob/main/src/transformers/models/mistral3/modeling_mistral3.py#L223-L248
```
echarlaix: thanks for adding a link! would you mind updating using any release branch like v5.2.0 (same comment for all links)
```suggestion
# Adopted from https://github.com/huggingface/transformers/blob/v5.2.0/src/transformers/models/mistral3/modeling_mistral3.py#L223-L248
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3307,6 +3307,66 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        selected_image_feature = image_features.hidden_states[vision_feature_layer]
+    else:
+        hs_pool = [image_features.hidden_states[layer_idx] for layer_idx in vision_feature_layer]
+        selected_image_feature = torch.cat(hs_pool, dim=-1)
+    image_features = self.multi_modal_projector.norm(selected_image_feature.squeeze(0))
+    return image_features
+
+
+# Adopted from https://github.com/huggingface/transformers/blob/main/src/transformers/models/mistral3/modeling_mistral3.py#L118-L124
+# Mistral3MultiModalProjector.forward() and Mistral3PatchMerger.forward() with norm and cycle block excluded.
```
echarlaix, line 3116: would you mind  adding a link to the cycle block excluded as well https://github.com/huggingface/transformers/blob/v5.2.0/src/transformers/models/mistral3/modeling_mistral3.py#L76-L94

optimum/exporters/openvino/model_configs.py
```diff
@@ -2053,6 +2059,138 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
...
+
+
+class Mistral3MultiModalProjectorOpenVINOConfig(OnnxConfig):
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyMistral3MultiModalProjectorInputGenerator,)
+    NORMALIZED_CONFIG_CLASS = NormalizedVisionConfig
+    _MODEL_PATCHER = Mistral3MultiModalProjectorPatcher
+
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        return {"image_features": {0: "num_patches"}}
```
echarlaix, line 2170: why not :
```suggestion
        return {"image_features": {0: "batch_size", 1: "sequence_length"}}
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -2053,6 +2059,138 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
         return LlavaNextVideoImageEmbeddingModelPatcher(self, model, model_kwargs)
 
 
+class Mistral3ConfigBehavior(str, enum.Enum):
+    LANGUAGE = "language"
+    VISION_EMBEDDINGS = "vision_embeddings"
+    TEXT_EMBEDDINGS = "text_embeddings"
+    MULTI_MODAL_PROJECTOR = "multi_modal_projector"
```
echarlaix, line 2160: here can you add a comment to explain that VISION_EMBEDDINGS + cycle block + MULTI_MODAL_PROJECTOR is  the equivalent of get_image_features (with links if possible)

echarlaix (commented):
thanks for the addition @kyoui-dev !

rkazants:
@kyoui-dev, please double-check that all newly added tests are passing locally on your machine:
{CEA5D74D-7014-441C-B491-4685ABF7D0E6}

optimum/exporters/openvino/model_configs.py
kyoui-dev: You're right. I'll remove it. Thanks for catching that!

optimum/exporters/openvino/model_configs.py
kyoui-dev: This is related to the `batch_size` discussion above — since the cycle block flattens all images' patches into a single 2D tensor `[total_merged_patches, dim]`, the projector input has only one dynamic dimension. So `{0: "num_patches"}` reflects the actual runtime shape.

optimum/exporters/openvino/model_patcher.py
kyoui-dev: Sure, I’ll update all links to use a release branch.

optimum/exporters/openvino/model_patcher.py
kyoui-dev: Sure, I’ll add that link as well. Thanks!

docs/source/openvino/models.mdx
kyoui-dev: Sure, I’ll update it.

optimum/exporters/openvino/model_configs.py
kyoui-dev: Sure, I'll add a comment with links.

optimum/exporters/openvino/model_configs.py
kyoui-dev: Good points! I'll refactor to inherit from DummyLLavaMultiModalProjectorInputGenerator and derive num_merged_patches from self.num_patches.
Regarding batch_size: in the original transformers implementation, Mistral3PatchMerger.forward() concatenates patches from all images via torch.cat(permuted_tensor, dim=0) into a flat 2D tensor [total_merged_patches, dim], because each image can have different spatial dimensions and thus produce a different number of merged patches — so they can't form a uniform batch axis. Since the [cycle block](https://github.com/huggingface/transformers/blob/main/src/transformers/models/mistral3/modeling_mistral3.py#L83-L92) runs in PyTorch at runtime, the projector submodel receives this already-flattened 2D input, which is why batch_size doesn't appear in the shape.

kyoui-dev:
@rkazants, I just ran the related tests locally and they passed on my end. Please let me know if anything else is needed.

tests/openvino/test_quantization.py
```diff
@@ -340,12 +340,14 @@ class OVQuantizerTest(unittest.TestCase):
...
-                "decoder": 52,
-            },
+            (
+                {"encoder": 30, "decoder": 52, "decoder_with_past": 61}
+                if is_transformers_version("<=", "4.45")
+                else {
+                    "encoder": 30,
+                    "decoder": 52,
+                }
+            ),
```
popovaan: There are some unrelated changes in the PR, probably due to code style applying? Please remove the changes from unrelated files.

kyoui-dev:
@popovaan,

I'm running the evaluation pipeline, but the smallest model for mistral3 is 24B, so it's taking quite a long time on my local machine. Is there another way I could run this? If not, I can still run it locally — it’ll just take a while.

```bash
Evaluate pipeline:   0%|                                                                                                                                              | 0/24 [00:00<?, ?it/s]Setting `pad_token_id` to `eos_token_id`:2 for open-end generation.
Evaluate pipeline:   4%|█████▎                                                                                                                        | 1/24 [2:47:16<64:07:11, 10036.15s/it]Setting `pad_token_id` to `eos_token_id`:2 for open-end generation.
Evaluate pipeline:   8%|██████████▎                                                                                                                 | 2/24 [22:30:16<280:24:41, 45885.54s/it]Setting `pad_token_id` to `eos_token_id`:2 for open-end generation.
Evaluate pipeline:  12%|███████████████▌                                                                                                            | 3/24 [37:40:36<290:55:56, 49874.13s/it]Setting `pad_token_id` to `eos_token_id`:2 for open-end generation.
Evaluate pipeline:  17%|████████████████████▋                                                                                                       | 4/24 [55:12:04<306:03:35, 55090.79s/it]Setting `pad_token_id` to `eos_token_id`:2 for open-end generation.
```

tests/openvino/test_seq2seq.py
```diff
@@ -729,7 +732,7 @@ def compare_outputs(inputs, ov_model, transformers_model, generation_config):
         transformers_model.config.eos_token_id = None
         ov_model.generation_config.do_sample = False
         # minicpmo diverges after 20 tokens
-        tokens_to_generate = 20 if model_arch == "minicpmo" else 30
+        tokens_to_generate = 20 if model_arch in ["minicpmo", "mistral3"] else 30
```
echarlaix: @kyoui-dev why is it needed ?

kyoui-dev: I initially added this because I was seeing a test failure during development, but I've since re-tested and it passes without the limitation now. I'll revert this change.

kyoui-dev:
@popovaan,

Here is my result from locally running the WWB benchmark evaluation. Please let me know if you'd like me to check anything else.

```bash
Evaluate pipeline:   0%|                                                                                                         | 0/4 [00:00<?, ?it/s]Setting `pad_token_id` to `eos_token_id`:2 for open-end generation.
Evaluate pipeline:  25%|███████████████████████▎                                                                     | 1/4 [32:21<1:37:05, 1941.67s/it]Setting `pad_token_id` to `eos_token_id`:2 for open-end generation.
Evaluate pipeline:  50%|█████████████████████████████████████████████▌                                             | 2/4 [4:44:42<5:23:32, 9706.10s/it]Setting `pad_token_id` to `eos_token_id`:2 for open-end generation.
Evaluate pipeline:  75%|███████████████████████████████████████████████████████████████████▌                      | 3/4 [8:14:05<3:03:30, 11010.61s/it]Setting `pad_token_id` to `eos_token_id`:2 for open-end generation.
Evaluate pipeline: 100%|███████████████████████████████████████████████████████████████████████████████████████████| 4/4 [12:25:05<00:00, 11176.40s/it]
Batches: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:03<00:00,  3.84s/it]
Batches: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  1.07it/s]
Batches: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  4.26it/s]
Batches: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  7.46it/s]
Similarity evaluation: 100%|█████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:05<00:00,  1.44s/it]
INFO:whowhatbench.wwb:Metrics for model: mistral3-openvino
INFO:whowhatbench.wwb:   similarity
0    0.951265
```

popovaan (approved):
Our team checked the similarity of this model, it’s ~0.99 for Ministral3-3B on a full WWB run, so the accuracy looks good.

popovaan:
@kyoui-dev Could you please update the branch and resolve conflicts?

@rkazants @echarlaix @IlyasMoutawwakil please review.

popovaan:
@kyoui-dev There are some issues with transformers 5.0, could you please take a look?
https://github.com/huggingface/optimum-intel/actions/runs/24864759126/job/72845975675?pr=1627
https://github.com/huggingface/optimum-intel/actions/runs/24864759126/job/72845975594?pr=1627
https://github.com/huggingface/optimum-intel/actions/runs/24864759126/job/72845975569?pr=1627

kyoui-dev:
@popovaan,

Thanks for the links! I’ve pushed an update. Could you please take another look?

AshutoshSinghIntel:
I tried the changes and encountered errors during evaluation. e.g. `ValueError: Tokenizer class TokenizersBackend does not exist or is not currently imported.`

To run the evaluation, I updated _tokenizer_config.json_ with following changes: 
1. "tokenizer_class": changed to "LlamaTokenizerFast" 
2. "extra_special_tokens": set to {}

Evaluation result: 
INFO:whowhatbench.wwb:   similarity
0    0.976109

@kyoui-dev Are you also changing above, for running the evaluation?

kyoui-dev:
Hi @AshutoshSinghIntel,

Thanks for checking that.

I ran the evaluation with `transformers==4.57.6`, and I didn’t encounter that issue. In my case, `tokenizer_config.json` was already set with the values you mentioned.

However, I tried exporting the model with Transformers v5, and it looks like the issue you mentioned can happen with that version. I’ll investigate the export behavior and check if we need to explicitly update `tokenizer_class` and `extra_special_tokens` during export.

Could you also confirm which transformers version you used?

AshutoshSinghIntel:
@kyoui-dev I see that `transformers - 4.56.2` was used. Thank you for looking into it.

kyoui-dev:
> I tried the changes and encountered errors during evaluation. e.g. `ValueError: Tokenizer class TokenizersBackend does not exist or is not currently imported.`
> 
> To run the evaluation, I updated _tokenizer_config.json_ with following changes:
> 
> 1. "tokenizer_class": changed to "LlamaTokenizerFast"
> 2. "extra_special_tokens": set to {}

@AshutoshSinghIntel, thanks for confirming the version.

In my case, when I exported the model artifacts with Transformers v5, `TokenizersBackend` was written into `tokenizer_config.json`. So this may happen if the artifacts were also exported with Transformers v5, while WWB evaluation was run with Transformers v4.

When the model/tokenizer is saved with Transformers v5, `tokenizer_config.json` can contain `"tokenizer_class": "TokenizersBackend"`. Since this tokenizer class is not available in Transformers v4, WWB running with `transformers==4.56.2` fails with:
```bash
ValueError: Tokenizer class TokenizersBackend does not exist or is not currently imported.
```

This also explains why manually changing `tokenizer_class` back to `LlamaTokenizerFast` and setting `extra_special_tokens` to `{}` makes the evaluation work in your environment.

I also tried exporting with Transformers v5 and running WWB evaluation with Transformers v5, but WWB still failed with a different compatibility issue:
```bash
ImportError: cannot import name 'AutoModelForVision2Seq' from 'transformers'
```

So for now, it looks like WWB is not fully compatible with the Transformers v5 VLM evaluation path yet. I think the safest option is to run both export and WWB evaluation with a Transformers v4 environment, or update WWB to fully support the Transformers v5 tokenizer/model API changes before evaluating v5-exported artifacts.

AshutoshSinghIntel:
Thanks @kyoui-dev , since you did not face any issues with transformers v4. I think, we can go ahead with next steps.

popovaan (approved):

rkazants (approved):

kyoui-dev:
Hi @IlyasMoutawwakil and @echarlaix,

Could you please take a look when you have a chance? I’d really appreciate your review so we can move this PR forward.

Thank you!

regisss (approved):
LGTM

rkazants (commented):
@kyoui-dev, can you please also check int8/int4 compression accuracy. For this, use wwb tool from OpenVINO GenAI.
If the default configs for int4/int8 provide low similarity, define custom configs.

Sorry. I really forgot about it. 
May be it makes sense to do this in a separate PR. how do you think?

kyoui-dev:
Hi @rkazants,

I agree with separating that into a follow-up PR.

Would it also be okay to include a small Ministral3 export fix in that follow-up PR? Although its model type is `mistral3`, its `text_config.model_type` is `ministral3`, so export currently fails.

Since the smallest Mistral3 model is 21B, running WWB with just 4 samples on my local machine would take about 5 days. So I was thinking of first adding the small fix for the Ministral3 issue, and then running the WWB benchmark with that model.

rkazants:
> Hi @rkazants,
> 
> I agree with separating that into a follow-up PR.
> 
> Would it also be okay to include a small Ministral3 export fix in that follow-up PR? Although its model type is `mistral3`, its `text_config.model_type` is `ministral3`, so export currently fails.
> 
> Since the smallest Mistral3 model is 21B, running WWB with just 4 samples on my local machine would take about 5 days. So I was thinking of first adding the small fix for the Ministral3 issue, and then running the WWB benchmark with that model.

Hi @kyoui-dev,

I see that its text part type is `mistral`: https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503/blob/main/config.json#L18
Does text-to-text export and inference work for this model? If it does not work, please correct text part export and inference in this PR? Regarding ministral, it will be a separate PR because it is a different model.

Best regards,
Roman

kyoui-dev:
@rkazants,

I added text-generation and text-generation-with-past to the Mistral3 task registration, and verified that both export and inference work with text-only input.

Export:
```bash
optimum-cli export openvino -m mistralai/Mistral-Small-3.1-24B-Instruct-2503 ./mistral3-text-generation --task text-generation
```

Inference:
```python

model_dir = "./mistral3-text-generation"

# Prepare input
inputs = processor(text=prompt, return_tensors="pt")

# Run inference

print(output_text)
```

kyoui-dev:
@rkazants,

I found one more Mistral3 export issue while testing the tiny model fixture.

When exporting Mistral3 without `torchvision` installed, the export itself completes successfully, but `preprocessor_config.json` / `processor_config.json` are not saved in the output directory:

```bash
optimum-cli export openvino \
  -m optimum-intel-internal-testing/tiny-random-mistral3 \
  ./mistral3-openvino \
  --task image-text-to-text \
  --trust-remote-code
```

After installing `torchvision`, the same export correctly saves those files.

The root cause does not seem to be the OpenVINO conversion itself. Rather, this looks like a Pixtral-family processor dependency issue exposed by Mistral3 export, combined with silent error handling in the preprocessor loading path.

The export flow goes through `load_preprocessors()` and then Optimum’s `maybe_load_preprocessors()`. That helper attempts to load objects such as `AutoTokenizer`, `AutoProcessor`, and `AutoImageProcessor`, but it silently skips failures while doing so. Later, the save step only calls `save_pretrained()` on the objects that were successfully loaded.

Mistral3 uses a Pixtral-family multimodal processor, and that processor path depends on the `torchvision`-backed image processor path. In particular, the Pixtral fast image processor imports `torchvision.transforms`, and the newer Pixtral processor path also relies on the `torch` / `torchvision` backend. Therefore, when `torchvision` is not installed, loading `AutoProcessor` / `AutoImageProcessor` fails. Since that failure is swallowed, the export continues successfully, but there is no processor object to save.

This makes the exported directory incomplete. Later usage such as:

```python
processor = AutoProcessor.from_pretrained("./mistral3-openvino")
```

cannot recover the Mistral3 processor config from the exported artifact.

Possible fixes would be to either:

- make `torchvision` an explicit dependency for Mistral3 / Pixtral image-text export, or
- make the exporter raise/warn when processor loading fails for `image-text-to-text`, instead of silently producing incomplete artifacts.

popovaan:
@kyoui-dev please fix code style in the PR.

kyoui-dev:
@popovaan,

Applied code style fixes. PTAL 🙏

popovaan:
The "text-generation" task seems to be not covered by test, please check this error:
https://github.com/huggingface/optimum-intel/actions/runs/25992307476/job/76442691823?pr=1627
https://github.com/huggingface/optimum-intel/actions/runs/25992307476/job/76442691885?pr=1627

rkazants (approved):

rkazants:
@kyoui-dev, please fix the following failures:
FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_find_untested_architectures - ValueError: For the task `text-generation`, the OpenVINO exporter supports {'mistral3'} which are not tested

other part looks good to me.

AshutoshSinghIntel:
Hi @kyoui-dev, I am blocked on this PR for enabling Ministral model. I see that there are pending comments, do you plan to finalize it?

kyoui-dev:
Hi @AshutoshSinghIntel,
Sorry for the delay. I'll push an update by end of day. Thanks for checking in!

kyoui-dev:
Hi @popovaan and @rkazants,

I’ve pushed an update. Please take another look when you have a chance.

Thank you!

kyoui-dev:
@rkazants,

I've synced the branch with the latest `main` and pushed follow-up fixes for the Mistral3 OpenVINO export config.

Please take another look when you have a chance.

Thank you!

rkazants (commented):
@kyoui-dev, your inference script does not work for transformers=5.0.0
It looks that not all preprocessor files are exported. Please check

kyoui-dev:
@rkazants,

I investigated the missing `processor_config.json` / `preprocessor_config.json` issue for the Mistral3 export.

The issue seems to be caused by a missing `Pillow` runtime dependency. The Pixtral/Mistral3 image processor requires `Pillow`, but `Pillow` is currently only included in the test dependencies. In an environment without `Pillow`, `maybe_load_preprocessors()` only returns the tokenizer. The image processor / processor loading failure is swallowed, so the export succeeds but only tokenizer files are saved. As a result, the exported artifact misses `processor_config.json` and `preprocessor_config.json`.

After installing `Pillow`, the same export command produces the expected processor artifacts. So I think the correct fix here is to add `Pillow` as a runtime dependency for this package/export path.

I verified locally that:

```bash
optimum-cli export openvino \
  -m optimum-intel-internal-testing/tiny-random-mistral3 \
  ./mistral3-openvino \
  --task image-text-to-text
  ```

saves the expected `processor_config.json` and `preprocessor_config.json` files once Pillow is installed.

**Comment by ash1shb1sht** [comment]
Hi @kyoui-dev, I recently tried exporting the `mistralai/Mistral-Small-3.2-24B-Instruct-2506` model to OpenVINO using optimum commit *7b35631*. It looks like the `processor_config.json` and `preprocessor_config.json` files are not being generated, even though Pillow is installed.

For comparison, when I export the sample model `optimum-intel-internal-testing/tiny-random-mistral3` using the same commit (7b35631), both of these files are generated as expected.

Could you please take a look when you have a chance? Thanks!

**Comment by kyoui-dev** [comment]
Hi @ash1shb1sht ,

Thanks for checking this and for the detailed repro.

I was able to reproduce the same behavior with `mistralai/Mistral-Small-3.2-24B-Instruct-2506`.

After looking into it, the difference seems to come from the processor/image processor assets available in the source checkpoint. The tiny model was created based on `mistralai/Mistral-Small-3.1-24B-Instruct-2503`, which provides the HF processor/image processor files, so the exporter can load the processor and save `processor_config.json` and `preprocessor_config.json` as expected.

For `mistralai/Mistral-Small-3.2-24B-Instruct-2506`, the current preprocessor loading path does not seem to be able to construct an HF processor/image processor object for this checkpoint. As a result, there is no processor object for the exporter to save, so `processor_config.json` and `preprocessor_config.json` are missing from the output directory.

Since this PR targets the HF-format Mistral3 case, I think full support for this 3.2 checkpoint is outside its scope and would be better handled separately, as it likely requires additional handling for its processor/image processor assets.

**Comment by rkazants** [comment]
Hi @kyoui-dev, can you please rebase your PR? Check that your PR works for https://huggingface.co/mistralai/Mistral-Small-3.2-24B-Instruct-2506. If you will add proper support, we can merge this PR.

Thanks for your effort.

**Comment by kyoui-dev** [comment]
@rkazants,

I've rebased the PR onto the latest main and added support for `mistralai/Mistral-Small-3.2-24B-Instruct-2506`. Please check.

**Comment by kyoui-dev** [comment]
@rkazants,

I've rebased the PR onto the latest main again. Please check.

**Comment by kyoui-dev** [comment]
@rkazants,

I've fixed the export test failure for the bfloat16 tiny checkpoint. Based on the previous CI run, I think that was the only failure related to this PR. Please check.

**Comment by kyoui-dev** [comment]
@rkazants,
I've rebased the PR onto the latest main again. Please check.

File: setup.py [hunk 10]
```diff
@@ -36,6 +36,7 @@
     "openvino>=2026.0",
     "openvino-tokenizers>=2026.0",
     "requests>=2.33,<3.0",
+    "mistral-common[image]",
```
rkazants: we don't need it. Please remove it

rkazants (changes requested):
new dependency model specific

setup.py
kyoui-dev: Removed it. Thanks!

rkazants (approved):

rkazants:
@kyoui-dev, please rebase to the latest main. It should help to resolve red CI

kyoui-dev:
@rkazants,

I've rebased the PR onto the latest main. Please check.

rkazants:
@anatyrova, still observe segfault in seq2seq. Could you please take a look?

anatyrova:
@kyoui-dev it looks like the mistral3 gives the segfault on CI (it is current plugin issue in openvino). please, try to skip mistral3, adding it to SEGFAULTING_ARCHITECTURES = {"gemma4"} in test_seq2seq.
```
SEGFAULTING_ARCHITECTURES = {"gemma4", "mistral3"}
``` 
it will only be fixed with the new OV.

anatyrova:
> @kyoui-dev it looks like the mistral3 gives the segfault on CI (it is current plugin issue in openvino). please, try to skip mistral3, adding it to SEGFAULTING_ARCHITECTURES = {"gemma4"} in test_seq2seq.
> 
> ```
> SEGFAULTING_ARCHITECTURES = {"gemma4", "mistral3"}
> ```
> 
> it will only be fixed with the new OV.

@kyoui-dev i discussed it with Roman, can you try to regenerate your tiny model mistral3? it very possibly might help. there are special agentic skills for it in optimum-intel, check them pls, it can be helpful. https://github.com/huggingface/optimum-intel/tree/main/.github/skills/tiny-model-creator

kyoui-dev:
Hi @anatyrova,

I regenerated the Mistral3 tiny model locally and updated the tests to use the locally generated model instead of the Hub fixture.

Please check.

tests/openvino/utils_tests.py
```diff
@@ -144,6 +144,56 @@ def _create_tiny_kokoro_model():
     return str(output_dir)
 
 
+def _create_tiny_mistral3_model():
```
echarlaix, line 147: @kyoui-dev would you mind removing `_create_tiny_mistral3_model` and update https://huggingface.co/optimum-intel-internal-testing/tiny-random-mistral3 instead so that we can use it in the tests

kyoui-dev: Hi @echarlaix,

Sure, I can remove it. 

However, it looks like I don't currently have permission to update the tiny model on HF. Could you please check whether I have permission to update `optimum-intel-internal-testing/tiny-random-mistral3`?

echarlaix: thanks @kyoui-dev ! would you mind opening a PR for this ? https://huggingface.co/optimum-intel-internal-testing/tiny-random-mistral3/discussions

kyoui-dev: @echarlaix,

Thanks! I've opened a PR to update the tiny Mistral3 model: [https://huggingface.co/optimum-intel-internal-testing/tiny-random-mistral3/discussions/1](https://huggingface.co/optimum-intel-internal-testing/tiny-random-mistral3/discussions/1). Could you please take a look?

Once it's merged, I'll open a follow-up PR to remove the local model generation code and use the updated HF model in the tests.

echarlaix: thanks a lot @kyoui-dev !  your PR is now merged

kyoui-dev: @echarlaix,

I've opened a follow-up PR #2017. Please check.

urakozz:
Oh really, after a half a year, congratulations!

PR #1632, opened 2026-03-04
[OpenVINO] Fix image plus video scenario in Qwen3VL.

popovaan:
# What does this PR do?

Fixes the bug when both image and video are passed to Qwen3VL. 

Ticket: CVS-182297

## Before submitting

### DIFF
--- a/tests/openvino/test_seq2seq.py
+++ b/tests/openvino/test_seq2seq.py
@@ -635,7 +635,7 @@ class OVModelForVisualCausalLMIntegrationTest(OVSeq2SeqTestMixin):
             ov_outputs = ov_model.generate(**inputs, generation_config=generation_config)
             # original minicpmv, internvl always skip input tokens in generation results, while transformers based approach provide them
             if model_arch in ["minicpmv", "minicpmo", "internvl_chat"]:
-                ov_outputs = ov_outputs[:, inputs["input_ids"].shape[1]:]
+                ov_outputs = ov_outputs[:, inputs["input_ids"].shape[1] :]
             with torch.no_grad():
                 transformers_outputs = transformers_model.generate(
                     **transformers_inputs, generation_config=generation_config, **additional_inputs


optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3695,7 +3695,7 @@ def get_multimodal_embeddings(
             image_mask_joint = image_mask[visual_pos_masks]
             video_mask_joint = video_mask[visual_pos_masks]
             for img_embed, vid_embed in zip(deepstack_image_embeds, deepstack_video_embeds):
-                embed_joint = img_embed.new_zeros(visual_pos_masks.sum(), img_embed.shape[-1]).to(img_embed.device)
+                embed_joint = np.zeros((visual_pos_masks.sum(), len(img_embed[0])))
```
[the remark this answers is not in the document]
popovaan: Done.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3695,7 +3695,7 @@ def get_multimodal_embeddings(
...
                 embed_joint[image_mask_joint, :] = img_embed
```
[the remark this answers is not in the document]
popovaan: Tested it, no type error happens.

tests/openvino/test_seq2seq.py
```diff
@@ -630,6 +630,21 @@ def test_find_untested_architectures(self):
 
     @parameterized.expand(SUPPORTED_ARCHITECTURES)
     def test_compare_to_transformers(self, model_arch):
+        def compare_outputs(inputs, ov_model, transformers_model, generation_config):
+            transformers_inputs = copy.deepcopy(inputs)
+            ov_outputs = ov_model.generate(**inputs, generation_config=generation_config)
```
rkazants, line 635: why we put generate call method to auxiliary function for comparation? I think we should avoid it.

popovaan: It is just a minor refactoring, there was a code duplication where outputs from transformers are compared with outputs from OV. This block is repeated a few times, so I made a separate method for it.

popovaan: I can remove this, it is not required for the fix, but the test looks weird in that case with a few identical repeated blocks.

rkazants: okay, let us leave it then

rkazants (approved):

tests/openvino/test_seq2seq.py
echarlaix, line 777: why not call  `compare_outputs` here as well ?

rkazants: it is minor thing. Let us fix it in a separate PR.

PR #1637, opened 2026-03-13
support videochat

xufang-lisa:
Conversion cmd-line for OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B:
`optimum-cli export openvino -m OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B ./VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B`

Inference of OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B using OpenVINO backend:

```
from transformers.video_utils import load_video

model_dir = "./VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B"

model = OVModelForVisualCausalLM.from_pretrained(model_dir, trust_remote_code=True, device="cpu")

# Prepare video input
video_path = "./253997_tiny.mp4"
input_video, _ = load_video(video_path, num_frames=8, backend="opencv")
question = "Describe this video in detail."

# preprocess inputs
tokenizer = AutoTokenizer.from_pretrained(model_dir, trust_remote_code=True)
inputs = model.preprocess_inputs(processor=None, tokenizer=tokenizer, text=question, video=input_video, config=model.config)

# Run inference
input_prompt_len = inputs["input_ids"].shape[-1]
generated = output_ids[:, input_prompt_len:]
output_text = tokenizer.decode(generated[0])

print(output_text)

```

## Before submitting


setup.py
```diff
@@ -63,6 +63,9 @@
     "vocos",
     "vector_quantize_pytorch",
     "openvino-genai",
+    "av",
+    "decord",
+    "imageio",
```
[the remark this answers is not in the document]
xufang-lisa: `av`, `decord` and` imageio` are imported in custom remote code of `videochat-flash-qwen`, they can't be removed.

optimum/exporters/openvino/convert.py
```diff
@@ -364,7 +364,8 @@ def export_pytorch(
             logger.info(f"Overriding {len(config.values_override)} configuration item(s)")
             for override_config_key, override_config_value in config.values_override.items():
                 logger.info(f"\t- {override_config_key} -> {override_config_value}")
-                setattr(model.config, override_config_key, override_config_value)
+                if hasattr(model, "config"):
+                    setattr(model.config, override_config_key, override_config_value)
```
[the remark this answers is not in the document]
xufang-lisa: Move the check of `.config` before logs.

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5306,205 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+        self,
+        task: str,
+        normalized_config: NormalizedTextConfig,
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        random_batch_size_range: Optional[Tuple[int, int]] = None,
+        **kwargs,
+    ):
+        self.task = task
+        self.batch_size = batch_size
+        self.hidden_size = normalized_config.mm_hidden_size
```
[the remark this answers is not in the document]
xufang-lisa: updated

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7640,6 +7640,140 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            past_key_values=None,
+            inputs_embeds=None,
+        ):
+            from transformers.cache_utils import DynamicCache
+
+            outputs, labels = self.model(
+                input_ids=None,
+                attention_mask=attention_mask,
+                position_ids=position_ids,
+                past_key_values=past_key_values,
```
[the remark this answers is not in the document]
xufang-lisa: done

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4805,885 @@ def preprocess_inputs(
...
+        """
+        original_width, original_height = original_size
+        best_fit = None
+        max_effective_resolution = 0
+        min_wasted_resolution = float("inf")
+
+        for width, height in possible_resolutions:
+            if max_resolutions != None and (width * height != patch_size * patch_size):
+                if (width * height + patch_size * patch_size) > max_resolutions:  # NOTE 要算一个global
+                    continue
```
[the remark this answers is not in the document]
xufang-lisa: done

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5306,205 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+        **kwargs,
+    ):
+        super().__init__(task, normalized_config, batch_size, num_channels, width, height, visual_seq_length, **kwargs)
+        if hasattr(normalized_config, "config") and hasattr(normalized_config.config, "mm_local_num_frames"):
+            self.num_frames = normalized_config.config.mm_local_num_frames
+        self.height = 224
+        self.width = 224
+        self.image_size = (self.height, self.width)
+        self.patch_size = 14
+        self.embed_dim = normalized_config.config.mm_hidden_size
```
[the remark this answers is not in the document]
xufang-lisa: done

rkazants (changes requested):
Please update documentation, avoid changes in openvino.py and convert.py files.
Check other PRs with new model support

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5306,206 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+        input_name: str,
+        framework: str = "pt",
+        int_dtype: str = "int64",
+        float_dtype: str = "fp32",
+    ):
+        shape = [self.batch_size, self.num_patches, self.hidden_size]
+        return self.random_float_tensor(shape, framework=framework, dtype=float_dtype)
+
+
+class VideoChatFlashQWENProjectorOpenVINOConfig(OnnxConfig):
```
```suggestion
class VideoChatFlashQwenProjectorOpenVINOConfig(OnnxConfig):
```

tests/openvino/test_seq2seq.py
```diff
@@ -875,13 +890,16 @@ def test_generate_utils(self, model_arch):
...
         )
         tokenizer = AutoTokenizer.from_pretrained(model_id, trust_remote_code=trust_remote_code)
-        question = "Describe image"
         preprocessors = self.get_preprocessors(model_arch)
-        inputs = model.preprocess_inputs(**preprocessors, text=question, image=self.IMAGE.resize((600, 600)))
-        # General case
-        outputs = model.generate(**inputs, max_new_tokens=10)
-        outputs = tokenizer.batch_decode(outputs[:, inputs["input_ids"].shape[1] :], skip_special_tokens=True)
-        self.assertIsInstance(outputs[0], str)
+
```
rkazants: the limitation of this test is that comparison with inference results generated by reference code is absent

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,889 @@ def preprocess_inputs(
...
+            if "qwen" in tokenizer.name_or_path.lower():
+                logger.info("Setting pad token to bos token for qwen model.")
+                tokenizer.pad_token_id = 151643
+        attention_masks = input_ids.ne(tokenizer.pad_token_id).long()
+        results["attention_mask"] = attention_masks
+
+        return results
+
+    def encode_video_image(self, images_list, video_idx_in_batch):
+        # video encoder编码后按图像的connector处理
```
rkazants: please correct comment

optimum/exporters/openvino/model_configs.py
xufang-lisa: Done.

tests/openvino/test_seq2seq.py
xufang-lisa: Done, comparison inference results with transformers has been added in `test_compare_to_transformers`.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Done.

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5306,210 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+class VideoChatFlashQwenConfigBehavior(str, enum.Enum):
+    LANGUAGE = "language"
+    VISION_EMBEDDINGS = "vision_embeddings"
+    VISION_PROJECTION = "vision_projection"
+    TEXT_EMBEDDINGS = "text_embeddings"
+
+
+@register_in_tasks_manager("videochat_flash_qwen", *["image-text-to-text"], library_name="transformers")
+class VideoChatFlashQwenOpenVINOConfig(BaseVLMOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.42.0"
```
peterchen-intel: Model from HF: Min?, Max 4.48.x
Model from modescope: Min 4.46?,  Max 4.57.x
```suggestion
    MIN_TRANSFORMERS_VERSION = "For both"
    MAX_TRANSFORMERS_VERSION = "4.57.99"
```
Throw exception in constructor

optimum/intel/openvino/quantization.py
```diff
@@ -1472,6 +1472,12 @@ def _quantize_ovbasemodel(
                 ov_model_name, pipeline_quantization_config.default_config
             )
             if config is None:
+                if immediate_save:
+                    # Unload skipped submodels early so their IR files are not held open on Windows.
+                    # This can avoid later _merge_move failures caused by locked .bin files.
+                    ov_model = self.model.ov_models[ov_model_name]
+                    if ov_model is not None:
+                        self.model._unload_ov_model(ov_model)
```
peterchen-intel, line 1485: Can do it in configuration file to skip weights compress for specific model.

peterchen-intel:
Update description by taking the template from https://github.com/huggingface/optimum-intel/pull/1551

optimum/exporters/openvino/model_configs.py
xufang-lisa: Update MIN_TRANSFORMERS_VERSION and MAX_TRANSFORMERS_VERSION.

optimum/intel/openvino/quantization.py
peterchen-intel: This model is not in the default configuration list, so it is skipped here for weights compression, and indeed it should be _unload_ov_model.   @rkazants

optimum/exporters/openvino/model_configs.py
xufang-lisa: Add exception in constructor.

rkazants:
Hi @echarlaix, @IlyasMoutawwakil, could you please review this PR?

Thanks,
Roman

optimum/exporters/openvino/model_configs.py
```diff
@@ -200,6 +200,9 @@
     Qwen3VLVisionEmbMergerPatcher,
     QwenModelPatcher,
     SanaTextEncoderModelPatcher,
+    VideochatFlashQwenLanguageModelPatcher,
+    VideochatFlashQwenVisionEmbeddingModelPatcher,
+    VideochatFlashQwenVisionProjectionModelPatcher,
```
```suggestion
    VideoChatFlashQwenLanguageModelPatcher,
    VideoChatFlashQwenVisionEmbeddingModelPatcher,
    VideoChatFlashQwenVisionProjectionModelPatcher,
```
for consistency

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5306,220 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+        width: int = DEFAULT_DUMMY_SHAPES["width"],
+        height: int = DEFAULT_DUMMY_SHAPES["height"],
+        visual_seq_length: int = DEFAULT_DUMMY_SHAPES["visual_seq_length"],
+        **kwargs,
+    ):
+        super().__init__(task, normalized_config, batch_size, num_channels, width, height, visual_seq_length, **kwargs)
+        self.num_frames = getattr(normalized_config.config, "mm_local_num_frames", 4)
+        self.embed_dim = getattr(normalized_config.config, "mm_hidden_size", 1408)
+        self.height = 224
+        self.width = 224
```
rkazants: please make a comment why they are fixed

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5306,220 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+        task: str,
+        normalized_config: NormalizedTextConfig,
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        random_batch_size_range: Optional[Tuple[int, int]] = None,
+        **kwargs,
+    ):
+        self.task = task
+        self.batch_size = batch_size
+        self.hidden_size = normalized_config.config.mm_hidden_size
+        self.num_patches = 64
```
rkazants, line 5413: please make a comment why it is fixed.

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5306,220 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+class VideoChatFlashQwenConfigBehavior(str, enum.Enum):
+    LANGUAGE = "language"
+    VISION_EMBEDDINGS = "vision_embeddings"
+    VISION_PROJECTION = "vision_projection"
+    TEXT_EMBEDDINGS = "text_embeddings"
+
+
+@register_in_tasks_manager("videochat_flash_qwen", *["image-text-to-text"], library_name="transformers")
+class VideoChatFlashQwenOpenVINOConfig(BaseVLMOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.45.0"
```
```suggestion
    MIN_TRANSFORMERS_VERSION = "4.49.0"
```
let use this limitation and remove below check with exception

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5306,220 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+        if self._behavior == VideoChatFlashQwenConfigBehavior.VISION_EMBEDDINGS and hasattr(config, "vision_config"):
+            self._config = config.vision_config
+            self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
+
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        if not self._behavior == VideoChatFlashQwenConfigBehavior.VISION_EMBEDDINGS:
+            return {}
+        return {
+            "hidden_states": {0: "batch_size", 1: "num_channels", 2: "num_frames", 3: "height", 4: "width"},
```
rkazants: strange that `num_channels` is dynamic, please double-check

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7640,6 +7640,138 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            residual = None
+
+            for idx, blk in enumerate(self.blocks):
+                if isinstance(hidden_states, tuple) and len(hidden_states) == 2:
+                    hidden_states, residual = hidden_states
+                hidden_states = blk(hidden_states, residual=residual)
+
+            if isinstance(hidden_states, tuple) and len(hidden_states) == 2:
+                hidden_states, residual = hidden_states
+                if residual is not None:
```
rkazants: why do we need this check? Is any model were `residial` is `None`

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7640,6 +7640,138 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            B, T, L, C = hidden_states.shape  # T: temporal; L: spatial
+            hidden_states = hidden_states.view([B, T * L, C])
+
+            # append cls token
+            cls_tokens = self.cls_token.expand(B, -1, -1)
+            hidden_states = torch.cat((cls_tokens, hidden_states), dim=1)
+
+            # add pos_embed
+            if self.sep_pos_embed:
+                raise NotImplementedError
```
rkazants: Not sure that this is needed. Please clean code in this patcher.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7640,6 +7640,138 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+
+class VideochatFlashQwenVisionProjectionModelPatcher(ModelPatcher):
+    def __init__(
+        self,
+        config: "OnnxConfig",
+        model: "PreTrainedModel",
+        model_kwargs: Dict[str, Any] = None,
+    ):
+        model.__orig_forward = model.forward
```
rkazants: why do you need this patching? You just need to use original names and patching is not needed

rkazants: please remove this patch

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -1,14 +1,16 @@
+import ast
```
rkazants: this module is quite old and release last time in 2017. Do we really need this one?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,926 @@ def preprocess_inputs(
...
+        """
+        assert embed_dim % 4 == 0
+        embed_dim_spatial = embed_dim // 4 * 3
+        embed_dim_temporal = embed_dim // 4
+
+        # spatial
+        grid_h = np.arange(grid_size, dtype=np.float32)
+        grid_w = np.arange(grid_size, dtype=np.float32)
+        grid = np.meshgrid(grid_w, grid_h)  # here w goes first
+        grid = np.stack(grid, axis=0)
```
rkazants: can you please torch? I think we mostly use torch in processing parts in optimum-intel. Let us be aligned.

rkazants: The same comment is for code below

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,926 @@ def preprocess_inputs(
...
+                            ) = _OVVideoChatFlashQwenForCausalLM.get_anyres_image_grid_shape(
+                                image_sizes[image_idx],
+                                self.config.image_grid_pinpoints,
+                                vision_tower_image_size,
+                                max_resolutions=None,
+                            )
+                        except Exception:
+                            logger.exception("Error while computing anyres image grid shape")
+                            raise
+                            # num_patch_width, num_patch_height = 2, 2
```
rkazants: why do we need try-catch block?

rkazants (commented):
Please clean-up code to remove extra checks with NotImplemented exceptions, try-catch blocks. I don't understand the meaning of these checks because now we don't expect any model that will fall into them.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,926 @@ def preprocess_inputs(
...
+        )
+        self.pos_embed.data.copy_(torch.from_numpy(pos_embed).float().unsqueeze(0))
+
+        img_pos_embed = _OVVideoChatFlashQwenForCausalLM.get_3d_sincos_pos_embed(
+            self.pos_embed.shape[-1], self.grid_size[1], 1, cls_token=True
+        )
+        self.img_pos_embed.data.copy_(torch.from_numpy(img_pos_embed).float().unsqueeze(0))
+
+    # Adopted from https://huggingface.co/OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B/blob/main/mm_projector_builder.py#L6
+    def bipartite_soft_matching(
```
rkazants, line 4959: please leave a comment what is this function doing? Describe this stage in the pipeline.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,926 @@ def preprocess_inputs(
...
+        x = merge(x * size, mode="sum")
+        size = merge(size, mode="sum")
+
+        x = x / size
+        return x, size
+
+    def get_vision_embeddings(self, images):
+        if isinstance(images, list):
+            raise NotImplementedError
+        else:
```
rkazants: please remove

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,926 @@ def preprocess_inputs(
...
+            # input: B T C H W
+            # output: B T*L C
```
rkazants: please leave a comment why it is needed

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,926 @@ def preprocess_inputs(
...
+                r_merge_list.append(tmp_p - target_num_token)
+                break
+            else:
+                r_merge_list.append(tmp_p // 2)
+                tmp_p = tmp_p - (tmp_p // 2)
+
+        head = self.num_attention_heads
+
+        dim = c // head
+        for r in r_merge_list:
```
rkazants, line 5100: what is the merge_list?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,926 @@ def preprocess_inputs(
...
+
+    # Adopted from https://huggingface.co/OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B/blob/main/mm_projector_builder.py#L96
+    def merge_tokens(self, x, target_num_token):
+        r"""
+        x = torch.randn(10, 2560, c)
+        x = merge_tokens(x, r_merge_list=[1280])
+        """
+        size = None
+        b, p, c = x.shape
+        tmp_p = p
```
rkazants: please use proper naming for variable. `tmp_p`?

optimum/exporters/openvino/model_configs.py
xufang-lisa: Done.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Done.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Done.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Done.

optimum/exporters/openvino/model_configs.py
xufang-lisa: `num_channels` has been removed.

optimum/exporters/openvino/model_patcher.py
xufang-lisa: `residual` is always `None` and has been removed. The type of `hidden_states` is `tensor` not `tuple` and this check has also been removed.

optimum/exporters/openvino/model_patcher.py
xufang-lisa: Removed.

optimum/exporters/openvino/model_patcher.py
xufang-lisa: This patcher was added only to change the input name. It has been removed.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Replace it with a simple string-to-list conversion.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Done.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: `ast.literal_eval` could potentially raise an exception within `get_anyres_image_grid_shape`, so there is an exception-handling block here. Now `ast.literal_eval` and try-cash have been replaced.  By the way, `grid_pinpoints` in the configuration file is a list not a string.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Done.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Removed.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Done.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Comment is added.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: It is renamed to `current_num_tokens`

xufang-lisa:
@rkazants A default `image_preprocess` implementation has been added to` _OVVideoChatFlashQwenForCausalLM` to provide image preprocessing when the processor is unavailable. Please take a look https://github.com/huggingface/optimum-intel/pull/1637/changes/e1dba190e1b2565e892752a06fc06bff86604aa9

optimum/exporters/openvino/model_patcher.py
peterchen-intel: Add the comments to explain why the patchers are required, and what have been patched by sharing to original code line. @xufang-lisa

xufang-lisa: Comments have been added in `VideoChatFlashQwenVisionEmbeddingModelPatcher`.

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5304,204 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        num_channels: int = DEFAULT_DUMMY_SHAPES["num_channels"],
+        width: int = DEFAULT_DUMMY_SHAPES["width"],
+        height: int = DEFAULT_DUMMY_SHAPES["height"],
+        visual_seq_length: int = DEFAULT_DUMMY_SHAPES["visual_seq_length"],
+        **kwargs,
+    ):
+        super().__init__(task, normalized_config, batch_size, num_channels, width, height, visual_seq_length, **kwargs)
+        self.num_frames = getattr(normalized_config.config, "mm_local_num_frames", 4)
+        self.embed_dim = getattr(normalized_config.config, "mm_hidden_size", 1408)
```
echarlaix: Shouldn't both always be present in the config ? If not wouldn't this be an issue as these could vary from model to model ?

optimum/exporters/openvino/model_configs.py
echarlaix: @xufang-lisa can you add a link to the original modeling code where this value is defined?

optimum/exporters/openvino/model_configs.py
echarlaix: @xufang-lisa can you add a link to the original code where this value is defined?

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5304,204 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+        # Then input image size and patch size for the vision encoder can not be got from the config, we set them to fixed values according to the original implementation.
+        self.height = 224
+        self.width = 224
+        self.image_size = (self.height, self.width)
+        self.patch_size = 14
```
echarlaix: @xufang-lisa can you add a link to the original code where this value is defined?

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5304,204 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+    LANGUAGE = "language"
+    VISION_EMBEDDINGS = "vision_embeddings"
+    VISION_PROJECTION = "vision_projection"
+    TEXT_EMBEDDINGS = "text_embeddings"
+
+
+@register_in_tasks_manager("videochat_flash_qwen", *["image-text-to-text"], library_name="transformers")
+class VideoChatFlashQwenOpenVINOConfig(BaseVLMOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.49.0"
+    MAX_TRANSFORMERS_VERSION = "4.57.99"
```
```suggestion
    MAX_TRANSFORMERS_VERSION = "4.57.6"
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -5303,6 +5304,204 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+                    self.width,
+                ],
+                framework=framework,
+                dtype=float_dtype,
+            )
+        elif input_name == "rotary_pos_emb":
+            grid_h, grid_w = self.height // self.patch_size, self.width // self.patch_size
+            grid_t = self.num_frames
+            return self.random_float_tensor(
+                [1, 1 + grid_h * grid_t * grid_w, self.embed_dim], framework=framework, dtype=float_dtype
```
echarlaix, line 5393: could you add a comment to explain why is there a fixed dimension of 1 ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4803,960 @@ def preprocess_inputs(
         return inputs
 
 
+class _OVVideoChatFlashQwenForCausalLM(OVModelForVisualCausalLM):
+    from transformers import AutoModel
```
echarlaix: would you mind moving to top file ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4803,960 @@ def preprocess_inputs(
...
+
+    auto_model_class = AutoModel
+    additional_parts = ["vision_projection"]
+    IMAGE_TOKEN_INDEX = -200
+    IGNORE_INDEX = -100
```
echarlaix, line 4813: where do these values come from ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4803,960 @@ def preprocess_inputs(
...
+            config=config,
+            device=device,
+            dynamic_shapes=dynamic_shapes,
+            ov_config=ov_config,
+            model_save_dir=model_save_dir,
+            quantization_config=quantization_config,
+            **kwargs,
+        )
+        num_frames = getattr(config, "mm_local_num_frames", 8)
+        self.num_attention_heads = 16
```
echarlaix: this should be extracted from the config no ?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4803,960 @@ def preprocess_inputs(
...
+        self.patch_size = 14
+        self.image_size = 224
```
echarlaix: why not check the exported model directly to infer these values?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4803,960 @@ def preprocess_inputs(
...
+
+            out[..., 1::2, :] = dst
+            out.scatter_(dim=-2, index=(2 * unm_idx).expand(n, unm_len, c), src=unm)
+            out.scatter_(dim=-2, index=(2 * src_idx).expand(n, r, c), src=src)
+
+            return out
+
+        return merge, unmerge
+
+    # Adopted from https://huggingface.co/OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B/blob/main/mm_projector_builder.py#L62
```
echarlaix: when the original method is the same can you replace with copied instead, that will be helpful for maintenance
```suggestion
    # Copied from https://huggingface.co/OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B/blob/main/mm_projector_builder.py#L62
```

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: @echarlaix Thank you for your comments. In the exported model, both `pos_embed` and `img_pos_embed` are provided through the same input `rotary_pos_emb`, and either one may be used at runtime. To support both cases, this input must remain dynamically shaped. For now, we cannot infer `patch_size` and `image_size` directly from the exported model.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Thanks for pointing this out. For the model we are testing, both fields are present in config, so using fallback defaults is unnecessary in this case. I've updated the code to read both values directly from config.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Added.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Added.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Added a comment.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Added a link.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Updated.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Sure, done. I've moved it to the top of the file.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Added links.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: @echarlaix Here `num_attention_heads` refers to the vision embedding model configuration. It can not be found in `config.json`, and it is a fixed value in original model. I renamed it to `mm_num_attention_head` and added a link to its source.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: Good point, thanks. I've updated this function and the other related functions to clearly distinguish whether the implementation has been modified.

tests/openvino/test_seq2seq.py
```diff
@@ -556,10 +556,22 @@ class OVModelForVisualCausalLMIntegrationTest(OVSeq2SeqTestMixin):
         SUPPORT_VIDEO += ["qwen3_vl"]
 
     if is_transformers_version(">=", "4.54.0"):
+        # the layers attribute of DynamicCache is used in videochat_flash_qwen model
+        SUPPORTED_ARCHITECTURES.append("videochat_flash_qwen")
+        SUPPORT_VIDEO.append("videochat_flash_qwen")
         # remote code models differs after transformers v4.54
         SUPPORTED_ARCHITECTURES = set(SUPPORTED_ARCHITECTURES) - {"llava-qwen2", "phi3_v", "phi4mm"}
```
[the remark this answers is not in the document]
xufang-lisa: This issue is outside the scope of this PR's diff

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,974 @@ def preprocess_inputs(
...
+            image_sizes.append(image_size)
+
+        if len(frames) >= 1:
+            results["images"] = frames
+            results["image_sizes"] = image_sizes
+
+        if tokenizer.pad_token_id is None:
+            if "qwen" in tokenizer.name_or_path.lower():
+                logger.info("Setting pad token to bos token for qwen model.")
+                tokenizer.pad_token_id = 151643
```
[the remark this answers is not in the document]
xufang-lisa: Replace `151643` with `tokenizer.bos_token_id`

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,974 @@ def preprocess_inputs(
...
+                        if mm_newline_position == "one_token":
+                            frame_feature = frame_feature.flatten(0, 1)
+                            if "unpad" in mm_patch_merge_type:
+                                frame_feature = torch.cat(
+                                    (frame_feature, self.model.image_newline[None].to(frame_feature.device)), dim=0
+                                )
+                            else:
+                                frame_feature = torch.cat(
+                                    (frame_feature, self.model.frame_newline[None].to(frame_feature.device)), dim=0
+                                )
```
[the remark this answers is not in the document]
xufang-lisa: Initialize `image_newline` and `frame_newline` in the constructor.

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,974 @@ def preprocess_inputs(
...
+            cur_new_input_embeds = []
+
+            for i in range(num_images + 1):
+                cur_new_input_embeds.append(cur_input_embeds_no_im[i])
+                if i < num_images:
+                    try:
+                        cur_image_features = image_features[cur_image_idx]
+                    except IndexError:
+                        print(f"cur_image_idx={cur_image_idx} is not ok")
+                        cur_image_features = image_features[cur_image_idx - 1]
```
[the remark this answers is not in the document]
xufang-lisa: Keep the error-tolerance fallback, add a validity check for `cur_image_idx`.

optimum/exporters/openvino/model_configs.py
rkazants: Can we take one of the image size sample from this array: https://huggingface.co/OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B/blob/main/config.json#L18 ?

I think this is array of supported image resolutions, isn't?

optimum/exporters/openvino/model_configs.py
rkazants: you need to find a way to access `InternVideo2VisionConfig` from which you can get patch_size.
Try to access from `normalized_config` or some members of it

optimum/exporters/openvino/model_configs.py
xufang-lisa: `image_grid_pinpoints` is a set of candidate resolutions used to align the real image size to a describe grid scheme.  The model uses the input image's original size together with these candidate resolutions to choose the most suitable resolution for computing the grid width and height.

optimum/exporters/openvino/model_configs.py
xufang-lisa: Yes, `InternVideo2VisionConfig` is the configuration object for the `vision_tower` submodule, and both `image_size` and `patch_size` are available from it. I have read these two values from the submodel config and written them into `model.config`.  Then `model.config` will be propagated into `normalized_config`.

optimum/exporters/openvino/convert.py
```diff
@@ -937,6 +937,15 @@ def _get_multi_modal_submodels_and_export_configs(
...
         model.config.num_img_tokens = model.model.embed_tokens_extend.image_embed.num_img_tokens
+    if model_type == "videochat_flash_qwen":
+        model.config.mm_num_attention_heads = model.model.get_vision_tower().config.num_attention_heads
+        # num_tome_tokens=64 comes from the upstream projector_type "tome16_mlp_hd64", which uses a fixed 64-token output.
+        # Source: https://huggingface.co/OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B/blob/main/mm_projector_builder.py#L135-L146
+        model.config.mm_projector_num_tome_tokens = 64
+        model.config.patch_size = model.model.get_vision_tower().config.patch_size
+        model.config.image_size = model.model.get_vision_tower().config.image_size
+        model.config.image_mean = model.model.get_vision_tower().image_processor.image_mean
+        model.config.image_std = model.model.get_vision_tower().image_processor.image_std
```
rkazants: @echarlaix, I think we should rather have it in model config class. The current part should not contain model specific logic. It is better to avoid if possible. How do you think?

optimum/exporters/openvino/model_configs.py
```diff
@@ -5396,6 +5397,205 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+            float_dtype=float_dtype,
+            behavior=behavior,
+            preprocessors=preprocessors,
+        )
+        self._orig_config = config
+
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        if not self._behavior == VideoChatFlashQwenConfigBehavior.VISION_EMBEDDINGS:
+            return {}
```
rkazants, line 5889: why no inputs? please double-check

optimum/exporters/openvino/model_patcher.py
```diff
@@ -7817,6 +7817,44 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            # append cls token
+            cls_tokens = self.cls_token.expand(B, -1, -1)
+            hidden_states = torch.cat((cls_tokens, hidden_states), dim=1)
+            hidden_states = hidden_states + rotary_pos_emb
+            hidden_states = hidden_states.reshape(B, -1, C)
+
+            for idx, blk in enumerate(self.blocks):
+                hidden_states = blk(hidden_states, residual=None)
+
+            return hidden_states
```
rkazants, line 8229: can you just call self._orig_forward(..., use_image=False), or with `use_image=True`?
This way we don't need to copy-paste all the code.

optimum/intel/openvino/quantization.py
```diff
@@ -1472,6 +1472,13 @@ def _quantize_ovbasemodel(
                 ov_model_name, pipeline_quantization_config.default_config
             )
             if config is None:
+                if immediate_save:
+                    # The submodels being quantized is unloaded after quantization,
+                    # so the skipped submodels should also be unloaded to avoid keeping their IR files open on Windows.
+                    # This can avoid later _merge_move failures caused by locked .bin files.
+                    ov_model = self.model.ov_models[ov_model_name]
+                    if ov_model is not None:
+                        self.model._unload_ov_model(ov_model)
```
rkazants, line 1485: @andrey-churkin, can you please check this part?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,958 @@ def preprocess_inputs(
...
+            image_size, possible_resolutions, max_resolutions=max_resolutions, patch_size=patch_size
+        )
+
+        return width // patch_size, height // patch_size
+
+    def get_text_embeddings(self, input_ids):
+        squeeze_batch_dim = False
+        if input_ids.ndim == 1:
+            input_ids = input_ids.unsqueeze(0)
+            squeeze_batch_dim = True
```
rkazants: is it possible to do `unsqueeze()` in `prepare_inputs_for_generation()` if needed? So this function is not needed

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,958 @@ def preprocess_inputs(
...
+
+        text_embed = super().get_text_embeddings(input_ids)
+
+        if squeeze_batch_dim and text_embed.ndim > 0 and text_embed.shape[0] == 1:
+            text_embed = text_embed[0]
+
+        text_embed = torch.from_numpy(text_embed) if isinstance(text_embed, np.ndarray) else text_embed
```
rkazants: why do we need this cast if OV returns `np.array`?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4804,958 @@ def preprocess_inputs(
...
+        self,
+        input_ids,
+        pixel_values=None,
+        attention_mask=None,
+        position_ids=None,
+        modalities=None,
+        image_sizes=None,
+        **kwargs,
+    ):
+        images = pixel_values
```
rkazants, line 5355: can you please avoid such huge copy-paste logics?

rkazants (changes requested):

optimum/exporters/openvino/model_patcher.py
xufang-lisa: @rkazants My understanding is that if `use_image` is alwarys set to `False`, the exported model can only process video inputs and can not process image inputs. Is my understanding correct?

optimum/exporters/openvino/model_configs.py
xufang-lisa: @rkazants For this config class, the inputs property is only meaningful for the `VISION_EMBEDDINGS` behavior. The other behaviors each have their own input schema. This guard prevents vision-embedding input definitions from being applied to non-vision behaviors.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: `prepare_inputs_for_generation()` always returns 2D `input_ids`. The wrapper around `get_text_embeddings` exists because, inside `get_multimodal_embeddings`, some call sites pass 2D tensors while others pass 1D tensors, so the wrapper handles both input shapes. 
Of course, this wrapper can be removed, and `unsqueeze()` can be added at the call sites instead.
I have already removed this wrapper.

optimum/intel/openvino/modeling_visual_language.py
xufang-lisa: I converted `np.array` to `torch.tensor` here because its return value is used as an input to `torch.cat`.
To stay consistent with the wrapper-removal refactor above, I moved this conversion to the call sites.

optimum/intel/openvino/modeling_visual_language.py
MaximProshin: @xufang-lisa , was it applied or not?

peterchen-intel: Done after today's decision.

tests/openvino/test_export.py
```diff
@@ -124,6 +124,9 @@ class ExportModelTest(unittest.TestCase):
     if is_transformers_version(">=", "4.57.0") and is_transformers_version("<", "5"):
         SUPPORTED_ARCHITECTURES.update({"qwen3_next": OVModelForCausalLM})
 
+    if is_transformers_version(">=", "4.49") and is_transformers_version("<", "4.57.6"):
```
rkazants: the problem here is that now we only have the matrix: 4.45, 4.57.6, and 5.0.0. And none of them will test this case.

optimum/intel/openvino/quantization.py
rkazants: @andrey-churkin, please check this change. Not sure about this fix!!!

rkazants (approved):
@xufang-lisa, please think of how to add the tests into pre-commit.
Now it is not tested because after moving to transformers v5.0.0, transformers version matrix was changed.

@xufang-lisa please ask to review @andrey-churkin. I worry about changes in quantiozation related file(s)

tests/openvino/test_export.py
xufang-lisa: Good catch, thanks! Since 4.57.6 should be included in this condition, I’ve changed it to <= "4.57.6".

xufang-lisa:
> @xufang-lisa, please think of how to add the tests into pre-commit. Now it is not tested because after moving to transformers v5.0.0, transformers version matrix was changed.
> 
> @xufang-lisa please ask to review @andrey-churkin. I worry about changes in quantiozation related file(s)

@rkazants Thanks for the review. I have updated the test condition in `test_export.py`.  So `videochat` is now covered and tested with `transformers 4.57.6`.

optimum/intel/openvino/quantization.py
andrey-churkin: LGTM, looks reasonable

andrey-churkin (approved):
Checked changes in `optimum/intel/openvino/quantization.py`, LGTM.

optimum/exporters/openvino/convert.py
regisss: If possible, I agree it would be better to move this to the model config class

optimum/exporters/openvino/model_configs.py
```diff
@@ -5764,6 +5765,205 @@ class SiglipTextOpenVINOConfig(SiglipTextOnnxConfig):
...
+    LANGUAGE = "language"
+    VISION_EMBEDDINGS = "vision_embeddings"
+    VISION_PROJECTION = "vision_projection"
+    TEXT_EMBEDDINGS = "text_embeddings"
+
+
+@register_in_tasks_manager("videochat_flash_qwen", *["image-text-to-text"], library_name="transformers")
+class VideoChatFlashQwenOpenVINOConfig(BaseVLMOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.49.0"
+    MAX_TRANSFORMERS_VERSION = "4.57.6"
```
rkazants, line 5861:
```suggestion
    MAX_TRANSFORMERS_VERSION = "4.57.99"
```

optimum/exporters/openvino/convert.py
xufang-lisa: Done. Move it to model config class.

optimum/exporters/openvino/model_configs.py
xufang-lisa: @rkazants  I checked the published `Transformers` versions, and within the `4.57.x` series, `4.57.6` is the latest available patch release. The next major release is `5.0.0`. So I think setting `MAX_TRANSFORMERS_VERSION = "4.57.6"` reflects the actual highest available `4.57.x` version while keeping the version constraint style consistent with other model configs in this file.

regisss (approved):
LGTM

PR #1641, opened 2026-03-19
[HybridAttention] Support past key values in attention_mask for stateful inference

apaniukov:
# What does this PR do?

GraniteMoeHybrid crashes with a shape mismatch error:
```
RuntimeError: Exception from src/inference/src/cpp/infer_request.cpp:224:
Exception from src/plugins/intel_cpu/src/node.cpp:787:
[CPU] Add node with name '__module.model/aten::add/Add_1' Check 'input_shape[j] == 1' failed at src/plugins/intel_cpu/src/shape_inference/custom/eltwise.cpp:52:
Eltwise shape infer input shapes dim index: 3 mismatch
```

The RuntimeError at the Add node that combines the causal mask with the position-expanded attention_mask:

 - port 0 (causal mask): [batch, 1, seq_len, seq_len]
 - port 1 (attn mask):   [batch, 1, 1, total_seq_len]
where seq_len != total_seq_len during 2nd+ generation steps when seq_len > 1 (e.g. second prefill stage).

The issue caused by patched `_update_causal_mask` method. Hybrid models that doesn't patch it (Qwen3-Next and LFM2) work fine.

Fix:
- Update `inputs` property to use `"past_sequence_length + sequence_length"` for `attention_mask` for `-with-past` conversion
- [unrelated] remove unnecessary `triu` operation

Fixes # (issue)

## Before submitting


rkazants (commented):
can you please add test to reproduce this case?

optimum/exporters/openvino/model_configs.py
```diff
@@ -4955,10 +4955,12 @@ def add_past_key_values(self, inputs_or_outputs: Dict[str, Dict[int, str]], dire
     def inputs(self) -> Dict[str, Dict[int, str]]:
         common_inputs = {
             "input_ids": {0: "batch_size", 1: "sequence_length"},
-            "attention_mask": {0: "batch_size", 1: "sequence_length"},
         }
         if self.use_past_in_inputs:
+            common_inputs["attention_mask"] = {0: "batch_size", 1: "past_sequence_length + sequence_length"}
```
rkazants, line 5002: how is it different from the current version? I think it is the same that means dynamic dimension

apaniukov: The current version can process two scenarios:
1) Prefill with no KV cache state (state len is 0)
2) Generation with a KV cache state (seq_len is 1)

The problematic scenario is prefill with a non-empty KV cache state. The attention mask for regular attention layer has to cover past cache size and current input sequence length, but it covers only sequence length.

rkazants (approved):

rkazants:
failed tests are not related to this fix in this PR

echarlaix:
@apaniukov the test added in this PR `test_hybrid_model_multi_step_generation` is currently failing with 

```python
FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_hybrid_model_multi_step_generation_1_lfm2 - RuntimeError: expected m1 and m2 to have the same dtype, but got: float != c10::BFloat16
FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_hybrid_model_multi_step_generation_2_qwen3_next - ValueError: The current version of Transformers does not allow for the export of the model. Maximum required is 4.57.6, got: 5.0.0
```

can you take a look?

echarlaix:
@apaniukov @rkazants any news on this ?

rkazants:
> @apaniukov @rkazants any news on this ?

@echarlaix, our plan is to set max transformers version to 4.99.99 for these two models and support them for transformers 5.0.0 in separate PRs.
I will open up PRs with these adjustements

PR #1644, opened 2026-03-22
[OpenVINO] Support Flux.2-Klein Models on OpenVINO

osabnis:
# What does this PR do?
This PR enables OpenVINO export for `black-forest-labs/FLUX.2-klein-4B` by making the Flux export path more robust to Flux2/Klein-specific config and signature differences.

**Main changes:**

1. `optimum/exporters/openvino/convert.py`
    * Added adaptive Flux text encoder model-type resolution (clip-text vs gemma2-text-encoder vs t5-encoder-model).
    * Added handling for Flux pipelines where text encoders are wrapped as *ForCausalLM (exports backbone .model when applicable).
    * Added guarded pooled_projection_dim setup only when the transformer forward signature expects pooled_projections.
2. `optimum/exporters/openvino/model_configs.py`
    * Added dynamic Flux RoPE ID width inference helper (_get_flux_ids_dim).
    * Updated Flux dummy input generators to create img_ids / txt_ids using inferred ID width (instead of hardcoded 3).
    * Updated FluxTransformerOpenVINOConfig to include/exclude pooled_projections dummy/input mapping based on model config.

These changes keep the existing Flux path intact while supporting Flux2-Klein variants.

Example of conversion cmd-line for `black-forest-labs/FLUX.2-klein-4B`:
```bash
optimum-cli export openvino --model black-forest-labs/FLUX.2-klein-4B ov_flux.2-klein-4b
```

Example of inference for `black-forest-labs/FLUX.2-klein-4B` using OpenVINO backend:
```bash
from pathlib import Path
import numpy as np
import torch
from optimum.intel import OVDiffusionPipeline

model_dir = Path("./ov_flux.2-klein-4b")

# deterministic seeds for reproducibility
np.random.seed(0)

pipe = OVDiffusionPipeline.from_pretrained(model_dir, compile=False)
pipe.to("GPU")
pipe.compile()

out = pipe(
    prompt="Astronaut on the moon",
    height=512,
    width=512,
    num_inference_steps=4,
    output_type="pil",
)

img = out.images[0]
img.save("flux2_klein_smoke.png")
print("Saved:", Path("flux2_klein_smoke.png").resolve())
print("Image size:", img.size)
```
Converted Model Output from above script:

image

## Before submitting


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1158,12 +1178,13 @@ def __init__(self, model: openvino.Model, parent_pipeline: OVDiffusionPipeline,
...
-        output_hidden_states: Optional[bool] = None,
-        return_dict: bool = False,
-    ):
+    self,
+    input_ids: Union[np.ndarray, torch.Tensor],
+    attention_mask: Optional[Union[np.ndarray, torch.Tensor]] = None,
+    output_hidden_states: Optional[bool] = None,
+    return_dict: bool = False,
+    **kwargs,
+):
```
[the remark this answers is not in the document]
osabnis: This doesn't lead to a Python syntax error - but fixing indentation makes sense.

optimum/exporters/openvino/convert.py
```diff
@@ -1286,28 +1308,79 @@ def get_sd3_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+            text_encoder_for_export = text_encoder.model
+
+        text_encoder_model_type = _resolve_flux_text_encoder_model_type(
+            text_encoder,
+            "clip-text",
+            getattr(pipeline, "tokenizer", None),
+        )
+
+        text_encoder_library_name = "diffusers"
+        if text_encoder_model_type in {"qwen3", "qwen2", "qwen"}:
```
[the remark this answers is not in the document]
osabnis: Removing the branch as it's not required.

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1728,9 +1803,10 @@ def _get_ov_class(pipeline_class_name: str, throw_error_if_not_exist: bool = Tru
     OV_IMAGE2IMAGE_PIPELINES_MAPPING["stable-diffusion-3"] = OVStableDiffusion3Img2ImgPipeline
 
 if is_diffusers_version(">=", "0.30.0"):
-    SUPPORTED_OV_PIPELINES.extend([OVStableDiffusion3InpaintPipeline, OVFluxPipeline])
+    SUPPORTED_OV_PIPELINES.extend([OVStableDiffusion3InpaintPipeline, OVFluxPipeline, OVFlux2KleinPipeline])
     OV_INPAINT_PIPELINES_MAPPING["stable-diffusion-3"] = OVStableDiffusion3InpaintPipeline
     OV_TEXT2IMAGE_PIPELINES_MAPPING["flux"] = OVFluxPipeline
+    OV_TEXT2IMAGE_PIPELINES_MAPPING["flux2-klein"] = OVFlux2KleinPipeline
```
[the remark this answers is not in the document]
osabnis: Makes sense - instead of globals, will add an explicit check.
Also Flux2KleinPipeline support was added in 0.37 - so will add a new check instead of gating it at 0.30.

optimum/exporters/openvino/__init__.py
```diff
@@ -15,7 +15,7 @@
 import optimum.exporters.openvino.model_configs
 
 from .__main__ import main_export
-from .convert import export, export_from_model, export_models, export_pytorch_via_onnx
+from .convert import export, export_from_model, export_models, export_pytorch_via_onnx, _resolve_flux_text_encoder_model_type
```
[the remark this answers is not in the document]
osabnis: Fixed.

tests/openvino/test_export.py
```diff
@@ -28,7 +28,8 @@
 
 from optimum.exporters.onnx.constants import SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED
 from optimum.exporters.onnx.model_configs import BertOnnxConfig
-from optimum.exporters.openvino import export_from_model, main_export
+from optimum.exporters.openvino import export_from_model, main_export, _resolve_flux_text_encoder_model_type
```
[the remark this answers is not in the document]
osabnis: Fixed.

anatyrova:
Hello @osabnis ! Thank you for committing.

Here are my comments:
1) _Tests:_ unit tests pass , but there are no integration/export tests verifying that existing FLUX.1 models still export correctly after the IDs dimension change from hardcoded 3 to dynamic _get_flux_ids_dim(). This is a regression risk. (the code previously hardcoded the number 3 for the ID dimension. The PR replaced this with a dynamic function. This change affects all Flux models, not just Klein. A question that hasn't been tested: does the old FLUX.1-dev still export correctly after this change?)

2) _Tests / Inference:_ SyntaxWarning: modeling_base.py:590: invalid escape sequence '\.' This means the code contains something like "\."; Python 3.12 now complains about such strings because \. is not a valid escape sequence. The correct escape sequence is r"\." (raw string). This is simply a warning. Nothing crashes.

3) _Export + Inference:_ Warning suggests scaling_factor is missing from VAE config after export. Inference still worked in my test, but this should be investigated — incorrect VAE scaling can cause artifacts in some cases.

4) _Inference:_ Flux2KleinPipeline.__call__ has image as the first positional argument (default None), not prompt. So calling pipe("cat") silently passes the string as image and leaves prompt=None, causing a ValueError. The PR should document that prompt must always be passed as a keyword argument.

optimum/exporters/openvino/convert.py
```diff
@@ -1286,28 +1308,76 @@ def get_sd3_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+        or "Qwen" in encoder_cls_name
+        or "Qwen" in tokenizer_cls_name
+    )
+    if looks_like_qwen:
+        return "qwen3-text-encoder"
+
+    return default_model_type
+
+
 def get_flux_models_for_export(pipeline, exporter, int_dtype, float_dtype):
```
anatyrova, line 1410: Code duplication in get_flux_models_for_export. The handling of text_encoder and text_encoder_2 is nearly identical : CausalLM unwrapping, _resolve_flux_text_encoder_model_type call, and config setup are copy-pasted with minor variable name differences. This should be extracted into a shared helper function.

optimum/exporters/openvino/model_configs.py
```diff
@@ -390,6 +390,34 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
             common_inputs = super().inputs
         return common_inputs
 
+@register_in_tasks_manager("qwen3-text-encoder", *["feature-extraction"], library_name="diffusers")
+class Qwen3TextEncoderOpenVINOConfig(Qwen3OpenVINOConfig):
```
anatyrova: Qwen3TextEncoderOpenVINOConfig inherits from the wrong base class. The existing pattern for text encoder configs is to inherit from CLIPTextOpenVINOConfig (see Gemma2TextEncoderOpenVINOConfig at model_configs.py:2444). This PR inherits from Qwen3OpenVINOConfig which is a decoder config -- inconsistent with the established convention

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -799,13 +802,28 @@ def _reshape_transformer(
             elif inputs.get_any_name() == "pooled_projections":
                 shapes[inputs] = [batch_size, self.transformer.config["pooled_projection_dim"]]
             elif inputs.get_any_name() == "img_ids":
+                ids_dim = inputs.get_partial_shape()[-1]
+                if hasattr(ids_dim, "is_dynamic") and ids_dim.is_dynamic:
+                    ids_dim = 3
```
anatyrova: Hardcoded 3 reappears in modeling_diffusion.py. _get_flux_ids_dim() was introduced to remove the hardcoded 3 for IDs dimension. However, in modeling_diffusion.py the fallback for dynamic shapes still uses ids_dim = 3. The fix is incomplete?

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1403,6 +1458,18 @@ def __init__(self, decoder: OVModelVaeDecoder, encoder: OVModelVaeEncoder):
...
         if hasattr(self.decoder.config, "latents_std_data"):
             self.latents_std = torch.tensor(self.decoder.config.latents_std_data)
+        # Flux2Klein compatibility: pipeline expects self.vae.bn.running_mean/running_var/eps
+        self.bn = None
+        bn_mean = getattr(self.decoder.config, "bn_running_mean_data", None)
+        bn_var = getattr(self.decoder.config, "bn_running_var_data", None)
+        bn_eps = float(getattr(self.decoder.config, "bn_eps", 1e-5))
+
+        if bn_mean is not None and bn_var is not None:
+            bn_state = type("BNState", (), {})()
```
anatyrova: BNState uses an unconventional pattern. bn_state = type("BNState", (), {})() -- this is hard to read. Use types.SimpleNamespace instead, it's the standard Python idiom for simple attribute containers and is already used elsewhere in the codebase

optimum/exporters/openvino/model_configs.py
```diff
@@ -2614,10 +2684,38 @@ class FluxTransformerOpenVINOConfig(SD3TransformerOpenVINOConfig):
     )
     _MODEL_PATCHER = FluxTransfromerModelPatcher
 
+    def __init__(self, *args, **kwargs):
```
anatyrova: DUMMY_INPUT_GENERATOR_CLASSES reassigned as instance variable. DUMMY_INPUT_GENERATOR_CLASSES is a class-level attribute across all other configs in this repo. FluxTransformerOpenVINOConfig.__init__ reassigns it per-instance depending on config. This breaks the established convention, bc. the conditional logic should be handled in the inputs property instead, consistent with how other configs work

anatyrova: The logic for pooled_projections is already in the inputs property below — [model_configs.py:2716-2717]
```
if not getattr(self, "_use_pooled_projections", True):
    common_inputs.pop("pooled_projections", None)
```

anatyrova (changes requested):

optimum/exporters/openvino/convert.py
```diff
@@ -532,21 +532,43 @@ def export_models(
...
+                    device=device,
+                    input_shapes=input_shapes,
+                    model_kwargs=model_kwargs,
+                    ov_config=ov_config,
+                    stateful=stateful[i] if isinstance(stateful, (list, tuple)) else stateful,
+                    patch_16bit_model=patch_16bit_model,
+                    library_name=library_name,
+                )
+            )
+        except Exception as e:
```
anatyrova: Internal API - fallback in export_models. export_models is a common function used by ALL models in the repository, not just Flux. A PR added a silent fallback to it on ONNX when prim::TupleConstruct is encountered. This changes the export behavior for all models without warning - an illogical change to the internal API.

optimum/exporters/openvino/convert.py
anatyrova: also, get_flux_models_for_export now directly mutates transformer.config by injecting pooled_projection_dim into it. This is a non-obvious side effect , and callers of this function don't expect their transformer object to be modified in place

rkazants (changes requested):
please provide tests to check inference results. Check test_diffusion.py

optimum/exporters/openvino/convert.py
osabnis: Thanks!
I refactored get_flux_models_for_export to remove duplication by extracting shared text encoder export logic into _add_flux_text_encoder_for_export, and reused it for both text_encoder and text_encoder_2.

About your second comment, how do you suggest I modify it to make it more obvious to the callers of this function?

optimum/exporters/openvino/model_configs.py
osabnis: Thanks for the catch - updated Qwen3TextEncoderOpenVINOConfig to inherit from CLIPTextOpenVINOConfig, matching the existing diffusers text-encoder config pattern (similar to Gemma2TextEncoderOpenVINOConfig).

optimum/intel/openvino/modeling_diffusion.py
osabnis: Good catch - thanks!
Fixed it.

optimum/intel/openvino/modeling_diffusion.py
osabnis: Agreed - added a SimpleNamespace implementation. I didn't find any other implementation in this codebase, so I have written something on my own.

optimum/exporters/openvino/model_configs.py
osabnis: Got it - I have removed the init function I had added and handled it within the inputs property as requested.

optimum/exporters/openvino/convert.py
osabnis: Removed the fallback option since we have fixed the issues and it is not needed anymore.

anatyrova:
pls rebase your pr to main when there are commits, add inference tests as Roman advised

optimum/exporters/openvino/convert.py
anatyrova:
```
import copy

transformer = pipeline.transformer
transformer_config = copy.copy(transformer.config)  # shallow copy is enough for scalar attrs
transformer_config.text_encoder_projection_dim = transformer_config.joint_attention_dim
transformer_config.requires_aesthetics_score = getattr(pipeline.config, "requires_aesthetics_score", False)
transformer_config.time_cond_proj_dim = None

if "pooled_projections" in transformer_forward_inputs and not hasattr(transformer_config, "pooled_projection_dim"):
    transformer_config.pooled_projection_dim = transformer_config.joint_attention_dim
```

try this. i added it and test passed:
```
python -m pytest tests/openvino/test_export.py::Flux2KleinSupportUnitTest -v 2>&1
```

but pls add inference test so this change could be confirmed.

rkazants:
We proceed this work in this PR https://github.com/huggingface/optimum-intel/pull/1809
All commits are moved to new PR

PR #1646, opened 2026-03-23
Improve numpy conversion handling for model inputs

MissLostCodes:
## Summary

This PR improves consistency and robustness of tensor-to-numpy conversion across OpenVINO model forward methods.

## Changes

- Introduced a helper function to standardize conversion logic
- Replaced repeated inline `.cpu().numpy()` patterns
- Improved handling of optional inputs like `token_type_ids`

## Motivation

- Reduce code duplication
- Improve readability and maintainability
- Ensure consistent handling of mixed input types (torch / numpy)

## Notes

- No functional changes intended
- Behavior remains identical, with safer input handling

### DIFF
--- a/optimum/intel/openvino/modeling.py
+++ b/optimum/intel/openvino/modeling.py
@@ -57,6 +57,7 @@ from .modeling_base import OVBaseModel
 from .modeling_sam import OVSamModel
 from .utils import _is_timm_ov_dir, ensure_numpy
 
+
 logger = logging.getLogger(__name__)
 
 


optimum/intel/openvino/utils.py
```diff
@@ -314,6 +314,16 @@ def np_to_pt_generators(np_object, device):
         return np_object
 
 
+def ensure_numpy(x):
+    if x is None:
+        return None
+    if isinstance(x, np.ndarray):
+        return x
```
echarlaix: why not remove as it should be equivalent  
```suggestion
```

optimum/intel/openvino/utils.py
```diff
@@ -314,6 +314,16 @@ def np_to_pt_generators(np_object, device):
...
+    if isinstance(x, torch.Tensor):
+        return x.cpu().numpy()
```
rkazants, line 319: let rather check that this is not numpy array and then convert it to numpy array.

MissLostCodes:
@rkazants @echarlaix 
That makes sense, thanks!

My intention was to ensure safe handling of mixed input types (torch / numpy / None), especially to avoid calling `.cpu()` on non-tensor inputs.

I’ve simplified the implementation to follow your suggestion while keeping the same behavior.

optimum/intel/openvino/utils.py
```diff
@@ -314,6 +314,14 @@ def np_to_pt_generators(np_object, device):
         return np_object
 
 
+def ensure_numpy(x):
+    if x is None:
+        return None
+    if not isinstance(x, np.ndarray):
+        return x.cpu().numpy()
+    return x
```
echarlaix: .cpu().numpy() should only be called when x is a torch.Tensor so would replace with 

```suggestion
    if isinstance(x, torch.Tensor):
        return x.cpu().numpy()
    if not isinstance(x, (np.ndarray, type(None))):
        raise TypeError(f"`x` must be a np.ndarray or torch.Tensor, got {type(x)}")
    return x
```

PR #1651, opened 2026-03-25
Added a new test to check whether some transformations is applied during conversion or completion

### LINKED ISSUE #1645: Add tests which check, that required transformations are applied
We need a new group of tests that check whether a specific set of transformations has been applied during conversion or inference for particular architectures. Initially, the tests should cover the AFMoE (aka Arcee Trinity) architecture and verify that the `MoEMatMulsFusion` transformation is applied during model compilation. The transformation log can be obtained by setting the environment variable` OV_ENABLE_PROFILE_PASS=1`. The tests should be scalable across different architectures and transformation sets.

alien-cyber:
This checks whether a specific set of transformations has been applied during conversion or compilation for particular architectures

- **The set of transformations** that needed to be checked for a particular architecture can be defined in a separate file called `expected_transformations.txt`
- **The test case runs for each architecture**
- **The model is loaded and compiled in a separate subprocess** so that the logs can be captured effectively
- **If a transformation is applied during conversion**, then the log definitely contains its name along with time and '+' symbol
- **We run a scan of the logs using an algorithm** to check if the transformations are applied based on that logic
- **Sometimes we may misspell the name of the transformation**, so I used `get_close_matches` library which does fuzzy search and returns the possible matches

Closes #1645

**Please review and let me know if any further improvements are needed.**

### DIFF
--- a/tests/openvino/transformations/test_transformations.py
+++ b/tests/openvino/transformations/test_transformations.py
@@ -17,6 +17,7 @@ import re
 
 from parameterized import parameterized
 from utils_tests import MODEL_NAMES, OPENVINO_DEVICE, REMOTE_CODE_MODELS
+from arch_to_model_class import ARCH_TO_MODEL_CLASS
 
 # Maps architecture name -> list of  transformation needed to be applied , as per expected_transformations.txt
 def _load_expected_transformations(path):
@@ -39,16 +40,16 @@ _CONFIG_PATH = os.path.join(
 ARCH_TO_EXPECTED_TRANSFORMATIONS = _load_expected_transformations(_CONFIG_PATH)
 
 
-def _capture_stderr_during(model_id, OPENVINO_DEVICE, trust_remote_code):
+def _capture_stderr_during(model_id, OPENVINO_DEVICE, trust_remote_code, model_class="OVModelForCausalLM"):
     #  Runs model loading in a subprocess to reliably capture OpenVINO C++ logs.
 
     code = textwrap.dedent(f"""
         import os
         os.environ["OV_ENABLE_PROFILE_PASS"] = "1"
 
-        from optimum.intel import OVModelForCausalLM
+        from optimum.intel import {model_class}
 
-        OVModelForCausalLM.from_pretrained(
+        {model_class}.from_pretrained(
             "{model_id}",
             export=True,
             compile=True,
@@ -130,11 +131,13 @@ class OVTransformationTest(unittest.TestCase):
     ):
         model_id = MODEL_NAMES[model_arch]
         trust_remote_code = model_arch in REMOTE_CODE_MODELS
+        model_class = ARCH_TO_MODEL_CLASS.get(model_arch)
 
         log_output = _capture_stderr_during(
             model_id,
             OPENVINO_DEVICE,
-            trust_remote_code
+            trust_remote_code,
+            model_class,
         )
 
         result = check_failed_transformations(
--- a/tests/openvino/transformations/test_transformations.py
+++ b/tests/openvino/transformations/test_transformations.py
@@ -1,18 +1,17 @@
-
-
-
-
 import os
-
 import sys
-
 import unittest
-
 
 os.environ["OV_ENABLE_PROFILE_PASS"] = "1"
 
-# we are adding this , so the parent directory (tests/openvino/) is in the python search path for utils.py to be imported
+# we are adding this , so the parent directory (tests/openvino/) is in the python search path for utils_test.py to be imported
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+
+import subprocess
+import textwrap
+import re
+from difflib import get_close_matches
 
 from parameterized import parameterized
 from utils_tests import MODEL_NAMES, OPENVINO_DEVICE, REMOTE_CODE_MODELS
@@ -20,8 +19,8 @@ from utils_tests import MODEL_NAMES, OPENVINO_DEVICE, REMOTE_CODE_MODELS
 
 
 
-# Maps architecture name -> list of  transformation needed to be applied , as per expected_transformations.txt
 
+# Maps architecture name -> list of  transformation needed to be applied , as per expected_transformations.txt
 def _load_expected_transformations(path):
     result = {}
     with open(path) as f:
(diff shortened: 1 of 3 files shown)


tests/openvino/transformations/test_transformations.py
```diff
@@ -0,0 +1,173 @@
+import os
+import sys
+import unittest
+
+os.environ["OV_ENABLE_PROFILE_PASS"] = "1"
```
MissLostCodes: i think setting OV_ENABLE_PROFILE_PASS at the top level might unintentionally leak into the global test environment . This variable is already being set inside the subprocess - line 49  . we can safely remove this line .

tests/openvino/transformations/test_transformations.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+    """)
+
+    result = subprocess.run(
+        [sys.executable, "-c", code],
+        capture_output=True,
+        text=True,
+    )
+
+    return result.stdout
```
MissLostCodes, line 72: I think this should use `stderr` instead of `stdout` since OpenVINO logs are emitted to stderr
even function name says capture stderr
```
- return result.stdout
+ return result.stderr
```

tests/openvino/transformations/test_transformations.py
```diff
@@ -0,0 +1,173 @@
...
+    return result
+
+
+_CONFIG_PATH = os.path.join(
+    os.path.dirname(__file__), "expected_transformations.txt"
+)
+ARCH_TO_EXPECTED_TRANSFORMATIONS = _load_expected_transformations(_CONFIG_PATH)
+
+
+def _capture_stderr_during(model_id, OPENVINO_DEVICE, trust_remote_code):
```
MissLostCodes: I think the subprocess code here may be quite expensive since it invokes `OVModelForCausalLM.from_pretrained(...)` inside the string, which can trigger model  export  on every test run.

tests/openvino/transformations/test_transformations.py
```diff
@@ -0,0 +1,173 @@
...
+
+        from optimum.intel import OVModelForCausalLM
+
+        OVModelForCausalLM.from_pretrained(
+            "{model_id}",
+            export=True,
+            compile=True,
+            device="{OPENVINO_DEVICE}",
+            trust_remote_code={trust_remote_code},
+        )
```
MissLostCodes, line 62: can add this param 
```
cache_dir="./ov_cache"
```

we could pass a `cache_dir` so that models are reused across runs instead of being re-exported each time.


This should help make the test more scalable.

tests/openvino/transformations/test_transformations.py
```diff
@@ -0,0 +1,173 @@
...
+
+        name = extract_transform_name(stripped)
+        if name:
+            applied_raw.append(name)
+            applied_norm.append(normalize(name))
+
+    remaining = {normalize(w): w for w in words}
+
+    for key in list(remaining.keys()):
+        if any(key in a for a in applied_norm):
```
MissLostCodes: I think substring matching here could cause false positives like `MatMul` matching `MatMulFusionExtra`
Using exact matching like `if key in applied_norm:` would be safer.

MissLostCodes (commented):
@popovaan @rkazants @alien-cyber   

Hi !

I had a few suggestions and potential issues:

- The test currently captures `stdout`, but OpenVINO logs are emitted to `stderr`, so this may miss the relevant transformation logs.
- Setting `OV_ENABLE_PROFILE_PASS` at the top level leaks into the global test environment  it seems sufficient to keep it scoped inside the subprocess.
- The subprocess call may be expensive since it can trigger model export for each test case. Adding a `cache_dir` (or similar reuse strategy) could help reduce overhead.
- The substring matching logic (`key in a`) could lead to false positives (e.g., `MatMul` matching `MatMulFusionExtra`). Exact matching would be safer.

Overall, the structure and idea of the test look great — addressing these points should make it more robust and CI-friendly.

tests/openvino/transformations/test_transformations.py
alien-cyber: Hi @MissLostCodes  
Thanks for the analysis.

- Now the string matching function is Fixed.
- stderr and stdout are captured in the same place.
- unnecessary OV_ENABLE_PROFILE_PASS is removed

Regarding cache_dir="./ov_cache", I believe it is not necessary in this context, as it mainly changes the default directory used for caching model files. The default Hugging Face cache should already handle reuse of downloaded model configuration.

alien-cyber:
Hi @popovaan 

This test cases initially checks for text-generation(OVModelForCausalLM) model classes.

If this structure is fine then can I start expanding for different architectures.

tests/openvino/transformations/test_transformations.py
```diff
@@ -0,0 +1,174 @@
...
+    for line in log.splitlines():
+        stripped = line.strip()
+
+        if not stripped:
+            continue
+
+        if not stripped.endswith('+'):  #  neglect '-' because those transformations are not applied
+            continue
+
+        name = extract_transform_name(stripped)
```
popovaan: Could you please change the logic a bit as follows:
Find the required transformations in the log together with the corresponding "+" or "-" status.
For each required transformation check:
1) It is present in the transformations set, otherwise error that the transformation name was not found.
2) It was applied, otherwise error this transformation wasn't applied.

tests/openvino/transformations/test_transformations.py
```diff
@@ -0,0 +1,174 @@
...
+
+    remaining = {normalize(w): w for w in words}
+
+    for key in list(remaining.keys()):
+        if key in applied_norm:
+            del remaining[key]
+
+    hints = {}
+    for key, original in remaining.items():
+        matches = get_close_matches(key, applied_norm, n=2, cutoff=0.8)
```
popovaan: I don't think we need to cover the case of a misspelled transformation name. Let's check for an exact transformation name match, otherwise raise an error indicating that the name was not found.

alien-cyber:
Hi @popovaan 

I’ve updated the logic based on your suggestions.
I also expanded the coverage for additional model classes.

architecture to model OVclass is predefined in arch_to_model_class.py , to expand for further architecture we have to update this file

Please review and let me know if any further improvements are needed.

PR #1653, opened 2026-03-27
[OpenVINO] Support Kokoro model

openvino-agent:
# What does this PR do?

Fixes 175565

Set-up environment:

```sh
pip install git+https://github.com/openvino-agent/optimum-intel.git@support_kokoro
pip install kokoro
```

Run exporting the model:

```sh
optimum-cli export openvino -m hexgrad/Kokoro-82M ov_Kokoro-82M --trust-remote-code
```

Run inference:

```python
import soundfile as sf
import torch
from optimum.intel import OVModelForTextToSpeechSeq2Seq

model_id = "hexgrad/Kokoro-82M"

# Export from Hub and save locally
ov_pipe = OVModelForTextToSpeechSeq2Seq.from_pretrained(model_id, trust_remote_code=True)

# Preprocess text into model inputs (G2P + tokenization + voice loading)
inputs = ov_pipe.preprocess_input(
    text="Hello, this is a test of Kokoro text-to-speech using OpenVINO.",
    voice="af_heart",
    speed=1.0,
)

# Generate audio waveform
speech = ov_pipe.generate(**inputs)

sf.write("speech.wav", speech.numpy(), samplerate=24000)
```

Fixes # (issue)

## Before submitting


optimum/exporters/openvino/convert.py
```diff
@@ -1454,3 +1454,119 @@ def _get_speecht5_tss_model_for_export(
...
+                    filename="config.json",
+                    token=kwargs.get("token"),
+                    revision=kwargs.get("revision"),
+                    cache_dir=kwargs.get("cache_dir"),
+                )
+            )
+        except Exception:
+            return False
+
+    with open(config_path) as f:
```
RyanMetcalfeInt8:
```suggestion
    with open(config_path, "r", encoding="utf-8") as f:
```

Otherwise, I get weird encoding errors on Windows:
```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 903: character maps to <undefined>
```

rkazants: please check now

tests/openvino/utils_tests.py
```diff
@@ -25,6 +29,121 @@
 from optimum.intel.utils.import_utils import is_transformers_version
 
 
+def _create_tiny_kokoro_model():
+    """Generate a tiny random Kokoro TTS model for testing and return its local path.
+
+    Falls back to the original Hub id if the `kokoro` package is not installed.
+    Result is cached on disk under the system temp dir, so subsequent calls are cheap.
+    """
```
IlyasMoutawwakil, line 36: why not have it on the hub ?

tests/openvino/test_exporters_cli.py
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import importlib.util
```
regisss: Why is this import needed?

optimum/intel/openvino/modeling_text2speech.py
```diff
@@ -160,24 +160,70 @@ class OVModelForTextToSpeechSeq2Seq(OVModelForSeq2SeqLM):
...
+        if getattr(config, "model_type", None) == "kokoro":
+            return _OVModelForKokoroTextToSpeech._from_pretrained(model_id, config, **kwargs)
+        elif getattr(config, "architectures", None) and "SpeechT5ForTextToSpeech" in config.architectures:
             return _OVModelForSpeechT5ForTextToSpeech._from_pretrained(model_id, config, **kwargs)
         else:
-            raise ValueError(f"{config.architectures} are not supported text-to-audio model using OpenVINO")
-
-            return super()._from_pretrained(model_id, config, **kwargs)
+            raise ValueError(
+                f"{getattr(config, 'architectures', None)} are not supported text-to-audio model using OpenVINO"
```
regisss: if `getattr(config, 'architectures', None)` returns `None`, it would rather be a model config issue than a model mismatch no?

Also why removing `return super()._from_pretrained(model_id, config, **kwargs)`?

tests/openvino/utils_tests.py
rkazants: The original model has restriction on minimal hidden size and number of intermediate layers. I managed to generate 200MB tiny-model only. Also, we have a discussion about moving to tes-time tiny-model generation due to limit of a number of downloads via HF API that leads to multiple failures in pre-commit permanently.
So this is the first model for which we applied another approach and I think we should use it for the future models as well.

tests/openvino/test_exporters_cli.py
openvino-agent: Removed

optimum/intel/openvino/modeling_text2speech.py
openvino-agent: `return super()._from_pretrained(model_id, config, **kwargs)` was removed because it is not reachable due to raising exception.

openvino-agent: Now `getattr(config, 'model_type')` is used

IlyasMoutawwakil:
please run styling

openvino-agent:
> please run styling

Done

regisss (approved):
LGTM

rkazants (approved):

RyanMetcalfeInt8 (approved):
LGTM

PR #1659, opened 2026-03-27
[OpenVINO] export and inference support for Ministral3 (ministral3) VLM

dhandhalyabhavik:
Jira: ITEP-87434 and CVS-184839

# What does this PR do?

Enable mistralai/Ministral-3-3B-Instruct-2512 (instruct and reasoning both) (model_type: mistral3) for OpenVINO export via optimum-cli and inference via OVModelForVisualCausalLM.

Installation instructions:
```bash
pip install transformers==5.0.0
pip install requests torchvision requests
pip install huggingface_hub==1.14.0
pip install nncf==3.1.0
```

Exporting command:
```
optimum-cli export openvino \
  --model mistralai/Ministral-3-3B-Instruct-2512 \
  --task image-text-to-text \
  --weight-format int4 \
  ministral-final-int4
```

Inference script:

```python

model_id = "/workspace/exported-models/ministral-final-int4"
Or wherever you exported:
model_id = "/workspace/exported-models/ministral-upstream-int4"

model = OVModelForVisualCausalLM.from_pretrained(model_id, device="CPU")
processor = AutoProcessor.from_pretrained(model_id, padding_side="left")

url = "https://media.istockphoto.com/id/1192867753/photo/cow-in-berchida-beach-siniscola.jpg?s=612x612&w=0&k=20&c=v0hjjniwsMNfJSuKWZuIn8pssmD5h5bSN1peBd1CmH4="

messages = [
    {
        "role": "system",
        "content": [
            {"type": "text", "text": "You are a helpful assistant."}
        ],
    },
    {
        "role": "user",
        "content": [
            {"type": "image", "url": url},
            {"type": "text", "text": "What is shown in this image?"},
        ],
    },
]

inputs = processor.apply_chat_template(
    messages,
    tokenize=True,
    return_dict=True,
    return_tensors="pt",
    add_generation_prompt=True,
)

output = model.generate(
    **inputs,
    max_new_tokens=50,
    do_sample=False,
)

generated_tokens = output[0, inputs["input_ids"].shape[1]:]
print(processor.decode(generated_tokens, skip_special_tokens=True))
```

### Before submitting

### DIFF
--- a/optimum/exporters/openvino/__main__.py
+++ b/optimum/exporters/openvino/__main__.py
@@ -384,7 +384,10 @@ def main_export(
             trust_remote_code=trust_remote_code,
         )
         quantization_config = getattr(config, "quantization_config", None)
-        quant_method = quantization_config.get("quant_method", None) if quantization_config else None
+        if isinstance(quantization_config, dict):
+            quant_method = quantization_config.get("quant_method", None)
+        else:
+            quant_method = getattr(quantization_config, "quant_method", None)
 
         # update config to load eagle3 models (both text-only and VLM variants)
         archs = getattr(config, "architectures", None)
@@ -451,6 +454,18 @@ def main_export(
             if "activation_checkpointing" in config.audio_processor["config"]:
                 config.audio_processor["config"]["activation_checkpointing"] = ""
             config._attn_implementation = "sdpa"
+            loading_kwargs["config"] = config
+
+        # Mistral3 FP8 checkpoints are dequantized to BF16 for OpenVINO export.
+        if model_type == "mistral3" and quant_method == "fp8":
+            logger.info(
+                "Detected FP8 weights in the Mistral3 checkpoint; dequantizing them before exporting the model "
+                "to OpenVINO."
+            )
+            if isinstance(quantization_config, dict):
+                quantization_config["dequantize"] = True
+            else:
+                quantization_config.dequantize = True
             loading_kwargs["config"] = config
         # there are some difference between remote and in library representation of past key values for some models,
         # for avoiding confusion we disable remote code for them
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -139,6 +139,8 @@ from optimum.exporters.openvino.model_patcher import (
     MiniCPMModelPatcher,
     MiniCPMVImageEmbeddingsModelPatcher,
     MiniCPMVResamplerModelPatcher,
+    Mistral3ImageEmbeddingModelPatcher,
+    Mistral3LanguageModelPatcher,
     MistralModelPatcher,
     MixtralModelPatcher,
     ModelPatcher,
@@ -323,6 +325,11 @@ def init_model_configs():
             "transformers",
             "Qwen3OmniMoeForConditionalGeneration",
         )
+
+    TasksManager._CUSTOM_CLASSES[("pt", "mistral3", "image-text-to-text")] = (
+        "transformers",
+        "Mistral3ForConditionalGeneration",
+    )
 
     if is_diffusers_available() and "fill" not in TasksManager._DIFFUSERS_TASKS_TO_MODEL_LOADERS:
         TasksManager._DIFFUSERS_TASKS_TO_MODEL_LOADERS["fill"] = "FluxFillPipeline"
@@ -5093,6 +5100,107 @@ class Llama4OpenVINOConfig(GotOCR2OpenVINOConfig):
         if self._behavior != VLMConfigBehavior.VISION_EMBEDDINGS:
             return super().patch_model_for_export(model, model_kwargs)
         return Llama4ImageEmbeddingsModelPatcher(self, model, model_kwargs)
+
+
+@register_in_tasks_manager("mistral3", *["image-text-to-text"], library_name="transformers")
+class Mistral3OpenVINOConfig(BaseVLMOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "5.0.0"
+
+    def __init__(
+        self,
+        config: "PretrainedConfig",
+        task: str = "feature-extraction",
+        int_dtype: str = "int64",
+        float_dtype: str = "fp32",
+        behavior: VLMConfigBehavior = VLMConfigBehavior.VISION_EMBEDDINGS,
+        preprocessors: Optional[List[Any]] = None,
+        **kwargs,
+    ):
+        super().__init__(
+            config=config,
+            task=task,
+            int_dtype=int_dtype,
+            float_dtype=float_dtype,
+            preprocessors=preprocessors,
+        )
+        self._orig_config = config
+        if self._behavior == VLMConfigBehavior.VISION_EMBEDDINGS and hasattr(config, "vision_config"):
+            self._config = config.vision_config
+            self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
+
+    def with_behavior(
+        self,
+        behavior: Union[str, VLMConfigBehavior],
+    ):
+        if isinstance(behavior, str) and not isinstance(behavior, VLMConfigBehavior):
+            behavior = VLMConfigBehavior(behavior)
+
+        if behavior == VLMConfigBehavior.TEXT_EMBEDDINGS:
+            model_type = self._orig_config.text_config.model_type
+            if model_type not in TasksManager._SUPPORTED_MODEL_TYPE:
+                model_type = "mistral"
+            return get_vlm_text_embeddings_config(
+                model_type, self._orig_config.text_config, self.int_dtype, self.float_dtype
+            )
+
+        if behavior == VLMConfigBehavior.LANGUAGE:
+            model_type = self._orig_config.text_config.model_type
+            if model_type not in TasksManager._SUPPORTED_MODEL_TYPE:
+                model_type = "mistral"
+            return get_vlm_text_generation_config(
+                model_type, self._orig_config.text_config, self.int_dtype, self.float_dtype,
+                model_patcher=Mistral3LanguageModelPatcher,
+            )
+
+        if behavior == VLMConfigBehavior.VISION_EMBEDDINGS:
+            return self.__class__(
+                self._orig_config,
+                task=self.task,
+                int_dtype=self.int_dtype,
+                float_dtype=self.float_dtype,
+                behavior=behavior,
+                preprocessors=self._preprocessors,
+            )
+
+        raise ValueError(f"Unsupported Mistral3 export behavior: {behavior}")
+
+    def get_model_for_behavior(self, model, behavior: Union[str, VLMConfigBehavior]):
+        if isinstance(behavior, str) and not isinstance(behavior, VLMConfigBehavior):
+            behavior = VLMConfigBehavior(behavior)
+
+        if behavior == VLMConfigBehavior.LANGUAGE:
+            return model
+
+        if behavior == VLMConfigBehavior.VISION_EMBEDDINGS:
+            return model
+
+        if behavior == VLMConfigBehavior.TEXT_EMBEDDINGS:
+            if hasattr(model, "model") and hasattr(model.model, "language_model"):
+                text_embedding = model.model.language_model.get_input_embeddings()
+                text_embedding.config = model.model.language_model.config
+            else:
+                text_embedding = model.get_input_embeddings()
+                text_embedding.config = model.config.text_config
+            return text_embedding
+
+        raise ValueError(f"Unsupported Mistral3 export behavior: {behavior}")
+
+    def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[Dict[str, Any]] = None):
+        model_kwargs = model_kwargs or {}
+        if self._behavior != VLMConfigBehavior.VISION_EMBEDDINGS:
+            return super().patch_model_for_export(model, model_kwargs)
+        return Mistral3ImageEmbeddingModelPatcher(self, model, model_kwargs)
+
+    @property
+    def outputs(self) -> Dict[str, Dict[int, str]]:
+        if self._behavior == VLMConfigBehavior.VISION_EMBEDDINGS:
+            return {"last_hidden_state": {0: "num_patches"}}
+        return super().outputs
+
+    def generate_dummy_inputs(self, framework: str = "pt", **kwargs) -> Dict:
+        if self._behavior == VLMConfigBehavior.VISION_EMBEDDINGS:
+            kwargs["batch_size"] = 1
+        return super().generate_dummy_inputs(framework, **kwargs)
 
 
 @register_in_tasks_manager(
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -3008,6 +3008,121 @@ def maira_vision_embed_forward(self, pixel_values):
     return self.get_image_features(pixel_values, vision_feature_layer, vision_feature_select_strategy)
 
 
+
+def _mistral3_vision_embed_forward(self, pixel_values):
+    """
+    Full vision pipeline for Mistral3 export: vision_tower + multi_modal_projector.
+    Inlines PixtralVisionModel + PatchMerger to keep all shapes derived from
+    pixel_values.shape, ensuring dynamic dimensions in the OpenVINO IR.
+    """
+    vision_tower = self.model.vision_tower
+    patch_size = vision_tower.patch_size
+    max_width = vision_tower.config.image_size // patch_size
+    patch_conv_dtype = vision_tower.patch_conv.weight.dtype
+    if pixel_values.dtype != patch_conv_dtype:
+        pixel_values = pixel_values.to(patch_conv_dtype)
+
+    patch_embeds = vision_tower.patch_conv(pixel_values)
+    h_patches = patch_embeds.shape[2]
+    w_patches = patch_embeds.shape[3]
+
+    patch_embeds = patch_embeds.flatten(2).transpose(1, 2)
+    patch_embeds = vision_tower.ln_pre(patch_embeds)
+
+    h_idx = torch.arange(h_patches, device=pixel_values.device)
+    w_idx = torch.arange(w_patches, device=pixel_values.device)
+    mesh_h, mesh_w = torch.meshgrid(h_idx, w_idx, indexing="ij")
+    position_ids = mesh_h.reshape(-1) * max_width + mesh_w.reshape(-1)
+
+    position_embeddings = vision_tower.patch_positional_embedding(patch_embeds, position_ids)
+
+    transformer_out = vision_tower.transformer(
+        patch_embeds,
+        attention_mask=None,
+        position_embeddings=position_embeddings,
+        output_hidden_states=True,
+        return_dict=True,
+    )
+
+    vision_feature_layer = self.config.vision_feature_layer
+    if isinstance(vision_feature_layer, int):
+        selected_image_feature = transformer_out.hidden_states[vision_feature_layer]
+    else:
+        hs_pool = [transformer_out.hidden_states[layer_idx] for layer_idx in vision_feature_layer]
+        selected_image_feature = torch.cat(hs_pool, dim=-1)
+
+    projector = self.model.multi_modal_projector
+    image_features = projector.norm(selected_image_feature.squeeze(0))
+
+    spatial_merge = projector.patch_merger.spatial_merge_size
+    d = image_features.shape[-1]
+
+    image_grid = image_features.view(h_patches, w_patches, d).permute(2, 0, 1).unsqueeze(0)
+    grid = torch.nn.functional.unfold(image_grid, kernel_size=spatial_merge, stride=spatial_merge)
+    grid = grid.view(d * spatial_merge**2, -1).t()
+
+    image_features = projector.patch_merger.merging_layer(grid)
+    image_features = projector.linear_1(image_features)
+    image_features = projector.act(image_features)
+    image_features = projector.linear_2(image_features)
+
+    return image_features
+
+
+class Mistral3ImageEmbeddingModelPatcher(ModelPatcher):
+    def __init__(
+        self,
+        config: "OnnxConfig",
+        model: "PreTrainedModel",
+        model_kwargs: Dict[str, Any],
+    ):
+        model.__orig_forward = model.forward
+        model.forward = types.MethodType(_mistral3_vision_embed_forward, model)
+        super().__init__(config, model, model_kwargs)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+        self._model.forward = self._model.__orig_forward
+
+
+class Mistral3LanguageModelPatcher(OVDecoderModelPatcher):
+    """
+    Patcher for the language model part of Mistral3 VLM.
+    Fixes sliding_window=None crash — MinistralModel.forward unconditionally creates
+    sliding window masks even when all layers use full_attention.
+    """
+
+    def _get_lang_model(self):
+        if hasattr(self._model, "model") and hasattr(self._model.model, "language_model"):
+            return self._model.model.language_model
+        if hasattr(self._model, "language_model"):
+            return self._model.language_model
+        if hasattr(self._model, "model") and hasattr(self._model.model, "config"):
+            return self._model.model
+        return None
+
+    def __enter__(self):
+        super().__enter__()
+        lang_model = self._get_lang_model()
+
+        if lang_model is not None and hasattr(lang_model, "config"):
+            cfg = lang_model.config
+            self._orig_sliding_window = getattr(cfg, "sliding_window", None)
+            if self._orig_sliding_window is None:
+                max_pos = getattr(cfg, "max_position_embeddings", 32768)
+                cfg.sliding_window = max_pos
+
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        lang_model = self._get_lang_model()
+
+        if lang_model is not None and hasattr(lang_model, "config"):
+            lang_model.config.sliding_window = self._orig_sliding_window
+
+        super().__exit__(exc_type, exc_value, traceback)
+
+
 class LlavaImageEmbeddingModelPatcher(ModelPatcher):
     def __init__(
         self,
--- a/optimum/exporters/openvino/utils.py
+++ b/optimum/exporters/openvino/utils.py
@@ -339,6 +339,7 @@ MULTI_MODAL_TEXT_GENERATION_MODELS = [
     "minicpmo",
     "videochat_flash_qwen",
     "qwen3_omni_moe",
+    "mistral3",
 ]
 
 SSM_MODELS = [
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -7378,11 +7378,34 @@ if is_transformers_version(">=", "5.2"):
 
 
 class _OVMistral3ForCausalLM(OVModelForVisualCausalLM):
-    def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
+    def get_vision_embeddings(self, pixel_values, input_ids=None, image_sizes=None, **kwargs):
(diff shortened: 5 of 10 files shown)


anatyrova:
Hello @dhandhalyabhavik ! Thank you for your PR. 

When I tried to do export & inference, I faced errors and would like to suggest you make corrections. Tests also need to be added.

1) There are no tests. (However, tests exist for all other VLM models in the repository):
- mistral3 is not added to the matrix tests/openvino/test_decoder.py
- There is no export test in tests/openvino/test_export.py
- There is no inference test in tests/openvino/test_modeling.py
It was the lack of tests that allowed three bugs in FP8 processing and an unregistered tokenizer to appear in PR undetected.

2) API concerns (more in code review)

3) code duplication (more in code review)

4) The transformers version is undocumented. This PR was tested on an unknown transformers version. Exporting on 4.57.6 (the last version compatible with the <4.58 package limitation) doesn't work without the above patch. You need to specify the exact version in the PR description and in MIN_TRANSFORMERS_VERSION.

5) Export warnings. TracerWarnings appear during tracing. This is normal for VLM models and does not block export, but it should be documented as a known limitation. *

6) _FP8 handling does not work in any version of transformers_ (code review)

7) _Tokenizer is not saved when exporting_ (code review)

optimum/exporters/openvino/__main__.py
```diff
@@ -367,6 +367,18 @@ def main_export(
                 config.audio_processor["config"]["activation_checkpointing"] = ""
             config._attn_implementation = "sdpa"
             loading_kwargs["config"] = config
+
+        # Handle FP8 quantized models (e.g. Ministral-3B FP8) by dequantizing to BF16
+        quant_cfg = getattr(config, "quantization_config", None)
+        if quant_cfg is not None and getattr(quant_cfg, "quant_method", None) == "fp8":
+            try:
```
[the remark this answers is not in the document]
anatyrova: getattr instead of .get() on a dictionary; config.quantization_config is a dict at load time, not an object. getattr(quant_cfg, "quant_method", None) on a dictionary always returns None. The condition never fires, and the entire FP8 block is skipped.

optimum/exporters/openvino/__main__.py
```diff
@@ -367,6 +367,18 @@ def main_export(
...
+                from transformers import FineGrainedFP8Config
+
+                loading_kwargs["quantization_config"] = FineGrainedFP8Config(dequantize=True)
+            except (ImportError, Exception):
+                # If FineGrainedFP8Config not available, strip quantization to avoid errors
+                config.quantization_config = None
```
anatyrova: config.quantization_config = None doesn't help; hasattr(config, "quantization_config") returns True even when the value is None — the attribute exists. Transformers calls get_hf_quantizer and crashes with AttributeError: 'NoneType' object has no attribute 'get'.

optimum/exporters/openvino/__main__.py
```diff
@@ -367,6 +367,18 @@ def main_export(
...
             config._attn_implementation = "sdpa"
             loading_kwargs["config"] = config
+
+        # Handle FP8 quantized models (e.g. Ministral-3B FP8) by dequantizing to BF16
+        quant_cfg = getattr(config, "quantization_config", None)
+        if quant_cfg is not None and getattr(quant_cfg, "quant_method", None) == "fp8":
+            try:
+                from transformers import FineGrainedFP8Config
+
+                loading_kwargs["quantization_config"] = FineGrainedFP8Config(dequantize=True)
```
anatyrova: FineGrainedFP8Config(dequantize=True) is not working . Even if the block were to work, merge_quantization_configs recreates the config from the dictionary via AutoQuantizationConfig.from_dict, which re-initializes FineGrainedFP8Config(activation_scheme="static") and fails with ValueError: Activation scheme static not supported.

optimum/exporters/openvino/__main__.py
anatyrova: Try using smth like this:

```
quant_cfg = getattr(config, "quantization_config", None)
if quant_cfg is not None:
    if isinstance(quant_cfg, dict):
        quant_method = quant_cfg.get("quant_method", None)
    else:
        quant_method = getattr(quant_cfg, "quant_method", None)
    if quant_method == "fp8":
        config.__dict__.pop("quantization_config", None)  # delete completely 
        loading_kwargs["config"] = config
```

optimum/exporters/openvino/utils.py
```diff
@@ -303,6 +303,7 @@ def get_submodels(model):
     "phi4_multimodal",
     "llama4",
     "minicpmo",
+    "mistral3",
```
anatyrova, line 342: The tokenizer isn't saved during export.
After a successful export, only the OV models and preprocessor_config.json are present in the output folder. The tokenizer files (tokenizer.json, tokenizer_config.json, tokenizer.model) are missing.
Reason: save_preprocessors in utils.py silently catches all exceptions:
```
except Exception as ex:
    logger.error(f"Saving {type(processor)} failed with {ex}")
```
For phi3_v and llava, the same function includes special preprocessing before saving, but not for mistral3. Because of this, Mistral3Processor.save_pretrained() fails silently and the tokenizer isn't saved. Inference from a local folder is impossible; the processor must be downloaded separately from HuggingFace Hub.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8319,3 +8319,128 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        model: "PreTrainedModel",
+        model_kwargs: Dict[str, Any],
+    ):
+        model.__orig_forward = model.forward
+        model.forward = types.MethodType(_mistral3_vision_embed_forward, model)
+        super().__init__(config, model, model_kwargs)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+        self._model.forward = self._model.__orig_forward
```
anatyrova, line 9467: Mistral3ImageEmbeddingModelPatcher (pattern __orig_forward + types.MethodType) duplicates with 8 other patchers in model_patcher.py

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8319,3 +8319,128 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        if lang_model is not None and hasattr(lang_model, "config"):
+            cfg = lang_model.config
+            self._orig_sliding_window = getattr(cfg, "sliding_window", None)
+            if self._orig_sliding_window is None:
+                max_pos = getattr(cfg, "max_position_embeddings", 32768)
+                cfg.sliding_window = max_pos
+
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
```
anatyrova, line 9850: logics in __enter__ & __exit__ duplicates - same code 2 times

maybe try this:

def _get_lang_model(self):
    if hasattr(self._model, "model") and hasattr(self._model.model, "language_model"):
        return self._model.model.language_model
    if hasattr(self._model, "language_model"):
        return self._model.language_model
    if hasattr(self._model, "model") and hasattr(self._model.model, "config"):
        return self._model.model
    return None

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4802,59 @@ def preprocess_inputs(
...
+        image_token_id = getattr(self.config, "image_token_index", getattr(self.config, "image_token_id", 10))
+        special_image_mask = input_ids == image_token_id
+        special_image_mask = special_image_mask.unsqueeze(-1).expand_as(inputs_embeds).to(inputs_embeds.device)
+        image_features = image_features.view(-1, image_features.shape[-1]).to(inputs_embeds.device, inputs_embeds.dtype)
+        inputs_embeds = inputs_embeds.masked_scatter(special_image_mask, image_features)
+
+        return inputs_embeds, attention_mask, position_ids
+
+    @staticmethod
+    def preprocess_inputs(
```
anatyrova, line 7274: preprocess_inputs duplicates w _OVLlama4ForCausalLM, _OVGemma3ForCausalLM, _OVQwen2VLForCausalLM & 2 more

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -4802,6 +4802,59 @@ def preprocess_inputs(
         return inputs
 
 
+class _OVMistral3ForCausalLM(OVModelForVisualCausalLM):
+    def get_vision_embeddings(self, pixel_values, input_ids=None, **kwargs):
+        if input_ids is not None and input_ids.shape[1] == 1:
```
anatyrova, line 7382: The meaning of the check is the same everywhere: shape[1] == 1 means we're in the decoding phase (generating one token at a time), not prefill—in this case, the image has already been processed and vision embeddings are not needed; we return None.

This behavior could have been factored into the OVModelForVisualCausalLM base class once, but instead it's copied in every subclass. PR adds an eighth copy.

Classes (& precise string) it copies:

_OVLlavaNextVideoForCausalLM | 1454
_OVInternVLForCausalLM | 1704
_OVMiniCPMVForCausalLM | 1928
_OVNanoLlavaForCausalLM | 2196
_OVGemma3ForCausalLM | 3856
_OVLlama4ForCausalLM | 4740

optimum/exporters/openvino/model_configs.py
```diff
@@ -4547,6 +4565,103 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
         return Llama4ImageEmbeddingsModelPatcher(self, model, model_kwargs)
 
 
+@register_in_tasks_manager("mistral3", *["image-text-to-text"], library_name="transformers")
+class Mistral3OpenVINOConfig(BaseVLMOpenVINOConfig):
+    MIN_TRANSFORMERS_VERSION = "4.50.0"
+
```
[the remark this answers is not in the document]
anatyrova: Mistral3OpenVINOConfig.__init__ (pattern of _orig_config) duplicates with LlavaOpenVINOConfig, Gemma3OpenVINOConfig, Idefics3OpenVINOConfig

optimum/exporters/openvino/__main__.py
```diff
@@ -367,6 +367,18 @@ def main_export(
                 config.audio_processor["config"]["activation_checkpointing"] = ""
             config._attn_implementation = "sdpa"
             loading_kwargs["config"] = config
+
+        # Handle FP8 quantized models (e.g. Ministral-3B FP8) by dequantizing to BF16
```
anatyrova: External API (affects all users, not just mistral3)

The FP8 dequantization block in `__main__.py` is applied globally to any model
with `quantization_config.quant_method == "fp8"`, not scoped to mistral3:

```python
quant_cfg = getattr(config, "quantization_config", None)
if quant_cfg is not None and getattr(quant_cfg, "quant_method", None) == "fp8":
    ...
```
This silently changes export behavior for all existing FP8 models without any
user-visible warning or opt-in flag. If intentional, it should be:

- documented in the PR description as a behavior change
- guarded with a log message so users know dequantization happened
- ideally exposed as an explicit CLI flag (--dequantize-fp8)

optimum/exporters/openvino/model_configs.py
```diff
@@ -288,6 +290,22 @@ def init_model_configs():
...
+    )
+
+    # Register "ministral3" text config type so Mistral3Config can instantiate its text sub-config
+    try:
+        from transformers.models.auto.configuration_auto import CONFIG_MAPPING
+
+        if "ministral3" not in CONFIG_MAPPING:
+            from transformers.models.ministral.configuration_ministral import MinistralConfig
+
+            CONFIG_MAPPING.register("ministral3", MinistralConfig, exist_ok=True)
```
anatyrova: Internal API (fragile dependencies)
Direct access to CONFIG_MAPPING internals (model_configs.py):


```
from transformers.models.auto.configuration_auto import CONFIG_MAPPING
CONFIG_MAPPING.register("ministral3", MinistralConfig, exist_ok=True)
```
This is a private transformers API that may change without notice. The except Exception: pass silently swallows failures, leaving the user with a confusing error later during export rather than here.

optimum/exporters/openvino/model_configs.py
```diff
@@ -4547,6 +4565,103 @@ def patch_model_for_export(self, model: PreTrainedModel, model_kwargs: Optional[
...
+            int_dtype=int_dtype,
+            float_dtype=float_dtype,
+            preprocessors=preprocessors,
+        )
+        self._orig_config = config
+        if self._behavior == VLMConfigBehavior.VISION_EMBEDDINGS and hasattr(config, "vision_config"):
+            self._config = config.vision_config
+            self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
+
+    def with_behavior(
```
anatyrova, line 5131: with_behavior() has no fallback (model_configs.py): if behavior doesn't match any of the three branches, the method returns None implicitly. Other VLM configs raise ValueError in this case ; should be consistent.

optimum/intel/openvino/modeling_visual_language.py
anatyrova, line 7274: Wrong type hint in preprocess_inputs: processor is annotated as Optional[AutoImageProcessor] but the method calls processor.apply_chat_template(), which is an AutoProcessor method. 

AutoImageProcessor is a class for image processing only (resize, pixel normalization). It doesn't have an apply_chat_template method.

AutoProcessor is a combined class for VLM models; it has image processing, a tokenizer, and an apply_chat_template method.

In other words, the annotation says one thing, but the code uses something else. If a developer looks at the function signature, they'll pass AutoImageProcessor and get an AttributeError: 'AutoImageProcessor' object has no attribute 'apply_chat_template' at runtime.

anatyrova (changes requested):

rkazants (changes requested):
Please provide proper PR desacription (see other examples), add tests for export, inference and quantization

optimum/exporters/openvino/__main__.py
dhandhalyabhavik: dequantizing is lossless so I didn't thought of adding anything, but you are right, there should be log message and PR description. Added it.

optimum/exporters/openvino/__main__.py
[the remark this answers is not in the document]
dhandhalyabhavik: replaced them with
```python3
if quant_method == "fp8":
```

optimum/exporters/openvino/__main__.py
dhandhalyabhavik: Good catch, I took your suggestion, removed fp8 handling part and removed quant from config, default openvino will handle during load time.

optimum/exporters/openvino/model_patcher.py
dhandhalyabhavik: Agreed it's duplicative boilerplate, but this is the existing pattern used by 6+ other patchers (Llava, Maira, MiniCPM, Phi3, etc.). Refactoring all of them into a generic class is out of scope for this PR. Happy to do it as a separate cleanup PR if desired.

optimum/exporters/openvino/model_patcher.py
dhandhalyabhavik: Good find, I took your suggestion and created `_get_lang_model()` helper.

optimum/intel/openvino/modeling_visual_language.py
dhandhalyabhavik: There are two possible solutions I can think of
1) Assign the method from another class explicitly, something like this
```python3
class _OVMistral3ForCausalLM(OVModelForVisualCausalLM):
   preprocess_inputs = _OVIdefics3ForCausalLM.preprocess_inputs
```

2) Adding common default `_default_preprocess_inputs` method as static and inherit it at all duplicate places, for now I have used this and added fix for ministral
```python3
# In base class or as a module-level function
@staticmethod
def _default_preprocess_inputs(text, image=None, processor=None, ...):
   ...

# In Mistral3:
preprocess_inputs = staticmethod(_default_preprocess_inputs)
```

optimum/exporters/openvino/__main__.py
dhandhalyabhavik: Removed int8 part based on your suggestion from other comments.

optimum/intel/openvino/modeling_visual_language.py
dhandhalyabhavik: Agreed this could be factored into the base class, but this is an existing pattern used by 7+ other subclasses (Llava, InternVL, MiniCPM, Gemma3, Llama4, etc.). Refactoring all of them is out of scope for this PR. Happy to do it as a separate cleanup PR if desired.

optimum/exporters/openvino/__main__.py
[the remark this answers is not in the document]
dhandhalyabhavik: removed it now

optimum/exporters/openvino/model_configs.py
dhandhalyabhavik: I am stuck on this, because transformers >= 5.0 solves the problem but I don't know if it will break other models. Let me think and solve it properly or else I will just add logs with #TODO as comment to modify once we move to versions >=5.0

optimum/exporters/openvino/__main__.py
dhandhalyabhavik: with transformer 5.0.0 we get two issues resolved
1) fp8 to fp16 dequantization automatically
2) 'ministral3' support - no longer we need to use private APIs to register it

optimum/exporters/openvino/__main__.py
dhandhalyabhavik: Removed it with proper verbose.

optimum/exporters/openvino/utils.py
dhandhalyabhavik: Transformer has `TokenizersBackend` which addresses this. I am able to see both tokenizers and preprocessor files.

optimum/exporters/openvino/__main__.py
dhandhalyabhavik: with transformer 5.0.0 it now addresses the issue.

optimum/exporters/openvino/model_configs.py
dhandhalyabhavik: moved to transformer 5.0.0 because it was solving multiple issues.

optimum/exporters/openvino/model_configs.py
dhandhalyabhavik: added `raise ValueError(f"Unsupported Mistral3 export behavior: {behavior}")`

optimum/intel/openvino/modeling_visual_language.py
dhandhalyabhavik: Addressed it using 'ProcessorMixin' which has `PixtralProcessor` which is better type hint than AutoImageProcessor.

optimum/exporters/openvino/model_configs.py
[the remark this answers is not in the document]
dhandhalyabhavik: Duplication requires refactoring at multiple places and its out of scope for this PR. If you want I can raise separate PR that refactors and remove duplication for all above listed models.

dhandhalyabhavik (commented):
@rkazants & @anatyrova 

Thank you so much for detailed review comments, I have resolved almost all of them. Kindly requesting you to review the changes I have submitted. Added tests as well.

Apologies for not following proper PR structure and format. I will make sure not to repeat this in future. Please review latest changes.

optimum/intel/openvino/modeling_visual_language.py
dhandhalyabhavik: BTW I've rebased this PR onto the latest optimum-intel main. A few of modifications were already addressed by upstream changes in huggingface_hub, so this branch now includes those fixes as well.

anatyrova:
```
pytest tests/openvino/test_modeling.py tests/openvino/test_export.py tests/openvino/test_exporters_cli.py tests/openvino/test_quantization.py -k mistral3 -v
```
tests passed (checked with the command), int4 export and inference run w/o problems.

dhandhalyabhavik:
@rkazants can you please review the PR.

dhandhalyabhavik:
@rkazants can you please review the PR.

dhandhalyabhavik:
@rkazants can you please review the PR.

dhandhalyabhavik:
@rkazants can you please review the PR.

CVS-181872
Optimum-intel: CVS-181889

optimum/exporters/openvino/__main__.py
```diff
@@ -367,6 +370,18 @@ def main_export(
...
+        if model_type == "mistral3" and quant_method == "fp8":
+            logger.info(
+                "Detected FP8 weights in the Mistral3 checkpoint; dequantizing them before exporting the model "
+                "to OpenVINO."
+            )
+            if isinstance(quantization_config, dict):
+                quantization_config["dequantize"] = True
+            else:
+                quantization_config.dequantize = True
+            loading_kwargs["config"] = config
```
rkazants, line 469: please avoid these changes in this file. We rarely change this file for new model adding

tests/openvino/test_modeling.py
```diff
@@ -216,6 +216,24 @@ def test_load_from_hub_and_save_decoder_model(self, use_cache):
         del model
         gc.collect()
 
+    @unittest.skipIf(is_transformers_version("<", "5.0.0"), reason="Mistral3 requires transformers>=5.0.0")
+    def test_export_and_infer_mistral3_visual_language_model(self):
```
rkazants: we have special test suite for VLM please use it. No need to implement separate test for inference

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3407,6 +3407,121 @@ def maira_vision_embed_forward(self, pixel_values):
     return self.get_image_features(pixel_values, vision_feature_layer, vision_feature_select_strategy)
 
 
+
```
rkazants, line 3011: please share a link to code-line what you exactly patch and why

dhandhalyabhavik: Upstream's vision pipeline encodes image sizes into a Python list `image_sizes` and does **looping, splitting, and concatenation** over that list in multiple places. We cannot bake this into ov-export so during export, tracing bakes those list values in as constants, so the graph isn't dynamic-shape friendly. This variable is used at multiple places in key components as follows:
- patch-embedding list + meshgrid positions: [pixtral L497](https://github.com/huggingface/transformers/blob/v5.0.0/src/transformers/models/pixtral/modeling_pixtral.py#L497)
- per-image block attention mask: [pixtral L518](https://github.com/huggingface/transformers/blob/v5.0.0/src/transformers/models/pixtral/modeling_pixtral.py#L518)
- per-image PatchMerger [mistral3 L84](https://github.com/huggingface/transformers/blob/v5.0.0/src/transformers/models/mistral3/modeling_mistral3.py#L84)

Patch implements following:

The vision submodel is always exported with a **single image**, so we drop the list logic and derive shapes directly:
- get `h`, `w` from `patch_embeds.shape` instead of `image_sizes`: [model_patcher.py L3424-L3426](https://github.com/dhandhalyabhavik/optimum-intel/blob/ministral-support/optimum/exporters/openvino/model_patcher.py#L3424-L3426)
- one image results in one attention block so block mask is unnecessary and we pass `attention_mask=None`: [model_patcher.py L3440](https://github.com/dhandhalyabhavik/optimum-intel/blob/ministral-support/optimum/exporters/openvino/model_patcher.py#L3440)
- PatchMerger then merges each 2×2 patch window into one token (no per-image `split`/`cat`), producing the vision submodule output.

The math is unchanged it's the same computation upstream does for a single image but without the data-dependent Python loops/splits, so it exports to a dynamic-shape OpenVINO graph.

tests/openvino/test_modeling.py
dhandhalyabhavik: Removed it, using default VLM test. It passed all 3 tests.

optimum/exporters/openvino/__main__.py
dhandhalyabhavik: Ministral 3B and 8B weights are in fp8, this block is necessary for us to dequantize tensors to fp16 before tracing. If you want I can remove model specific condition and make it generic. Without this model is generating garble values. Correct me if I wrong but in our current codebase, optimum-intel doesn't natively dequantize fp8 to fp16/bf16.

popovaan:
There's a tiny model available for Ministral3 https://huggingface.co/optimum-intel-internal-testing/tiny-random-ministral3
Please use it in tests and add export, seq2seq and quantization tests.

popovaan:
Please update docs/source/openvino/models.mdx

popovaan:
Please fix code style
https://github.com/huggingface/optimum-intel/actions/runs/34982094981/job/104794160327?pr=1659

popovaan:
Could you please measure WWB similarity for the converted model with fp16, int4 and in8 weights and share results?
Here's the instruction on how to run WWB: https://github.com/openvinotoolkit/openvino.genai/tree/master/tools/who_what_benchmark#compare-visual-language-models-with-image-inputs-vlms
Running the last command don't use " --genai", it will trigger inference through optimum-intel.

dhandhalyabhavik:
@popovaan WWB scores

| Model | FP16 | INT4 | INT8 |
|---|---:|---:|---:|
| Ministral-3-3B-Reasoning-2512 | 0.984973 | 0.988140 | 0.987629 |
| Ministral-3-3B-Instruct-2512 | 0.951441 | 0.950368 | 0.949989 |

popovaan:
> @popovaan WWB scores
> 
> Model	FP16	INT4	INT8
> Ministral-3-3B-Reasoning-2512	0.984973	0.988140	0.987629
> Ministral-3-3B-Instruct-2512	0.951441	0.950368	0.949989

0.951441 is low for FP16 precision, we expect 0.97+.

Could you please try recollecting the ground truth using ` --torch-dtype float32` in WWB and measure similarity for Ministral-3-3B-Instruct-2512 again? If it doesn't improve accuracy, please try investigating why is it low.

dhandhalyabhavik:
> > @popovaan WWB scores
> > Model	FP16	INT4	INT8
> > Ministral-3-3B-Reasoning-2512	0.984973	0.988140	0.987629
> > Ministral-3-3B-Instruct-2512	0.951441	0.950368	0.949989
> 
> 0.951441 is low for FP16 precision, we expect 0.97+.
> 
> Could you please try recollecting the ground truth using ` --torch-dtype float32` in WWB and measure similarity for Ministral-3-3B-Instruct-2512 again? If it doesn't improve accuracy, please try investigating why is it low.

@popovaan 

It gave 0.97436.

tests/openvino/test_export.py
```diff
@@ -307,7 +309,8 @@ def test_export(self, model_type: str):
         model_kwargs = None
         if model_type == "speecht5":
             model_kwargs = {"vocoder": "fxmarty/speecht5-hifigan-tiny"}
-        self._openvino_export(model_type, model_kwargs=model_kwargs)
+        patch_16bit_model = model_type == "mistral3"
```
rkazants, line 312: disagree with this code. It should not be related to model type

optimum/exporters/openvino/__main__.py
```diff
@@ -467,6 +470,18 @@ def main_export(
                 config.audio_processor["config"]["activation_checkpointing"] = ""
             config._attn_implementation = "sdpa"
             loading_kwargs["config"] = config
+
+        # Mistral3 FP8 checkpoints are dequantized to BF16 for OpenVINO export.
+        if model_type == "mistral3" and quant_method == "fp8":
```
rkazants, line 475: we do not support fp8 quants. We mostly require model of bf16 and fp32. For this model, bf16 precision exists so we expect this precision.

Avoid changes in the common part. __main__ and ministral has no smth new from arch perspective to change common parts, It should be easily intergrated as other vlm models.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -3040,6 +3040,129 @@ def maira_vision_embed_forward(self, pixel_values):
     return self.get_image_features(pixel_values, vision_feature_layer, vision_feature_select_strategy)
 
 
+def _mistral3_vision_embed_forward(self, pixel_values):
```
rkazants, line 3043: provide comments fow shich patch -why it is needed and share a link to the original code what is exactly patched.

rkazants (changes requested):
Need to address left comments

PR #1678, opened 2026-04-08
[NNCF] Update test references for NNCF 3.1.0

andrey-churkin:
# What does this PR do?

- Updates the minimum required version of NNCF to 3.1.0
- Updates test references to align with the desired changes in NNCF

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ INSTALL_REQUIRE = [
     "optimum-onnx@git+https://github.com/huggingface/optimum-onnx.git@main",
     "transformers>=4.45,<4.58",
     "setuptools",
-    "nncf>=2.19.0",
+    "nncf>=3.1.0",
     "openvino>=2025.4.0",
     "openvino-tokenizers>=2025.4.0",
 ]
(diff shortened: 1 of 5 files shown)


tests/openvino/test_exporters_cli.py
```diff
@@ -539,7 +539,7 @@ class OVCLIExportTestCase(unittest.TestCase):
             "text-generation-with-past",
             "opt125m",
             "mxfp4",
-            {"model": {"int8": 4, "f4e2m1": 72, "f8e8m0": 72}},
+            {"model": {"int8": 0, "f4e2m1": 72, "f8e4m3": 2, "f8e8m0": 74}},
```
rkazants, line 542: why was it changed and now there is no int8 coeffs?

setup.py
```diff
@@ -31,7 +31,7 @@
     "optimum-onnx@git+https://github.com/huggingface/optimum-onnx.git@main",
     "transformers>=4.45,<4.58",
     "setuptools",
-    "nncf>=2.19.0",
+    "nncf>=3.1.0",
```
rkazants, line 34: why is it recommended to install 3.1.0 by default? Do we have any model requires it? Please recommend.

tests/openvino/test_exporters_cli.py
andrey-churkin: We recently merged a new feature into NNCF. Please see the https://github.com/openvinotoolkit/nncf/pull/3886. The references have been updated to reflect the current state.

setup.py
andrey-churkin: I bumped the minimum nncf version to 3.1.0 because there is a test that passes only with nncf ≥ 3.1.0, and the default behavior of the --backup-precision parameter was extended in that release. Please see the updated docs.

rkazants (commented):
let us move to the latest OV. After that we will merge this PR

rkazants (commented):
@andrey-churkin, please rebase your PR. We recently merged PR to move to the latest OpenVINO

PR #1681, opened 2026-04-13
[OPENVINO] Fix  : Sentence-transformers compatibility: remove deprecated _text_length usage Fixes #1680

### LINKED ISSUE #1680: Bug: sentence-transformers export fails due to removed _text_length attribute in version 5.4.0
**Description**

Exporting `sentence-transformers` models using the OpenVINO exporter fails with an `AttributeError` related to the `_text_length` attribute.

**This is leading multiple failing tests :** 
```
FAILED tests/openvino/test_exporters_cli.py::OVCLIExportTestCase::test_exporters_cli_full_quantization_10_feature_extraction - subprocess.CalledProcessError: Command 'optimum-cli export openvino --task feature-extraction --model optimum-intel-internal-testing/stsb-bert-tiny-safetensors --quant-mode int8 --library sentence_transformers --dataset c4 --num-samples 1 /tmp/tmp8apu_q83' returned non-zero exit status 1.
```
```
FAILED tests/openvino/test_exporters_cli.py::OVCLIExportTestCase::test_exporters_cli_sentence_transformers - AttributeError: type object 'SentenceTransformer' has no attribute '_text_length'. Did you mean: '_input_length'?
```

**Reproduction**
Reproduce the bug in this [notebook](https://colab.research.google.com/drive/1ydaxdDx8BosGlFmgoR3JlTyDt26kxhTD?usp=sharing)

Run the following command:
```
optimum-cli export openvino \
  --task feature-extraction \
  --model optimum-intel-internal-testing/stsb-bert-tiny-safetensors \
  --quant-mode int8 \
  --library sentence_transformers \
  --dataset c4 \
  --num-samples 1 \
  ov_model
```

**Error**
```
  File "/usr/local/bin/optimum-cli", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/optimum/commands/optimum_cli.py", line 219, in main
    service.run()
  File "/usr/local/lib/python3.12/dist-packages/optimum/commands/export/openvino.py", line 485, in run
    _main_quantize(
  File "/usr/local/lib/python3.12/dist-packages/optimum/exporters/openvino/__main__.py", line 681, in _main_quantize
    model = model_cls.from_pretrained(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/optimum/intel/openvino/modeling_base.py", line 613, in from_pretrained
    return super().from_pretrained(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/optimum/modeling_base.py", line 407, in from_pretrained
    return from_pretrained_method(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/optimum/intel/openvino/modeling_sentence_transformers.py", line 97, in _from_pretrained
    model = super()._from_pretrained(
            ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/optimum/intel/openvino/modeling_base.py", line 534, in _from_pretrained
    model = cls(
            ^^^^
  File "/usr/local/lib/python3.12/dist-packages/optimum/intel/openvino/modeling_sentence_transformers.py", line 30, in __init__
    self._text_length = MethodType(SentenceTransformer._text_length, self)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'SentenceTransformer' has no attribute '_text_length'. Did you mean: '_input_length'?
```

**Root Cause**

The exporter binds the private `_text_length` method from `SentenceTransformer`. However, this attribute has been removed in recent versions of the sentence-transformers library.

The binding appears to be unused in the current implementation (dead binding).

**Expected Behavior**

The export should complete successfully without relying on removed or private attributes.

**Environment**
optimum-intel: current main
sentence-transformers: latest(5.4.0)

MissLostCodes:
# What does this PR do?

This PR removes the usage of the `_text_length` attribute from `SentenceTransformer`, which is no longer available in recent versions of the library.

**Details**
- `_text_length` is a private API and has been removed in newer versions of `sentence-transformers`, causing an AttributeError during export.
- The code was binding this method using `MethodType`, but the bound method is not used elsewhere in the class.
- Since this is a **dead binding** and relies on a removed private API, it has been removed.

**Notes**
- While newer versions expose `_input_length`, it is also a private/internal API and not guaranteed to be stable.
- The fix avoids relying on private attribute and keeps the implementation compatible across versions.

**Result**
- Fixes export failure for sentence-transformers models.
- Improves compatibility with newer versions of the library.

Complete Testing [Notebook](https://colab.research.google.com/drive/1m11RYsSgp1eIDZVBN_DD96xIrE2UP7dZ?usp=sharing)

Fixes #1680

### DIFF
--- a/optimum/intel/openvino/modeling_sentence_transformers.py
+++ b/optimum/intel/openvino/modeling_sentence_transformers.py
@@ -27,7 +27,6 @@ class OVSentenceTransformer(OVModel):
         super().__init__(model, config, **kwargs)
 
         self.encode = MethodType(SentenceTransformer.encode, self)
-        self._text_length = MethodType(SentenceTransformer._text_length, self)
         self.default_prompt_name = None
         self.truncate_dim = None
         self.tokenizer = tokenizer


optimum/intel/openvino/modeling_sentence_transformers.py
```diff
@@ -27,7 +27,6 @@ def __init__(self, model=None, config=None, tokenizer=None, **kwargs):
         super().__init__(model, config, **kwargs)
 
         self.encode = MethodType(SentenceTransformer.encode, self)
-        self._text_length = MethodType(SentenceTransformer._text_length, self)
```
echarlaix, line 30: wouldn't that be an issue for previous sentence-transformers version (<= v5.3) https://github.com/huggingface/sentence-transformers/blob/v5.3.0/sentence_transformers/SentenceTransformer.py#L1075
also shouldn't we do the same with _input_length https://github.com/huggingface/sentence-transformers/blob/c500af50eb597aba2523024d98a842cde064c3fe/sentence_transformers/sentence_transformer/model.py#L636 or isn't it needed ?

echarlaix (commented):
Thanks @MissLostCodes

optimum/intel/openvino/modeling_sentence_transformers.py
MissLostCodes: From what i have observed ... this is a dead binding
_text_length is assigned but never used anywhere . So , removing it does not affect any version.

MissLostCodes: Regarding _input_length, since the original _text_length binding was not used anywhere, I avoided introducing a dependency on another private method , it makes it version independent .

echarlaix: @MissLostCodes this will result in an error when calling `.encode()`(a test to ensure this is working is currently missing)

for example :
```python
model_id = "sentence-transformers/all-MiniLM-L6-v2"
model = OVSentenceTransformer.from_pretrained(model_id)

sentences = [
    "The weather is lovely today.",
    "It's so sunny outside!",
    "He drove to the stadium.",
]

embeddings = model.encode(sentences)
```

MissLostCodes: Thanks for pointing this out — I tested the example you provided.

Calling `model.encode()` currently fails with:

```
AttributeError: 'OVSentenceTransformer' object has no attribute 'is_singular_input'
```

I think the issue is not limited to `_text_length`; the `encode` binding itself is problematic.

encode() depends on several additional internal methods (_input_length, is_singular_input, _resolve_prompt, _can_flatten_inputs, etc.).

This means the current encode binding is fragile and version-dependent.
I think the correct fix here would be to avoid binding `encode()` directly, rather than trying to patch missing private methods. 
To maintian user facing encode we need to define encode separately .

echarlaix: @MissLostCodes sentence-transformers support openvino models (+ export phase for model that aren't already) since v4.1 so I think it makes more sense to deprecate `OVSentenceTransformer` and to inherit from `SentenceTransformer` in the meantime, opened https://github.com/huggingface/optimum-intel/pull/1683

MissLostCodes: That makes sense — thanks for the clarification🙌

rkazants:
@echarlaix, shall we close this PR?
@altnnatra please take a look for the context.

PR #1684, opened 2026-04-15
 Add support for transformers v5.5

echarlaix:
Adds compatibility with transformers v5.5 and bumps the supported range to `>=4.57,<5.6`.

- `attention_mask` is now created explicitly when needed, since `generate()` no longer creates it since v5.2
- `mm_token_type_ids` is now computed and passed to `get_rope_index` as required by transformers v5.3 for Qwen2VL, Qwen2.5VL, and Qwen3VL models
- `position_ids` generation fixed for v5.3 decode step where `input_ids` is already sliced to 1 token
  - cache_position handling for SSM models (>=v4.54): SSM models no longer receive cache_position from generate(). It is now inferred from cache_params when absent
- AutoModelForVision2Seq → AutoModelForImageTextToText (>=v4.54)


tests/openvino/test_modeling.py
```diff
@@ -1131,13 +1130,16 @@ class OVModelForMaskedLMIntegrationTest(unittest.TestCase):
...
-    if is_transformers_version("<", "4.51.0"):
+    if is_transformers_version("<", "4.51"):
         SUPPORTED_ARCHITECTURES += ("nystromformer",)
 
     # TODO: add fix for v5 and update MAX_TRANSFORMERS_VERSION accordingly
     if is_transformers_version("<", "5"):
         SUPPORTED_ARCHITECTURES += ("data2vec-text", "flaubert", "xlm")
 
+    if is_transformers_version("!=", "4.52"):
+        SUPPORTED_ARCHITECTURES += ("convbert",)
```
echarlaix, line 1141: failing in v5.2 https://github.com/huggingface/transformers/blob/v5.2.0/src/transformers/modeling_utils.py#L2315 fixed in v5.3 since https://github.com/huggingface/transformers/commit/a64996e33be9d2a32b199a5e944403ecbca9fb18

setup.py
```diff
@@ -28,8 +28,8 @@
 
 INSTALL_REQUIRE = [
     "torch>=2.1",
-    "optimum-onnx@git+https://github.com/huggingface/optimum-onnx.git@transformers-v5",
-    "transformers>=4.45,<5.1",
+    "optimum-onnx@git+https://github.com/huggingface/optimum-onnx.git@xadupre/transformers5",
+    "transformers>=4.57,<5.5",
```
echarlaix: setting min transformers version to v4.57, will replicate everywhere once validated cc @rkazants

setup.py
```diff
@@ -28,8 +28,8 @@
 
 INSTALL_REQUIRE = [
     "torch>=2.1",
-    "optimum-onnx@git+https://github.com/huggingface/optimum-onnx.git@transformers-v5",
-    "transformers>=4.45,<5.1",
+    "optimum-onnx@git+https://github.com/huggingface/optimum-onnx.git@transformers-v5.5",
```
echarlaix:
```suggestion
    "optimum@git+https://github.com/huggingface/optimum",
```
should be updated before merging (once https://github.com/huggingface/optimum-intel/pull/1690 merged and https://github.com/huggingface/optimum/pull/2430 for v5.4)

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -875,6 +875,8 @@ def forward(
 
         # Add the attention_mask inputs when needed
         if "attention_mask" in self.input_names:
+            if attention_mask is None:
```
echarlaix, line 902: before v5.2 attention_mask was created in `generate` by calling `_prepare_attention_mask_for_generation`  https://github.com/huggingface/transformers/blob/v5.1.0/src/transformers/generation/utils.py#L2530 not the case of encoder_decoder model since v5.2 https://github.com/huggingface/transformers/blob/v5.2.0/src/transformers/generation/utils.py#L2555

tests/openvino/test_seq2seq.py
```diff
@@ -611,10 +598,15 @@ class OVModelForVisualCausalLMIntegrationTest(OVSeq2SeqTestMixin):
         SUPPORTED_ARCHITECTURES += ["internvl_chat", "minicpmv"]
 
     # TODO: add fix for v5 and update MAX_TRANSFORMERS_VERSION accordingly
-    if is_transformers_version("<", "5"):
-        SUPPORTED_ARCHITECTURES += ("llava_next_video",)
-    else:
-        UNSUPPORTED_ARCHITECTURES.update({"got_ocr2", "idefics3", "llama4", "llava_next_video", "smolvlm"})
+    _is_model_supported = {
```
echarlaix: TODO: extend for all architectures (waiting for validation min required transformers version set to 4.57 / can be done in a following PR if needed)

tests/openvino/test_seq2seq.py
```diff
@@ -153,20 +151,18 @@ class OVModelForSeq2SeqLMIntegrationTest(OVSeq2SeqTestMixin):
...
-    # TODO: add fix for v5 and update MAX_TRANSFORMERS_VERSION accordingly
-    if is_transformers_version("<", "5"):
-        SUPPORTED_ARCHITECTURES += ("mt5",)
-    else:
-        UNSUPPORTED_ARCHITECTURES.add("mt5")
+    # config loading failing coming from type mismatch coming from transformers v5.4
+    # known issues with marian on OpenVINO 2025.3.x and 2025.4.x
+    # TODO: add fix for v5 and update MAX_TRANSFORMERS_VERSION accordingly (mt5)
+    _is_model_supported = {
+        "m2m_100": is_transformers_version("!=", "5.4"),
```
echarlaix, line 164: issue coming from mismatch in https://github.com/huggingface/transformers/blob/v5.4.0/src/transformers/models/m2m_100/configuration_m2m_100.py#L68 (introduced in v5.4, fixed in 5.5)

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1070,7 +1061,7 @@ def _reorder_cache(
     INPUTS_DOCSTRING,
 )
 class OVModelForVision2Seq(OVModelForSeq2SeqLM):
-    auto_model_class = transformers_auto_class
+    auto_model_class = AutoModelForImageTextToText
```
IlyasMoutawwakil, line 1099: shouldn't we deprecate it and create OVModelForImageTextToText

regisss (approved):
LGTM

rkazants (commented):
please clean-up preview models (gemma 4, qwen3.5) job because they should come to regular jobs

optimum/intel/openvino/modeling_seq2seq.py
echarlaix: good point! updated in https://github.com/huggingface/optimum-intel/pull/1684/changes/ca3d05747fb59a6ab8a38b8f3b62f6f4a03dedcc

IlyasMoutawwakil (approved):
LGTM !

.github/workflows/test_openvino.yml
```diff
@@ -39,7 +39,7 @@ jobs:
             "*quantization*",
             "*transformation*",
           ]
-        transformers-version: ["4.45.0", "4.57.6", "latest"]
+        transformers-version: ["4.57.6", "5.3.0", "latest"]
```
rkazants: may be let us leave ` ["4.45.0", "4.57.6", "latest"]`?

tests/openvino/test_quantization.py
```diff
@@ -1066,58 +1068,48 @@ class OVWeightCompressionTest(unittest.TestCase):
...
         )
 
-    if is_transformers_version(">=", "4.49.0") and is_transformers_version("<=", "4.57.6"):
+    if is_transformers_version("==", "4.57.6"):
+        SUPPORTED_ARCHITECTURES_WITH_AUTO_COMPRESSION.append((OVModelForSpeechSeq2Seq, "qwen3_asr", True))
+
+    if is_transformers_version("<=", "4.57.6"):
         SUPPORTED_ARCHITECTURES_WITH_AUTO_COMPRESSION.append((OVModelForVisualCausalLM, "videochat_flash_qwen", True))
 
     if is_transformers_version(">=", "5.2.0") and is_transformers_version("<", "5.3.0"):
```
rkazants: can we use smart logic for adding cases based on min and max versions in config?

rkazants (changes requested):
We loose support for critical models: 
minicpm | MiniCPMOpenVINOConfig | 4.53.3 | 469
minicpm3 | MiniCPM3OpenVINOConfig | 4.53.3 | 478
chatglm | ChatGLM2OpenVINOConfig | 4.55.4 | 518
qwen | QwenOpenVINOConfig | 4.55.4 | 822
arctic | ArcticOpenVINOConfig | 4.53.3 | 1219
llava-qwen2 | LlavaQwen2OpenVINOConfig | 4.53.3 | 2133
minicpmo | MiniCPMOOpenVINOConfig | 4.51.3 | 2752
phi3_v | Phi3VisionOpenVINOConfig | 4.53.3 | 2769
phi4_multimodal | Phi4MMOpenVINOConfig | 4.53.3 | 2904
deepseek / deepseek_v2 | DeepseekOpenVINOConfig | 4.53.3 | 391

tests/openvino/test_quantization.py
echarlaix: can be added in a following pr

.github/workflows/test_openvino.yml
echarlaix: @rkazants we only test min/max supported transformers version (4.57.6 and 5.5)
https://github.com/huggingface/optimum-intel/blob/transformers-v5.5/.github/workflows/test_openvino.yml#L42 having more tests can result in rate limit errors as this launches x8 tests each time (can be an issue especially for concurrent PRs), would make more sense to have nightly tests for missing transformers release

.github/workflows/test_openvino_preview_models.yml
```diff
@@ -61,18 +61,14 @@ jobs:
...
-
       - name: Gemma 4 Unified Validation
-        if: ${{ matrix.test-pattern != '*quantization*' }}
+        if: ${{ matrix.test-pattern != '*decoder*' }}
         run: |
           uv pip install "transformers>=5.10,<5.11"
           pytest tests/openvino/${{ matrix.test-pattern }} -k gemma4_unified --durations=0
 
-      - name: Qwen3.5 and Qwen3.5-Moe Validation
+      - name: Legacy remote-code models validation
```
rkazants: let us have for remote-code-models separate CI group test_openvino_remote_code.yml. There we can install model specific dependencies for them.
In this current group we will test model support from future transformers version

echarlaix: added in https://github.com/huggingface/optimum-intel/pull/1684/changes/7722c9de93397386669c3382b5560e77c93e1861

.github/workflows/test_openvino_legacy_remote_code_models.yml
```diff
@@ -0,0 +1,56 @@
...
+  HF_TOKEN: ${{ secrets.HF_HUB_READ_TOKEN }}
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        test-pattern:
+          [
+            "*decoder*",
```
rkazants, line 33: export?

.github/workflows/test_openvino_legacy_remote_code_models.yml
```diff
@@ -0,0 +1,56 @@
+name: Legacy Remote Code Models Validation
```
rkazants, line 1: please include this into precommit. Now it seems not be included

tests/openvino/test_seq2seq.py
rkazants, line 1: will test_seq2seq still work for legacy remote-code models?

optimum/exporters/openvino/model_configs.py
```diff
@@ -95,11 +95,8 @@
     ArcticModelPatcher,
     BaichuanModelPatcher,
     BigBirdPegasusModelPatcher,
-    BlenderbotModelPatcher,
-    BlenderbotSmallModelPatcher,
```
rkazants, line 99: why were these patchers removed?

rkazants: same question for patchers removals below.

tests/openvino/test_seq2seq.py
echarlaix: yes as you can see in https://github.com/huggingface/optimum-intel/actions/runs/29125895141/job/86471294618

optimum/exporters/openvino/model_configs.py
echarlaix: BlenderbotSmallModelPatcher was only applied for transformers version < 4.56, but BlenderbotSmallOpenVINOConfig.MIN_TRANSFORMERS_VERSION was now upgraded to 4.57 so this patcher will never be applied

.github/workflows/test_openvino_legacy_remote_code_models.yml
echarlaix: added

.github/workflows/test_openvino_legacy_remote_code_models.yml
echarlaix: currently failing with 
```ValueError: Parameters iterable is empty (hint: use `parameterized.expand([], skip_on_empty=True)` to skip this test when the input is empty)```
can be added in a following PR to not block merging (this workflow is not related to transformers v5.5 support directly, these remote code models are already legacy models and we don't have any tests for v4.51)

PR #1691, opened 2026-04-16
[OpenVINO] Support LFM2-MoE

popovaan:
# What does this PR do?

Conversion cmd-line for `LiquidAI/LFM2-24B-A2B`:
```sh
optimum-cli export openvino -m LiquidAI/LFM2-24B-A2B ./LFM2-24B-A2B --task text-generation-with-past
```

Inference of `LiquidAI/LFM2-24B-A2B` using OpenVINO backend:

```python
import torch

model_dir="./LFM2-24B-A2B"

# Prepare input

# Run inference

print(output_text)
```

## Before submitting


rkazants (approved):

rkazants:
@popovaan, can you please fix it:

```
FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_find_untested_architectures - ValueError: For the task `text-generation`, the OpenVINO exporter supports {'lfm2_moe'} which are not tested
```

rkazants:
@echarlaix, @IlyasMoutawwakil, can you please review this PR?

tests/openvino/test_decoder.py
MissLostCodes, line 31:
```suggestion
    LFM2MoeOpenVINOConfig,
)
```

MissLostCodes: Without this import, `lfm2_moe` cannot be gated using its MIN_TRANSFORMERS_VERSION.

MissLostCodes:
```suggestion
              if is_transformers_version("<",  str(LFM2MoeOpenVINOConfig.MIN_TRANSFORMERS_VERSION)):
            supported_architectures -= {"lfm2_moe"}
```

MissLostCodes: Add `MIN_TRANSFORMERS_VERSION` check for lfm2_moe.

MissLostCodes (commented):
Fixed : 
```
tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_find_untested_architectures - ValueError: For the task text-generation, the OpenVINO exporter supports {'lfm2_moe'} which are not tested ===== 1 failed, 62 passed, 108 skipped, 1990 warnings in 538.40s (0:08:58)
```

MissLostCodes:
> @popovaan, can you please fix it:
> 
> ```
> FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_find_untested_architectures - ValueError: For the task `text-generation`, the OpenVINO exporter supports {'lfm2_moe'} which are not tested
> ```
@rkazants @popovaan 
I’ve suggested a fix for this by adding missing version gating for `lfm2_moe` in `test_find_untested_architectures`.
 [here](https://github.com/huggingface/optimum-intel/pull/1691#pullrequestreview-4128619666)

tests/openvino/test_decoder.py
```diff
@@ -308,7 +313,8 @@ def test_find_untested_architectures(self):
             supported_architectures -= {"bitnet"}
         if is_transformers_version("<", str(LFM2OpenVINOConfig.MIN_TRANSFORMERS_VERSION)):
             supported_architectures -= {"lfm2"}
-
+              if is_transformers_version("<",  str(LFM2MoeOpenVINOConfig.MIN_TRANSFORMERS_VERSION)):
```
MissLostCodes:
```suggestion
        if is_transformers_version("<",  str(LFM2MoeOpenVINOConfig.MIN_TRANSFORMERS_VERSION)):
```

MissLostCodes: Indentation fix

MissLostCodes (commented):
indentation fix

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8496,3 +8496,88 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+                conv_layer._orig_forward = conv_layer.slow_forward
+                conv_layer.slow_forward = types.MethodType(lfm2_short_conv_forward_patched, conv_layer)
+
+            if isinstance(layer, Lfm2MoeDecoderLayer) and isinstance(layer.feed_forward, Lfm2MoeSparseMoeBlock):
+                sparse_moe_block = layer.feed_forward
+                if isinstance(sparse_moe_block.experts, Lfm2MoeExperts):
+                    lfm2_moe_experts = sparse_moe_block.experts
+                    lfm2_moe_experts._orig_forward = lfm2_moe_experts.forward
+                    lfm2_moe_experts.forward = types.MethodType(lfm2_moe_experts_forward, lfm2_moe_experts)
+
```
echarlaix, line 8565: why is it needed ? why not setting expert implementation to batched_mm instead as in
https://github.com/huggingface/optimum-intel/blob/71e0584cbcfdd2293ae5685461345701dc102575/optimum/exporters/openvino/model_patcher.py#L416

tests/openvino/test_decoder.py
```diff
@@ -98,6 +99,9 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
     if is_transformers_version(">=", "4.57.0") and is_transformers_version("<", "5"):
         SUPPORTED_SSM_ARCHITECTURES += ("qwen3_next",)
 
+    if is_transformers_version(">=", "5.1.0"):
```
echarlaix: why not instead, as MIN_TRANSFORMERS_VERSION is set to 5.0?
```suggestion
    if is_transformers_version(">=", "5.0"):
```

echarlaix (commented):
Thanks a lot @popovaan

tests/openvino/test_exporters_cli.py
```diff
@@ -179,6 +179,14 @@ class OVCLIExportTestCase(unittest.TestCase):
             ]
         )
 
+    if is_transformers_version(">=", "5.1"):
```
MissLostCodes:
```suggestion
    if is_transformers_version(">=", "5.0"):
```

MissLostCodes: version mismatch

MissLostCodes (commented):
version mismatch

optimum/exporters/openvino/model_patcher.py
popovaan: I just checked how model works with `self._model.set_experts_implementation("batched_mm")` and openvino MoE optimizations do not work with it. They require the model to have a specific pattern of operations where MoE experts weights go to Transpose operation and then to MatMul. But with `self._model.set_experts_implementation("batched_mm")` something seems to have changed, and the pattern no longer matches. So I returned the patching of lfm2_moe_experts_forward, otherwise we have bad performance.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8496,3 +8496,88 @@ def __exit__(self, exc_type, exc_value, traceback):
                 sparse_moe_block = decoder_layer.mlp
                 decoder_layer.mlp.forward = decoder_layer.mlp._orig_forward
                 del sparse_moe_block.down_projs, sparse_moe_block.gate_projs, sparse_moe_block.up_projs
+
+
+def lfm2_moe_experts_forward(
```
echarlaix, line 8504: would you mind adding a link to the original code as well https://github.com/huggingface/transformers/blob/v5.0.0/src/transformers/models/lfm2_moe/modeling_lfm2_moe.py#L167

optimum/exporters/openvino/model_patcher.py
echarlaix: in such case wouldn't it make sense to create an `openvino_experts_forward` that could be added to `ALL_EXPERTS_FUNCTIONS` instead of patching each models independently, here with `lfm2_moe_experts_forward` or with `batched_mm_experts_forward`

popovaan: Yes, that would be a good solution, but there are plans to rework how we currently handle MoE in OpenVINO and to use `GroupedMatMul`: https://github.com/openvinotoolkit/openvino/pull/35215/changes
So we will probably need to use "grouped_mm" patching eventually. Therefore, we would like to keep the current approach as a temporary solution, as it will need to be changed anyway once `GroupedMatMul` is implemented.

optimum/exporters/openvino/model_patcher.py
popovaan: Sure, done.

echarlaix (approved):
LGTM, thanks @popovaan

tests/openvino/test_decoder.py
```diff
@@ -705,7 +711,7 @@ def test_beam_search(self, model_arch):
         # LFM2 fails with beam search, issue link: https://github.com/huggingface/transformers/issues/42257
         # CVS-177964 GraniteMoeHybrid fails due to lack support of Beam search for hybrid models in OpenVINO
         # For this support, we expect changes in IRs to have connected beam_idx with Mamba/Linear attention states
-        if model_arch in ["lfm2", "granitemoehybrid"]:
+        if model_arch in ["lfm2", "lfm2_moe", "granitemoehybrid"]:
```
echarlaix, line 712: @popovaan why disabling the test for lfm2_moe ? (looks like the lfm2 issue https://github.com/huggingface/transformers/issues/42257 was fixed in v5)

popovaan: Yes, the transformers issue has been fixed, but the output with beam search doesn't match transformers. There seems to be a common issue with how SSM states are handled during beam search, as the "granitemoehybrid" output shows the same mismatch. We would prefer to address this in a separate PR since this model is urgent. I've created a ticket for that: CVS-185664

PR #1694, opened 2026-04-21
[NOT FOR MERGE][DEBUG] CVS-183350: Add debug instrumentation for CI reproduction

yuxu42:
## Purpose
This PR adds debug instrumentation to investigate CVS-183350 (OpenVINO 2026.1.0 inference precision regression) that only reproduces on CI hardware but not locally.

## Changes
- Add `setUpClass` to print CI CPU info (model name, flags)
- Add per-test debug output for opt/pegasus/xglm showing:
  - Actual max diff values
  - Tolerance threshold
  - Logits samples

## Expected CI Behavior
Tests for opt and pegasus will **fail**, but will print diagnostic information showing:
- CI CPU model (likely Intel Xeon Cascade Lake/Ice Lake)
- Actual numerical diff (expected > 1e-4)
- Logits comparison

This data will help root cause the OpenVINO CPU plugin regression.

## Note
This is a debug PR - not intended to merge. Will be reverted after collecting CI diagnostics.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,8 +33,8 @@ INSTALL_REQUIRE = [
     "setuptools",
     "huggingface-hub>=0.23.2,<2.0",
     "nncf>=2.19.0",
-    "openvino>=2025.3.0,<2026.1",
-    "openvino-tokenizers>=2025.3.0,<2026.1",
+    "openvino>=2026.1",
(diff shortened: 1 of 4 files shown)


rkazants:
Looks no longer needed PR

PR #1689, opened 2026-04-15
[OpenVINO] Support Qwen3.5, Qwen3.5-MoE and Qwen3.6

rkazants:
# What does this PR do?

Re-created PR https://github.com/huggingface/optimum-intel/pull/1634

Fixes 181271, 181280, 182003

Installation instructions:

```sh
pip install -U git+https://github.com/rkazants/optimum-intel.git@support_qwen3_5
pip install --pre -U openvino openvino-tokenizers nncf --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
pip install transformers==5.2.0
pip install requests torchvision opencv-python
```

Exporting cmd-line:

```sh
optimum-cli export openvino -m Qwen/Qwen3.5-0.8B Qwen3.5-0.8B
```

Inference script:

```python
from transformers import AutoProcessor
from transformers.video_utils import load_video
from huggingface_hub import hf_hub_download
from optimum.intel.openvino import OVModelForVisualCausalLM

model_dir = "Qwen/Qwen3.5-0.8B"

processor = AutoProcessor.from_pretrained(model_dir)
model = OVModelForVisualCausalLM.from_pretrained(model_dir)

# Prepare video input
video_path = hf_hub_download(
                repo_id="raushan-testing-hf/videos-test",
                filename="sample_demo_1.mp4",
                repo_type="dataset",
            )
input_video, _ = load_video(video_path, num_frames=10, backend="opencv")

messages = [
    {"role": "user", "content": [
        {"type": "video"},
        {"type": "text", "text": "Why is this video funny?"},
    ]}
]
text = processor.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
inputs = processor(text=[text], videos=[input_video], return_tensors="pt")

# Run inference
output_ids = model.generate(**inputs, max_new_tokens=100)
output_text = processor.decode(output_ids[0], skip_special_tokens=True)

print(output_text)
```

## Before submitting
- [N/A] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you make sure to update the documentation with your changes?
- [] Did you write any new necessary tests?

PR #1705, opened 2026-04-29
[OpenVINO] Fix beam_idx wiring in patch_stateful_hybrid_ssm for hybrid SSM/attention models

urakozz:
Hello team!

I believe I found a bug in hybrid_ssm. I was experimenting converting Qwen 3.6 models to IR (stripping off visual layer) and running them on OVMS and here is a small patch that addresses the issue

# What does this PR do?

Fixes `fuse_cache_reorder` usage: SSM states are recurrent accumulators, not per-beam caches — gathering them by `beam_idx` breaks the IR after `make_stateful` hides those SSM parameters as internal variables. So `kv_input_names` would be the correct subset to pass

With an additional check for VLM dispatch https://github.com/huggingface/optimum-intel/pull/1689/changes#diff-bd6225d5672fbb5fa75d6a97bffe1a23ace426dd21e0bc07376a6d29e5a7aaea this bug will most likely be reproducedon new Qwen models.

# Root cause:
`fuse_cache_reorder`: When the original code passes SSM tensors to fuse_cache_reorder for hybrid models, it inserts Gather ops on SSM input edges. After `make_stateful` subsequently hides those SSM tensors as internal variables, the resulting IR graph has Gather ops referring to beam_idx in ways inconsistent with the final input parameter list. OVMS rejects the IR with:

```
Model references undeclared parameters: opset1::Parameter beam_idx () -> (i32[?])
```

This bug doesn't surface for pure-SSM models (mamba, falcon_mamba) because they have no K/V tensors, so the for-loop in fuse_cache_reorder iterates zero times and the Gather ops are never inserted. It only surfaces for hybrid models with both attention and SSM layers (e.g., qwen3_5_text, also recently merged lfm2 family).

# Repro for the original bug:
Convert any model whose inner architecture is in SSM_MODELS and contains both attention and SSM layers. With current main branch (or #1689 branch):
- Export: succeeds
- OVMS load: fails with "Model references undeclared parameters: beam_idx"

With this PR:

- Export: succeeds
- OVMS load: succeeds

Relates to #1689 so @rkazants this might be relevant for your `patch_stateful` change

UPD:
This fixes the `beam_idx` wiring bug for hybrid SSM models. Note: separate issues exist for LFM2 family related to `attention_mask` handling. This PR addresses the `beam_idx` issue specifically (so far)

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/stateful.py
+++ b/optimum/exporters/openvino/stateful.py
@@ -301,7 +301,9 @@ def patch_stateful_hybrid_ssm(ov_model: ov.Model):
     # hybrid models can contain transformer blocks as well
     # so KV tensors must be handled properly
     batch_dim = 0
-    fuse_cache_reorder(ov_model, not_cache_inputs, cache_inputs, batch_dim)
+    # Only KV tensors need beam-search reorder; SSM states are not per-beam
+    if kv_input_names:
+        fuse_cache_reorder(ov_model, not_cache_inputs, kv_input_names, batch_dim)
     make_stateful(ov_model, not_cache_inputs, cache_inputs, cache_outputs, batch_dim)
 
 


rkazants (commented):
WIll it fix this one https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_decoder.py#L723 ?
If yes, please remove it and we will have test coverage.

urakozz:
@rkazants not yet I'm afraid

I just tried lfm2, ovms with build from main branch: 
```
Model references undeclared parameters: 
opset1::Parameter beam_idx () -> (i32[?])
opset1::Parameter attention_mask () -> (i64[?,?])
```

and from my fix branch:
```
Model references undeclared parameters: 
opset1::Parameter attention_mask () -> (i64[?,?])
```

For qwen 3.6 my fix on top of your PR branch was enough to run it in ovms, it was failing on the template processing (from ovms genai). And debugging C++ felt a bit painful

Let me check if I can manage to run LFM, then we could try to uncomment tests

rkazants:
> @rkazants not yet I'm afraid
> 
> I just tried lfm2, ovms with build from main branch:
> 
> ```
> Model references undeclared parameters: 
> opset1::Parameter beam_idx () -> (i32[?])
> opset1::Parameter attention_mask () -> (i64[?,?])
> ```
> 
> and from my fix branch:
> 
> ```
> Model references undeclared parameters: 
> opset1::Parameter attention_mask () -> (i64[?,?])
> ```
> 
> For qwen 3.6 my fix on top of your PR branch was enough to run it in ovms, it was failing on the template processing (from ovms genai). And debugging C++ felt a bit painful
> 
> Let me check if I can manage to run LFM, then we could try to uncomment tests

Please check, then it will be a great fix.

Thanks,
Roman

urakozz:
@rkazants I felieve I need your read on this. I was checking today of the topology and connections between optimum-intel, openvino, genai and OVMS. Working on new failing test here I realised an important caveat I didn't put attention to earlier.

My understanding today's evening after 3 days of trial-and-error debugging is that optimum uses openvino library directly, while OVMS operates through genai that makes its own conversion before reaching openvino (that guys: ContinuousBatchingPipeline -> SDPAToPagedAttention), right?

**My findings today:**
- patch in this PR on top of #1689  breaks test `test_compare_to_transformers_40_lfm2_moe` (model compilation looks important), but makes qwen 3.5 run in the OVMS 2026.1
- commenting out this conversion https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_patcher.py#L7285 eliminates `opset1::Parameter attention_mask () -> (i64[?,?])` error and LFM2 starts in OVMS
- both fail at the first http request for different reasons

**Bottom line:**
I feel it sounds like sort of hacks to trick genai rather then workaround. If model compiles here against openvino, I guess then genai needs to be adjusted to consume IR model and talk to openvino from its pipelines

**One more thing:**
To make myself and this PR useful I can change it to remove this skipTest https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_decoder.py#L472 , that discussion was resolved and merged long time ago and no new tests are failing when I comment it out

Overall does this match your read of the layering, or am I missing something?

PR #1731, opened 2026-05-08
optimum-intel requirements should include requests

andrey-churkin:
# What does this PR do?

- Add the missing `request` module
- Add a GitHub workflow to test that `optimum-intel` works after being installed from source using `pip install -e .`

Fixes:
- CVS-185672

## Before submitting

### DIFF
--- a/.github/workflows/test_installation.yml
+++ b/.github/workflows/test_installation.yml
@@ -1,0 +1,54 @@
+name: Test Installation
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - v*-release
+  pull_request:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+env:
+  UV_TORCH_BACKEND: cpu
+  UV_SYSTEM_PYTHON: true
+  TRANSFORMERS_IS_CI: true
+  HF_TOKEN: ${{ secrets.HF_HUB_READ_TOKEN }}
+
+jobs:
+  build:
+
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: |
+          pip install --upgrade pip uv
+          uv pip install -e .
+
+      - name: Login with fork PRs CI token
+        if: ${{ env.HF_TOKEN == '' }}
+        run: |
+          python tests/scripts/login_with_ci_token.py
+
+      - name: Test installation
+        run: |
+          optimum-cli export openvino \
+            -m Qwen/Qwen3-VL-2B-Instruct \
+            temp_model
--- a/.github/workflows/test_user_installation.yml
+++ b/.github/workflows/test_user_installation.yml
@@ -40,6 +40,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install --upgrade pip uv
+          # test user installation
           uv pip install .
 
       - name: Login with fork PRs CI token
@@ -50,5 +51,5 @@ jobs:
       - name: Test Model Exporting
         run: |
           optimum-cli export openvino \
-            -m Qwen/Qwen3-VL-2B-Instruct \
+            -m google/gemma-3-270m \
             temp_model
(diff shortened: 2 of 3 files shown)


.github/workflows/test_installation.yml
```diff
@@ -0,0 +1,54 @@
+name: Test Installation
```
rkazants:
```suggestion
name: Test User Installation
```

.github/workflows/test_installation.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: |
+          pip install --upgrade pip uv
+          uv pip install -e .
```
rkazants:
```suggestion
          uv pip install .
```

.github/workflows/test_installation.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Login with fork PRs CI token
+        if: ${{ env.HF_TOKEN == '' }}
+        run: |
+          python tests/scripts/login_with_ci_token.py
+
+      - name: Test installation
```
rkazants:
```suggestion
      - name: Test Model Exporting
```

.github/workflows/test_user_installation.yml
rkazants, line 1: please rename file to `test_user_installation.yml`

rkazants (approved):

.github/workflows/test_installation.yml
IlyasMoutawwakil: why does this require a workflow on its own ?

andrey-churkin: I think the main reason is that we want to install optimum-intel via `pip install .` instead of `pip install .[tests]`, because some modules may be missing in `INSTALL_REQUIRE` but present in the test requirements. This way, we might forget to specify something in `INSTALL_REQUIRE`, as was the case with the `requests` module.

andrey-churkin: Thanks, fixed

.github/workflows/test_installation.yml
andrey-churkin: Thanks, fixed

.github/workflows/test_installation.yml
andrey-churkin: Thanks, fixed

.github/workflows/test_installation.yml
IlyasMoutawwakil: makes sense !

IlyasMoutawwakil: i thin we can still use uv for faster installation tho, also a smaller model for the export would be nice

rkazants: @IlyasMoutawwakil, I propose to use `Qwen/Qwen2.5-0.5B` for this test.

.github/workflows/test_user_installation.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Login with fork PRs CI token
+        if: ${{ env.HF_TOKEN == '' }}
+        run: |
+          python tests/scripts/login_with_ci_token.py
+
+      - name: Test Model Exporting
+        run: |
+          optimum-cli export openvino \
+            -m Qwen/Qwen3-VL-2B-Instruct \
+            temp_model
```
rkazants, line 55:
```suggestion
          optimum-cli export openvino \
            -m google/gemma-3-270m \
            temp_model
```

.github/workflows/test_user_installation.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: |
+          pip install --upgrade pip uv
+          uv pip install .
```
rkazants, line 44:
```suggestion
          # test user installation
          uv pip install .
```

.github/workflows/test_user_installation.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Login with fork PRs CI token
+        if: ${{ env.HF_TOKEN == '' }}
+        run: |
+          python tests/scripts/login_with_ci_token.py
+
+      - name: Test Model Exporting
+        run: |
+          optimum-cli export openvino \
+            -m google/gemma-3-270m \
```
rkazants:
```suggestion
            -m Qwen/Qwen2.5-0.5B \
```

PR #1733, opened 2026-05-13
Support LTX2 text-to-audio-video pipeline

anatyrova:
## Summary
This PR adds support for the LTX-2 text-to-audio-video pipeline in optimum-intel.
It is recommended to use OpenVINO >= 2026.3, because it has a fix to save RAM memory.
```
uv pip install diffusers==0.38.0
uv pip install --upgrade --pre openvino openvino-tokenizers --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly

# uninstall nncf to save weights in bf16 format
uv pip uninstall nncf

optimum-cli export openvino --model Lightricks/LTX-2 --task text-to-video output_dir
```
Inference script for text-to-audio-video:
```
import torch
import scipy.io.wavfile
import subprocess

from optimum.intel.openvino import OVLTX2Pipeline
from diffusers.utils import export_to_video

f32_config = {
    "INFERENCE_PRECISION_HINT": "f32"
}

pipe = OVLTX2Pipeline.from_pretrained(
    "/home/anatyrova/optimum-intel/.claude/worktree/add-ltx2/output_dir_bf16",
    ov_config=f32_config,
)

prompt = (
    "Close-up shot, shallow depth of field, cinematic realism. "
    "A young woman, mid-20s, dark straight hair pulled back loosely, wearing a light beige linen shirt, "
    "stands by a sunlit window in a quiet café. Warm golden morning light filters through sheer curtains, "
    "casting soft shadows on a wooden table. A white ceramic cup of coffee sits in front of her. "
    'She slowly wraps both hands around the cup, lifts it slightly, closes her eyes for a brief moment, '
    'then exhales. She says quietly: "Finally." '
    "Subtle ambient café sounds — distant murmur of voices, a spoon clinking against porcelain, "
    "soft acoustic guitar playing low in the background. "
    "The camera holds still, then gently pushes in toward her face as she opens her eyes and almost smiles."
)
negative_prompt = "blurry, low quality, watermark, text, logo"

output = pipe(
    prompt=prompt,
    negative_prompt=negative_prompt,
    num_frames=97,
    height=480,
    width=704,
    num_inference_steps=30,  
    guidance_scale=3.0,
    audio_guidance_scale=7.0,
    generator=torch.manual_seed(42),
)

video = output.frames[0]
audio = output.audio

# Save video without audio
video_path = "/home/anatyrova/optimum-intel/.claude/worktree/add-ltx2/ltx2_bf16.mp4"
export_to_video(video, video_path, fps=24)
print(f"Video saved: {len(video)} frames, {len(video)/24:.1f}s")

# Save audio as WAV (24kHz stereo — vocoder output_sampling_rate)
audio_np = audio[0].cpu().float().numpy()
sample_rate = 24000
print(f"Audio shape: {audio_np.shape}, sample rate: {sample_rate}")
audio_wav_path = "/home/anatyrova/optimum-intel/.claude/worktree/add-ltx2/ltx2_bf16_au.wav"
scipy.io.wavfile.write(audio_wav_path, sample_rate, audio_np.T)
print(f"Audio saved: {audio_wav_path}")
```

https://github.com/user-attachments/assets/0bde743d-f934-40ba-934f-7e16a7f533fc

```
prompt = ("A pair of metallic elevator doors slide open inside a spaceship corridor as thin mist rolls out from floor vents. The camera begins in a stationary wide shot, revealing a tall figure in a white uniform stepping forward through the haze. Blue accent lights line the corridor walls, casting geometric patterns on the polished floor. As the figure walks toward the camera, their footsteps echo with a hollow metallic ring. The camera glides sideways in a smooth tracking shot, following their stride past illuminated wall panels and sealed doorways. A low mechanical hum fills the background, punctuated by occasional electronic beeps from nearby systems.")
```

https://github.com/user-attachments/assets/a49efc1f-76e8-4a53-aa88-b6ffdb91dcd1

```
# command to merge video with audio
ffmpeg -i video.mp4 -i audio.wav -c:v copy -c:a aac -shortest output.mp4
```
wwb similarity bf16: 0.9842785


anatyrova:
export works, but inference does not. debugging

tests/openvino/utils_tests.py
```diff
@@ -349,6 +349,7 @@ def _create_tiny_kokoro_model():
     "sana": "optimum-intel-internal-testing/tiny-random-sana",
     "sana-sprint": "optimum-intel-internal-testing/tiny-random-sana-sprint",
     "ltx-video": "optimum-intel-internal-testing/tiny-random-ltx-video",
+    "ltx2-video": "optimum-intel-internal-testing/tiny-random-ltx2-video",
```
anatyrova: this is not actually on optimum-intel-internal-testing. yet

anatyrova:
@likholat check this pls

anatyrova:
for now, i tested doing export and inference for both int4,int8. they work, but i get blurry videos. i am searching, what could be the problem. maybe it's broken text conditioning, input/output formats that differ, **kwargs silently dropping critical parameters.

regisss:
@anatyrova It seems the quality check failed, can you run `make style` from the root of the repo please?

anatyrova:
> @anatyrova It seems the quality check failed, can you run `make style` from the root of the repo please?

fixed, thank you

regisss:
I think you need to rebase your branch on the latest main to make the doc build test pass

optimum/exporters/openvino/convert.py
```diff
@@ -1093,7 +1101,11 @@ def get_diffusion_models_for_export_ext(
     elif is_sana:
         models_for_export = get_sana_models_for_export(pipeline, exporter, int_dtype, float_dtype)
     elif is_ltx_video:
-        models_for_export = get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype)
+        is_ltx2 = pipeline.__class__.__name__.startswith("LTX2")
+        if is_ltx2:
+            models_for_export = get_ltx2_video_models_for_export(pipeline, exporter, int_dtype, float_dtype, task=task)
```
IlyasMoutawwakil: why does the task make any diff here ?

optimum/exporters/openvino/convert.py
```diff
@@ -1173,6 +1185,152 @@ def get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+    # mask creation is fully traceable (no data-dependent boolean ops).
+    # SDPA handles float masks correctly on real tokens; padding-position
+    # values differ from standard boolean masks but are unused downstream.
+    text_encoder = pipeline.text_encoder
+    text_encoder.config.output_hidden_states = True
+    orig_forward = text_encoder.forward
+
+    def text_encoder_forward(input_ids, attention_mask=None, **kwargs):
+        if attention_mask is not None and attention_mask.dim() == 2:
+            import torch
```
IlyasMoutawwakil: import at the top please

optimum/exporters/openvino/convert.py
```diff
@@ -1173,6 +1185,152 @@ def get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+            causal_mask = causal_mask.expand(bsz, 1, seq_len, seq_len).clone()
+            causal_positions = torch.tril(torch.ones(seq_len, seq_len, dtype=torch.float32))
+            causal_mask = causal_mask * causal_positions[None, None, :, :]
+            causal_mask = (1.0 - causal_mask) * torch.finfo(torch.float32).min
+            attention_mask = {"full_attention": causal_mask, "sliding_attention": causal_mask}
+        outputs = orig_forward(input_ids=input_ids, attention_mask=attention_mask, output_hidden_states=True)
+        result = {"last_hidden_state": outputs.hidden_states[-1]}
+        for i, hs in enumerate(outputs.hidden_states):
+            result[f"hidden_states.{i}"] = hs
+        return result
```
IlyasMoutawwakil: do we not support export gemma3 as an encoder ?

optimum/exporters/openvino/convert.py
```diff
@@ -1173,6 +1185,152 @@ def get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+
+    text_encoder.forward = text_encoder_forward
+
+    export_config_constructor = TasksManager.get_exporter_config_constructor(
+        model=text_encoder,
+        exporter=exporter,
+        library_name="diffusers",
+        task="feature-extraction",
+        model_type="gemma3-text-encoder",
```
IlyasMoutawwakil, line 1195: why not use transformers' gemma3 export path ?

optimum/exporters/openvino/convert.py
```diff
@@ -1173,6 +1185,152 @@ def get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+        task="semantic-segmentation",
+        model_type="ltx2-video-transformer",
+    )
+    transformer_export_config = transformer_config_constructor(
+        transformer.config, int_dtype=int_dtype, float_dtype=float_dtype
+    )
+    models_for_export["transformer"] = (transformer, transformer_export_config)
+
+    # VAE Decoder
+    vae_decoder = copy.deepcopy(pipeline.vae)
```
IlyasMoutawwakil, line 1234: a deepcopy doubles the memory here

optimum/exporters/openvino/convert.py
```diff
@@ -1173,6 +1185,152 @@ def get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+    if hasattr(pipeline.vae, "latents_mean") and pipeline.vae.latents_mean is not None:
+        vae_decoder.register_to_config(latents_mean_data=pipeline.vae.latents_mean.tolist())
+    if hasattr(pipeline.vae, "latents_std") and pipeline.vae.latents_std is not None:
+        vae_decoder.register_to_config(latents_std_data=pipeline.vae.latents_std.tolist())
+    if hasattr(pipeline, "audio_vae") and pipeline.audio_vae is not None:
+        if hasattr(pipeline.audio_vae, "latents_mean") and pipeline.audio_vae.latents_mean is not None:
+            vae_decoder.register_to_config(audio_latents_mean_data=pipeline.audio_vae.latents_mean.tolist())
+        if hasattr(pipeline.audio_vae, "latents_std") and pipeline.audio_vae.latents_std is not None:
+            vae_decoder.register_to_config(audio_latents_std_data=pipeline.audio_vae.latents_std.tolist())
```
IlyasMoutawwakil, line 1243: why not like LTX above ?

optimum/exporters/openvino/convert.py
```diff
@@ -1173,6 +1185,152 @@ def get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+        model_type="ltx2-vae-decoder",
+    )
+    vae_decoder_export_config = vae_decoder_config_constructor(
+        vae_decoder.config, int_dtype=int_dtype, float_dtype=float_dtype
+    )
+    vae_decoder_export_config.runtime_options = {"ACTIVATIONS_SCALE_FACTOR": "8.0"}
+    models_for_export["vae_decoder"] = (vae_decoder, vae_decoder_export_config)
+
+    # Audio VAE decoder and vocoder for text-to-audio-video task
+    if task == "text-to-audio-video" and hasattr(pipeline, "audio_vae") and pipeline.audio_vae is not None:
```
IlyasMoutawwakil: you don't really need the task, the pipeline contain all the necessary information normally (ie if it has the audio vae na dits not none)

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1707,6 +1826,534 @@ class OVLTXPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, LTXPipel
...
+        tokenizer: Optional[CLIPTokenizer] = None,
+        device: str = "CPU",
+        compile: bool = True,
+        compile_only: bool = False,
+        dynamic_shapes: bool = True,
+        ov_config: Optional[Dict[str, str]] = None,
+        model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
+        quantization_config: Optional[Union[OVWeightQuantizationConfig, Dict]] = None,
+        **kwargs,
+    ):
```
IlyasMoutawwakil, line 1924: hmm i'm not sure if things changed since last time but this not how the diffusion pipeline class is supposed to be used (ie rewriting most of it), we try to keep a big chunk of the loading logic in the OVDiffusionPipeline which is reusable across many classes

IlyasMoutawwakil (changes requested):
overall looks good, seem a bit in isolation of previous work, we do have a lot of utils and integration around very similar pipelines (LTX v1 for example) let's try to reuse what we have as much as possible

IlyasMoutawwakil:
please also check copilot review and address and resolve its comments when addressed

optimum/exporters/openvino/convert.py
anatyrova: in get_ltx2_video_models_for_export the task parameter controls whether vocoder and audio_vae get exported

we have two tasks: t2v, t2audio-v

optimum/exporters/openvino/convert.py
IlyasMoutawwakil: as in through the transformers library

optimum/exporters/openvino/convert.py
anatyrova: actually , i used the file's pattern, and in other functions you can also see the same situation, eg. _save_kokoro_config_and_assets

optimum/exporters/openvino/convert.py
IlyasMoutawwakil: but the thing is that this diverges from what we do usually, normally we export everything that's in the pipeline, for example for image to image, some text encoder could be skipped but we don't skip it, when asked to export a diffusion pipeline, we export all its avaialable components to later support it being loaded with the task of choice.

optimum/exporters/openvino/convert.py
```diff
@@ -1236,6 +1240,129 @@ def get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+    def text_encoder_forward(input_ids, attention_mask=None, **kwargs):
+        if attention_mask is not None and attention_mask.dim() == 2:
+            import torch
+
+            bsz, seq_len = attention_mask.shape
+            causal_mask = attention_mask[:, None, None, :].to(dtype=torch.float32)
+            causal_mask = causal_mask.expand(bsz, 1, seq_len, seq_len).clone()
+            causal_positions = torch.tril(torch.ones(seq_len, seq_len, dtype=torch.float32))
+            causal_mask = causal_mask * causal_positions[None, None, :, :]
+            causal_mask = (1.0 - causal_mask) * torch.finfo(torch.float32).min
```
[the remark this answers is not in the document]
anatyrova: resolved in the next commit

optimum/exporters/openvino/convert.py
anatyrova: we register gemma3-text-encoder as a diffusers feature-extraction config specifically for LTX2's pipeline. for standalone gemma3 from transformers (e.g. text-generation), it's only registered for image-text-to-text right now so exporting a text-only gemma3 via the CLI isn't supported unfortunately

optimum/exporters/openvino/convert.py
```diff
@@ -1236,6 +1240,129 @@ def get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+            attention_mask = {"full_attention": causal_mask, "sliding_attention": causal_mask}
+        outputs = orig_forward(input_ids=input_ids, attention_mask=attention_mask, output_hidden_states=True)
+        result = {"last_hidden_state": outputs.hidden_states[-1]}
+        for i, hs in enumerate(outputs.hidden_states):
+            result[f"hidden_states.{i}"] = hs
+        return result
+
+    text_encoder.forward = text_encoder_forward
```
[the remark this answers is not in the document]
anatyrova: would be valid if this were a library function called on a user's own pipeline object that they keep using after but here the pipeline is loaded internally just for export

optimum/exporters/openvino/convert.py
anatyrova: the transformers gemma3 path solves a different problem (LLM/VLM export). this is a diffusers-embedded encoder that needs all hidden states unlike existing g3, so a different export shape

optimum/exporters/openvino/convert.py
anatyrova: hmm i used the same logic as other diffusion models (sd3, sana, ltx_video)

optimum/exporters/openvino/convert.py
anatyrova: yes, it differs from ltx above, because ltx's fields are always present, so the compact dict form is safe. ltx2's fields are optional (may be missing or none) and conditionally include audio VAE stats, so it needs hasattr guards instead

optimum/exporters/openvino/convert.py
anatyrova: w/o task checking audio VAE also exported, even though the user asked for text-to-video, so i would not remove it 

edited: i removed it in the next commit

optimum/exporters/openvino/model_patcher.py
```diff
@@ -9937,3 +9937,420 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_value, traceback):
         super().__exit__(exc_type, exc_value, traceback)
         self._model.forward = self._model._orig_forward
+
+
+def _ltx2_connector_forward_patched(self, hidden_states, attention_mask=None, attn_mask_binarize_threshold=-9000.0):
```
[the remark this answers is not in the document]
anatyrova: solved in next commit

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1308,6 +1315,106 @@ def forward(
...
+        if num_frames is not None:
+            model_inputs["num_frames"] = num_frames
+        if height is not None:
+            model_inputs["height"] = height
+        if width is not None:
+            model_inputs["width"] = width
+        if fps is not None:
+            if not isinstance(fps, torch.Tensor):
+                fps = torch.tensor(float(fps))
+            model_inputs["fps"] = fps
```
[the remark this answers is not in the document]
anatyrova: this is an ov inference request, not a pt forward pass, device placement is managed by the ov runtime, not by tensor .device. the torch.tensor(float(fps)) is immediately passed to the ov request which converts it to the appropriate format regardless

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1308,6 +1315,106 @@ def forward(
...
+            "attention_mask": attention_mask,
+        }
+
+        ov_outputs = self.request(model_inputs, share_inputs=True).to_dict()
+
+        results = []
+        for key, value in ov_outputs.items():
+            results.append(torch.from_numpy(value))
+
+        return tuple(results)
```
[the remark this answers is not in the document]
anatyrova: resolved in the next commit

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1663,6 +1780,332 @@ class OVLTXPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, LTXPipel
...
+        class _Cfg:
+            sample_rate = 16000
+            mel_hop_length = 160
+            mel_bins = 64
+            latent_channels = 8
+
+        self.config = _Cfg()
+
+    def decode(self, latents, return_dict=False):
+        return (torch.zeros(1),)
```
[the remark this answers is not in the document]
anatyrova: the _DummyAudioVae.decode() is never called during video-only inference, the diffusers pipeline only invokes audio_vae.decode() after producing audio_latents from the transformer which doesn't happen in the video-only path. the dummy is only accessed for latents_mean/latents_std attributes, so the fix would be defensive coding but isn't addressing a real failure path

optimum/intel/openvino/modeling_diffusion.py
anatyrova: i don't think we can reuse it, because ltx2 needs LTX2AudioVideoRotaryPosEmbed attached to the transformer object before LTX2Pipeline.__init__ runs and audio_vae , vocoder need wrapping

optimum/exporters/openvino/convert.py
anatyrova: i did it for saving some disk memory. but you right that consistency matters. will change to export all components regardless the given task.

optimum/exporters/openvino/convert.py
```diff
@@ -1173,6 +1185,154 @@ def get_ltx_video_models_for_export(pipeline, exporter, int_dtype, float_dtype):
...
+                torch.ones(seq_len, seq_len, dtype=torch.float32, device=attention_mask.device)
+            )
+            causal_mask = causal_mask * causal_positions[None, None, :, :]
+            causal_mask = (1.0 - causal_mask) * torch.finfo(torch.float32).min
+            attention_mask = {"full_attention": causal_mask, "sliding_attention": causal_mask}
+        outputs = orig_forward(input_ids=input_ids, attention_mask=attention_mask, output_hidden_states=True)
+        result = {"last_hidden_state": outputs.hidden_states[-1]}
+        for i, hs in enumerate(outputs.hidden_states):
+            result[f"hidden_states.{i}"] = hs
+        return result
```
IlyasMoutawwakil: if the patch is very big we can always put it in model_patcher.py with a proper patcher, separation of concerns. we only did vae encoder/decoder patching originally in these decomposition functions.

optimum/exporters/openvino/input_generators.py
```diff
@@ -864,6 +864,184 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+    ):
+        super().__init__(task, normalized_config, batch_size, num_channels, width, height, **kwargs)
+        self.num_frames = num_frames
+        self.frame_rate = frame_rate
+        self.vae_scale_factors = normalized_config.config.vae_scale_factors
+        self.audio_in_channels = normalized_config.config.audio_in_channels
+        self.audio_scale_factor = normalized_config.config.audio_scale_factor
+        self.cross_attention_dim = normalized_config.config.cross_attention_dim
+        self.caption_channels = normalized_config.config.caption_channels
+        self.encoder_seq_length = 128
```
IlyasMoutawwakil: why not the seq passed to the generator ?

optimum/exporters/openvino/input_generators.py
```diff
@@ -864,6 +864,184 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+
+class LTX2ConnectorsDummyInputGenerator(DummyVisionInputGenerator):
+    SUPPORTED_INPUT_NAMES = ("text_encoder_hidden_states", "attention_mask")
+
+    def __init__(
+        self,
+        task: str,
+        normalized_config: NormalizedVisionConfig,
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        sequence_length: int = 128,
```
IlyasMoutawwakil: why not the default dummy shape ?

optimum/exporters/openvino/convert.py
anatyrova: you right, i will create a patcher for that.

optimum/exporters/openvino/input_generators.py
anatyrova: i will change it to self.encoder_seq_length = kwargs.get("sequence_length", DEFAULT_DUMMY_SHAPES["sequence_length"])

thank you

anatyrova:
@IlyasMoutawwakil can you review pls?

optimum/exporters/openvino/convert.py
```diff
@@ -1062,7 +1066,11 @@ def _get_submodels_and_export_configs(
 
 
 def get_diffusion_models_for_export_ext(
-    pipeline: "DiffusionPipeline", int_dtype: str = "int64", float_dtype: str = "fp32", exporter: str = "openvino"
+    pipeline: "DiffusionPipeline",
+    int_dtype: str = "int64",
+    float_dtype: str = "fp32",
+    exporter: str = "openvino",
+    task: Optional[str] = None,
```
IlyasMoutawwakil: task is unused now

optimum/exporters/openvino/input_generators.py
```diff
@@ -864,6 +864,184 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+    SUPPORTED_INPUT_NAMES = ("pixel_values", "pixel_mask", "sample", "latent_sample", "timestep")
+
+    def __init__(
+        self,
+        task: str,
+        normalized_config: NormalizedVisionConfig,
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        num_channels: int = DEFAULT_DUMMY_SHAPES["num_channels"],
+        width: int = DEFAULT_DUMMY_SHAPES["width"],
+        height: int = DEFAULT_DUMMY_SHAPES["height"],
```
IlyasMoutawwakil: do we not wanna use smaller dims for the vae export as well ?

optimum/exporters/openvino/input_generators.py
```diff
@@ -864,6 +864,184 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+    def __init__(
+        self,
+        task: str,
+        normalized_config: NormalizedVisionConfig,
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        num_channels: int = DEFAULT_DUMMY_SHAPES["num_channels"],
+        width: int = 16,
+        height: int = 8,
+        num_frames: int = 2,
+        frame_rate: int = 24,
```
IlyasMoutawwakil, line 902: these need explicit comments indicating why we use them instead of defaults (faster export or other)

optimum/exporters/openvino/input_generators.py
```diff
@@ -864,6 +864,184 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+        self.sequence_length = (self.sequence_length // num_registers) * num_registers
+        self.caption_channels = normalized_config.config.caption_channels
+        text_proj_in_factor = getattr(normalized_config.config, "text_proj_in_factor", 1)
+        self.input_channels = self.caption_channels * text_proj_in_factor
+
+    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+        if input_name == "text_encoder_hidden_states":
+            return self.random_float_tensor([self.batch_size, self.sequence_length, self.input_channels])
+        if input_name == "attention_mask":
+            import torch
```
IlyasMoutawwakil, line 974: import at the top please

optimum/exporters/openvino/model_patcher.py
```diff
@@ -10512,3 +10512,454 @@ def __enter__(self):
...
+    audio_text_embedding, _ = self.audio_connector(audio_text_emb_proj, add_attn_mask)
+
+    return video_text_embedding, audio_text_embedding, binary_attn_mask.squeeze(-1)
+
+
+class _LTX2AttnProcessorWithEps:
+    """
+    Attention processor that replaces SDPA with manual attention + epsilon
+    to work around OpenVINO CPU plugin numerical issues with zero attention weights.
+    """
```
IlyasMoutawwakil, line 10068: no better solution than this ? this is a big chunk of code that will need a version after version update to keep things working

optimum/exporters/openvino/model_patcher.py
```diff
@@ -10512,3 +10512,454 @@ def __enter__(self):
...
+
+class _LTX2TraceSafeAttnProcessor:
+    """
+    Attention processor with trace-safe mask handling and RoPE.
+    Replaces prepare_attention_mask (which has data-dependent branches) and
+    apply_split_rotary_emb (which uses in-place addcmul_ on views) with
+    trace-safe equivalents.
+    """
+
+    def __call__(
```
IlyasMoutawwakil, line 10236: same comment, it looks like a "fix" now but this becomes a liability in one or two diffusers/transformers versions

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1707,6 +1826,534 @@ class OVLTXPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, LTXPipel
...
+    def __init__(self, latents_mean=None, latents_std=None):
+        self.dtype = torch.float32
+        self.mel_compression_ratio = 4
+        self.temporal_compression_ratio = 4
+
+        class _Cfg:
+            sample_rate = 16000
+            mel_hop_length = 160
+            mel_bins = 64
+            latent_channels = 8
```
IlyasMoutawwakil: is it normal that all of this is hardcoded ? is it not in the diffusers config ?

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1707,6 +1826,534 @@ class OVLTXPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, LTXPipel
...
+
+class OVModelAudioVaeDecoder(OVPipelinePart):
+    """OV wrapper for LTX2 audio VAE decoder (mel spectrogram decoder)."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.mel_compression_ratio = getattr(self.config, "mel_compression_ratio", 4) if hasattr(self, "config") else 4
+        self.temporal_compression_ratio = (
+            getattr(self.config, "temporal_compression_ratio", 4) if hasattr(self, "config") else 4
+        )
```
IlyasMoutawwakil: if it's supposed to be in the config and used during inference it could make sense to fail loudly than to use the wrong value

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1707,6 +1826,534 @@ class OVLTXPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, LTXPipel
...
+        self.transformer = (
+            OVModelTransformerLTX2(transformer, self, DIFFUSION_MODEL_TRANSFORMER_SUBFOLDER)
+            if transformer is not None
+            else None
+        )
+        if self.transformer is None:
+            raise ValueError("`transformer` model should be provided for pipeline work")
+
+        # Create rope objects for coordinate preparation (used by pipeline, not by OV model)
+        from diffusers.models.transformers.transformer_ltx2 import LTX2AudioVideoRotaryPosEmbed
```
IlyasMoutawwakil, line 1945: wdym used by pipeline but not the ov model ? should we export it as well as a component ?

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1707,6 +1826,534 @@ class OVLTXPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, LTXPipel
...
+
+        if self.is_dynamic and not self._compile_only:
+            self.reshape(batch_size=-1, height=-1, width=-1, num_images_per_prompt=-1)
+
+        if compile and not self._compile_only:
+            self.compile()
+
+    @property
+    def _execution_device(self):
+        return torch.device("cpu")
```
IlyasMoutawwakil, line 2060: why isnt this in the parent OVDiffusionPipleine ?

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1707,6 +1826,534 @@ class OVLTXPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, LTXPipel
...
+        """Return list of OVPipelinePart components that are loaded."""
+        return [c for c in self.components.values() if c is not None]
+
+    def clear_requests(self):
+        for component in self._ov_components:
+            component.clear_requests()
+
+    def compile(self):
+        for component in self._ov_components:
+            component.compile()
```
IlyasMoutawwakil: same question

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1707,6 +1826,534 @@ class OVLTXPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, LTXPipel
...
+            "transformer": self.transformer,
+            "vocoder": ov_vocoder,
+        }
+        LTX2Pipeline.__init__(self, **diffusers_pipeline_args)
+
+        # Set attributes that OVDiffusionPipeline.reshape() checks but LTX2 doesn't have
+        self.unet = None
+        self.text_encoder_2 = None
+        self.text_encoder_3 = None
+        self.tokenizer_2 = None
```
IlyasMoutawwakil: why define them if .reshape() is overridden just under it

optimum/exporters/openvino/model_patcher.py
anatyrova: this code patches the sdpa problem in openvino cpu plugin. the same approach you can see https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_patcher.py#L6124 in speecht5_attention_forward 

w/o it, the video differs from diffusers original, and the audio and video don't match up(tested in ov nightly)

i just hope in future it won't be needed, and i'll remove it, but for now it is a must

optimum/exporters/openvino/model_patcher.py
anatyrova: yes, i understand.  but it's necessary to create a right graph, since it replaces prepare_attention_mask (which has if/else branches on tensor values that torch.jit.trace puts into a single path) and apply_split_rotary_emb (which uses in-place addcmul_ on views that breaks the tracer) with equivalent ops.  you can see same approach in Qwen2VL, Gemma 3.

maybe this would not be needed if optimum-intel used torch.export instead of torch.jit.trace

optimum/intel/openvino/modeling_diffusion.py
anatyrova: it means that rope are computed on-the-fly in python by the pipeline before being passed as inputs to the transformer

optimum/intel/openvino/modeling_diffusion.py
anatyrova: i cannot remove them,cause the parent's save_pretrained (and potentially other inherited methods) access self.tokenizer_2 etc. since OVLTX2Pipeline doesn't override save_pretrained, those attributes need to exist

optimum/intel/openvino/modeling_diffusion.py
anatyrova: good point, it It could be moved to the parent, but that's a broader refactor beyond this PR's scope? i don't think i could do this rn

optimum/intel/openvino/modeling_diffusion.py
anatyrova: again, could be generalized in the parent if all pipelines used _ov_components, but that's a broader refactor

optimum/intel/openvino/modeling_diffusion.py
anatyrova: these are only in the dummy stub (compatibility shim for the generic OVDiffusionPipeline dispatch path). they're never used for real inference, just to satisfy parent class attribute checks. the real OVModelAudioVaeDecoder reads from the model config

optimum/intel/openvino/modeling_diffusion.py
anatyrova: these values aren't saved in the model config, they come from a hardcoded LATENT_DOWNSAMPLE_FACTOR =4 constant in diffusers, so the default of 4 is always correct and failing loudly would just always fail

optimum/intel/openvino/modeling_diffusion.py
IlyasMoutawwakil: okay but the comment is misleading, i'm speaking based on the comment

anatyrova: okay, i should change it

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: > i just hope in future it won't be needed

we've been hoping for more than a year @rkazants

IlyasMoutawwakil: i opened many sdpa issues in the openvino repo asking for it to match torch sdpa behaviour

IlyasMoutawwakil: nevermind the rant please, let's just make sure report the openvino issues in openvino and link them here instead of simply and continuously patching them

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: yeh, more reason to switch to hf exporters i guess

IlyasMoutawwakil (approved):
LGTM ! still stand by my preference for generality, but no worries if it looks like too much work.

IlyasMoutawwakil:
please make sure all tests are passing before merging.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -9921,6 +9921,457 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            gates = 2.0 * torch.sigmoid(gate_logits)
+            hidden_states = hidden_states * gates.unsqueeze(-1)
+            hidden_states = hidden_states.flatten(2, 3)
+
+        hidden_states = attn.to_out[0](hidden_states)
+        hidden_states = attn.to_out[1](hidden_states)
+        return hidden_states
+
+
+class LTX2TextEncoderPatcher(ModelPatcher):
```
rkazants, line 10309: please leave a comment. Why this patching is needed, and what exactly is patched with a reference to original code base and code-line.
Same comment for other patchers.

Double-check that all of these patchings are needed.

tests/openvino/test_exporters_cli.py
```diff
@@ -143,6 +144,106 @@ class OVCLIExportTestCase(unittest.TestCase):
...
+            [
+                ("text-to-video", "ltx2"),
+            ]
+        )
+
+    if is_transformers_version(">=", "4.48.0"):
+        SUPPORTED_ARCHITECTURES.extend(
+            [
+                ("text-generation-with-past", "cohere2"),
+            ]
```
rkazants: why all of these ltx2-unrelated code changes. Please double check

tests/openvino/test_export.py
```diff
@@ -129,13 +130,62 @@ class ExportModelTest(unittest.TestCase):
         in get_supported_model_for_library("transformers") | get_supported_model_for_library("diffusers")
     }
 
+    if is_diffusers_version(">=", "0.38.0"):
+        SUPPORTED_ARCHITECTURES.update({"ltx2": OVLTX2Pipeline})
+
+    if is_diffusers_version(">=", "0.37.0"):
+        SUPPORTED_ARCHITECTURES.update({"flux.2-klein": OVFlux2KleinPipeline})
```
rkazants: why it is changed here?

rkazants (approved):

PR #1735, opened 2026-05-13
Improve performance with acceptable accuracy for Qwen/Qwen3.6-35B-A3B

andreyanufr:
# What does this PR do?

Change default compression config for Qwen/Qwen3.6-35B-A3B.

wwb --target-model Qwen3.6-35B-A3B-Int4/ --gt-data models/gt_qwen_3_6_35B.csv --model-type visual-text --ov-config 
similarity 0.876817

### DIFF
--- a/optimum/intel/openvino/configuration.py
+++ b/optimum/intel/openvino/configuration.py
@@ -486,7 +486,7 @@ _DEFAULT_4BIT_WQ_CONFIGS = {
             "lm_model": {
                 "bits": 4,
                 "sym": False,
-                "group_size": 128,
+                "group_size": 64,
             },
             "text_embeddings_model": {"bits": 8, "sym": True, "weight_only": True},
             "vision_embeddings_merger_model": {"bits": 8, "sym": True, "weight_only": True},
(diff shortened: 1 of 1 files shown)


MaximProshin:
@andrey-churkin , @rkazants , please help to merge it asap. It reduces the accuracy but improves GPU performance 2x according to OV GPU team.

SearchSavior:
Hello @MaximProshin,

In a general case, how do we select the optimal 'group_size'? Maybe there is something about a given architecture to look out for- I know in your past PRs maybe the result comes from studying changes to wwb across export variations?

MaximProshin:
> Hello @MaximProshin,
> 
> In a general case, how do we select the optimal 'group_size'? Maybe there is something about a given architecture to look out for- I know in your past PRs maybe the result comes from studying changes to wwb across export variations?

Default is 128. A smaller group size is better for the accuracy but slightly impacts performance (though not that much as the ratio or the ignored scope), so we usually start from the default and if the accuracy is not ok, the group size value is one that we play with to improve the accuracy.

rkazants (approved):

PR #1741, opened 2026-05-15
Fix MoE models patching to enable ConvertTiledMoeBlockToGatherMatmuls transformation.

popovaan:
# What does this PR do?

Fixes CVS-186759

## Before submitting
- Did you write any new necessary tests? - Will be covered in https://github.com/huggingface/optimum-intel/pull/1651

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -7906,7 +7906,7 @@ class GptOssModelPatcher(OVDecoderModelPatcher):
     def __exit__(self, exc_type, exc_value, traceback):
         super().__exit__(exc_type, exc_value, traceback)
 
-        if is_transformers_version(">=", "4.55.0") and is_transformers_version("<", "5"):
+        if is_transformers_version(">=", "4.55.0"):
             from transformers.models.gpt_oss.modeling_gpt_oss import GptOssExperts
 
             GptOssExperts.forward = self.original_gpt_oss_forward
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -9089,12 +9089,12 @@ def lfm2_moe_experts_forward(
         num_experts, -1, hidden_dim
     )  # (num_experts, num_tokens, hidden_dim)
 
-    gate_proj, up_proj = self.gate_up_proj.transpose(1, 2).chunk(2, dim=-2)
+    gate_proj, up_proj = self.gate_up_proj.chunk(2, dim=-2)
 
     gate = torch.bmm(hidden_states_expanded, gate_proj.transpose(1, 2))
     up = torch.bmm(hidden_states_expanded, up_proj.transpose(1, 2))
-    next_states = torch.sigmoid(gate) * up
-    next_states = torch.bmm(next_states, self.down_proj.transpose(1, 2).transpose(1, 2))
+    next_states = self.act_fn(gate) * up
+    next_states = torch.bmm(next_states, self.down_proj.transpose(1, 2)
 
     next_states = next_states.view(num_experts, num_tokens, hidden_dim)
     next_states = next_states * dense_routing_weights.transpose(0, 1).view(num_experts, num_tokens)[..., None]
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -9094,7 +9094,7 @@ def lfm2_moe_experts_forward(
     gate = torch.bmm(hidden_states_expanded, gate_proj.transpose(1, 2))
     up = torch.bmm(hidden_states_expanded, up_proj.transpose(1, 2))
     next_states = self.act_fn(gate) * up
-    next_states = torch.bmm(next_states, self.down_proj.transpose(1, 2)
+    next_states = torch.bmm(next_states, self.down_proj.transpose(1, 2))
 
     next_states = next_states.view(num_experts, num_tokens, hidden_dim)
     next_states = next_states * dense_routing_weights.transpose(0, 1).view(num_experts, num_tokens)[..., None]
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -161,8 +161,8 @@ def batched_mm_experts_forward_patched(
     num_experts = self.gate_up_proj.size(0)
 
     if not self.is_transposed:
-        gate_up_proj = self.gate_up_proj.transpose(1,2)
-        down_proj = self.down_proj.transpose(1,2)
+        gate_up_proj = self.gate_up_proj.transpose(1, 2)
+        down_proj = self.down_proj.transpose(1, 2)
     else:
         gate_up_proj = self.gate_up_proj
         down_proj = self.down_proj
@@ -210,6 +210,7 @@ def batched_mm_experts_forward_patched(
 
 def patch_batched_mm(patcher):
     from transformers.integrations.moe import ALL_EXPERTS_FUNCTIONS
+
     patcher.original_gpt_oss_forward = ALL_EXPERTS_FUNCTIONS._global_mapping["batched_mm"]
     ALL_EXPERTS_FUNCTIONS._global_mapping["batched_mm"] = batched_mm_experts_forward_patched
     patcher._model.set_experts_implementation("batched_mm")
@@ -217,6 +218,7 @@ def patch_batched_mm(patcher):
 
 def set_original_batched_mm(patcher):
     from transformers.integrations.moe import ALL_EXPERTS_FUNCTIONS
+
     ALL_EXPERTS_FUNCTIONS._global_mapping["batched_mm"] = patcher.original_gpt_oss_forward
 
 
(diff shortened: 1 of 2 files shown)


popovaan:
@rkazants @echarlaix @IlyasMoutawwakil please review.

IlyasMoutawwakil:
hi can you please explain the reason for the rewrite ? i assume from the title that its a specific pattern for openvino op fusion ?
btw you dont have to patch/overwrite the batched_mm iml, we can simply register a new impl for ov export purposes and use it by default (with set_experts_impl)

rkazants (approved):

rkazants:
Hi @echarlaix, @IlyasMoutawwakil,

Can you please review it? This is quite urgent fix.

Thanks,
Roman

popovaan:
> hi can you please explain the reason for the rewrite ? i assume from the title that its a specific pattern for openvino op fusion ?

Hi! Yes, this is exactly the reason for the rewrite. OpenVINO MoE optimizations expect a specific MoE pattern which is fused in a optimized operation, otherwise we have a quite large performance regression.

> btw you dont have to patch/overwrite the batched_mm iml, we can simply register a new impl for ov export purposes and use it by default (with set_experts_impl)

Sure, I will use this approach.

popovaan:
@IlyasMoutawwakil 
When I try to add an ov specific MoE implementation to a `ALL_EXPERTS_FUNCTIONS._global_mapping` and then set `patcher._model.set_experts_implementation("ov_batched_mm")` I get the following error:

```
  File "/home/panas/venv/new_py3_12/lib/python3.12/site-packages/transformers/modeling_utils.py", line 1937, in get_correct_experts_implementation
    raise ValueError(message)
ValueError: Specified `experts_implementation="ov_batched_mm"` is not supported. The only possible arguments are `experts_implementation="eager"`, `"experts_implementation=grouped_mm"` and `"experts_implementation=batched_mm"`.
```
What is the correct way to register the new MoE impl to avoid this error?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -150,6 +150,78 @@ def sdpa_mask_without_vmap(batch_size, q_length=None, kv_length=None, q_offset=0
         UNSUPPORTED_OPS_PATCHING_SPEC.pop(idx)
 
 
+def batched_mm_experts_forward_patched(
```
echarlaix, line 160: @popovaan would you mind adding a link to the original code + add comments to explain / highlight the differences

optimum/exporters/openvino/model_patcher.py
```diff
@@ -150,6 +150,78 @@ def sdpa_mask_without_vmap(batch_size, q_length=None, kv_length=None, q_offset=0
...
+
+    patcher.original_gpt_oss_forward = ALL_EXPERTS_FUNCTIONS._global_mapping["batched_mm"]
+    ALL_EXPERTS_FUNCTIONS._global_mapping["batched_mm"] = batched_mm_experts_forward_patched
+    patcher._model.set_experts_implementation("batched_mm")
+
+
+def set_original_batched_mm(patcher):
+    from transformers.integrations.moe import ALL_EXPERTS_FUNCTIONS
+
+    ALL_EXPERTS_FUNCTIONS._global_mapping["batched_mm"] = patcher.original_gpt_oss_forward
```
echarlaix: why not

```
ALL_EXPERTS_FUNCTIONS.register("ov_batched_mm", batched_mm_experts_forward_patched)
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -437,14 +509,21 @@ def __enter__(self):
                     _mixtral_sparse_moe_block_forward, layer.block_sparse_moe
                 )
         else:
-            self._model.set_experts_implementation("batched_mm")
+            from transformers.models.mixtral.modeling_mixtral import MixtralExperts
+
+            self.original_moe_forward = MixtralExperts.forward
+            MixtralExperts.forward = lfm2_moe_experts_forward
```
echarlaix, line 544:
```suggestion
            self._model.set_experts_implementation("ov_batched_mm")
```

echarlaix (commented):
Thanks for the addition @popovaan

optimum/exporters/openvino/model_patcher.py
popovaan: When I set it like this:
```
ALL_EXPERTS_FUNCTIONS.register("ov_batched_mm", batched_mm_experts_forward_patched)
patcher._model.set_experts_implementation("ov_batched_mm")
```

It results in error:
```
ValueError: Specified `experts_implementation="ov_batched_mm"` is not supported. The only possible arguments are `experts_implementation="eager"`, `"experts_implementation=grouped_mm"` and `"experts_implementation=batched_mm"`.
```

optimum/exporters/openvino/model_patcher.py
popovaan: Done.

optimum/exporters/openvino/model_patcher.py
popovaan: Setting custom MoE implementation `patcher._model.set_experts_implementation("ov_batched_mm")` doesn't work, as I mentioned in the previous comment it fails with this error:
```
ValueError: Specified `experts_implementation="ov_batched_mm"` is not supported. The only possible arguments are `experts_implementation="eager"`, `"experts_implementation=grouped_mm"` and `"experts_implementation=batched_mm"`.
```

Looks like it's due to this check that expects specifically one of `["eager", "grouped_mm", "batched_mm"]`: https://github.com/huggingface/transformers/blob/08810b1e278938278c50153ee1edfd7a20a759da/src/transformers/modeling_utils.py#L1932

popovaan:
@regisss could you please review?

optimum/exporters/openvino/model_patcher.py
echarlaix: @popovaan could it come from https://github.com/huggingface/transformers/blob/v5.0.0/src/transformers/modeling_utils.py#L1932 where possible experts implementation are set to  ["eager", "grouped_mm", "batched_mm"] when we could check `ALL_EXPERTS_FUNCTIONS` instead ? cc @IlyasMoutawwakil

popovaan: Yes, the error comes from this line.

echarlaix: what do you think of patching `get_correct_experts_implementation` ?

popovaan: Yes, that worked. Done.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -152,6 +152,108 @@ def sdpa_mask_without_vmap(batch_size, q_length=None, kv_length=None, q_offset=0
...
+                raise e
+            applicable_experts = "eager"
+
+    return applicable_experts
+
+
+def register_ov_batched_mm(patcher):
+    from transformers.integrations.moe import ALL_EXPERTS_FUNCTIONS
+
+    patcher.get_correct_experts_implementation_orig = patcher._model.get_correct_experts_implementation
```
echarlaix: looks like it should be fixed in 5.7 https://github.com/huggingface/transformers/blob/v5.7.0/src/transformers/modeling_utils.py#L1994 (so we can only patch for lower version)

optimum/exporters/openvino/model_patcher.py
```diff
@@ -152,6 +152,108 @@ def sdpa_mask_without_vmap(batch_size, q_length=None, kv_length=None, q_offset=0
...
+    patcher._model.get_correct_experts_implementation = types.MethodType(
+        get_correct_experts_implementation_patched, patcher._model
+    )
+
+    ALL_EXPERTS_FUNCTIONS.register("ov_batched_mm", batched_mm_experts_forward_patched)
+    patcher._model.set_experts_implementation("ov_batched_mm")
+
+
+def set_original_get_correct_experts_implementation(patcher):
+    patcher.get_correct_experts_implementation = patcher.get_correct_experts_implementation_orig
```
echarlaix: shouldn't it be 
```suggestion
    patcher._model.get_correct_experts_implementation = patcher.get_correct_experts_implementation_orig
```
but shouldn't we need to either call set_experts_implementation to original or remove unpatching ? (otherwise will fail no?)

optimum/exporters/openvino/model_patcher.py
```diff
@@ -1759,7 +1868,10 @@ def __enter__(self):
                     _phi_moe_sparse_moe_block_forward, layer.block_sparse_moe
                 )
         else:
-            self._model.set_experts_implementation("batched_mm")
+            from transformers.models.phimoe.modeling_phimoe import PhimoeExperts
+
+            self.original_moe_forward = PhimoeExperts.forward
+            PhimoeExperts.forward = lfm2_moe_experts_forward
```
echarlaix, line 1871: shouldn't we do the same here ? (use ov_batched_mm or similar)

echarlaix: can be done in a following PR

echarlaix (approved):
Thanks a lot for iterating @popovaan

optimum/exporters/openvino/model_patcher.py
popovaan: Yes, indeed it is not needed. I will remove unpatching.

optimum/exporters/openvino/model_patcher.py
echarlaix: thanks a lot

optimum/exporters/openvino/model_patcher.py
popovaan: Yes, added check to patch it only for transformers lower than 5.7.

optimum/exporters/openvino/model_patcher.py
IlyasMoutawwakil: hi yes i hardcoded the list of supported impls long ago and it only got fixed / became dynamically checked in last transformers version 🥲

IlyasMoutawwakil: was fixed in https://github.com/huggingface/transformers/pull/45577

PR #1589, opened 2026-01-19
Transformers v5

echarlaix:
with https://github.com/huggingface/optimum/pull/2408/

List of architectures for which are deprecating support for transformers >= v5
(remote code modeling incompatible with v5)
- codegen2
- exaone
- decilm
- internlm2
- orion
- aquila2
- jais
- baichuan2
- internvl_chat
- minicpmv
(remote code models)
- aquila
- xverse
- internlm

List of architectures for which we removed support for transformers >= v5 for which support can be added later on
- xlm
- data2vec-text
- flaubert
- marian
- ~granitemoe~
- zamba2
- ~phimoe~ 
- drbx
- ~eagle3~
- mt5
- ~llava~
- idefics3
- ~gemma3~
- smolvlm
- got_ocr2
- llava_next_video
- llama4
- bitnet
- qwen3_next
- exaone4

PR #1744, opened 2026-05-21
[OpenVINO] Fix long context inference for Phi3.5 and Phi4

rkazants:
# What does this PR do?

Fix for the export was already done in this PR by providing the dedicated patch in the common patcher class: https://github.com/huggingface/optimum-intel/pull/1589

Fixes 170599

## Before submitting


tests/openvino/test_decoder.py
```diff
@@ -88,6 +88,7 @@ class OVModelForCausalLMIntegrationTest(unittest.TestCase):
 
     if is_transformers_version(">=", "4.49") and is_transformers_version("<", "5"):
         SUPPORTED_SSM_ARCHITECTURES += ("zamba2",)
+        SUPPORTED_ARCHITECTURES += ("phi3", "phi3-longrope")
```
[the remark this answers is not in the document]
echarlaix:
```suggestion
        SUPPORTED_ARCHITECTURES += ("phi3-longrope",)
```

tests/openvino/test_decoder.py
```diff
@@ -1029,3 +1031,54 @@ def test_hybrid_model_multi_step_generation(self, model_arch):
         del transformers_model
         del ov_model
         gc.collect()
+
+    def test_phi3_longrope_support(self):
```
echarlaix, line 1040: test_phi3_longrope_support failing with 
```
FAILED tests/openvino/test_decoder.py::OVModelForCausalLMIntegrationTest::test_phi3_longrope_support - AssertionError: False is not true : OpenVINO and PyTorch outputs do not match for LongRoPE test with inputs > original_max_pos.
```

echarlaix (commented):
Thanks for the addition!

optimum/exporters/openvino/model_patcher.py
```diff
@@ -395,10 +296,21 @@ def eager_mask_without_vmap(*args, **kwargs) -> Optional[torch.Tensor]:
 # Adapted from https://github.com/huggingface/transformers/blob/3c307e380ad07ca16903a39e09a47d532cb782d9/src/transformers/models/phimoe/modular_phimoe.py#L57
 def _longrope_forward(self, x, position_ids=None, layer_type=None):
```
echarlaix: an additional fix is needed for phimoe for transformers < v5 
https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/phimoe/modeling_phimoe.py#L161

echarlaix: added in https://github.com/huggingface/optimum-intel/pull/1744/changes/f4992572a1ae7cb6b61145ff1f5f3f9d070bce1a feel free to update @rkazants

tests/openvino/test_decoder.py
rkazants: Surprisingly, reference results generated by transformers v4 and v5 for long rope case are different (in real model as well). Currently, OV results match v4 transformers in both environments with v4 and v5.
I excluded transformers v5 case.

Issue #1721, opened 2026-05-05
Qwen3.6 (qwen3_5 / qwen3_5_moe) text-generation export blocked — only image-text-to-text task registered


bobvious:
## Summary

Both `Qwen/Qwen3.6-35B-A3B` (MoE) and `Qwen/Qwen3.6-27B` (dense) are published by the Qwen team as unified VL variants with `architectures: ['Qwen3_5MoeForConditionalGeneration']` / `['Qwen3_5ForConditionalGeneration']`. There are no separate text-only releases.

The OpenVINO exporter on `main` (`1.27.0.dev0+d4dd21a` as of 2026-05-05) registers these parent architectures for `image-text-to-text` only, so attempting a text-generation export aborts:

```
$ optimum-cli export openvino \
    --model Qwen/Qwen3.6-35B-A3B \
    --task text-generation-with-past \
    --weight-format int4 \
    --group-size 64 \
    --ratio 1.0 \
    out/

ValueError: Asked to export a qwen3_5_moe model for the task text-generation-with-past,
but the Optimum OpenVINO exporter only supports the tasks image-text-to-text for qwen3_5_moe.
Please use a supported task. Please open an issue at https://github.com/huggingface/optimum-intel/issues
if you would like the task text-generation-with-past to be supported in the OpenVINO export for qwen3_5_moe.
```

Same error for `Qwen/Qwen3.6-27B` (`qwen3_5`).

## Repro

```bash
pip install \
  "optimum-intel @ git+https://github.com/huggingface/optimum-intel.git@main" \
  "transformers @ git+https://github.com/huggingface/transformers.git@main"

optimum-cli export openvino \
  --model Qwen/Qwen3.6-27B \
  --task text-generation-with-past \
  --weight-format int4 \
  out/
```

(transformers main is required separately — release < 5.7 doesn't recognize the `qwen3_5` / `qwen3_5_moe` model_types.)

## Root cause

`optimum/exporters/openvino/model_configs.py` has the registrations split:

- `qwen3_5_moe` (line ~6174) → `image-text-to-text` only — corresponds to the unified `*_ForConditionalGeneration` architecture HF publishes
- `qwen3_5_moe_text` (line ~6165) → `text-generation`, `text-generation-with-past` — works but only routable from inside the VL config's LANGUAGE behavior code path, not as an export entrypoint
- Same split for `qwen3_5` / `qwen3_5_text`

[...]

PR #1756, opened 2026-05-30
[OpenVINO] Export DFlash for OpenVINO

ofirzaf:
# What does this PR do?

We implement the support to export DFlash draft models for speculative decoding with OpenVINO.
Also, we implement hidden_states annotations in exported OV models to better support operations that require hidden_states as outputs from OV models (like DFlash/Eagle3) methods, that will be applied automatically to all models exported for text generation the graph doesn't change as this is only annotations.

Commands to export DFlash model with this PR:
```bash
optimum-cli export openvino \
  --model z-lab/Qwen3.6-Coder-35B-A3B-DFlash \
  --task text-generation-with-past \
  --weight-format int4 \
  --all-layers \
  --trust-remote-code \
  qwen3.6-35b-a3b-dflash-int4-ov
```

## Before submitting


optimum/commands/export/openvino.py
```diff
@@ -294,6 +294,15 @@ def parse_args_openvino(parser: "ArgumentParser"):
...
     )
+    optional_group.add_argument(
+        "--dflash-target-model",
+        type=str,
+        default=None,
+        help=(
+            "Target model ID or local path used when exporting DFlash draft models. Only the target token "
+            "embedding and lm_head weights are loaded."
+        ),
+    )
```
openvino-agent: do not add any new options. They are not general.

optimum/commands/export/openvino.py
```diff
@@ -479,6 +488,7 @@ def run(self):
                 library_name=library_name,
                 variant=self.args.variant,
                 model_kwargs=self.args.model_kwargs,
+                dflash_target_model=self.args.dflash_target_model,
```
rkazants: not needed

optimum/exporters/openvino/__main__.py
```diff
@@ -189,6 +189,35 @@ def update_config_for_eagle3(config):
...
+            "AutoModelForCausalLM": moduler_path + "--model_patcher.Qwen3DFlashForCausalLM",
+        }
+
+    config.dflash_target_model = dflash_target_model
+    config.dflash_target_cache_dir = cache_dir
+    config.dflash_target_revision = revision
+    config.dflash_target_token = token
+    config.dflash_target_local_files_only = local_files_only
+    config.tie_word_embeddings = False
+    return config
```
rkazants: looks duplicate for eagle3. Avoid duplication just try to re-use exisintg functionality for model loading

optimum/exporters/openvino/__main__.py
```diff
@@ -358,6 +388,15 @@ def main_export(
...
             loading_kwargs["config"] = update_config_for_eagle3(config)
+        elif isinstance(archs, list) and len(archs) > 0 and archs[0] == "DFlashDraftModel":
+            loading_kwargs["config"] = update_config_for_dflash(
+                config,
+                dflash_target_model=dflash_target_model,
+                cache_dir=cache_dir,
+                revision=revision,
+                token=token,
+                local_files_only=local_files_only,
+            )
```
rkazants: generalize with eagle 3

optimum/exporters/openvino/convert.py
```diff
@@ -83,6 +84,9 @@
 
 logger = logging.getLogger(__name__)
 
+HIDDEN_STATES_RT_INFO_KEY = "hidden_states_decoder_layers"
+HIDDEN_STATE_TENSOR_NAME_TEMPLATE = "ov.hidden_states.decoder_layer_{}"
```
rkazants: please remove it, it should be in internal functions special for dflash

optimum/exporters/openvino/model_configs.py
```diff
@@ -421,6 +421,55 @@ class Qwen2MoEOpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
     _MODEL_PATCHER = Qwen2MoEPatcher
 
 
+class DFlashDummyInputGenerator(DummyInputGenerator):
```
rkazants: can we re-use InputGenerator for Eagle3

rkazants (commented):
please avoid duplicate weights among dflash and main models.
Let us have weights in the main model and avoid any changes in the main IR.
Dflash will have `input_embeds` input instead of `input_ids`

optimum/exporters/openvino/convert.py
ofirzaf: This is part of the hidden_states annotation mechanism which is beneficial for all methods, speculative decoding or others, that will need to find the the output hidden_states from any of the decoder layers. It is generic and can be applied (IMO, should be applied by default) to any model exported that support recording of the hidden_states.

optimum/commands/export/openvino.py
ofirzaf: Removed

optimum/commands/export/openvino.py
ofirzaf: Removed

optimum/exporters/openvino/__main__.py
ofirzaf: Unified in a single function

optimum/exporters/openvino/__main__.py
ofirzaf: Done

optimum/exporters/openvino/model_configs.py
ofirzaf: I don't see a way to reuse it cleanly. It comes down to that we have a contract we need to keep between the lengths of the `position_ids`, `inputs_embeds` and `hidden_states`.

ofirzaf:
@rkazants hi, please review the changes and comments. Thanks

optimum/exporters/openvino/model_configs.py
rkazants: Please double-check one more time, based on model type you can use different shapes for these inputs inside this class

optimum/exporters/openvino/model_configs.py
```diff
@@ -439,8 +487,41 @@ class Qwen3OpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
...
+
     @property
     def inputs(self) -> Dict[str, Dict[int, str]]:
+        if self.dflash:
+            common_inputs = super().inputs
+            common_inputs.pop("attention_mask", None)
+            common_inputs.pop("input_ids", None)
+            common_inputs["inputs_embeds"] = {0: "batch_size", 1: "block_size", 2: "hidden_size"}
+            common_inputs["hidden_states"] = {0: "batch_size", 1: "context_length", 2: "target_hidden_size"}
+            common_inputs["position_ids"] = {0: "batch_size", 1: "position_sequence_length"}
```
rkazants: where is there no attention mask? We need to have it for multi-batch. Please add it

optimum/exporters/openvino/model_configs.py
```diff
@@ -439,8 +487,41 @@ class Qwen3OpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
...
+            self.DUMMY_INPUT_GENERATOR_CLASSES = (DFlashDummyInputGenerator, GemmaDummyPastKeyValuesGenerator)
+            self.MIN_TRANSFORMERS_VERSION = "4.57.0"
+
     @property
     def inputs(self) -> Dict[str, Dict[int, str]]:
+        if self.dflash:
+            common_inputs = super().inputs
+            common_inputs.pop("attention_mask", None)
+            common_inputs.pop("input_ids", None)
+            common_inputs["inputs_embeds"] = {0: "batch_size", 1: "block_size", 2: "hidden_size"}
```
rkazants: why hidden_size is here? This is not dynamic dimension. Clean-up other static dimensions

optimum/exporters/openvino/model_configs.py
```diff
@@ -439,8 +487,41 @@ class Qwen3OpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
     _MODEL_PATCHER = OVDecoderModelPatcher
 
+    def __init__(
```
rkazants, line 427: what is model_type of Dflash model? What transformers versions are compatible for exporting

tests/openvino/test_export.py
rkazants, line 1: add test case into test_genai.py same as for eagle 3.
Create tiny model for dflash and let me share it on optimum-intel-internal-testing

tests/openvino/test_export.py
```diff
@@ -65,6 +68,24 @@
 
 logger = logging.get_logger()
 
+HIDDEN_STATES_RT_INFO_KEY = "hidden_states_decoder_layers"
+
+
```
rkazants: please avoid implementing new test suite special for dflash. Use exising test suite for eagle 3 and just test dflash case there. Otherwise, it becomes not mainable code and hard to track what each test suite validates

optimum/exporters/openvino/convert.py
rkazants: disagree that we should have the global vars here that needed only for dflash (1% case of all). Put them into dflash specific rountine

rkazants (changes requested):
The main point: remove dflash specific code from common part into dflash methods,
do not implement new test suite and re-use existing for validation, add test case into test_genai.py,

rkazants:
update documentation is also required

optimum/exporters/openvino/model_configs.py
ofirzaf: Done

optimum/exporters/openvino/model_configs.py
ofirzaf: Added

optimum/exporters/openvino/model_configs.py
ofirzaf: Removed

optimum/exporters/openvino/model_configs.py
ofirzaf: Added checks

tests/openvino/test_export.py
ofirzaf: Added

optimum/exporters/openvino/convert.py
ofirzaf: Like I stated in a previous comment, this isn't dflash specific this is general for every model export and can be later utilized by other methods that require hidden_states from exported models.
Let me know if you still want me to move it to a specific routine

tests/openvino/test_export.py
ofirzaf: The tests added in this file are for the annotations export which is not specific for DFlash or SD.
Do you mean that I should add a DFlash case in the Eagle3 test in `test_genai.py`?

ofirzaf:
Hi @rkazants, I made some changes and have some questions. Please respond. Thanks

docs/source/openvino/models.mdx
```diff
@@ -201,4 +201,7 @@ Here is the list of the supported architectures :
 
 ## [AngelSlim] (https://github.com/Tencent/AngelSlim)
 - Qwen3-VL-4B-Instruct_eagle3
-- Qwen3-1.7B_eagle3
\ No newline at end of file
+- Qwen3-1.7B_eagle3
+
+## [DFlash](https://github.com/z-lab/dflash)
+- All Qwen3-based DFlash draft models
```
rkazants: let us only mention Qwen3,5 and Qwen3,6.
As I understand, for others like Qwen3-next, Qwen3, you did not validate.

ofirzaf: All the draft models are based on Qwen3 architecture so the export should work for all published drafts until now.
Regarding what we tested to work with GenAI in the full pipeline: Qwen3/3.5/3.6.

Do you want me to change?

optimum/exporters/openvino/convert.py
rkazants: Kind of risky change because you can override existing tensor names for some models on which some users already relide their logic. All IRs can get backward-incompatible. Let us change only for Qwen3.5 and Qwen3,6.
Also, I worried that we support many models of different archs and not all of them a stack of transformers blocks. They can be mix with Convolution layers, encoders, TTS, diffusers. So we can impact some of them or logic overwriting tensor names will break for some arch.

optimum/exporters/openvino/model_configs.py
```diff
@@ -439,8 +439,59 @@ class Qwen3OpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
...
+            dflash_config = getattr(config, "dflash_config", {}) or {}
+            if not dflash_config.get("target_layer_ids", []):
+                raise ValueError("DFlash export requires non-empty dflash_config['target_layer_ids'].")
+            # DFlash draft checkpoints still advertise model_type="qwen3"; the
+            # architecture and dflash_config fields identify the draft variant.
+            self.DUMMY_INPUT_GENERATOR_CLASSES = (
+                DummyTextInputGenerator,
+                Eagle3VLMDummyGenerator,
+                Eagle3DummyGenerator,
+                GemmaDummyPastKeyValuesGenerator,
```
rkazants, line 461: do you need all of these generators? I think - not. Please clean-up this list.

optimum/exporters/openvino/model_configs.py
```diff
@@ -439,8 +439,59 @@ class Qwen3OpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
...
+            )
+            self.MIN_TRANSFORMERS_VERSION = "4.57.0"
```
rkazants: does it support transformers `>=5.0.0`? I think - not. Let us set `MAX_TRANSFORMERS_VERSION=4.57.99` because this is remote-code-model and for some future transfomrers version it will not work.

tests/openvino/test_genai.py
```diff
@@ -587,6 +588,70 @@ def test_compare_outputs(self, model_arch):
         torch.testing.assert_close(transformers_output, genai_output, rtol=1e-2, atol=1e-3)
 
 
+@pytest.mark.skipif(OPENVINO_DEVICE == "NPU", reason="DFlash test is not yet supported on NPU")
```
rkazants: this new suite is mostly a copy-paste of eagle3. Please parameterize existing test suite eagle3 with minimal changes and test dflash there. The ideas and API is quite same for both methods.

tests/openvino/utils_tests.py
```diff
@@ -358,6 +358,13 @@ def _create_tiny_kokoro_model():
 
 EAGLE3_MODELS = {"qwen3_eagle3": ("AngelSlim/Qwen3-1.7B_eagle3", "Qwen/Qwen3-1.7B")}
 
+DFLASH_MODELS = {
+    "qwen3_dflash": (
+        os.getenv("OV_DFLASH_DRAFT_MODEL_ID", "optimum-intel-internal-testing/tiny-random-qwen3-dflash"),
+        os.getenv("OV_DFLASH_TARGET_MODEL_ID", "optimum-intel-internal-testing/tiny-random-qwen3"),
```
rkazants: please send me tiny model for dflash and I upload to `optimum-intel-internal-testing`

rkazants (changes requested):
Please update PR description. The code-snippets are out-dated there.

Also, now the main concern is that we change IRs for all models, let us change only Qwen3.5,     Qwen3.6. Or better to think how avoid any changes in IRs for currently supported models. I think you can walkthough model graph in run-time and find required hidden states.

docs/source/openvino/models.mdx
rkazants: should work but not validated. This does not work for documentation

tests/openvino/utils_tests.py
ofirzaf: Sent to you over Teams

optimum/exporters/openvino/model_configs.py
ofirzaf: I validated export of DFlash draft with `transformers==5.2.0`.
Let me know if you still want to limit the version

optimum/exporters/openvino/model_configs.py
ofirzaf: - `DummyTextInputGenerator` is required for `[attention_mask, position_ids]`
- `Eagle3VLMDummyGenerator` is required for `inputs_embeds`
- `Eagle3DummyGenerator` is required for `hidden_states`
- `GemmaDummyPastKeyValuesGenerator` is required for `past_key_values`

tests/openvino/test_genai.py
ofirzaf: done

optimum/exporters/openvino/model_configs.py
ofirzaf: done

docs/source/openvino/models.mdx
ofirzaf: done

optimum/exporters/openvino/convert.py
```diff
@@ -138,12 +142,94 @@ def _save_model(
...
+        return
+
+    # HF hidden_states usually contains embeddings first, then one tensor after each decoder block.
+    offset = 1 if len(temporary_results) >= num_hidden_layers + 1 else 0
+    hidden_layer_results = temporary_results[offset : offset + num_hidden_layers]
+    if len(hidden_layer_results) == num_hidden_layers:
+        layers = {}
+        for layer_idx, result in enumerate(hidden_layer_results):
+            tensor_name = HIDDEN_STATE_TENSOR_NAME_TEMPLATE.format(layer_idx)
+            result.input_value(0).get_tensor().add_names({tensor_name})
```
rkazants: still changed

ofirzaf: Fixed

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8644,6 +8676,336 @@ def forward(
         )
 
 
+def _dflash_apply_rotary_pos_emb(q, k, cos, sin, unsqueeze_dim=1):
```
rkazants, line 8167: I believe the same routine already exists defined in transformers or optimum-intel. Please re-use it, let us avoid duplication

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8644,6 +8676,336 @@ def forward(
...
+
+    full_mask = full_mask[None, None, :, :].expand(query_states.shape[0], 1, -1, -1)
+
+    if attention_mask is None:
+        return full_mask
+    # Keep the last kv_len columns: the kept keys are the caller mask's final entries.
+    return attention_mask[:, :, :, -kv_len:] + full_mask
+
+
+class Qwen3DFlashAttention(nn.Module):
```
rkazants: same comment about duplication, also check classes for Eagle 3 and re-use it

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8644,6 +8676,336 @@ def forward(
...
+            scaling=self.scaling,
+            sliding_window=self.sliding_window,
+            **kwargs,
+        )
+        attn_output = attn_output.reshape(bsz, q_len, -1)
+        attn_output = self.o_proj(attn_output)
+        return attn_output, attn_weights
+
+
+class Qwen3DFlashDecoderLayer(nn.Module):
```
rkazants, line 8332: same comment about duplication, also check classes for Eagle 3 and re-use it

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -553,7 +553,10 @@ def prepare_inputs(
             else:
                 position_ids = np.cumsum(attention_mask, axis=1) - 1
                 position_ids[attention_mask == 0] = 1
-            if past_key_values:
+            dflash_hidden_states = kwargs.get("hidden_states", None)
```
rkazants: it is strange that we always treat `hidden_states` from kwargs as `dflash` hidden states. This is common logic and can affect other models.
Please distinguish dflash case preciously.

tests/openvino/test_export.py
```diff
@@ -263,6 +309,145 @@ def test_export(self, model_type: str):
             model_kwargs = {"vocoder": "fxmarty/speecht5-hifigan-tiny"}
         self._openvino_export(model_type, model_kwargs=model_kwargs)
 
+    @parameterized.expand(("llama", "gemma4", "qwen3_5", "qwen3_5_moe", "gpt_oss"))
```
rkazants: please do not touch other models. It is strange that implement tests for existing models. Please check how it is implemented export tests for eagle 3 and try to embed dflash there. Now I see a lot new tests, seems AI generated that will be hard to maintain.
Embed dflash export validation into existing test suites.

optimum/exporters/openvino/convert.py
```diff
@@ -138,12 +141,16 @@ def _save_model(
...
+    if getattr(config, "dflash", False):
+        model = _add_dflash_mode_to_rt_info(model, config._config)
 
     save_model(model, path, compress_to_fp16)
     del model
     gc.collect()
 
 
+
+
```
rkazants: not needed

optimum/exporters/openvino/convert.py
```diff
@@ -416,10 +425,10 @@ def ts_patched_forward(*args, **kwargs):
         if allow_skip_tracing_check(library_name, model_type):
             ts_decoder_kwargs["trace_kwargs"] = {"check_trace": False}
 
+        torch_export = os.getenv("OPENVINO_DYNAMO_EXPORT", "false").lower() == "true"
```
rkazants: why changed? DO not change this logic, please

rkazants (commented):
@ofirzaf, please correct PR description. cmd-line for export is still out-dated.
For best user experience, use should run the followgin command for dflash export:

this is just pseudo-cmd-line, please provide a real one
```
optimum-cli export openvino -m dflash_model_id ov_dflash --trust-remote-code
```

I think that exported dflash can be used for both text-only and vlm scenarious. No need to specify task here.

optimum/exporters/openvino/convert.py
```diff
@@ -464,6 +472,8 @@ def ts_patched_forward(*args, **kwargs):
         if stateful:
             patch_stateful(model.config, ov_model)
 
+        _add_hidden_states_rt_info(model, ov_model, config, torch_export)
```
rkazants: please avoid calling this routine for all text-generation models. Just call it for dflash supported

optimum/exporters/openvino/convert.py
```diff
@@ -979,6 +989,24 @@ def _add_eagle3_mode_to_rt_info(model: Model):
     return model
 
 
+def _add_dflash_mode_to_rt_info(model: Model, hf_config):
```
rkazants: move all of thes auxiliary internal functions into `utils_dflash.py`

optimum/exporters/openvino/model_configs.py
```diff
@@ -439,8 +439,59 @@ class Qwen3OpenVINOConfig(TextDecoderWithPositionIdsOnnxConfig):
...
+            config=config,
+            task=task,
+            int_dtype=int_dtype,
+            float_dtype=float_dtype,
+            use_past=use_past,
+            use_past_in_inputs=use_past_in_inputs,
+            preprocessors=preprocessors,
+        )
+        archs = getattr(config, "architectures", None)
+        self.dflash = isinstance(archs, list) and len(archs) > 0 and archs[0] == "DFlashDraftModel"
```
regisss, line 447: Maybe easier to check if the config has a `dflash_config` field?

tests/openvino/utils_tests.py
```diff
@@ -358,6 +358,13 @@ def _create_tiny_kokoro_model():
 
 EAGLE3_MODELS = {"qwen3_eagle3": ("AngelSlim/Qwen3-1.7B_eagle3", "Qwen/Qwen3-1.7B")}
 
+DFLASH_MODELS = {
+    "qwen3_dflash": (
+        "optimum-intel-internal-testing/tiny-random-qwen3-dflash",
+        "optimum-intel-internal-testing/tiny-random-qwen3",
```
regisss, line 395: Why do we need the non-dflash model here? To compare the outputs of both models in the tests?

tests/openvino/test_export.py
```diff
@@ -66,6 +71,47 @@
 logger = logging.get_logger()
 
 
+def _find_output_by_locator(model, locator):
```
regisss: Better to have such method defined in a utils file like `utils_tests.py`

tests/openvino/test_export.py
```diff
@@ -66,6 +71,47 @@
...
+    matches = [op for op in model.get_ops() if op.get_friendly_name() == locator["producer"]]
+    if len(matches) != 1:
+        raise AssertionError(f"Producer {locator['producer']!r} resolved to {len(matches)} OpenVINO nodes")
+    output_index = locator["output_index"]
+    if output_index >= len(matches[0].outputs()):
+        raise AssertionError(f"Producer {locator['producer']!r} has no output {output_index}")
+    return matches[0].output(output_index)
+
+
+class _HiddenStateReference(torch.nn.Module):
```
regisss: Same

tests/openvino/test_export.py
```diff
@@ -66,6 +71,47 @@
...
+            attention_mask=attention_mask,
+            position_ids=position_ids,
+            output_hidden_states=True,
+            use_cache=False,
+            return_dict=True,
+        )
+        return tuple(outputs.hidden_states)
+
+
+def _export_hidden_state_reference(model, input_ids, attention_mask, position_ids):
```
regisss: Same

optimum/exporters/openvino/model_patcher.py
ofirzaf: I could not find a transformers/optimum-intel helper that can be used directly here. The helper in transformers expects `cos`, `sin`, `q`, `k` to be in the same length of the `position_ids`. In DFlash `q` and `k` are not in the same length, `q` length is the block length, 1 + mask tokens,  and `k` length  is the target hidden states length + block length.

optimum/exporters/openvino/model_patcher.py
ofirzaf: `Qwen3DecoderLayer` is not a suitable base because its initializer creates a standard `Qwen3Attention`.  DFlash would then need to replace it with `Qwen3DFlashAttention`, unnecessarily initializing and discarding a second attention module. Its forward also assumes a single hidden-state sequence.

The Eagle3 class are also not compatible. Eagle3 concatenate the input embeddings and target hidden states  and uses Llama attention.

optimum/exporters/openvino/convert.py
ofirzaf: This annotation is useful for any method that needs hidden_states from the middle of the model, not just DFlash but Eagle3 and DSpark also for example.
Furthermore, a lot of new models get Eagle3/DFlash support and since this only adds an `rt_info` entry and doesn't change the IR, wouldn't it make sense to have it on by default?

In case you still want to limit this call to DFlash supported models only, how do you suggest to do that?

optimum/exporters/openvino/model_configs.py
ofirzaf: I saw Eagle3 export uses the same check so I went for the same style. Should I change?

tests/openvino/utils_tests.py
ofirzaf: `tiny-random-qwen3-dflash` is a draft model and `tiny-random-qwen3` is a target model. Need both for testing, similar to Eagle3 testing.

optimum/exporters/openvino/convert.py
ofirzaf: Reverted

optimum/exporters/openvino/convert.py
ofirzaf: Fixed

optimum/intel/openvino/modeling_decoder.py
ofirzaf: Fixed

optimum/exporters/openvino/model_patcher.py
ofirzaf: I refactored `Qwen3DFlashAttention` to subclass Transformers’ `Qwen3Attention`. `forward` remains custom to support DFlash asymmetric `q` and `k`, cache handling and attention mask handling

Eagle3 attention is not compatible since it handles hidden_states and inputs differently.

optimum/exporters/openvino/convert.py
ofirzaf: This function specifically is very similar to the eagle3 function above it. 
I guess you refer to the other functions that relate to the hidden_states annotations, moved to `utils_annotations.py`

tests/openvino/test_export.py
ofirzaf: Moved

tests/openvino/test_export.py
ofirzaf: These tests test the implementation of the annotations. This is a list of common models that also support DFlash/Eagle3 and I thought it is correct to test that the new annotation method is able to find their hidden_states correctly. Do you want me to remove it or narrow down the list?

I didn't find any Eagle3 export test.

Tests that were added in this PR are annotations test for a few common models and a test for annotations + quantization, and genai test that was fused with the Eagle3 genai test.

ofirzaf:
@rkazants @regisss 

I addressed your comments, please review. Thanks!

@rkazants, the command in the description is up-to date. Regarding the task, I haven't tested what happens when you remove the task from the command. Regarding using with VLM, this is still not tested, but DFlash should be agnostic to that and act exactly the same.

optimum/exporters/openvino/model_configs.py
regisss: Let's keep the same logic here then, should be good enough 👍

optimum/exporters/openvino/convert.py
```diff
@@ -382,6 +385,8 @@ def export_pytorch(
 
         dummy_inputs = config.rename_ambiguous_inputs(dummy_inputs)
         dummy_inputs, dict_inputs = remove_none_from_dummy_inputs(dummy_inputs)
+        output_names = list(config.outputs.keys())
+
```
rkazants: revert this change

optimum/exporters/openvino/convert.py
```diff
@@ -451,7 +456,6 @@ def ts_patched_forward(*args, **kwargs):
 
         ov_model.validate_nodes_and_infer_types()  # TODO: remove as unnecessary validation?
 
-        output_names = list(config.outputs.keys())
```
rkazants, line 387: please revert

optimum/exporters/openvino/convert.py
```diff
@@ -979,6 +985,24 @@ def _add_eagle3_mode_to_rt_info(model: Model):
     return model
 
 
+def _add_dflash_mode_to_rt_info(model: Model, hf_config):
+    """
+    Add DFlash metadata.
+    """
+    try:
+        model.set_rt_info("True", ["dflash_mode"])
```
rkazants, line 945: I think you don't need `dflash_mode`. If `dflash` is set below, it means that the model is ready for dflash scenario. Please remove `dflash_mode`

optimum/exporters/openvino/convert.py
```diff
@@ -979,6 +985,24 @@ def _add_eagle3_mode_to_rt_info(model: Model):
     return model
 
 
+def _add_dflash_mode_to_rt_info(model: Model, hf_config):
+    """
+    Add DFlash metadata.
+    """
+    try:
```
rkazants, line 944: avoid to use try-catch block. If it is use, it means internal design is not complete and we can expect some undefined behaviour. Properlly check inputs and write-up logic out of try-catch block

optimum/exporters/openvino/convert.py
```diff
@@ -979,6 +985,24 @@ def _add_eagle3_mode_to_rt_info(model: Model):
     return model
 
 
+def _add_dflash_mode_to_rt_info(model: Model, hf_config):
```
rkazants: please set input types. For `hf_config` it is node defined.

optimum/exporters/openvino/convert.py
```diff
@@ -464,6 +468,8 @@ def ts_patched_forward(*args, **kwargs):
         if stateful:
             patch_stateful(model.config, ov_model)
 
+        add_hidden_states_rt_info(model, ov_model, config)
```
rkazants: I don't think you nee to add here. We should have common entry point when rt_info is set into ov.Model. There you should add calling logic for setting dflash rt info.

optimum/exporters/openvino/model_patcher.py
```diff
@@ -71,6 +73,36 @@
...
+    Qwen3MLP = _UnavailableQwen3Module
+    Qwen3Attention = _UnavailableQwen3Module
+    Qwen3RMSNorm = _UnavailableQwen3Module
+    Qwen3RotaryEmbedding = _UnavailableQwen3Module
+
+    def qwen3_eager_attention_forward(*args, **kwargs):
+        raise ImportError("DFlash export requires transformers >= 4.57.")
+
+    def qwen3_rotate_half(*args, **kwargs):
+        raise ImportError("DFlash export requires transformers >= 4.57.")
```
rkazants: please remove it, because there is another logic existing that will check transformers version

optimum/exporters/openvino/model_patcher.py
```diff
@@ -71,6 +73,36 @@
...
+    Qwen3PreTrainedModel = PreTrainedModel
+
+    class _UnavailableQwen3Module(nn.Module):
+        def __init__(self, *args, **kwargs):
+            raise ImportError("DFlash export requires transformers >= 4.57.")
+
+    Qwen3MLP = _UnavailableQwen3Module
+    Qwen3Attention = _UnavailableQwen3Module
+    Qwen3RMSNorm = _UnavailableQwen3Module
+    Qwen3RotaryEmbedding = _UnavailableQwen3Module
```
rkazants: please remove `_UnavailableQwen3Module` and assign `object` for definition of classes above

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8644,6 +8676,322 @@ def forward(
...
+
+    full_mask = full_mask[None, None, :, :].expand(query_states.shape[0], 1, -1, -1)
+
+    if attention_mask is None:
+        return full_mask
+    # Keep the last kv_len columns: the kept keys are the caller mask's final entries.
+    return attention_mask[:, :, :, -kv_len:] + full_mask
+
+
+class Qwen3DFlashAttention(Qwen3Attention):
```
rkazants, line 8230: please add a comment with a link from where this class definition is borrowed. Need this as a reference

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8644,6 +8676,322 @@ def forward(
...
+            scaling=self.scaling,
+            sliding_window=self.sliding_window,
+            **kwargs,
+        )
+        attn_output = attn_output.reshape(bsz, q_len, -1)
+        attn_output = self.o_proj(attn_output)
+        return attn_output, attn_weights
+
+
+class Qwen3DFlashDecoderLayer(nn.Module):
```
rkazants, line 8332: need a comment with a link to reference code

optimum/exporters/openvino/model_patcher.py
```diff
@@ -8644,6 +8676,322 @@ def forward(
...
+        hidden_states = residual + hidden_states
+
+        residual = hidden_states
+        hidden_states = self.post_attention_layernorm(hidden_states)
+        hidden_states = self.mlp(hidden_states)
+        hidden_states = residual + hidden_states
+        return hidden_states
+
+
+class Qwen3DFlashDraftModel(Qwen3PreTrainedModel):
```
rkazants, line 8378: need a comment with a link to reference code

optimum/exporters/openvino/model_patcher.py
```diff
@@ -9512,6 +9867,9 @@ def patched_forward(
 
                 outputs["present_key_values"] = present_key_values
 
+            if hidden_states is not None:
+                outputs["hidden_states"] = hidden_states
+
```
rkazants: It seems that you changed IR by adding new outputs, right? Can we avoid this?

optimum/exporters/openvino/utils_annotations.py
```diff
@@ -0,0 +1,239 @@
...
+    hidden_states_rt_info_key = "hidden_states_decoder_layers"
+
+    if "text-generation" not in getattr(config, "task", ""):
+        return
+    if os.getenv("OPENVINO_DYNAMO_EXPORT", "false").lower() == "true":
+        logger.warning(
+            "Skipping hidden-state RT-info annotation for Dynamo export; "
+            "the locator discovery supports only the TorchScript conversion path."
+        )
+        return
```
rkazants: not needed, please remove. Now let us not support dflash for torch.export. It is a separate activity for adding torch.export.

optimum/exporters/openvino/utils_annotations.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+    if locators is None:
+        raise ValueError("Could not resolve unambiguous hidden-state boundaries.")
+
+    return {
+        "layers": {str(layer_id): locator for layer_id, locator in locators.items()},
+    }
+
+
+def add_hidden_states_rt_info(source_model, ov_model, config: "OnnxConfig"):
```
rkazants: now, there is no "OnnxConfig". We removed dependency from optimum-onnx

optimum/exporters/openvino/convert.py
```diff
@@ -979,6 +985,24 @@ def _add_eagle3_mode_to_rt_info(model: Model):
     return model
 
 
+def _add_dflash_mode_to_rt_info(model: Model, hf_config):
+    """
+    Add DFlash metadata.
```
rkazants: please clarify in the comment to which model do you set rt_into: main or draft? Explain why it is needed

optimum/intel/openvino/modeling_decoder.py
```diff
@@ -482,6 +482,7 @@ def prepare_inputs(
         **kwargs,
     ) -> Dict:
         batch_size = input_ids.shape[0]
+        is_dflash = getattr(self.config, "dflash_config", None) is not None
```
rkazants: please add a comment: why this logic is needed? is this logic for main or draft model?

tests/openvino/test_export.py
rkazants: I disagree to have it here. Move these new tests into test_dflash.py.
Also, similar to eagle3, add dflash into test_export_cli.py

rkazants (changes requested):
Please also set rt_info only for qwen3_5 model type. Do not affect other archs by these changes

rkazants:
> @rkazants, the command in the description is up-to date. Regarding the task, I haven't tested what happens when you remove the task from the command. Regarding using with VLM, this is still not tested, but DFlash should be agnostic to that and act exactly the same.

Please simplify cmd-line in the description. First-of-all, you defined Dflash model class inside optimum-intel, so it is no longer remote-code, so you don't need this option. Also, I do mind to have `--disable-convert-tokenizer`, please avoid. Exporting should work out-of-the-box.

optimum/exporters/openvino/convert.py
ofirzaf: Are you sure? All similar functions use try-catch: `_add_eagle3_mode_to_rt_info`, `_add_runtime_options_to_rt_info`

optimum/intel/openvino/modeling_decoder.py
ofirzaf: Removed logic from `modeling_decoder.py`

optimum/exporters/openvino/convert.py
ofirzaf: Done

optimum/exporters/openvino/utils_annotations.py
ofirzaf: Done

optimum/exporters/openvino/utils_annotations.py
ofirzaf: Done

optimum/exporters/openvino/model_patcher.py
ofirzaf: Done

optimum/exporters/openvino/model_patcher.py
ofirzaf: Done

optimum/exporters/openvino/model_patcher.py
ofirzaf: Done

optimum/exporters/openvino/model_patcher.py
ofirzaf: Done

optimum/exporters/openvino/model_patcher.py
ofirzaf: Done

optimum/exporters/openvino/convert.py
ofirzaf: Done

optimum/exporters/openvino/convert.py
ofirzaf: Done

optimum/exporters/openvino/convert.py
ofirzaf: Done

tests/openvino/test_export.py
ofirzaf: Done

optimum/exporters/openvino/model_patcher.py
ofirzaf: Removed. Left over from old annotation logic

optimum/exporters/openvino/convert.py
ofirzaf: Fixed

tests/openvino/test_dflash.py
rkazants, line 1: this test suite is not triggered now in CI. Please run it in this job https://github.com/huggingface/optimum-intel/blob/main/.github/workflows/test_openvino_remote_code.yml

rkazants (approved):
left one comment about running test_dflash in CI. Other looks good to me. Thanks!

optimum/exporters/openvino/model_configs.py
```diff
@@ -424,8 +424,61 @@ class Qwen3OpenVINOConfig(TextDecoderWithPositionIdsOpenVINOConfig):
...
+        if self.dflash:
+            model_type = getattr(config, "model_type", "")
+            if model_type != "qwen3":
+                raise ValueError(
+                    f"DFlash export supports only Qwen3-based draft models, got model_type={model_type!r}."
```
```suggestion
                    f"DFlash export supports only Qwen3-based draft models, got model_type={model_type}."
```

optimum/exporters/openvino/model_configs.py
```diff
@@ -424,8 +424,61 @@ class Qwen3OpenVINOConfig(TextDecoderWithPositionIdsOpenVINOConfig):
...
+            float_dtype=float_dtype,
+            use_past=use_past,
+            use_past_in_inputs=use_past_in_inputs,
+            preprocessors=preprocessors,
+        )
+        archs = getattr(config, "architectures", None)
+        self.dflash = isinstance(archs, list) and len(archs) > 0 and archs[0] == "DFlashDraftModel"
+        if self.dflash:
+            model_type = getattr(config, "model_type", "")
+            if model_type != "qwen3":
```
echarlaix, line 450: shouldn't we exclude the case where `model_type` is an empty string ?

optimum/exporters/openvino/model_configs.py
```diff
@@ -424,8 +424,61 @@ class Qwen3OpenVINOConfig(TextDecoderWithPositionIdsOpenVINOConfig):
...
+                raise ValueError("DFlash export requires non-empty dflash_config['target_layer_ids'].")
+            # DFlash draft checkpoints still advertise model_type="qwen3"; the
+            # architecture and dflash_config fields identify the draft variant.
+            self.DUMMY_INPUT_GENERATOR_CLASSES = (
+                DummyTextInputGenerator,
+                Eagle3VLMDummyGenerator,
+                Eagle3DummyGenerator,
+                GemmaDummyPastKeyValuesGenerator,
+            )
+            self.MIN_TRANSFORMERS_VERSION = "4.57.0"
```
echarlaix: we should be able to remove as `MIN_TRANSFORMERS_VERSION` already set to 4.57 by default 
```suggestion
```

optimum/exporters/openvino/model_patcher.py
echarlaix: @ofirzaf would you mind adding link to the original transformers code as well so that we can easily fix incompatibilities following transformers release https://github.com/huggingface/transformers/blob/v5.14.0/src/transformers/models/qwen3/modeling_qwen3.py#L222

optimum/exporters/openvino/model_patcher.py
echarlaix: @ofirzaf looks missing here, also would you mind adding link to the original transformers code as well so that we can easily fix incompatibilities following transformers release

optimum/exporters/openvino/model_patcher.py
echarlaix: also did you try for different transformers version (like v5.5 or v5.10) and if yes did you observed any issues ?

echarlaix (commented):
thanks for the addition @ofirzaf !

optimum/exporters/openvino/model_patcher.py
ofirzaf: Will add. Yeah, had no issues exporting DFlash with v5.2.0 and v5.5.4

optimum/exporters/openvino/model_patcher.py
ofirzaf: Oops, good catch!

optimum/exporters/openvino/model_configs.py
ofirzaf: What do you mean? If `model_type == ""` then we should proceed instead of throwing an error?

ofirzaf:
@echarlaix thanks for the review! Left 1 question for you

optimum/exporters/openvino/model_configs.py
echarlaix: yes to only raise an error when the `model_type` is set to an incompatible architecture (to avoid raising when `model_type` not present in the config) not a big deal in any case, we can be kept as it is

echarlaix (approved):
LGTM, thanks a lot @ofirzaf !

yangsu2022:
@rkazants @MaximProshin should we support Qwen3.6-35B-A3B with the `text-generation-with-past` task? DFlash depends on it.  Reference [Qwen3.6 (qwen3_5 / qwen3_5_moe) text-generation export blocked — only image-text-to-text task registered](https://github.com/huggingface/optimum-intel/issues/1721)

ofirzaf:
@yangsu2022 I think it will make more sense to be able to load VLM exported models with `LLMPipeline` or `OVModelForCausalLM` for those who want text only so you only export once.
We intend to extend DFlash support to VLM so this dependency is temporary

echarlaix:
@ofirzaf since the release of openvino 2026.3 (which triggers the following test), we have 

```
FAILED tests/openvino/test_genai.py::LLMPipelineWithSpeculativeDecodingTestCase::test_compare_outputs_1_qwen3_dflash - RuntimeError: Exception from src/core/src/pass/graph_rewrite.cpp:298:
[TotalSequenceLengthPattern] END: node: opset8::Gather Gather_3396428 (opset3::ShapeOf ShapeOf_3396425[0]:i64[4], opset1::Constant Constant_3396426[0]:i64[1], opset1::Constant Constant_3396427[0]:i64[]) -> (i64[1]) CALLBACK HAS THROWN: Check 'replacement' failed at src/common/transformations/src/transformations/paged_attention/total_sequence_length_pattern.cpp:108:
TotalSequenceLengthPattern transformation failed to determine the dimension value after the Gather operation. Most probably, the required dimension is dynamic: opset1::Concat __module.layers.0.self_attn/aten::cat/Concat_3 (opset8::Gather Gather_3399477[0]:f32[?,2,?,8], opset1::VariadicSplit __module.layers.0.self_attn/aten::split_with_sizes/VariadicSplit[0]:f32[?,2,?,8]) -> (f32[?,2,?,8])
```

would you mind taking a look ?

ofirzaf:
@echarlaix sure. Can you provide exact versions used? DFlash was not included in 2026.3 release so this might be the issue. Right now it is only in the nightly build.

echarlaix:
@ofirzaf v2026.3.0 is used as you can see in https://github.com/huggingface/optimum-intel/actions/runs/31163030328/job/92817468149 the test you added wasn't triggered before the release https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/test_genai.py#L584

ofirzaf:
@echarlaix we better bump it to 2026.4 or something like 2026.3.999 as it is still not in an official release. Let me know if you want me to open a PR on this

echarlaix:
@ofirzaf would you mind taking care of this ?

PR #1762, opened 2026-06-03
[OpenVINO] Add Image-to-Video support for LTX-Video

goyaladitya05:
# What does this PR do?
This PR adds image-to-video pipeline support for LTX-Video, enabling users to create videos conditioned on an input image combined with a text prompt. Adding image conditioning provides a strong visual anchor, improving control over composition and style.

Image-to-Video sample : https://gist.github.com/goyaladitya05/e615487b9d95163802be7b7875af1136
ltx_i2v_output (4)

Text-to-Video sample: https://gist.github.com/goyaladitya05/f86e8d252e0830b072f41d6e5c518ee0
ltx_t2v_output (3)

GenAI PR: https://github.com/openvinotoolkit/openvino.genai/pull/3907

## Before submitting


optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1657,11 +1666,51 @@ class OVSanaSprintPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, S
...
     main_input_name = "prompt"
     export_feature = "text-to-video"
     auto_model_class = LTXPipeline
 
+    def __init__(self, *args, **kwargs):
+        # LTX-Video transformer cross-attention saturates in bf16 on bf16-capable CPUs,
+        # producing corduroy artifacts amplified by CFG. Default to fp32 unless overridden.
+        ov_config = kwargs.get("ov_config") or {}
+        if "INFERENCE_PRECISION_HINT" not in ov_config:
+            kwargs["ov_config"] = {**ov_config, "INFERENCE_PRECISION_HINT": "f32"}
```
goyaladitya05: I tested without this block, but the outputs were drastically worse. Hence this is required/preferred.

goyaladitya05: These were the outputs I got without forcing the transformer to fp32 ->

Image to Video

ltx_i2v_output (3)

Text to Video
ltx_t2v_output (2)

goyaladitya05:
cc @sgonorov @likholat

rkazants:
@anatyrova

goyaladitya05:
All the CI failures are either 429 rate limit errors, or unreleted failures. Could we please re-run them?

goyaladitya05:
Hey @echarlaix @anatyrova, could you please have a look?

optimum/exporters/openvino/model_configs.py
```diff
@@ -2472,7 +2471,7 @@ def inputs(self):
             "width": {},
             "height": {},
             "num_frames": {},
-            "timestep": {0: "batch_size"},
+            "timestep": {0: "batch_size", 1: "video_sequence_length"},
```
anatyrova, line 2642: adding "video_sequence_length" parameter could potentially do something to t2v inference, but i checked and wwb similarity for t2v with this pr is 0.978353 (fp16)

optimum/intel/openvino/modeling_diffusion.py
anatyrova: in the original ltx-video, do they also use fp32 in the transformer? because it is strange if they don't. could you check that pls?

goyaladitya05: Hi Altana, I'm comparing it with different dtypes, and will share the results soon. I've already attached the default outputs (bf16), which were significantly worse. will update here soon. 

Point worth noting - The default outputs were worse on both sides, optimum and genai.

goyaladitya05: Following this, I ran more tests.
1. The outputs were still grey/bad/noisy on genai master using Text2Video (pre-existing).
2. I had been using a bf16 capable CPU (Intel Sapphire Rapids). I isolate this to be an issue with how OpenVINO's CPU plugin handles bf16. 
3. I ran diffusers with bf16 on the same CPU, and it worked fine.
4. I regenerated samples in fp16 using an Intel GPU, and they were fine(samples attached). Hence I've removed inference precision hint.

ltx_i2v_output
ltx_t2v_output

goyaladitya05: Ref: https://github.com/openvinotoolkit/openvino/issues/36715

goyaladitya05:
Hi @anatyrova, is there anything we require from my end to merge this PR?

goyaladitya05:
hey @anatyrova, could you please re-run the failing tests?

goyaladitya05:
I tested t2v from master and this PR, and got the same numbers (fp32, CPU,10 samples).
HF vs Optimum 0.986. This PR does not create any regressions for the existing code.

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -847,6 +852,8 @@ def _reshape_vae_encoder(
                     "Could not identify `in_channels` from the VAE encoder configuration, to statically reshape the VAE encoder please provide a configuration."
                 )
                 self.is_dynamic = True
+        if self.__class__.__name__ == "OVLTXImageToVideoPipeline":
```
anatyrova: class name string check should use isinstance or class attributes. string-based class checks are brittle and make inheritance difficult

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1701,11 +1710,43 @@ class OVSanaSprintPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, S
...
     auto_model_class = LTXPipeline
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.transformer is not None:
+            self.transformer.__class__ = OVModelLTXTransformer
+
+
+class OVLTXImageToVideoPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, LTXImageToVideoPipeline):
+    main_input_name = "image"
```
anatyrova, line 1933: consider adding input validation for image (raise error if image is None)

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -1701,11 +1710,43 @@ class OVSanaSprintPipeline(OVDiffusionPipeline, OVTextualInversionLoaderMixin, S
     auto_model_class = SanaSprintPipeline
 
 
+class OVModelLTXTransformer(OVModelTransformer):
+    def forward(self, hidden_states, timestep=None, **kwargs):
+        # T2V passes timestep as [B]; if the exported IR expects [B, S], broadcast here.
+        # The rank check makes this backward-compatible with older 1D-timestep exports.
+        if timestep is not None and timestep.ndim == 1 and self._timestep_rank == 2:
+            timestep = timestep.unsqueeze(-1).expand(-1, hidden_states.shape[1]).contiguous()
```
anatyrova, line 1909: is .contiguous() necessary here? check if it's not important, remove it to avoid unnecessary memory copying

anatyrova: also would be good to add a comment explaining the expected tensor shapes, because hidden_states.shape[1] is fragile if the tensor layout changes

anatyrova:
@goyaladitya05 add those test cases pls:
```
def test_image_resize_behavior(self, model_arch: str):
    # Test that input images are properly resized when dimensions don't match

def test_num_frames_validation(self, model_arch: str):
    # Test error handling when num_frames > 1 is passed for image-to-video
```

optimum/intel/openvino/modeling_diffusion.py
goyaladitya05: .contiguous() is needed, expand() returns a stride-0 view (the broadcast dim has no backing memory), and OVModelPart.forward passes inputs with share_inputs=True, which wraps the tensor's buffer directly. Materializing it before the call keeps us off that path rather than relying on OV's handling of a non-contiguous buffer.

Also, this pattern is already used for the broadcast in model_patcher.py:7713-7729 (expand(...).contiguous()), and modeling_decoder.py:1538 similarly calls .contiguous() on input_ids

optimum/intel/openvino/modeling_diffusion.py
goyaladitya05: added

goyaladitya05:
Hi @anatyrova, I have addressed all the review comments, and added more tests cases.

anatyrova:
> Hi @anatyrova, I have addressed all the review comments, and added more tests cases.

great,thanks! overall LGTM

have you compared your video to the one from diffusers LTXImageToVideoPipeline? i mean doing an inference with original model and then with your pr's one. if it is visually very similar, then ok, i don't have any more comments.

goyaladitya05:
> great,thanks! overall LGTM
> 
> have you compared your video to the one from diffusers LTXImageToVideoPipeline? i mean doing an inference with original model and then with your pr's one. if it is visually very similar, then ok, i don't have any more comments.

Yes @anatyrova , they are very similar.

diffusers ->
ltx_i2v_diffusers

optimum (this PR) ->
ltx_i2v_ov_fp32

anatyrova (approved):

tests/openvino/test_diffusion.py
```diff
@@ -1206,3 +1207,194 @@ def test_static_shape_generation(self, model_arch):
...
+            "prompt": ["a dog running in a field"] * batch_size,
+            "negative_prompt": ["static, motionless"] * batch_size,
+        }
+
+        inputs["height"] = height
+        inputs["width"] = width
+
+        inputs["num_inference_steps"] = 2
+        inputs["num_frames"] = num_frames
+        inputs["guidance_scale"] = 4.5
```
echarlaix: why not
```suggestion
        inputs = {
            "image": _generate_images(height=height, width=width, batch_size=batch_size),
            "prompt": ["a dog running in a field"] * batch_size,
            "negative_prompt": ["static, motionless"] * batch_size,
            "height": height,
            "width": width,
            "num_inference_steps": 2,
            "num_frames": num_frames,
            "guidance_scale": 4.5,
        }
```

tests/openvino/test_diffusion.py
```diff
@@ -1206,3 +1207,194 @@ def test_static_shape_generation(self, model_arch):
...
+        ov_pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch], device=OPENVINO_DEVICE)
+
+        self.assertEqual(ov_pipeline.auto_model_class, auto_pipeline.__class__)
+
+    @parameterized.expand(SUPPORTED_ARCHITECTURES, skip_on_empty=True)
+    @require_diffusers
+    def test_num_videos_per_prompt(self, model_arch: str):
+        pipeline = self.OVMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch], device=OPENVINO_DEVICE)
+
+        for batch_size in [1]:
```
echarlaix: could make sense to add a test for batch_size > 1

echarlaix (approved):
Looks great, thanks for the addition @goyaladitya05

regisss:
It seems one test failed: https://github.com/huggingface/optimum-intel/actions/runs/29340825411/job/87879405687?pr=1762#step:11:483

tests/openvino/test_diffusion.py
goyaladitya05: done

tests/openvino/test_diffusion.py
goyaladitya05: added

goyaladitya05:
> It seems one test failed: https://github.com/huggingface/optimum-intel/actions/runs/29340825411/job/87879405687?pr=1762#step:11:483

fixed @regisss

goyaladitya05:
@regisss could you please re-run tests?

goyaladitya05:
Hi @regisss @anatyrova, the remaining test failures are unreleted to my changes. Is there any other blocker to merge this PR?

regisss (approved):

PR #1782, opened 2026-06-10
[OpenVINO] Fix for long audio inference for Qwen3-ASR

popovaan:
# What does this PR do?

PR fixes the inference of long audios for Qwen3-ASR.

Ticket: CVS-187744

## Before submitting
- This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case). N/A
- Did you make sure to update the documentation with your changes? N/A
- [x] Did you write any new necessary tests? - Updated test to cover long audio case.

### DIFF
--- a/optimum/intel/openvino/modeling_seq2seq.py
+++ b/optimum/intel/openvino/modeling_seq2seq.py
@@ -15,6 +15,8 @@ import logging
 import os
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Union
+from copy import deepcopy
+
 
 import numpy as np
 import openvino
@@ -66,6 +68,14 @@ else:
 
     transformers_auto_class = AutoModelForVision2Seq
 
+from ..utils import (
+    is_accelerate_available)
+if is_accelerate_available():
+    from accelerate.hooks import AlignDevicesHook, add_hook_to_module
+
+import inspect
+from torch.nn import functional as F
+from torch import nn
 
 core = Core()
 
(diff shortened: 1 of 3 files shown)


optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -871,6 +875,36 @@ def __init__(
...
+        chunk_list = input_features.T.split(chunk_lengths.tolist(), dim=0)
+        padded_feature = nn.utils.rnn.pad_sequence(chunk_list, batch_first=True).transpose(1, 2)
+        padded_feature = padded_feature.unsqueeze(1)
+
+        audio_features = []
+        inputs = {}
+        for idx, input_feature in enumerate(padded_feature):
+            inputs["input_features"] = input_feature
+
+            last_hidden_state = torch.from_numpy(self.request(inputs)["last_hidden_state"]).to(self.device)
```
as-suvorov: Can we infer padded_features in batch?

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -887,6 +921,16 @@ def forward(
         if "attention_mask" in self.input_names:
             inputs["attention_mask"] = attention_mask
 
+        # Qwen3-ASR requires input_features chunking before passing to encoder for processing of long audios.
+        if getattr(self.config, "model_type", None) == "qwen3_asr":
+            input_features = inputs["input_features"]
+            audio_features = []
+            for idx, input_feature in enumerate(input_features):
+                audio_feature = self.chanked_forward(input_feature, self.config.n_window)
```
as-suvorov: Can we infer input_features in batch?

optimum/intel/openvino/modeling_seq2seq.py
popovaan: Yes, done.

optimum/intel/openvino/modeling_seq2seq.py
popovaan: Encoder excepts 3-dim input, to support batching here we need to patch encoder to support 4-dimentional input.

as-suvorov: 1st encoder dimension is batch. Feel free to skip it as input_features batching can introduce huge number of batches.

rkazants (approved):

optimum/intel/openvino/modeling_seq2seq.py
popovaan: Currently it works more like (chunk_num, num_mel_bins, seq_len), so the first dimension is the number of chunks in a batch. I suppose we can flatten all chunks from all batches pass in one inference, but I'm not sure if that will work fine, as original implementation also processes each batch separately: https://github.com/QwenLM/Qwen3-ASR/blob/c17a131fe028b2e428b6e80a33d30bb4fa57b8df/qwen_asr/core/transformers_backend/modeling_qwen3_asr.py#L1124-L1130

There's by the way a comment that encoder doesn't support batch inference: https://github.com/QwenLM/Qwen3-ASR/blob/c17a131fe028b2e428b6e80a33d30bb4fa57b8df/qwen_asr/core/transformers_backend/modeling_qwen3_asr.py#L1122

as-suvorov: @popovaan Ok, thanks for details

popovaan:
@echarlaix could you please review this PR?

as-suvorov (approved):

PR #1787, opened 2026-06-11
Remove safetensors requirements 

echarlaix:


setup.py
```diff
@@ -28,7 +28,6 @@
 
 INSTALL_REQUIRE = [
     "torch>=2.1",
-    "safetensors<0.8.0",
```
echarlaix, line 31: incompatibility between safetensors v0.8.0 and nncf v3.2.0 https://github.com/openvinotoolkit/nncf/blob/v3.2.0/src/nncf/tensor/functions/numpy_io.py#L35

PR #1788, opened 2026-06-12
Add CohereAsrOpenVINOConfig for CohereLabs/cohere-transcribe-03-2026

mlukasze:
## Summary

Add OpenVINO export support for the `cohere_asr` model type, enabling inference of `CohereLabs/cohere-transcribe-03-2026` and compatible models via `OVModelForSpeechSeq2Seq`.

## Changes

- **model_configs.py**: Register `CohereAsrOpenVINOConfig` for the `cohere_asr` model type under the `automatic-speech-recognition` / `automatic-speech-recognition-with-past` tasks; extends `AudioToTextOpenVINOConfig` and handles nested encoder/decoder config extraction, plus a dynamic `length` encoder input (see Bugfix update below)
- **model_patcher.py**: `CohereAsrModelPatcher` fixes a `torch.jit.trace` export crash (positional-encoding materialization) and, as of the bugfix update below, a trace-time cache-length constant bug that broke the stateful with-past decoder
- **input_generators.py**: `CohereAsrDummyAudioInputGenerator` / `CohereAsrDummySeq2SeqDecoderTextInputGenerator` for correct dummy-input shapes
- **modeling_seq2seq.py**: `OVEncoder.forward()` accepts an optional `length` input (default: full `input_features` length) for architectures like `cohere_asr` whose feature extractor doesn't pad to a fixed frame count
- **docs/source/openvino/models.mdx**: Add `cohere_asr` to the supported architectures list
- **tests/openvino/test_seq2seq.py**: Add `cohere_asr` to `OVModelForSpeechSeq2SeqIntegrationTest.SUPPORTED_ARCHITECTURES` (requires transformers ≥ 4.57.0), plus end-to-end inference tests added in the bugfix update below
- **tests/openvino/test_export.py**, **test_exporters_cli.py**, **utils_tests.py**: Register tiny test model and op count fixtures for `cohere_asr`

## Export

```bash
pip install optimum[openvino] transformers>=4.57.0
```

```bash
optimum-cli export openvino \
  --model CohereLabs/cohere-transcribe-03-2026 \
  --task automatic-speech-recognition-with-past \
  --weight-format fp16 --trust-remote-code \
  ./cohere-transcribe-openvino
```

> **Note:** use `--task automatic-speech-recognition-with-past` (not the bare `automatic-speech-recognition`
> task) to get a cache-enabled, stateful decoder export -- see "Bugfix update" below. `--trust-remote-code`
> is required since `cohere_asr` ships custom modeling/processing code on the Hub.

## Inference

```python
import numpy as np
from optimum.intel import OVModelForSpeechSeq2Seq

model_id = "./cohere-transcribe-openvino"
model = OVModelForSpeechSeq2Seq.from_pretrained(model_id, trust_remote_code=True)

# 16 kHz mono audio as float32 numpy array (any length; no need to pad/truncate to 30s)
audio = np.zeros(16000, dtype=np.float32)
inputs = processor.feature_extractor(audio, return_tensors="pt")
predicted_ids = model.generate(**inputs)
transcription = processor.tokenizer.batch_decode(predicted_ids, skip_special_tokens=True)
print(transcription)
```

Tested on `CohereLabs/cohere-transcribe-03-2026`: CPU WER 1.20% vs reference 1.25% on 50 LibriSpeech clean samples.

---

## Bugfix update (follow-up to review feedback / real-inference testing)

Three bugs were found running real inference against the exported IR, following this PR's own
documented usage exactly (`optimum-cli export ... --task automatic-speech-recognition`, then the
Inference snippet above):

1. **Encoder input_features hardcoded to Whisper's static 3000-frame shape.** `CohereAsrOpenVINOConfig`
   extended `WhisperOpenVINOConfig` and inherited its dummy-input generator, forcing
   `input_features` to a static `nb_max_frames=3000` last dimension. The Hub remote code's
   `CohereAsrFeatureExtractor` does not pad to 3000 frames -- it pads only to the batch's natural
   max sample length and additionally returns a `length` tensor required by the Conformer
   encoder's internal masking. Fix: `CohereAsrOpenVINOConfig` now extends
   `AudioToTextOpenVINOConfig` directly, with a dynamic `length` input threaded through
   `OVEncoder.forward()`.

2. **Decoder exported without a KV-cache/stateful contract**, and the with-past export itself was
   broken. `automatic-speech-recognition-with-past` was already registered for `cohere_asr`, but
   exporting with that task crashed on real multi-step `generate()` -- root-caused to
   `_get_cache_seq_length()` in the model's remote code, which does
   `int(past_key_values.get_seq_length())`. The explicit `int()` cast bakes the KV-cache length as
   a `torch.jit.trace`-time constant, breaking dynamic decode-step shapes after the first step.
   Fix: `CohereAsrModelPatcher` now monkeypatches this free function for the duration of
   tracing/export only, keeping the cache length dynamic. With this fix,
   `--task automatic-speech-recognition-with-past` produces a genuinely stateful decoder
   (`beam_idx` input + KV-cache sinks) and multi-step `generate()` succeeds. The Export section
   above has been updated to use `-with-past` explicitly, since silently overriding a user's
   explicit `--task` choice isn't this project's convention.

3. **Exported dir allegedly not self-contained (tokenizer files).** Investigated and does **not**
   reproduce in the current environment/Hub revision: `AutoProcessor.from_pretrained(export_dir,
   trust_remote_code=True)` succeeds without any manual file copying, since `CohereAsrProcessor`'s
   tokenizer loads directly from `tokenizer.model` (sentencepiece), which `optimum-cli export`
   already copies. Added a regression-guard test. This does not appear to be a general
   `optimum-cli export openvino` limitation nor a `cohere_asr`-specific one at this time.

**New tests** added to `tests/openvino/test_seq2seq.py`
(`OVModelForSpeechSeq2SeqIntegrationTest`), exercising the actual inference call path (not just
export success):
- `test_cohere_asr_generate_non_30s_multiple_audio` -- `generate()` end-to-end on non-30s-multiple
  synthetic audio (3s/7s/11s), asserting dynamic encoder shapes and no crash.
- `test_cohere_asr_with_past_decoder_is_stateful` -- asserts the with-past decoder is stateful
  (`beam_idx` input + KV-cache sinks) and multi-step `generate()` succeeds.
- `test_cohere_asr_exported_processor_is_self_contained` -- asserts `AutoProcessor.from_pretrained`
  succeeds on a freshly exported dir without manual file copying.

**Real-model validation** (`CohereLabs/cohere-transcribe-03-2026`, `--task
automatic-speech-recognition-with-past`):
- Encoder inputs: `input_features=[?,?,?]`, `length=[?]` (previously `input_features=[?,?,3000]`
  static).
- Decoder: `input_ids` / `encoder_hidden_states` / `beam_idx`, 32 KV-cache sinks (previously not
  stateful, no `beam_idx`).
- `generate()` verified on 3s/7s/11s audio, on **CPU**, **GPU.0 (iGPU)**, and **GPU.1 (Arc Pro B60
  dGPU)** -- no shape-mismatch or cache errors on any device.

**WER regression check:** re-running the WER methodology referenced above against the *current*
Hub revision of `CohereLabs/cohere-transcribe-03-2026` surfaces an unrelated, pre-existing model
issue: the current revision's `.generate()` produces garbled/repetitive output regardless of audio
content (confirmed via a from-scratch manual eager-PyTorch decode loop, bypassing OpenVINO
entirely -- same garbled output). The original 1.20%/0.45% WER reference numbers were produced
against an older remote-code revision whose feature extractor returned different decoder-input
semantics (`decoder_input_ids`/`language`), which the current revision no longer honors -- this is
Hub-side model/remote-code drift, unrelated to and predating these export fixes. No WER
regression is attributable to this PR: the encoder-shape and decoder-statefulness fixes here do not
change the (pre-existing, orthogonal) output-quality issue in either direction. Filed for separate
follow-up investigation if needed.

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -9911,22 +9911,47 @@ class CohereAsrModelPatcher(OVSeq2SeqModelPatcher):
 
     The remote-code Conformer encoder's relative positional encoding
     (`RelPositionalEncoding` in `modeling_cohere_asr.py`) lazily materializes a `pe` buffer in
-    `_materialize_pe`, a method decorated with `torch._dynamo.disable(...)`. `torch.jit.trace`
-    (used by the OpenVINO exporter) re-invokes the traced module a second time to check that
-    the captured graph is deterministic (`check_trace`). The dynamo-disable wrapper hooks into
-    the same CPython frame-evaluation API used by the tracer, so the lazily-cached `pe` buffer
-    (and/or its recomputation) is not observed consistently across the two invocations,
-    producing constant nodes that differ (including differing dtypes for internal index
-    tensors), so `torch.jit.trace` aborts with:
+    `_materialize_pe`, a method decorated with `torch._dynamo.disable(...)`. The exact patched
+    code, from `modeling_cohere_asr.py` at Hub revision `b1eacc2686a3d08ceaae5f24a88b1d519620bc09`
+    of `CohereLabs/cohere-transcribe-03-2026` (line numbers as of that revision)::
+
+        @_dynamo_disable
+        def _materialize_pe(self, length: int, device: torch.device, dtype: torch.dtype):
+            needed_size = 2 * length - 1
+            if hasattr(self, "pe") and self.pe.size(1) >= needed_size:
+                if self.pe.device != device:
+                    self.pe = self.pe.to(device=device)
+                if self.pe.dtype != dtype:
+                    self.pe = self.pe.to(dtype=dtype)
+                return
+            effective_length = max(length, self.max_len)
+            positions = torch.arange(
+                effective_length - 1, -effective_length, -1, dtype=torch.float32, device=device
+            ).unsqueeze(1)
+            pe = self._create_pe(positions=positions, dtype=dtype)
+            if hasattr(self, "pe"):
+                self.pe = pe
+            else:
+                self.register_buffer("pe", pe, persistent=False)
+        # (modeling_cohere_asr.py, lines ~199-215)
+
+    `torch.jit.trace` (used by the OpenVINO exporter) re-invokes the traced module a second time
+    to check that the captured graph is deterministic (`check_trace`). The dynamo-disable
+    wrapper hooks into the same CPython frame-evaluation API used by the tracer, so the
+    lazily-cached `pe` buffer (and the `hasattr(self, "pe") and self.pe.size(1) >= needed_size`
+    early-return above) is not observed consistently across the two invocations, producing
+    constant nodes that differ (including differing dtypes for internal index tensors), so
+    `torch.jit.trace` aborts with:
 
         "Tensor-valued Constant nodes differed in value across invocations."
 
-    Fix: replace `_materialize_pe` with a dynamo-free implementation that *always* recomputes
-    the `pe` buffer from scratch (no lazy "already materialized?" caching, no
-    `torch._dynamo.disable`). Because `_create_pe` is a pure, deterministic function of
-    `length`/`device`/`dtype`, a fresh eager recompute on every call is guaranteed to produce
-    bit-identical constants across `torch.jit.trace`'s two check-trace invocations, regardless
-    of the actual sequence length seen at trace time (dummy inputs) vs. `pos_enc.max_len`.
+    Fix: replace `_materialize_pe` with a dynamo-free implementation (`_materialize_pe_eager`
+    below) that *always* recomputes the `pe` buffer from scratch (no lazy "already
+    materialized?" caching, no `torch._dynamo.disable`). Because `_create_pe` is a pure,
+    deterministic function of `length`/`device`/`dtype`, a fresh eager recompute on every call
+    is guaranteed to produce bit-identical constants across `torch.jit.trace`'s two check-trace
+    invocations, regardless of the actual sequence length seen at trace time (dummy inputs) vs.
+    `pos_enc.max_len`.
     """
 
     def __enter__(self):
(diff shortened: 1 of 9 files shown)


tests/openvino/test_seq2seq.py
```diff
@@ -444,6 +444,97 @@ def test_pipeline(self, model_arch):
         gc.collect()
 
 
+class CohereAsrTest(unittest.TestCase):
```
rkazants: use existing test suite for ASR, no need to implement a new one. Having common test suite gives confidence to check optimum-intel API

rkazants (changes requested):
need to reimplement tests and provide proper PR description with python code snippets how to set-up, export and infer

optimum/exporters/openvino/model_patcher.py
```diff
@@ -10980,6 +10980,79 @@ def __exit__(self, exc_type, exc_value, traceback):
             del self._model._orig_forward
 
 
+class CohereAsrModelPatcher(OVSeq2SeqModelPatcher):
+    """
+    Model patcher for CohereLabs/cohere-transcribe (Conformer encoder + Transformer decoder).
```
rkazants, line 9910: share code line that is patched

mlukasze:
Rebased \`fix/cohere-asr-with-tests\` onto latest \`main\` (which had since gained #1684 / "Add transformers v5.5 support", refactoring the architecture-support test lists from per-version \`if is_transformers_version(...)\` blocks to a single \`get_supported_model_for_library\`-based filter). This caused a conflict — resolved by:

- `tests/openvino/test_export.py` / `tests/openvino/test_exporters_cli.py`: added `cohere_asr` to the new unconditional architecture list/dict (matching the new pattern used for `qwen3_asr` and others), instead of reintroducing the now-obsolete version-conditional block.
- `optimum/intel/openvino/modeling_seq2seq.py`: dropped a stale duplicate `AutoModelForVision2Seq`/`AutoModelForImageTextToText` compatibility shim + duplicate `torch.nn` imports that were already superseded upstream (verified `AutoModelForImageTextToText` is imported and used directly elsewhere in the file; no functional change).

All 4 commits replayed cleanly; verified `cohere_asr` still resolves as a supported architecture under the new filter mechanism, and force-pushed (`f13d947e` → `28d8dbd3`). No functional changes to this PR's diff — only conflict resolution against upstream's own refactor.

optimum/exporters/openvino/model_patcher.py
mlukasze: Addressed — both docstrings now quote the exact patched source lines from `modeling_cohere_asr.py` at the pinned Hub revision (`b1eacc2686a3d08ceaae5f24a88b1d519620bc09`), with line references, instead of only describing the issue in prose. See commit bc5ea4ac.

mlukasze:
## Follow-up: investigated two user-reported issues

A user testing this PR reported two problems. I reproduced both in an isolated venv (`optimum-intel` from this branch, `transformers` 5.5.4, `openvino`/`openvino-genai` 2026.2.1) to determine root cause.

### 1. Export failure: `MatMul` dim mismatch on `encoder_decoder_proj` (1024 vs 1280) with `--task automatic-speech-recognition-with-past`

**Not reproducible on current branch HEAD** (`bc5ea4a` at the time of testing). Already fixed by `28d8dbd3` (`CohereAsrDummySeq2SeqDecoderTextInputGenerator` now derives the dummy hidden size from the encoder's own `d_model`). The user was very likely testing an intermediate revision of this PR before that fix landed, after upgrading only `optimum` (not `optimum-intel`). No action needed here — confirmed export succeeds end-to-end with `--weight-format fp16 --trust-remote-code`.

### 2. Inference crash: `AttributeError: 'NoneType' object has no attribute 'shape'` (`modeling_seq2seq.py:1006`, `input_ids.shape`)

**Reproduced exactly**, but only when going through `transformers.pipeline("automatic-speech-recognition", model=ov_model, ...)` (which is what the user's script does) — calling `model.generate(**inputs)` directly works fine.

Root cause: `AutomaticSpeechRecognitionPipeline.__init__` picks its internal mode via:
```python
if model.config.model_type == "whisper":
    self.type = "seq2seq_whisper"
elif model.__class__.__name__ in MODEL_FOR_SPEECH_SEQ_2_SEQ_MAPPING_NAMES.values():
    self.type = "seq2seq"
else:
    self.type = "ctc"
```
`MODEL_FOR_SPEECH_SEQ_2_SEQ_MAPPING_NAMES` only ever contains native PyTorch class names (e.g. `CohereAsrForConditionalGeneration`), never `OVModelForSpeechSeq2Seq`. Since `cohere_asr`'s `model_type` isn't `"whisper"`, the pipeline falls through to `"ctc"` and calls `model(**inputs)` directly instead of `.generate(...)`, so `decoder_input_ids` is never built → `None` → crash. This is a **generic, pre-existing optimum-intel gap**, not specific to `cohere_asr` — it affects *any* non-Whisper `OVModelForSpeechSeq2Seq` architecture (e.g. `qwen3_asr` too), and equally affects `optimum.intel.pipelines.pipeline()` since it just delegates to `transformers.pipeline()`.

**Fix pushed to this branch** (`a665334`): register `OVModelForSpeechSeq2Seq`'s class name into `MODEL_FOR_SPEECH_SEQ_2_SEQ_MAPPING_NAMES` once at module import, architecture-independent (benefits every current/future non-Whisper OV ASR model, not just `cohere_asr`). Added a fast regression test (`test_pipeline_autodetection_registers_non_whisper_architectures`) that doesn't require downloading a model.

### Validation performed
- Confirmed RED→GREEN on the new regression test.
- No regression: existing Whisper tests (`test_compare_to_transformers`) still pass.
- End-to-end (`optimum-cli export` → `transformers.pipeline("automatic-speech-recognition", ...)` with the real `CohereLabs/cohere-transcribe-03-2026` model) validated on **CPU, GPU.0, GPU.1** — pipeline correctly detects `type == "seq2seq"` and completes generation without crashing on all three.
- **NPU**: model compilation currently fails (`Upper bounds are not specified...` / non-broadcastable dims) — this is a known OpenVINO NPU compiler limitation around fully-dynamic input shapes in this encoder's conv frontend, unrelated to the two bugs above and to this fix; would need static/bounded reshaping to target NPU, tracked as a separate follow-up rather than blocking this PR.
- Also note: the existing `test_pipeline`/`test_compare_to_transformers` cohere_asr tests reference a `mlukasze/tiny-random-cohere-asr` fixture that doesn't yet exist on the Hub, so they've never actually executed in CI — flagging as a known gap; ran validation instead against the real gated model via `OPTIMUM_INTEL_COHERE_ASR_TEST_MODEL` env override.

### Side discovery (not addressed here, flagging for awareness)
`transformers` 5.5.4 now ships a **native** `cohere_asr` implementation (`transformers.models.cohere_asr.modeling_cohere_asr`) with a different forward contract (`attention_mask`/`output_attention_mask`) than the Hub's remote code (`length`). This PR's export config still targets the Hub remote-code interface, so `--trust-remote-code` remains necessary for now. Worth a follow-up look once the native implementation stabilizes.

mlukasze:
Full write-up: https://gist.github.com/mlukasze/3dd2060d09025b8a9116858d150da6d6

Issue #1785, opened 2026-06-11
Gemma-4 export broken in v2.0.0 — TorchScript tracing fails after ONNX path removal


PlanteAmigor:
**Model:** `google/gemma-4-E2B-it` (also likely affects other Gemma-4 variants)
**Task:** `image-text-to-text`
**Format:** `int4`

### Steps to reproduce

```bash
optimum-cli export openvino --model ./gemma/E2B --task image-text-to-text --weight-format int4 ./E2B-int4-ov
```

### Result

Export fails with TorchScript tracing error during vision encoder tracing:

```
RuntimeError: Couldn't get TorchScript module by tracing.
Exception: Tracing failed sanity checks!
encountered an exception while running the trace with test inputs.
Exception:
        The following operation failed in the TorchScript interpreter.
        Traceback of TorchScript (most recent call last):
        .../openvino/frontend/pytorch/patch_model.py(497): <lambda>
        .../openvino/frontend/pytorch/patch_model.py(49): new_forward
        .../transformers/models/gemma4/modeling_gemma4.py(185): forward
```

The full traceback involves `apply_rotary_pos_emb`, `q_norm`, and other Gemma-4 custom ops that fail under TorchScript tracing.

### Root cause

Gemma-4 support was added in #1688 when the ONNX export path (`export_pytorch_via_onnx`) was still available. PR #1753 removed the ONNX dependency and `export_pytorch_via_onnx`, now forcing all exports through PyTorch → TorchScript → OpenVINO. The TorchScript tracer cannot handle Gemma-4's custom operations (e.g., `apply_rotary_pos_emb_gemma4`, `q_norm`, rotary position embedding with custom `unsqueeze_dim` parameter).

The old export path (optimum-intel-1.27.0 with ONNX) works correctly for this model.

### Expected behavior

Gemma-4 should export successfully with `optimum-cli export openvino --task image-text-to-text`.

### Environment

- optimum-intel: 2.0.0
- transformers: 5.x
- openvino: 2026.2.0

PlanteAmigor:
Confirmed fixed by PR #1794. Tested with `optimum-intel==2.1.0.dev0+cd89a25` (PR #1794 branch) + `transformers==5.12.1`:

 ```
 optimum-cli export openvino --model ./gemma/E2B/ --task image-text-to-text --weight-format int4 ./E2B-int4-ov/
 ```

 Export completed successfully. NNCF quantization stats:
[...]

PR #1794, opened 2026-06-16
Fix Gemma4 shared KV export compatibility for transformer 5.12.0 with older support as well

dhandhalyabhavik:
# What does this PR do?

ov export for gemma4 models failed with recent transformers version 5.12.0 because it has changed internal model implementation of Gemma4. They started using `shared_kv_states` along with `past_key_values`. Older 5.5.0 implementation of gemma4 had only `past_key_values` with `shared_kv_states` inside it.

This patch adds support for latest transformers as well as old transformers.

Fixes # (issue)
#1785 

## Before submitting

### Reference links
https://github.com/huggingface/transformers/blob/v5.12.1/src/transformers/models/gemma4/modeling_gemma4.py#L1234

### DIFF
--- a/optimum/exporters/openvino/model_configs.py
+++ b/optimum/exporters/openvino/model_configs.py
@@ -1293,6 +1293,7 @@ class Gemma4TextOpenVINOConfig(Gemma3TextOpenVINOConfig):
     DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, Gemma4DummyPastKeyValuesGenerator)
     DUMMY_PKV_GENERATOR_CLASS = Gemma4DummyPastKeyValuesGenerator
     MIN_TRANSFORMERS_VERSION = "5.5"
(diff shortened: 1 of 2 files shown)


rkazants (approved):

rkazants (commented):
please set `MAX_TRANSFORMERS_VERSION=5.12.99` here https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_configs.py#L1289

the same we need here https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_configs.py#L3844 both min and max version. It helps to avoid bug issue report from the community

rkazants (approved):

rkazants (changes requested):
not needed PR because PR with transformers v5.14 support is being prepared with generic approach for models support: https://github.com/huggingface/optimum-intel/pull/1886

PR #1801, opened 2026-06-18
[OpenVINO] Support FunAudioLLM/Fun-ASR Model

openvino-agent:
# What does this PR do?

Installation:

```sh
pip install transformers==4.57.6
pip install funasr
pip install -U git+https://github.com/openvino-agent/optimum-intel.git@worktree-funasr
pip install torchaudio
```

Export cmd-line:

```sh
optimum-cli export openvino -m FunAudioLLM/Fun-ASR-Nano-2512 Fun-ASR-Nano-2512 --trust-remote-code
```

Inference:

```python
import json
import logging
import os
import warnings

import soundfile as sf

from optimum.intel import OVModelForSpeechSeq2Seq

MODEL_ID = "FunAudioLLM/Fun-ASR-Nano-2512"
OV_DIR = "./Fun-ASR-Nano-2512"

# Load the model to OpenVINO format (tokenizer IR is exported for fun_asr) ---
ov_model = OVModelForSpeechSeq2Seq.from_pretrained(OV_DIR, trust_remote_code=True, device="CPU")

# Decode generated ids back to text with the bundled Qwen3 tokenizer files.
from transformers import AutoTokenizer  # noqa: E402
from huggingface_hub import snapshot_download  # noqa: E402

model_path = snapshot_download(MODEL_ID, allow_patterns=["Qwen3-0.6B/*", "example/*"])
tokenizer = AutoTokenizer.from_pretrained(os.path.join(model_path, "Qwen3-0.6B"))
example_dir = os.path.join(model_path, "example")

# Bundled example clips, keyed by language. The `language` value is the FunASR prompt language tag
# (a Chinese word interpolated into the prompt), per the model card usage example.
# https://huggingface.co/FunAudioLLM/Fun-ASR-Nano-2512/tree/main/example
EXAMPLE_WAVS = {
    "zh": {"wav": os.path.join(example_dir, "zh.mp3"), "language": "中文"},
    "en": {"wav": os.path.join(example_dir, "en.mp3"), "language": "英文"},
    "ja": {"wav": os.path.join(example_dir, "ja.mp3"), "language": "日文"},
    "ko": {"wav": os.path.join(example_dir, "ko.mp3"), "language": "韩文"},
    "yue": {"wav": os.path.join(example_dir, "yue.mp3"), "language": "粤语"},
}

# Reference transcriptions produced by the original FunASR PyTorch pipeline, used to validate the
# OpenVINO output. Generated once with `funasr.AutoModel(...).generate(...)` and dumped to JSON.
REFERENCES_FILE = "funasr_references.json"
REFERENCES = {}
if os.path.exists(REFERENCES_FILE):
    with open(REFERENCES_FILE, encoding="utf-8") as f:
        REFERENCES = {lang: item["text"] for lang, item in json.load(f).items()}

# --- Run the OpenVINO pipeline for each language using preprocess_input ---
num_match = 0
for lang, item in EXAMPLE_WAVS.items():
    waveform, sampling_rate = sf.read(item["wav"], dtype="float32")
    inputs = ov_model.preprocess_input(waveform, sampling_rate, language=item["language"])

    generated_ids = ov_model.generate(**inputs, max_new_tokens=256)

    # Decode only the generated tokens (skip the prompt).
    prompt_len = inputs["decoder_input_ids"].shape[1]
    transcription = tokenizer.batch_decode(generated_ids[:, prompt_len:], skip_special_tokens=True)[0]

    print(f"\n[{lang}] input_features={tuple(inputs['input_features'].shape)}")
    print(f"  OpenVINO  : {transcription!r}")
```

Fixes # (issue)

## Before submitting


regisss:
When running the script written in the first post, I get:
```
The attention mask is not set and cannot be inferred from input because pad token is same as eos token. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.

[zh] input_features=(1, 94, 560)
  OpenVINO  : '!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!'

[en] input_features=(1, 120, 560)
  OpenVINO  : '!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!'

[ja] input_features=(1, 120, 560)
  OpenVINO  : '!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!'

[ko] input_features=(1, 77, 560)
  OpenVINO  : '!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!'

[yue] input_features=(1, 86, 560)
  OpenVINO  : '!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!'
```

Are these the expected outputs?

LauraGPT:
No, the repeated `!` output is not expected for these examples. The OpenVINO path should decode to the same ASR text as the original FunASR pipeline, not punctuation-only strings.

From the current patch, I would focus the debugging on the standalone preprocessing / generation handoff rather than treating this as a valid model output:

- The test added in `tests/openvino/test_seq2seq.py` compares OV generation against the original FunASR output only after feeding `input_features` and `decoder_input_ids` captured from `funasr_model.model.inference_prepare`. That validates the decoder/export path for exact FunASR-prepared tensors.
- The script in the PR body exercises a different path: `ov_model.preprocess_input(...)` and then `ov_model.generate(**inputs)`. That path is not currently end-to-end compared against the FunASR reference output; shape equality for `input_features` is not enough.
- In `_preprocess_input_funasr`, `decoder_input_ids` are padded with `0`, while the wrapper also sets `audio_token_id = 0`. For batched or padded prompts this can create extra audio placeholder positions unless a decoder mask is carried through and/or padding uses the real `pad_token_id`.
- In `prepare_inputs_for_generation`, `decoder_attention_mask` is created but not returned in the dict, so it is dropped before `forward`. That also lines up with the warning about missing attention masks when `pad_token_id == eos_token_id`.

Suggested next check/fix: add a small test that runs the exact public API from the PR body for at least `example/zh.mp3`:

```python
inputs = ov_model.preprocess_input(waveform, sampling_rate, language="中文")
generated_ids = ov_model.generate(**inputs, max_new_tokens=64)
```

and compare the decoded text to the original `funasr.AutoModel(...).generate(...)` output. If that fails while the captured-input test passes, the bug is almost certainly in `_preprocess_input_funasr` feature values and/or the masks passed from `preprocess_input` into `generate`, not in the expected transcript.

LauraGPT:
I prepared a small follow-up branch with the mask handoff fix I mentioned above:

https://github.com/LauraGPT/optimum-intel/tree/funasr/optimum-1801-mask-fix

Commit: https://github.com/LauraGPT/optimum-intel/commit/56654e61b178de5789a07bd58dbb61d4982a7be5

What it changes:

- `preprocess_input()` now returns both `attention_mask` for padded audio features and `decoder_attention_mask` for the FunASR chat prompt.
- Prompt padding now uses the configured `pad_token_id` / `eos_token_id` instead of `0`, because `0` is also the FunASR audio placeholder token id.
- `prepare_inputs_for_generation()` now keeps `decoder_attention_mask` in the returned model inputs instead of dropping it.
- `FunASROpenVINOConfig` exports the decoder mask input (`decoder_attention_mask`, mapped to the OV decoder `attention_mask`) so the IR can actually consume the mask.
- The FunASR test now asserts the standalone preprocessing masks and that the generation-prep path preserves `decoder_attention_mask`.

Local verification I could run in this lightweight environment:

```text
git diff --check
python -m py_compile optimum/intel/openvino/modeling_seq2seq.py optimum/exporters/openvino/model_configs.py tests/openvino/test_seq2seq.py
```

Both passed. I could not run the end-to-end OpenVINO/FunASR test locally because this environment is missing `openvino`, `torch`, `funasr`, `torchaudio`, and `soundfile`; `pytest tests/openvino/test_seq2seq.py::FunASRTest::test_compare_to_transformers -q` fails at collection with `ModuleNotFoundError: No module named 'openvino'`.

If you can run the PR-body script again after cherry-picking this commit, the important check is whether the warning disappears and whether `example/zh.mp3` decodes to the same text as the original `funasr.AutoModel(...).generate(...)` reference instead of repeated `!`.

LauraGPT:
Follow-up on the FunASR OpenVINO repeated-`!` decoding issue: I opened a PR directly against the contributor branch so the mask handoff fix can be merged into this PR without manual cherry-picking:

https://github.com/openvino-agent/optimum-intel/pull/4

The patch is the same focused fix as the branch above: preserve audio and decoder masks through standalone preprocessing/generation, avoid padding FunASR prompt ids with the audio placeholder token id, and export the decoder mask input. The PR is currently open and mergeable against `worktree-funasr`.

regisss:
The code quality check failed, please run `make style`

tests/openvino/test_seq2seq.py
```diff
@@ -546,6 +547,104 @@ def test_compare_to_transformers(self, model_arch):
         gc.collect()
 
 
+class FunASRTest(unittest.TestCase):
```
IlyasMoutawwakil: please move to a seperate file

optimum/intel/utils/modeling_utils.py
```diff
@@ -499,3 +533,112 @@ def from_pretrained(
         model.config = config
 
         return model
+
+
+class _FunASRAudioEncoder(torch.nn.Module):
```
IlyasMoutawwakil: i'd prefer keep third party library logic in separate files, we can have a modeling_funasr.py file

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1304,15 +1321,249 @@ class OVModelForSpeechSeq2Seq(OVModelForSeq2SeqLM):
...
+    def from_pretrained(cls, model_id, export: bool = False, config: Optional["PretrainedConfig"] = None, **kwargs):
+        # FunASR models (e.g. Fun-ASR-Nano) are loaded via the funasr library and have no root
+        # config.json; transformers' AutoConfig / optimum's library inference cannot handle them.
+        # We detect them (either the original funasr repo, or an already-exported OpenVINO dir whose
+        # config.json carries `export_model_type == "fun_asr"`) and load the config ourselves,
+        # bypassing the standard inference path.
+        if config is None and cls._is_funasr_source(model_id, **kwargs):
+            return cls._from_pretrained_funasr(model_id, export=export, **kwargs)
+
+        return super().from_pretrained(model_id, export=export, config=config, **kwargs)
```
IlyasMoutawwakil: so hmm OVModelForSpeechSeq2Seq is a replacement class for AutoModelForSpeechSeq2Seq, i don't think it makes sense to push a third party lib in the same path

IlyasMoutawwakil (changes requested):
let's move the third party lib logic and special handling outside of the transformers logic path

tests/openvino/test_seq2seq.py
openvino-agent: created common test suite

optimum/intel/utils/modeling_utils.py
openvino-agent: done

optimum/exporters/openvino/model_patcher.py
```diff
@@ -10302,6 +10302,175 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+        model.forward = patched_decoder_forward
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+
+        if hasattr(self._model, "_orig_forward"):
+            self._model.forward = self._model._orig_forward
+            del self._model._orig_forward
+
```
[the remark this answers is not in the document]
openvino-agent: not needed

optimum/intel/openvino/modeling_seq2seq.py
openvino-agent: Done

openvino-agent:
> let's move the third party lib logic and special handling outside of the transformers logic path

Done, please review again

optimum/exporters/openvino/model_patcher.py
```diff
@@ -10497,6 +10497,175 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+                    scores = torch.matmul(q_h, k_h.transpose(-2, -1))
+                    attn_weights = torch.softmax(scores, dim=-1)
+                    out = torch.matmul(attn_weights, v_h)
+                    n_batch = out.size(0)
+                    out = out.transpose(1, 2).contiguous().view(n_batch, -1, att.h * att.d_k)
+                    return att.linear_out(out) + fsmn_memory
+
+                return patched_sanm_forward
+
+            attn.forward = make_sanm_forward(attn)
```
IlyasMoutawwakil: why do you use a function to create the patched forward ? it's the same forward if you added self (attn) to th signature and bound it to the object

optimum/exporters/openvino/model_patcher.py
```diff
@@ -10497,6 +10497,175 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+                        scores = torch.matmul(q, k.transpose(-2, -1)) / (att.d_k**0.5)
+                        attn_weights = torch.softmax(scores, dim=-1)
+                        x = torch.matmul(attn_weights, v)
+                        n_batch = x.size(0)
+                        x = x.transpose(1, 2).contiguous().view(n_batch, -1, att.h * att.d_k)
+                        return att.linear_out(x)
+
+                    return patched_adaptor_forward
+
+                attn.forward = make_adaptor_forward(attn)
```
IlyasMoutawwakil: same remark

optimum/exporters/openvino/model_patcher.py
```diff
@@ -10497,6 +10497,175 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        encoder_wrap.forward = patched_encoder_forward
+
+    def _patch_decoder(self):
+        # self._model is the _FunASRForSpeechSeq2Seq wrapper
+        model = self._model
+        llm = model.llm
+        audio_token_id = getattr(model.config, "audio_token_id", 0)
+
+        # Force eager attention for stable OpenVINO tracing.
+        llm.config._attn_implementation = "eager"
```
IlyasMoutawwakil: please use the public api (set_attn_impl)

IlyasMoutawwakil: so the llm is a transformers model ? 🤔

optimum/exporters/openvino/model_patcher.py
```diff
@@ -10497,6 +10497,175 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            return tuple(flat_output)
+
+        model.forward = patched_decoder_forward
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
+
+        if hasattr(self._model, "_orig_forward"):
+            self._model.forward = self._model._orig_forward
+            del self._model._orig_forward
```
IlyasMoutawwakil, line 9885: what about the other patched components ?

optimum/intel/openvino/modeling_funasr.py
```diff
@@ -0,0 +1,469 @@
...
+
+    Structure: WavFrontend (fbank features, handled by the processor) -> audio encoder (SenseVoice) ->
+    audio adaptor -> spliced into the Qwen3 LLM input embeddings at audio placeholder positions -> Qwen3 LLM.
+
+    The wrapper exposes a transformers-style interface (`get_encoder`, `config.is_encoder_decoder`, an
+    `audio_token_id` placeholder marker) so it flows through the standard speech-seq2seq export path.
+    """
+
+    # Audio placeholder token id used by FunASR (zero-id tokens are inserted as audio slots).
+    AUDIO_TOKEN_ID = 0
```
IlyasMoutawwakil: why not in config ?

optimum/intel/openvino/modeling_funasr.py
```diff
@@ -0,0 +1,469 @@
...
+            return getattr(cfg, "export_model_type", None) == "fun_asr"
+        except Exception:
+            return False
+    return False
+
+
+def _get_ov_model_for_funasr_class():
+    """Deferred class construction to avoid circular import with modeling_seq2seq."""
+    from .modeling_seq2seq import OVModelForSpeechSeq2Seq
+
```
IlyasMoutawwakil: there's no circular import as long as import from this file are scoped (which they are)

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1321,11 +1344,9 @@ def _prepare_decoder_input_ids_for_generation(
                 decoder_input_ids = None
 
             if decoder_input_ids is None:
-                # Fallback to default behavior if no decoder_input_ids provided
                 return super()._prepare_decoder_input_ids_for_generation(
                     batch_size, model_input_name, model_kwargs, decoder_start_token_id, device
                 )
-            # Return decoder_input_ids as-is without prepending decoder_start_token_id
```
IlyasMoutawwakil, line 1331: not sure why these comments were removed

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1467,14 +1488,16 @@ def _from_pretrained(
         config: "PretrainedConfig",
         **kwargs,
     ):
-        if "WhisperForConditionalGeneration" in getattr(config, "architectures", []):
+        if "WhisperForConditionalGeneration" in (getattr(config, "architectures", None) or []):
             return _OVModelForWhisper._from_pretrained(model_id, config, **kwargs)
-        elif getattr(config, "model_type", None) == "qwen3_asr":
-            # Ensure is_encoder_decoder is set for proper model loading
+        if getattr(config, "model_type", None) == "fun_asr":
+            from .modeling_funasr import _get_funasr_class
```
IlyasMoutawwakil: you can import here the class directly, no need for a deferred constructor

optimum/intel/utils/modeling_utils.py
```diff
@@ -172,13 +172,17 @@ def _infer_library_from_model_name_or_path(
     cache_dir: str = HUGGINGFACE_HUB_CACHE,
     token: Optional[Union[bool, str]] = None,
 ):
+    from ..openvino.modeling_funasr import _is_funasr_model
```
IlyasMoutawwakil, line 175: remove and import at the top

tests/openvino/test_seq2seq.py
```diff
@@ -449,13 +449,13 @@ def test_pipeline(self, model_arch):
         gc.collect()
 
 
-class Qwen3ASRTest(unittest.TestCase):
+class OVASRTest(unittest.TestCase):
     """
-    Test Qwen3ASR model type.
-    Compares OpenVINO model output to original PyTorch transformers model output.
+    Test ASR model types (Qwen3-ASR, FunASR).
+    Compares OpenVINO model output to original PyTorch model output.
```
IlyasMoutawwakil: in a separate file please, with their own ci testing group, that way third party libraries also don't fail because of their transformers version constraints

tests/openvino/test_seq2seq.py
```diff
@@ -471,80 +471,158 @@ def _generate_audio_data(self):
...
         gc.collect()
 
+    def _get_pt_reference(self, model_arch):
+        """
+        Obtain PyTorch reference: input_features, decoder_input_ids, pt_text, gen_kwargs, decode_fn.
+        Returns a dict with keys consumed by _test_common_asr.
+        """
+        model_id = MODEL_NAMES[model_arch]
+
+        if model_arch == "fun_asr":
```
IlyasMoutawwakil: there are two model types tested here and each has its own single mega branch 😅

IlyasMoutawwakil (changes requested):
please be careful with ai generated code and review it thoroughly, llms have tunnel vision when implementing so they're mostly blind to the bigger picture and ignore a library's code style and practices, etc.

LauraGPT:
I took a quick pass over the failing check logs on `c70504e`.

The current failures do not look like a single Fun-ASR adapter/style failure:

- `build (*seq2seq*, 4.57.6)`: `OVModelForPix2StructIntegrationTest::test_compare_to_transformers_0_pix2struct` fails with `AssertionError: False is not true`.
- `build (*quantization*, 4.57.6)`: several image-text quantization cases fail with `PIL.UnidentifiedImageError: cannot identify image file`.
- `build (*export*, 4.57.6)`: multiple `image-text-to-text` export CLI cases fail through `CalledProcessError`.
- An older `build` run also had `HF_HUB_OFFLINE` + incomplete cached snapshots for `katuni4ka/*` and `echarlaix/t5-small-openvino`, which looks like a CI/cache issue rather than this PR's Fun-ASR logic.

So the actionable next split seems to be: rerun/refresh the offline-cache failure, then triage the remaining Pix2Struct and image-text quantization/export regressions separately.

LauraGPT:
Follow-up from the FunASR side: I refreshed the small source-branch PR on the current `worktree-funasr` head:

https://github.com/openvino-agent/optimum-intel/pull/5

Its scope is now limited to the two still-open review requests here: FunASR and Qwen3-ASR are moved from `test_seq2seq.py` into separate test files, each with its own regular/nightly CI group and model-specific dependency installation. The refreshed head is `2e5f882`; GitHub reports it mergeable and clean.

Both end-to-end tests pass locally on Python 3.10, Transformers 4.57.6, and OpenVINO 2026.2.1, including PyTorch-reference comparison and OpenVINO export. Ruff, Black, Python compilation, YAML parsing, and `git diff --check` also pass.

LauraGPT:
Small follow-up on the review-structure PR: openvino-agent/optimum-intel#5 has been updated to `ad7d6c576d989f4191f029268125c4568b51075d` with the project-pinned quality tools applied (`black~=23.1`, `ruff==0.4.4`).

The touched FunASR/OpenVINO files now pass focused `black --check`, `ruff check`, `py_compile`, and `git diff --check` in my local verification. That PR is still clean and waiting to be merged into `openvino-agent:worktree-funasr`; after it lands, #1801 should pick up the review-structure cleanup and the CI can rerun on the updated source branch.

LauraGPT:
I opened a draft mirror PR at #1856 from `LauraGPT:codex/funasr-review-cleanup` so there is a branch I can keep updating directly while this original review thread remains the source of context.

It includes the review-cleanup work from openvino-agent/optimum-intel#5 plus one extra line-ending cleanup so `git diff --check huggingface/main...HEAD` is clean. No need to switch review targets unless maintainers prefer the directly updateable branch.

I also sampled the current failing Actions logs here; the representative failures I saw were unrelated to the FunASR diff (`HF_HUB_OFFLINE` incomplete cached snapshots and existing image-text quantization jobs failing with `PIL.UnidentifiedImageError`).

optimum/exporters/openvino/model_patcher.py
openvino-agent: Simplified patching

optimum/exporters/openvino/model_patcher.py
openvino-agent: done

optimum/exporters/openvino/model_patcher.py
openvino-agent: done. Yes, this is qwen3 from transformers

optimum/exporters/openvino/model_patcher.py
rkazants: Done

optimum/intel/openvino/modeling_funasr.py
rkazants: done

optimum/intel/openvino/modeling_funasr.py
rkazants: Done

optimum/intel/openvino/modeling_seq2seq.py
rkazants: Done

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1304,13 +1321,22 @@ class OVModelForSpeechSeq2Seq(OVModelForSeq2SeqLM):
...
+            return _OVModelForFunAsr._from_pretrained_funasr(model_id, export=export, **kwargs)
+
+        return super().from_pretrained(model_id, export=export, config=config, **kwargs)
+
     def _prepare_decoder_input_ids_for_generation(
         self, batch_size, model_input_name, model_kwargs, decoder_start_token_id, device=None
     ):
         """
         For qwen3_asr: skip prepending decoder_start_token_id since the full prompt
         (including chat template tokens) is already provided as decoder_input_ids.
```
rkazants, line 1339:
```suggestion
        (including chat template tokens) is already provided as decoder_input_ids.
        This matches the PyTorch model behavior where input_ids is used as-is.
```

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1321,11 +1348,9 @@ def _prepare_decoder_input_ids_for_generation(
                 decoder_input_ids = None
 
             if decoder_input_ids is None:
```
rkazants, line 1350:
```suggestion
            if decoder_input_ids is None:
                # Fallback to default behavior if no decoder_input_ids provided
```

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1325,7 +1352,6 @@ def _prepare_decoder_input_ids_for_generation(
                 return super()._prepare_decoder_input_ids_for_generation(
                     batch_size, model_input_name, model_kwargs, decoder_start_token_id, device
                 )
```
rkazants, line 1354:
```suggestion
                )
            # Return decoder_input_ids as-is without prepending decoder_start_token_id
```

optimum/intel/openvino/modeling_seq2seq.py
rkazants: reverted

optimum/intel/utils/modeling_utils.py
rkazants: not possible, circular import

tests/openvino/test_seq2seq.py
openvino-agent: Done

LauraGPT:
Current CI triage for head `5e1da4a052c5`:

- 33 checks pass, 2 are skipped, and 2 fail.
- The FunASR/Qwen remote-code job, `build (*asr*)`, passes.
- `build (*seq2seq*)` fails in the unrelated Gemma4 comparison after upgrading to Transformers 5.5.0 and the OpenVINO 2026.4 nightly. Python exits with SIGSEGV in `test_seq2seq.py::...test_compare_to_transformers`; no FunASR test is involved.
- `build (*quantization*, latest)` fails two existing quantization cases with `KeyError: instruction` at `optimum/intel/openvino/quantization.py:849`; no FunASR test is involved.
- All 24 review threads are currently resolved.

The current red checks therefore do not indicate a FunASR regression. A maintainer rerun once the nightly/dataset failures are stable should be sufficient; I do not see a justified FunASR code change from these logs.

LauraGPT:
One small source-branch cleanup from reviewing the current `5e1da4a052c5` head: `FunASROpenVINOConfig` assigns `MIN_TRANSFORMERS_VERSION` twice (`4.57.0`, then `4.57.6`).

I opened [openvino-agent/optimum-intel#7](https://github.com/openvino-agent/optimum-intel/pull/7) directly against the current `worktree-funasr` head. It removes only the stale `4.57.0` assignment and preserves `4.57.6`, matching the dedicated FunASR CI dependency. GitHub reports it clean/mergeable: 1 commit, 1 file, 1 deletion.

Validated with an AST single-assignment assertion, `py_compile`, Black, project-pinned Ruff 0.4.4, and `git diff --check`.

The two upstream red jobs remain the unrelated Gemma4 nightly SIGSEGV and existing quantization dataset `KeyError: instruction` described above; this cleanup does not claim to fix or mask them.

LauraGPT:
I resolved the five current merge conflicts against `huggingface/main` (`037438e`) in openvino-agent/optimum-intel#8. The source PR is clean/mergeable and also folds in the earlier #6/#7 cleanups.

CI-equivalent local verification on Python 3.10 / Transformers 4.57.6 / OpenVINO 2026.2.1:
- exact remote-code export command: 3 passed
- exact remote-code ASR command: 2 passed
- exact remote-code quantization command: 4 passed
- project-pinned Black/Ruff checks passed

The rebase testing also found and fixed a flaky cache resolver that could select a one-file PR snapshot instead of the cached `main` revision. Once source PR #8 is merged into `worktree-funasr`, this PR should update to the tested head `cff702862264bcc5128bbad0ae0cfb166cd8fe0b`; I will recheck mergeability and CI immediately.

tests/openvino/test_asr.py
```diff
@@ -0,0 +1,203 @@
...
+        duration = 120
+        t = np.linspace(0, 1.0, sample_rate * duration, endpoint=False)
+        audio_data = (0.5 * np.sin(2 * np.pi * 440 * t)).astype(np.float32)
+        return audio_data, sample_rate
+
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    @pytest.mark.skipif(
+        is_transformers_version("!=", "4.57.6"),
+        reason="requires transformers==4.57.6.",
+    )
```
IlyasMoutawwakil, line 49: please give a proper reason as in why it isn't supported

tests/openvino/test_asr.py
```diff
@@ -0,0 +1,203 @@
...
+        def _capture(*args, **kw):
+            inputs_embeds, contents, batch, source_ids, meta = orig_prepare(*args, **kw)
+            captured["speech"] = batch["speech"]
+            captured["source_ids"] = source_ids
+            return inputs_embeds, contents, batch, source_ids, meta
+
+        gen_kwargs = {"max_new_tokens": 64}
+
+        core.inference_prepare = _capture
+        with redirect_stdout(buf), redirect_stderr(buf):
```
IlyasMoutawwakil, line 130: why the redirect ?

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1467,14 +1494,16 @@ def _from_pretrained(
...
     ):
-        if "WhisperForConditionalGeneration" in getattr(config, "architectures", []):
+        if "WhisperForConditionalGeneration" in (getattr(config, "architectures", None) or []):
             return _OVModelForWhisper._from_pretrained(model_id, config, **kwargs)
-        elif getattr(config, "model_type", None) == "qwen3_asr":
-            # Ensure is_encoder_decoder is set for proper model loading
+        if getattr(config, "model_type", None) == "fun_asr":
+            from .modeling_funasr import _OVModelForFunAsr
+
             config.is_encoder_decoder = True
```
IlyasMoutawwakil, line 1502: why not return the right class here instead of overriding the public from_pretrained

optimum/intel/openvino/modeling_seq2seq.py
```diff
@@ -1304,6 +1321,16 @@ class OVModelForSpeechSeq2Seq(OVModelForSeq2SeqLM):
...
 
+    @classmethod
+    def from_pretrained(cls, model_id, export: bool = False, config: Optional["PretrainedConfig"] = None, **kwargs):
+        from .modeling_funasr import _is_funasr_source, _OVModelForFunAsr
+
+        # the original FunASR model has no config file, so _from_pretrained() dispatch does not work
+        if config is None and _is_funasr_source(model_id, **kwargs):
+            return _OVModelForFunAsr._from_pretrained_funasr(model_id, export=export, **kwargs)
+
+        return super().from_pretrained(model_id, export=export, config=config, **kwargs)
```
IlyasMoutawwakil, line 1332: see comment above

optimum/exporters/openvino/model_configs.py
```diff
@@ -4030,6 +4049,72 @@ def add_past_key_values(self, inputs_or_outputs: Dict[str, Dict[int, str]], dire
...
+        decoder_num_layers="num_hidden_layers",
+        num_attention_heads="num_attention_heads",
+        # Use num_key_value_heads for KV cache shape generation (GQA)
+        decoder_num_attention_heads="num_key_value_heads",
+        feature_size="num_mel_bins",
+        allow_new=True,
+    )
+    _MODEL_PATCHER = FunASRModelPatcher
+    MIN_TRANSFORMERS_VERSION = "4.57.0"
+    MIN_TRANSFORMERS_VERSION = "4.57.6"
```
IlyasMoutawwakil: did you mean max ?

IlyasMoutawwakil (approved):
just a couple nits, otherwise LGTM

optimum/intel/openvino/modeling_seq2seq.py
LauraGPT: `_from_pretrained` can dispatch an already-exported FunASR IR because that directory has the generated `config.json`. The original FunASR Hub source is different: it has `config.yaml` plus `configuration.json`, but no root `config.json`, so Optimum/AutoConfig fails before `_from_pretrained` is reached. The public override is currently only the pre-config source detector; exported models still take this `_from_pretrained` branch. Would you prefer a dedicated public `OVModelForFunASR` entry point for the no-config source and removal of `OVModelForSpeechSeq2Seq.from_pretrained(source)` support, or should the narrow pre-config dispatch stay on the replacement class?

optimum/intel/openvino/modeling_seq2seq.py
LauraGPT: This is the pre-config half of the same dispatch question. The source repo has no root `config.json`, so the standard loader cannot reach `_from_pretrained`; I have asked in the thread above whether you prefer a dedicated public FunASR class or this narrow source detector on `OVModelForSpeechSeq2Seq`.

tests/openvino/test_asr.py
LauraGPT: This is a shared Qwen3-ASR/FunASR remote-code parity test. Qwen3-ASR is registered only for Transformers 4.57.6 (`MIN` and `MAX` are both 4.57.6), and the dedicated remote-code workflow pins that version while the latest-version matrix excludes these cases. I will replace the generic reason with that concrete compatibility/CI reason.

tests/openvino/test_asr.py
LauraGPT: The FunASR loader writes per-tensor checkpoint/progress output to both stdout and stderr. These redirects keep the parity-test log readable; they do not catch exceptions, so load failures still fail the test. I will make that reason explicit next to the redirect.

optimum/exporters/openvino/model_configs.py
LauraGPT: Yes. I checked the introducing commit (`4dee07f`, titled `Set min and max versions`): the second assignment is a typo and should be `MAX_TRANSFORMERS_VERSION = "4.57.6"`, preserving the intended supported range from 4.57.0 through 4.57.6. I will correct that in the source-branch update.

LauraGPT:
The clear review fixes are ready in the source-branch PR https://github.com/openvino-agent/optimum-intel/pull/9, based directly on the current f3ce062 head. While validating, I also found that moving the export config to the separate funasr library made the export and quantization filters silently drop all FunASR cases; #9 restores those cases.

Local CI-equivalent validation on Python 3.10 / Transformers 4.57.6 / OpenVINO 2026.2.1 is green: export 2/2, ASR parity 2/2, and quantization 4/4, plus Black 23.x, Ruff 0.4.4, py_compile, and diff checks. I intentionally left the public dispatch API unchanged while awaiting the preference in the inline thread.

optimum/exporters/openvino/model_configs.py
rkazants: fixed.

tests/openvino/test_asr.py
rkazants: added comment. Without this, it will clutter test logs

tests/openvino/test_asr.py
rkazants: Added. We will only maintain these remote-code models for 4.57 version to reduce our effort.

LauraGPT: The current shared minor-version gate is broader than the two architecture configs.

At this head:

- `Qwen3ASROpenVINOConfig.MIN_TRANSFORMERS_VERSION` and `MAX_TRANSFORMERS_VERSION` are both `4.57.6`, and its custom task registration is also guarded by `is_transformers_version("==", "4.57.6")`.
- `FunASROpenVINOConfig` supports `4.57.0` through `4.57.6`.
- The dedicated remote-code workflow installs only `transformers==4.57.6`.

Therefore the new `>=4.57,<4.58` decorator runs the Qwen3-ASR case on `4.57.0-4.57.5`, even though that architecture is not registered or declared supported there. A per-architecture runtime skip preserves the intended maintenance boundary without unnecessarily narrowing FunASR:

```python
if model_arch == "qwen3_asr" and not is_transformers_version("==", "4.57.6"):
    pytest.skip("Qwen3-ASR remote-code parity is supported only with transformers==4.57.6")
if model_arch == "fun_asr" and (
    is_transformers_version("<", "4.57.0") or is_transformers_version(">", "4.57.6")
):
    pytest.skip("FunASR remote-code parity is supported with transformers>=4.57.0,<=4.57.6")
```

I checked the boundary table explicitly: the current condition admits `4.57.0` and `4.57.5` for both cases, while only the FunASR config admits those versions.

Issue #1630, opened 2026-03-03
Phi-3.5-vision-instruct invalid EOS token


przepeck:
Hello, 
I have a problem with exporting the model microsoft/Phi-3.5-vision-instruct.
Model loads and generates correctly, but it doesn't stop after answering and generates random output unrelated to the prompt.
After some troubleshooting I found that EOS token id in generation_config.json is 2, while it should be 32000. After changing it manually it works correctly.
Export command:
```
optimum-cli export openvino --model microsoft/Phi-3.5-vision-instruct --weight-format int4 --sym --ratio 1.0 --group-size -1 --trust-remote-code models\microsoft\Phi-3.5-vision-instruct
```
I've used optimum from main with transformers==4.53.3 (https://github.com/huggingface/optimum-intel/issues/1573).

Python: 3.10.7

Kind regards

dhandhalyabhavik:
I have added temporary fix right now, But this issue needs resolution from Microsoft or transformer's side. @rkazants please provide your guidance on if this is the correct way to handle it right now.

PR #1804, opened 2026-06-19
Fix Phi-3.5-vision eos token temporary fix

dhandhalyabhavik:
# What does this PR do?

As pointed our correctly in issue #1630 with default config provided by `microsoft/Phi-3.5-vision-instruct` results into infinite generation loop. The issue is mismatch in stale `eos_token_id=2` in `generation_config.json` while tokenizer's actual eos token is `32000`. This fix modifies `eos_token_id` to be `[2, 32000]`. I have added `#TODO` to remove it in future when upstream config gets fixed.

Fixes # (issue)
#1630 

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/convert.py
+++ b/optimum/exporters/openvino/convert.py
@@ -761,6 +761,13 @@ def export_from_model(
         generation_config = getattr(model, "generation_config", None)
         if generation_config is not None:
             try:
+                # Todo: delete this check once phi-3 eos token get fixed in generation_configs.json
+                # Phi-3.5-vision ships stale eos_token_id=2 but its tokenizer eos is 32000, so add it as a stop token.
+                if model_type == "phi3_v":
+                    t = next((e for p in (preprocessors or []) if (e := getattr(p, "eos_token_id", None)) is not None), None)
+                    g = generation_config.eos_token_id
+                    if t is not None and t not in (g if isinstance(g, list) else [g]):
+                        generation_config.eos_token_id = (g if isinstance(g, list) else [g]) + [t]
                 generation_config.save_pretrained(output)
             except Exception as exception:
                 logger.warning(
--- a/tests/openvino/test_exporters_cli.py
+++ b/tests/openvino/test_exporters_cli.py
@@ -928,6 +928,26 @@ class OVCLIExportTestCase(unittest.TestCase):
 
         self._openvino_export(MODEL_NAMES[model_type], task, model_kwargs=model_kwargs, loading_kwargs=loading_kwargs)
 
+    @unittest.skipUnless(
+        is_transformers_version(">=", "4.40.0") and is_transformers_version("<=", "4.53.3"),
+        reason="Phi-3.5-vision export is only supported for transformers 4.40.0 - 4.53.3",
+    )
+    def test_export_phi3_v_eos_token_id(self):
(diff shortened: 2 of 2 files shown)


optimum/exporters/openvino/convert.py
```diff
@@ -751,6 +751,13 @@ def export_from_model(
         generation_config = getattr(model, "generation_config", None)
         if generation_config is not None:
             try:
+                # Todo: delete this check once phi-3 eos token get fixed in generation_configs.json
```
rkazants, line 764: can you please provide discussion raised in Github or HF about this issue? And refer with the link.

rkazants (changes requested):
please implement test

optimum/exporters/openvino/convert.py
dhandhalyabhavik: I found following discussion on HF,

https://huggingface.co/microsoft/Phi-3.5-vision-instruct/discussions/40

No discussion on GitHub transformers repo.

The script `sample_inference.py` on HF official repo replaces EOS token at generation, checkout this line - https://huggingface.co/microsoft/Phi-3.5-vision-instruct/blob/main/sample_inference.py#L58

dhandhalyabhavik:
> please implement test

@rkazants Done, added EOS specific tests with issue number. If upstream HF gets updated then we can later on remove the temp fix. Rebased the repo to inherit latest changes.

optimum/exporters/openvino/convert.py
```diff
@@ -761,6 +761,13 @@ def export_from_model(
...
+                # Phi-3.5-vision ships stale eos_token_id=2 but its tokenizer eos is 32000, so add it as a stop token.
+                if model_type == "phi3_v":
+                    t = next((e for p in (preprocessors or []) if (e := getattr(p, "eos_token_id", None)) is not None), None)
+                    g = generation_config.eos_token_id
+                    if t is not None and t not in (g if isinstance(g, list) else [g]):
+                        generation_config.eos_token_id = (g if isinstance(g, list) else [g]) + [t]
```
rkazants, line 770: what is `t`? what is `g`? it is not maintainable code. Please give proper variable naming

optimum/exporters/openvino/convert.py
```diff
@@ -761,6 +761,13 @@ def export_from_model(
         generation_config = getattr(model, "generation_config", None)
         if generation_config is not None:
             try:
+                # Todo: delete this check once phi-3 eos token get fixed in generation_configs.json
```
rkazants, line 764: create Github issue for this model in their repo and describe reproducer for it. Refer to this Github issue in our comment, othervise, we just forget why this WA is needed

tests/openvino/test_exporters_cli.py
```diff
@@ -928,6 +928,26 @@ def test_export(self, task: str, model_type: str):
 
         self._openvino_export(MODEL_NAMES[model_type], task, model_kwargs=model_kwargs, loading_kwargs=loading_kwargs)
 
+    @unittest.skipUnless(
+        is_transformers_version(">=", "4.40.0") and is_transformers_version("<=", "4.53.3"),
```
rkazants, line 932: the problem with this is that it is not validated in pre-commit. Please add new GHA workflow `Legacy models support` and create a step to run exactly this test

PR #1805, opened 2026-06-20
Fix test_exporters_cli_full_quantization

rkazants:
# What does this PR do?

After tiny-whisper model update https://huggingface.co/optimum-intel-internal-testing/tiny-random-whisper/tree/main, we need to adjust test.

## Before submitting


echarlaix:
@rkazants thanks for the fix !

PR #1811, opened 2026-06-22
Update optimum-intel CI tests from OV 25.3 to last OV version

anatyrova:
# What does this PR do?

- Moved CI to any last openvino version (previously it was 25.3, now it is 26.2.1). 

- Changed/added transformations to test_transformations.py acc. to existing in OV 2026.2.1

## Before submitting


.github/workflows/test_openvino.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       - if: ${{ matrix.test-pattern == '*modeling*' || matrix.test-pattern == '*quantization*' || matrix.test-pattern == '*transformation*' }}
         name: Install OpenVINO
         run: |
-          uv pip install openvino==2025.3.0 openvino-tokenizers==2025.3.0
+          uv pip install -U openvino openvino-tokenizers
```
echarlaix: we should be able to remove completely 

```suggestion
```

anatyrova: in setup.py, we set openvino >=2026.0 which might be too low, and some unexpected errors may appear. so i suggest leaving uv pip install -U openvino openvino-tokenizers to always use last version

echarlaix: the latest openvino version will already be installed (pip will install the highest version available at the time of installation so v2026.2.1 as of today) so why not removing it completely?

setup.py
```diff
@@ -60,7 +60,6 @@
     "peft",
     "datasets>=1.4.0,<4.0.0",
     "tbb",
-    "langchain-huggingface",
```
anatyrova, line 63: removing this unused in tests library removed segfault during collecting locally and tests passed, but in ci for preview models segfault stayed (after collecting). maybe because of some changes in ov

.github/workflows/test_openvino.yml
echarlaix: @anatyrova let me know

echarlaix (approved):

.github/workflows/test_openvino.yml
anatyrova: yes, i discussed it with Roman, and we think it's better to keep uv pip install -U openvino openvino-tokenizers, even though we set >=2026.0 in setup.py. just to avoid unexpected installation of eg.26.1/26.0

echarlaix: @anatyrova we have `uv pip install .[tests]` above in "Install dependencies" which will install the highest version available at the time of installation so openvino==2026.2.1 as you can see in the logs https://github.com/huggingface/optimum-intel/actions/runs/28230122798/job/84048691916?pr=1811 
could you extend on what you mean by avoid unexpected installation, when would that happen ?

echarlaix: also at the moment applied for only a set of test-pattern (modeling / quantization / transformation) which shouldn't be the case

rkazants: @echarlaix, occasionally it might be a case that some older version of OpenVINO already installed, for example, `2026.1` release will not be updated to latest one having just `pip install .[tests]`. I know it may be not possible that some allocated machine contains installed OV but anyway let us make it more transparent and robust.

rkazants: > also at the moment applied for only a set of test-pattern (modeling / quantization / transformation) which shouldn't be the case

can you elaborate? do you mean we miss to install latest OV for some test suite, for example, diffusers?

echarlaix: @rkazants I'm talking about the `if: ${{ matrix.test-pattern == '*modeling*' || matrix.test-pattern == '*quantization*' || matrix.test-pattern == '*transformation*' }}`

echarlaix: https://github.com/anatyrova/optimum-intel/blob/e3dcdcf50654fba195f6af35200b05cf12d3fc0d/.github/workflows/test_openvino.yml#L68

echarlaix: it was previously added as we had seg fault with latest openvino release for these specific tests but can be removed now

anatyrova: now i am thinking, it is actually may not be needed, since before you've set it for 25.3 to cure segfaults, but now everything is on last ov

so, can i just remove those lines? 
```
- if: ${{ matrix.test-pattern == '*modeling*' || matrix.test-pattern == '*quantization*' || matrix.test-pattern == '*transformation*' }}
        name: Install OpenVINO
        run: |
          uv pip install -U openvino openvino-tokenizers
```

echarlaix: yes works for me, leaving the choice to you @anatyrova

PR #1817, opened 2026-06-24
fix fork PR rate limiting with HF model caching

echarlaix:
CI jobs triggered by fork PRs cannot access the `HF_HUB_READ_TOKEN` secret, resulting in rate limit being reached trying loading the 200 models.
In this PR we add a new workflow to build and maintain a github actions model cache `hf-models` that is shared across workflows 

This workflow is triggered by `workflow_dispatch` or two PR labels: `update-hf-cache` (will only download model that are not already cached) and `update-hf-cache-force` (forcing download of all models, can be useful when models have been updated)
models that are larger than 100MB are excluded to avoid reaching the 10 GB limit constraint

### DIFF
--- a/.github/workflows/create_model_cache.yml
+++ b/.github/workflows/create_model_cache.yml
@@ -16,6 +16,7 @@ permissions:
 
 env:
   HF_TOKEN: ${{ secrets.HF_HUB_READ_TOKEN }}
+  UV_SYSTEM_PYTHON: true
 
 jobs:
   create-cache:
--- a/.github/workflows/create_model_cache.yml
+++ b/.github/workflows/create_model_cache.yml
@@ -1,0 +1,69 @@
+name: Create Model Cache
+
+on:
+  pull_request:
+    types: [labeled]
+  workflow_dispatch:
+    inputs:
+      force_update:
+        description: "Force the (re-)download, overriding the cached versions if exist"
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+  actions: write  # needed to delete the stale cache entry
+
+env:
+  HF_TOKEN: ${{ secrets.HF_HUB_READ_TOKEN }}
+
+jobs:
+  create-cache:
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.label.name == 'create-model-cache' ||
+      github.event.label.name == 'create-model-cache-force'
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6.0.2
+
+      - name: Setup Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.10"
+
+      - name: Check HF token is available
+        run: |
+          if [ -z "$HF_TOKEN" ]; then
+            echo "HF_TOKEN secret not available — cannot run on fork PRs"
+            exit 1
+          fi
+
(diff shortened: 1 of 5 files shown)


.github/workflows/test_openvino.yml
```diff
@@ -87,6 +94,25 @@ jobs:
...
+      - name: Save cache
+        if: >-
+          steps.hf-cache-restore.outputs.cache-hit != 'true' &&
+          env.HF_TOKEN != '' &&
+          matrix.test-pattern == '*decoder*' &&
+          matrix.transformers-version == 'latest'
+        uses: actions/cache/save@v4
+        with:
+          path: ~/.cache/huggingface/hub
+          key: hf-models
```
echarlaix: moving to dedicated workflow so that we can update the cache anytime (in case models are being updated)

.github/workflows/create_model_cache.yml
```diff
@@ -0,0 +1,70 @@
...
+env:
+  HF_TOKEN: ${{ secrets.HF_HUB_READ_TOKEN }}
+  UV_SYSTEM_PYTHON: true
+
+jobs:
+  create-cache:
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.label.name == 'create-hf-cache' ||
+      github.event.label.name == 'create-hf-cache-force'
```
regisss: Both labels will do exactly the same right?

echarlaix: `create-hf-cache-force` will force download for all models (even the ones that are already cached), this can be useful in case some models are being updated on the hub so that we have their latest versions

tests/scripts/download_cached_models.py
```diff
@@ -0,0 +1,109 @@
...
+sys.path.insert(0, str(Path(__file__).parents[1] / "openvino"))
+
+from utils_tests import MODEL_NAMES
+
+
+# github actions repo cache limit is 10 GB so using 9 GB
+CACHE_LIMIT = 9 * 10**9
+
+# checkpoints always excluded from the cache
+SKIP_MODEL_NAMES = {
```
echarlaix: currently hardcoded, could make sense to replace with a size threshold

.github/workflows/create_model_cache.yml
echarlaix: @regisss will add more info in the description right now!

echarlaix: also renamed both label to update-hf-cache/ update-hf-cache-force

.github/workflows/test_openvino.yml
```diff
@@ -57,6 +57,14 @@ jobs:
           pip install --upgrade pip uv
           uv pip install .[tests] librosa diffusers
 
+      - name: Restore HF models cache
+        id: hf-cache-restore
+        if: env.HF_TOKEN == ''
```
echarlaix, line 62: cache won't be restored when doesn't have to be, could also make sense to always restore

regisss (approved):
Super cool!!

tests/scripts/download_cached_models.py
echarlaix: modified in https://github.com/huggingface/optimum-intel/pull/1817/changes/f507d43ff9de0a15536f39cbe56c544dbb3a6e86, filtering 100MB models

PR #1839, opened 2026-06-26
Bump minimum transformers version to 4.57 

echarlaix:


setup.py
```diff
@@ -30,7 +30,7 @@
     "torch>=2.1",
     "safetensors<0.8.0",
     "optimum~=2.2.0",
-    "transformers>=4.45,<5.1",
+    "transformers>=4.57,<5.1",
```
echarlaix, line 33: officially removed in https://github.com/huggingface/optimum-intel/pull/1684 that should soon be merged

rkazants (changes requested):
please do not merge it because we have new models requests that need transformers < 4.57

echarlaix:
@rkazants which models are you talking about ? It was previously discussed that only the 3 latest transformers versions should be officially supported. Something we can do is add support like for preview models (without officially supporting this version of transformers)

PR #1841, opened 2026-06-30
Qwen3 omni code pred opt

byungilm:
Personal test only draft

##  summary
Introduces a single-step CodePredictor export path for Qwen3-Omni (PoC), gated by the QWEN3_OMNI_CP_STEP=1 environment variable, plus two related quantization/loading fixes.

### detail
Single-step CodePredictor (the main change)

model_configs.py: New Qwen3OmniCodePredictorStepLMConfigHelper with I/O (inputs_embeds, temperature, top_k, seed, step_idx) → (token, codec_embed), stateless (use_past=False). The CODE_PREDICTOR behavior now branches: env-on → step helper/patcher, env-off → original unrolled path (default, unchanged).
model_patcher.py: New Qwen3OmniCodePredictorStepPatcher that exports one decoder pass instead of unrolling all num_code_groups-1 steps. No KV cache is exported; the GenAI speech_pipeline.cpp loops in C++, feeding the full embed sequence each step. step_idx selects the per-step lm_head/codec embedding so body weights are shared. Gumbel-max sampling stays in-graph for bit-exact reproducibility. Result: the compiled graph holds a single decoder block (~1/N of the fp32 intermediate memory).
Supporting quantization/loading fixes

modeling_text2speech.py: Route qwen3_omni reload through OVModelForVisualCausalLM, so INT4 weight compression can load/quantize/save every submodel (mirrors the working INT8 path).
quantization.py: For qwen3_omni, set default_config=None so only the language model is weight-compressed; speech/vocoder submodels (talker, code_predictor, code2wav, embeddings) are kept in original precision (FP16), since INT8 corrupts the audio.

Fixes # (issue)
CVS-180272

## Before submitting


byungilm:
Not for review accuracy issue observed.

rkazants:
These changes were incorporated into initial PR with qwen3-omni support.
cc @sgonorov

PR #1842, opened 2026-06-30
Remove login ci token to avoid rate limit errors for fork PRs

echarlaix:
Replace login() and save token directly instead to avoid whoami api call that causes rate limiting when many CI jobs run in parallel

### DIFF
--- a/tests/scripts/login_with_ci_token.py
+++ b/tests/scripts/login_with_ci_token.py
@@ -14,3 +14,4 @@ token_path.parent.mkdir(parents=True, exist_ok=True)
 token_path.write_text(CI_TOKEN)
 
 assert get_token() == CI_TOKEN, f"Token was not saved correctly to {constants.HF_TOKEN_PATH}"
+print(get_token(), CI_TOKEN)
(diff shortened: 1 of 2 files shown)


tests/scripts/login_with_ci_token.py
```diff
@@ -1,9 +1,14 @@
...
+from huggingface_hub import constants
 
 
 # fmt: off
 # not critical
 CI_TOKEN = "".join(['h', 'f', '_', 'X', 'N', 'C', 'O', 'g', 'A', 'O', 'g', 'S', 'H', 'W', 'Y', 'B', 'H', 'g', 'L', 'J', 'n', 'C', 'U', 'f', 'i', 'O', 'D', 'n', 'm', 'N', 'n', 'V', 'Q', 'Q', 'N', 'y', 'c'])
 # fmt: on
 
-login(token=CI_TOKEN)
+# save token directly to avoid whoami api call that causes rate limiting when many CI jobs run in parallel
```
echarlaix, line 11: when using `_write_secret` getting 

```ImportError: cannot import name '_write_secret' from 'huggingface_hub.utils._auth' (/opt/hostedtoolcache/Python/3.10.20/x64/lib/python3.10/site-packages/huggingface_hub/utils/_auth.py)```

PR #1848, opened 2026-07-02
[skills] add  (optimum & create tiny model) agents skills

Mohamed-Ashraf273:
# What does this PR do?

 add  (optimum & create tiny mode) agents skills

## Before submitting


helena-intel:
Thanks for this PR! Can you please add instructions to make sure that the tiny model outputs do not collapse and do not contain NaN? When tiny model outputs collapse (e.g. output is 0,0,0,0,0,0,0,...) comparing outputs becomes less meaningful and we miss important issues. This is a known issue with some existing tiny models; we should prevent this for new models. Sometimes preventing this takes a bit of experimenting. For example, keep more parameters, or initialize the model with a bit more variance, e.g. 

```
    model.lm_head.weight.normal_(mean=0.0, std=0.2)
```

Mohamed-Ashraf273:
> Thanks for this PR! Can you please add instructions to make sure that the tiny model outputs do not collapse and do not contain NaN? When tiny model outputs collapse (e.g. output is 0,0,0,0,0,0,0,...) comparing outputs becomes less meaningful and we miss important issues. This is a known issue with some existing tiny models; we should prevent this for new models. Sometimes preventing this takes a bit of experimenting. For example, keep more parameters, or initialize the model with a bit more variance, e.g.
> 
> ```
>     model.lm_head.weight.normal_(mean=0.0, std=0.2)
> ```

Thanks for the suggestion! I added checks for NaN/Inf values, collapsed logits, and repeated-token outputs, along with guidance for increasing model capacity or initialization variance when needed.

Mohamed-Ashraf273:
@rkazants 
Could you please take a look?

.github/agents/model-analysis.agent.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+3. Inspect the real Transformers architecture and print relevant modules:
+
+```python
+from transformers import AutoConfig, AutoModelForCausalLM
+
+model_id = "<model_id>"
+config = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
+model = AutoModelForCausalLM.from_config(config, trust_remote_code=True)
```
echarlaix: setting trust_remote_code=True will run arbitrary code so shouldn't be default, also we should specify that we need user explicit confirmation that they trust that source instead

.github/skills/tiny-model-creator/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+the original model, especially for trust-remote-code architectures. Create and
+validate the tiny model using a compatible version; do not silently substitute
+another model implementation because the active Transformers version lacks the
+architecture.
+
+Inspect precision fields under every relevant configuration level. Remote
+models may use `dtype`, `torch_dtype`, or both, including separate values in
+vision, text, audio, or projector sub-configs.
+
+## Step 2 — Write a reusable constructor
```
echarlaix, line 42: shouldn't we also add a memory constraint as well ? Also this would ensure that we don't end up with a model with almost the same parameters with a smaller number of layers for example

.github/agents/model-analysis.agent.md
Mohamed-Ashraf273: I’ve disabled `trust_remote_code` by default and added an explicit requirement to confirm that the user trusts the source and accepts the arbitrary-code-execution risk before enabling it.

.github/skills/tiny-model-creator/SKILL.md
Mohamed-Ashraf273: Agreed. I’ve added explicit parameter-count and model-memory budgets, verified after construction, so reducing the layer count cannot be offset by widening other dimensions and leave the model nearly as large.

.github/skills/optimum-model-enabler/model-patching-patterns.md
```diff
@@ -0,0 +1,58 @@
...
+    gate = torch.bmm(repeated, self.gate_projs)
+    up = torch.bmm(repeated, self.up_projs)
+    expert_output = torch.bmm(self.act_fn(gate) * up, self.down_projs)
+    expert_output = expert_output * routing.transpose(0, 1).unsqueeze(-1)
+    return expert_output.sum(0).reshape(batch, sequence, hidden)
+```
+
+Adapt parameter layout and routing semantics to the real model. Do not apply
+this template when experts are not represented by compatible stacked tensors.
+
```
rkazants, line 56: Please add that it needs to re-use existing patches, functions in optimum-intel and avoid duplicated semantically same patches. It relates MoE, RoPE, different type of attentions and other blocks.

Mohamed-Ashraf273: Addressed. I added guidance to search for and reuse or extend existing Optimum Intel patches and helpers, avoiding semantically duplicate implementations across MoE, RoPE, attention variants, and other architecture blocks.

.github/skills/optimum-model-enabler/SKILL.md
```diff
@@ -0,0 +1,232 @@
...
+- `with_behavior()` returns a non-`None` export configuration;
+- its inputs and outputs match that submodel;
+- unsupported enum values are excluded instead of being advertised.
+
+Do not consider successful conversion sufficient when a behavior silently
+maps to the wrong component or produces a `None` configuration.
+
+## Step 4 — Add repository tests
+
+Source support without tests is incomplete. Update the applicable files:
```
popovaan, line 263: Please add also information that if the transformers version required by the pytorch model is not supported by Optimum-Intel then the model tests should be added to a dedicated pipeline "Preview models support" and tested with the required transformers version: https://github.com/huggingface/optimum-intel/blob/a3b6012a4c02f4147260da4d4601bb6a3c0d2bb0/.github/workflows/test_openvino_preview_models.yml#L64

So that means other models won't work for this version, but this specific new model should work and be tested.

.github/skills/optimum-model-enabler/SKILL.md
```diff
@@ -0,0 +1,232 @@
...
+
+- `tests/openvino/test_decoder.py` for decoder-only generation;
+- `tests/openvino/test_seq2seq.py` for seq2seq and VLM integration;
+- `tests/openvino/test_export.py` for model/API export coverage;
+- `tests/openvino/test_exporters_cli.py` for CLI export coverage;
+- `tests/openvino/test_quantization.py` for compression and quantization;
+- `tests/openvino/utils_tests.py` for fixtures and expected IR details.
+
```
popovaan, line 272: For models with MoE blocks transformations tests should be added that check presence of "ConvertTiledMoeBlockToGatherMatmuls": https://github.com/huggingface/optimum-intel/blob/a3b6012a4c02f4147260da4d4601bb6a3c0d2bb0/tests/openvino/test_transformations.py#L260

Similarly for models with RoPE blocks tests are needed that check "RoPEFusion": https://github.com/huggingface/optimum-intel/blob/a3b6012a4c02f4147260da4d4601bb6a3c0d2bb0/tests/openvino/test_transformations.py#L51

Mohamed-Ashraf273: Good point, thank you! I’ll update the skill to explicitly require the relevant MoE/RoPE transformation tests and reuse the existing test patterns.

rkazants (approved):
LGTM

.github/skills/optimum-model-enabler/model-patching-patterns.md
```diff
@@ -0,0 +1,63 @@
...
+    routing.scatter_(1, selected.reshape(-1, selected.shape[-1]), scores.reshape(-1, scores.shape[-1]))
+
+    flat = hidden_states.reshape(-1, hidden)
+    repeated = flat.unsqueeze(0).expand(self.config.num_experts, -1, -1)
+    gate = torch.bmm(repeated, self.gate_projs)
+    up = torch.bmm(repeated, self.up_projs)
+    expert_output = torch.bmm(self.act_fn(gate) * up, self.down_projs)
+    expert_output = expert_output * routing.transpose(0, 1).unsqueeze(-1)
+    return expert_output.sum(0).reshape(batch, sequence, hidden)
+```
```
IlyasMoutawwakil, line 52: why not just set_experts_implementation("batched_mm") instead of re implementing it ?

IlyasMoutawwakil (approved):

PR #1849, opened 2026-07-03
Deprecate broken contextual dataset and replace it with textvqa

echarlaix:
Deprecate broken https://huggingface.co/ucla-contextual dataset and replace it with https://huggingface.co/datasets/lmms-lab/textvqa

### DIFF
--- a/optimum/intel/openvino/quantization.py
+++ b/optimum/intel/openvino/quantization.py
@@ -847,8 +847,11 @@ class OVCalibrationDatasetBuilder:
                     break
 
                 instruction = item[dataset_metadata["inputs"]["instruction"]]
-                image_url = item[dataset_metadata["inputs"]["image_url"]]
-                image = Image.open(requests.get(image_url, stream=True).raw).convert("RGB")
+                if "image_url" in dataset_metadata["inputs"]:
+                    image_url = item[dataset_metadata["inputs"]["image_url"]]
+                    image = Image.open(requests.get(image_url, stream=True).raw).convert("RGB")
+                else:
+                    image = item[dataset_metadata["inputs"]["image"]].convert("RGB")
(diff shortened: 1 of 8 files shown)


optimum/intel/openvino/quantization.py
```diff
@@ -292,7 +292,14 @@ def build_from_quantization_config(self, config: OVQuantizationConfigBase) -> OV
                 )
 
             if isinstance(self.model, OVModelForVisualCausalLM):
-                dataset_metadata = PREDEFINED_VISUAL_LM_DATASETS[config.dataset]
+                dataset_name = config.dataset
+                if dataset_name == "contextual":
+                    logger.warning(
+                        "The `contextual` calibration dataset is deprecated because its images are no longer "
+                        "reachable, and will be removed in a future release. Using `textvqa` instead."
```
rkazants, line 299: @andrey-churkin, @andreyanufr, do you agree with this replacement? Will it be suitable for calibration?

rkazants (approved):

optimum/intel/openvino/quantization.py
echarlaix: used https://huggingface.co/datasets/lmms-lab/textvqa but we can replace with any dataset that you think is more suitable

andreyanufr: Perhaps it's worth adding another dataset as an option: nlphuji/flickr30k.

It seems to me that textvqa is too simple and not diverse in terms of questions.

rkazants: let us merge temporarily textvqa. In the meantime, @andreyanufr, can you check which dataset to use in terms of quality and calibrarion time, etc.?

PR #1857, opened 2026-07-07
Fix export for encoder-decoder models with unnamed tensor

echarlaix:
Fix https://github.com/huggingface/optimum-intel/actions/runs/28857771168/job/85588501704?pr=1855 happening with openvino-nightly when calling `output.get_any_name` raise RuntimeError (happening for tensors when no name associated)

```
FAILED tests/openvino/test_modeling_basic.py::OVModelBasicIntegrationTest::test_pipeline_4_hf_internal_testing_tiny_random_t5 - RuntimeError: Check '!get_names().empty()' failed at src/core/src/descriptor/tensor.cpp:102:
Attempt to get a name for a Tensor without names
```

### DIFF
--- a/optimum/exporters/openvino/stateful.py
+++ b/optimum/exporters/openvino/stateful.py
@@ -253,8 +253,12 @@ def insert_state_for_nodes(model: ov.Model, nodes):
     outputs = sum((node.outputs() for node in nodes), [])
     for output in outputs:
         consumers = output.get_target_inputs()
-        # FIXME: get_any_name is not reliable as tensor may not have any names
-        variable_id = output.get_any_name()
+        if output.get_names():
+            variable_id = output.get_any_name()
+        else:
+            variable_id = f"{output.get_node().get_friendly_name()}.{output.get_index()}"
+            # set tensor name when tensor has no name
+            output.get_tensor().set_names({variable_id})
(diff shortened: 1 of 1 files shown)


optimum/exporters/openvino/stateful.py
```diff
@@ -253,8 +253,12 @@ def insert_state_for_nodes(model: ov.Model, nodes):
     outputs = sum((node.outputs() for node in nodes), [])
     for output in outputs:
         consumers = output.get_target_inputs()
-        # FIXME: get_any_name is not reliable as tensor may not have any names
-        variable_id = output.get_any_name()
+        if output.get_names():
+            variable_id = output.get_any_name()
+        else:
+            variable_id = f"{output.get_node().get_friendly_name()}.{output.get_index()}"
```
IlyasMoutawwakil, line 259: is the index always necessary ?

IlyasMoutawwakil (approved):
LGTM !

optimum/exporters/openvino/stateful.py
echarlaix: would be needed in case where a node has multiple (unnamed) outputs, maybe there is a cleaner way to do it though cc @rkazants @popovaan

rkazants (approved):

PR #1860, opened 2026-07-07
Fix gemma4_unified chat template processing + test coverage

mmikolajcz:
# What does this PR do?

Add chat template processing for gemma4_unified architecture and test coverage for both base and instruction-tuned support.

Fixes # (issue)

## Before submitting

### DIFF
--- a/tests/openvino/utils_tests.py
+++ b/tests/openvino/utils_tests.py
@@ -209,7 +209,7 @@ HUB_MODEL_NAMES = {
     "gemma3n_text": "optimum-intel-internal-testing/tiny-random-gemma3n-text",
     "gemma4": "optimum-intel-internal-testing/tiny-random-gemma4",
     "gemma4_moe": "optimum-intel-internal-testing/tiny-random-gemma4-moe",
-    "gemma4_unified": "optimum-intel-internal-testing/tiny-random-gemma4-unified",
+    "gemma4_unified": "optimum-intel-internal-testing/tiny-random-gemma4-unified-it",
     "falcon": "optimum-intel-internal-testing/really-tiny-falcon-testing",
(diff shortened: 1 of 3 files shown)


tests/openvino/utils_tests.py
```diff
@@ -209,7 +209,7 @@ def _create_tiny_kokoro_model():
     "gemma3n_text": "optimum-intel-internal-testing/tiny-random-gemma3n-text",
     "gemma4": "optimum-intel-internal-testing/tiny-random-gemma4",
     "gemma4_moe": "optimum-intel-internal-testing/tiny-random-gemma4-moe",
-    "gemma4_unified": "optimum-intel-internal-testing/tiny-random-gemma4-unified",
+    "gemma4_unified": "optimum-intel-internal-testing/tiny-random-gemma4-unified-it",
```
rkazants, line 212: let us have two versions non-it and it and test both

mmikolajcz: I wonder, wouldn't it be more effective to have single tiny_random and slightly override config so template/non-template behavior would be tested and apply it also for broader scope of models?

Testing both gemma4_unified variants is fine, however there are more models that could have bugged pre-processing so it will duplicate lot of tiny-random models + might cause some backward compatibility issues for tests if existing tiny random images would require update.

For example, other gemma4 variants are tested by `-it` variant by default (https://huggingface.co/optimum-intel-internal-testing/tiny-random-gemma4-31B) and they can have different problem that base model without chat template might fail in preprocessing. Same for Qwen3.5 `-base` and probably many more.

Looking on modelling_visual_language.py there are couple of examples where chat template is optional

mitruska:
For the record, the issue has been addressed in another PR:
- https://github.com/huggingface/optimum-intel/pull/1895

PR #1863, opened 2026-07-13
Add transformers v5.10 support

echarlaix:
Add support for transformers v5.10:

- `CLIPTextTransformer` was removed in https://github.com/huggingface/transformers/pull/44431, so `text_encoder.text_model` no longer exists since transformers v5.6
- `gemma4_text_attention_forward`: transformers v5.6 (huggingface/transformers#45788) replaced `past_key_values.shared_layers` / `self.kv_shared_layer_index` with a `shared_kv_states` dict passed through `kwargs`
- `rot_pos_emb()` in `_OVQwen2VLForCausalLM` and `_OVQwen3OmniMoeForCausalLM` now calls `self._rotary_pos_emb(pos_ids)` directly when `is_transformers_version(">=", "5.9")`, since the upstream signature changed from taking `max_grid_size` to taking `pos_ids` directly


.github/workflows/test_openvino_slow.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-22.04", "windows-2022"]
-        transformers-version: ["4.57.6", "latest"]
+        transformers-version: ["5.8.0", "latest"]
```
rkazants: we still need to check 4.57.6 that is because we have remote-code models requiring this version

echarlaix: @rkazants yes this is used for testing purpose only, I'll remove before merging

.github/workflows/test_openvino.yml
```diff
@@ -39,7 +39,7 @@ jobs:
             "*quantization*",
             "*transformation*",
           ]
-        transformers-version: ["4.57.6", "latest"]
+        transformers-version: ["4.57.6", "5.8.0", "latest"]
```
echarlaix: added for tests, will remove

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -3160,9 +3160,12 @@ def rot_pos_emb(self, grid_thw):
             wpos_ids = wpos_ids.flatten()
             pos_ids.append(torch.stack([hpos_ids, wpos_ids], dim=-1).repeat(t, 1))
         pos_ids = torch.cat(pos_ids, dim=0)
-        max_grid_size = grid_thw[:, 1:].max()
-        rotary_pos_emb_full = self._rotary_pos_emb(max_grid_size)
-        rotary_pos_emb = rotary_pos_emb_full[pos_ids].flatten(1)
+        if is_transformers_version(">=", "5.9"):
```
echarlaix, line 3269: since https://github.com/huggingface/transformers/pull/45396/

regisss (approved):
LGTM!

optimum/exporters/openvino/model_patcher.py
```diff
@@ -5674,6 +5670,10 @@ def gemma4_text_attention_forward(
 ) -> tuple:
     from transformers.models.gemma4.modeling_gemma4 import apply_rotary_pos_emb as apply_rotary_pos_emb_gemma4
 
+    # since transformers >= v5.6 (PR #45788) `shared_kv_states` dict and is passed and `kv_shared_layer_index` removed
```
```suggestion
    # since transformers >= v5.6 (PR #45788) `shared_kv_states` dict passed and `kv_shared_layer_index` removed
```

optimum/exporters/openvino/model_patcher.py
```diff
@@ -5674,6 +5670,10 @@ def gemma4_text_attention_forward(
...
+    shared_kv_states = kwargs.pop("shared_kv_states", None)
+    legacy_shared_kv_states = hasattr(self, "kv_shared_layer_index")
```
rkazants: shall we have if-else condition on transformers version? Does this change work for transfoemrs <5.6?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -5684,8 +5684,12 @@ def gemma4_text_attention_forward(
     query_states = apply_rotary_pos_emb_gemma4(query_states, cos, sin, unsqueeze_dim=2)
     query_states = query_states.transpose(1, 2)
 
-    if self.is_kv_shared_layer and past_key_values is not None:
-        key_states, value_states = past_key_values.shared_layers[self.kv_shared_layer_index]
+    if self.is_kv_shared_layer and (not legacy_shared_kv_states or past_key_values is not None):
```
```suggestion
    if self.is_kv_shared_layer and (past_key_values is not None or shared_kv_states is not None):
```
I think this change looks more clear

rkazants (approved):

optimum/exporters/openvino/model_patcher.py
echarlaix: applied your comment in https://github.com/huggingface/optimum-intel/pull/1863/changes/3bbb9a0fd0813fe7c2f62121c4afbf3cb7080b36

PR #1872, opened 2026-07-15
[OpenVINO] Fix passing of token_types_ids for Gemma3.

popovaan:
# What does this PR do?

`is_prefill` / `is_first_iteration` flag is traced incorrectly which results in not using of `token_types_ids` in the model. This PR fixes this issue. 

Fixes CVS-190458, CVS-172722

## Before submitting


popovaan:
@rkazants @echarlaix could you please review the PR?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4825,13 +4825,264 @@ def patched_forward(*args, **kwargs):
         self.patched_forward = patched_forward
 
 
+# The patching is needed avoid usage of is_first_iteration flag. In original code this flag is set when past_key_values are not initilazed,
```
echarlaix: `is_first_iteration` will be removed from `create_causal_mask_mapping` starting transformers v5.6
 https://github.com/huggingface/transformers/pull/45454/

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4874,6 +5133,9 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_value, traceback):
         super().__exit__(exc_type, exc_value, traceback)
 
+        if is_transformers_version(">=", "4.57.6"):
+            self._model.model.forward = self._model.orig_lm_forward
+
```
echarlaix: also I see in the tests that we currently remove `token_type_ids` from inputs https://github.com/huggingface/optimum-intel/blob/109314ca20f32838f02f5a10b9475bd12bd818c6/tests/openvino/test_seq2seq.py#L792 could make sense to change this as well

optimum/exporters/openvino/model_patcher.py
```diff
@@ -4866,6 +5117,14 @@ def forward(
...
             return result
 
+        if is_transformers_version("==", "4.57.6"):
+            model.orig_lm_forward = model.model.forward
+            model.model.forward = types.MethodType(patched_lm_forward_4_57_6, model.model)
+
+        if is_transformers_version(">=", "5.0"):
+            model.orig_lm_forward = model.model.forward
+            model.model.forward = types.MethodType(patched_lm_forward_5_0, model.model)
+
```
echarlaix: wouldn't it be more robust instead to use `_create_gemma4_unified_bidirectional_mask_dict` (or similar)  https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_patcher.py#L8661 to create an attention_mask like in https://github.com/huggingface/optimum-intel/blob/transformers-v5.10-support/optimum/exporters/openvino/model_patcher.py#L4851

echarlaix (commented):
Thanks a lot for the quick fix

echarlaix (approved):

optimum/exporters/openvino/model_patcher.py
popovaan: Agree, added passing of `token_type_ids` in this test for `gemma3`.

optimum/exporters/openvino/model_patcher.py
popovaan: Replaced this code with the fix from this branch https://github.com/huggingface/optimum-intel/pull/1863, so looks robust now.

optimum/exporters/openvino/model_patcher.py
popovaan: Yes, looks like this patching won't be needed for transformers v5.6+.

rkazants (approved):

PR #1892, opened 2026-07-23
Trigger openvino tests for release branches for pull_request

echarlaix:


.github/workflows/test_openvino.yml
rkazants, line 1: please update other workflows. Thanks!

echarlaix: added in https://github.com/huggingface/optimum-intel/pull/1892/changes/823290a9377956ff228cd70ab0715f3db29c5b70

PR #1893, opened 2026-07-23
Trigger openvino tests for release branches for pull_request

echarlaix:


.github/workflows/test_openvino.yml
rkazants, line 1: please update other yml files as well
for example, https://github.com/huggingface/optimum-intel/blob/main/.github/workflows/test_openvino_preview_models.yml#L11

and others.

rkazants (approved):

.github/workflows/test_openvino.yml
echarlaix: yes updated @rkazants !

PR #1895, opened 2026-07-24
[OpenVINO] Fix chat template applying for Gemma4Unified.

popovaan:
# What does this PR do?

Currently preprocess_inputs() for Gemma4Unified doesn't apply chat template which leads to significant drop of similarity for google/gemma-4-12B-it.

Fixes CVS-190652

## Before submitting

### DIFF
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -5601,7 +5601,7 @@ class _OVGemma4UnifiedForCausalLM(_OVGemma3ForCausalLM):
         if getattr(tokenizer, "chat_template", None) is None:
             if image is not None:
                 # If chat template is not available we need to add image token manually, as there's a requirement
-                # that image token number should match the number ofprovided images:
+                # that image token number should match the number of provided images:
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -5598,12 +5598,25 @@ def preprocess_inputs(
             raise ValueError("Video input is not supported")
         if audio is not None:
             raise ValueError("Audio input is not supported")
-        # The gemma4_unified processor has no chat template, so build the prompt directly.
-        # An image is referenced by the processor's image token placeholder.
+        if getattr(tokenizer, "chat_template", None) is None:
```
rkazants, line 5601: will be awesome to have a link to the reference code where they recommend to attach image_token manually when chat template is not available.

rkazants (approved):

optimum/intel/openvino/modeling_visual_language.py
popovaan: There's no direct recommendation in the code, but I added a comment explaining why it is needed and added a link to the check that will fail in the inputs validator of original model if we don't add image token.

popovaan:
@echarlaix could you please review this PR?

PR #1898, opened 2026-07-27
Sync built docs to the hf-doc-build bucket

echarlaix:
build_main_documentation.yml workflow switched to syncing built docs into a bucket https://github.com/huggingface/doc-builder/blob/main/.github/workflows/build_main_documentation.yml#L212

optimum intel documentation https://huggingface.co/docs/optimum-intel/index is not updated [since 15/04](https://github.com/huggingface/doc-builder/pull/780/changes)

cc @mishig25


regisss (approved):
LGTM!

PR #1902, opened 2026-07-30
[OpenVINO] Add audio input support for Gemma 4 models

as-suvorov:
# What does this PR do?

This PR adds OpenVINO export and inference support for audio inputs in Gemma 4 models, including:

- Standard Gemma 4 checkpoints with an audio tower, such as `google/gemma-4-e2b-it` and `google/gemma-4-E4B-it`.
- Gemma 4 Unified checkpoints with encoder-free audio projection, such as `google/gemma-4-12B-it`.

Audio support is enabled from the checkpoint configuration. If `audio_config` is absent, the audio component is not exported or loaded, preserving compatibility with text/vision-only Gemma 4 and MoE checkpoints.

### Standard Gemma 4: E2B and E4B

The exported `audio_embeddings` component combines:

1. `model.model.audio_tower`
2. `model.model.embed_audio`

Its OpenVINO contract is:

| Tensor | Shape | Type | Description |
| --- | --- | --- | --- |
| `input_features` | `[num_audios, audio_sequence_length, 128]` | FP32 | 128-bin acoustic features produced by the processor |
| `input_features_mask` | `[num_audios, audio_sequence_length]` | boolean | Marks valid acoustic frames before padding |
| `last_hidden_state` | `[num_audios, audio_output_sequence_length, text_hidden_size]` | floating point | Audio embeddings projected into the language-model hidden space |
| `attention_mask` | `[num_audios, audio_output_sequence_length]` | boolean | Validity mask after audio-tower subsampling |

The audio tower contains two stride-2 convolutional subsampling stages. Approximately 100 acoustic frames per second become 25 projected audio tokens per second. The runtime filters padded output positions using the returned mask and scatters valid embeddings into positions where `input_ids == audio_token_id`.

The projection width follows the checkpoint text hidden size, for example 1536 for E2B and 2560 for E4B.

### Gemma 4 Unified: 12B

Gemma 4 Unified uses an encoder-free audio path. The processor splits a 16 kHz waveform into non-overlapping 640-sample, 40 ms frames:

| Tensor | Shape | Type | Description |
| --- | --- | --- | --- |
| `input_features` | `[num_audios, num_frames, 640]` | FP32 | Raw waveform frames |
| `last_hidden_state` | `[num_audios, num_frames, text_hidden_size]` | floating point | Frames projected into the language-model hidden space |

The exported component wraps `model.model.embed_audio`. Since the projection does not change the temporal resolution, the processor mask is applied by the runtime after projection.

Both architectures therefore produce approximately 25 audio tokens per second, but standard Gemma 4 reaches that rate through acoustic feature extraction and convolutional subsampling, while Unified projects 40 ms waveform frames directly.

## Export

Install this checkout in editable mode so `optimum-cli` uses the implementation from the current repository:

```bash
python -m pip install -e .
```

Gemma 4 E2B E4B:

```bash
python -m pip install "transformers==5.7.0"
optimum-cli export openvino \
  --model google/gemma-4-e2b-it \
  --task image-text-to-text \
  models/gemma-4-e2b-it
```

Gemma 4 Unified 12B:

```bash
python -m pip install "transformers==5.10.0"
optimum-cli export openvino \
  --model google/gemma-4-12B-it \
  --task image-text-to-text \
  models/gemma-4-12b-it
```

A checkpoint with audio weights produces these additional files:

```text
openvino_audio_embeddings_model.xml
openvino_audio_embeddings_model.bin
```

## Inference example

```python
from optimum.intel import OVModelForVisualCausalLM

MODEL_ID = "google/gemma-4-12b-it"

# Load model
processor = AutoProcessor.from_pretrained(MODEL_ID)
model = OVModelForVisualCausalLM.from_pretrained(MODEL_ID, device="CPU")

audio1 = "https://storage.openvinotoolkit.org/models_contrib/speech/2021.2/librispeech_s5/how_are_you_doing_today.wav"
audio2 = "https://raw.githubusercontent.com/google-gemma/cookbook/refs/heads/main/apps/sample-data/journal1.wav"

prompt = "Transcribe the following speech segment in its original language. Follow these specific instructions for formatting the answer:\n* Only output the transcription, with no newlines.\n* When transcribing numbers, write the digits, i.e. write 1.7 and not one point seven, and write 3 instead of three."
conversations = [
    [
        {
            "role": "user",
            "content": [
                {"type": "text", "text": prompt},
                {"type": "audio", "audio": audio},
            ],
        }
    ]
    for audio in (audio1, audio2)
]

# Process input
inputs = processor.apply_chat_template(
    conversations,
    tokenize=True,
    return_dict=True,
    return_tensors="pt",
    add_generation_prompt=True,
    processor_kwargs={"padding": True},
).to(model.device)
input_length = inputs["input_ids"].shape[-1]

# Generate output
outputs = model.generate(**inputs, max_new_tokens=512, do_sample=False)
responses = processor.batch_decode(outputs[:, input_length:], skip_special_tokens=False)

# Parse output
for index, response in enumerate(responses, start=1):
    print(f"Audio {index}: {processor.parse_response(response)}")

Audio 1: {'content': 'How are you doing today?', 'role': 'assistant'}
Audio 2: {'content': 'I woke up early today feeling refreshed. The morning light was beautiful and I enjoyed a nice cup of coffee.', 'role': 'assistant'}
```

### E2B

| OpenVINO component | Omitted format | `int8` | `int4` |
| --- | ---: | ---: | ---: |
| Language model | 2182.0 MiB | 2182.0 MiB (0%) | 1354.5 MiB (-37.9%) |
| Text embeddings | 1536.0 MiB | 384.5 MiB (-75.0%) | 384.5 MiB (-75.0%) |
| Per-layer text embeddings | 2240.8 MiB | 2240.5 MiB (-0.01%) | 2240.5 MiB (-0.01%) |
| Audio embeddings | 588.5 MiB | 308.0 MiB (-47.7%) | 179.6 MiB (-69.5%) |
| Vision embeddings | 353.3 MiB | 163.0 MiB (-53.9%) | 163.0 MiB (-53.9%) |
| **Total export size** | **6.789 GiB (baseline)** | **5.205 GiB (-23.3%)** | **4.271 GiB (-37.1%)** |
| **WWB similarity** | **0.9802 (baseline)** | **0.9736 (-0.0066)** | **0.9626 (-0.0176)** |

### Unified 12B

| OpenVINO component | Omitted format | `int8` | `int4` |
| --- | ---: | ---: | ---: |
| Language model | 11370.3 MiB | 11370.3 MiB (0%) | 6572.7 MiB (-42.2%) |
| Text embeddings | 960.8 MiB | 960.5 MiB (-0.03%) | 960.5 MiB (-0.03%) |
| Audio embeddings | 4.7 MiB | 2.4 MiB (-49.7%) | 2.4 MiB (-49.7%) |
| Vision embeddings | 111.7 MiB | 47.8 MiB (-57.3%) | 47.8 MiB (-57.3%) |
| **Total export size** | **12.206 GiB (baseline)** | **12.141 GiB (-0.5%)** | **7.456 GiB (-38.9%)** |
| **WWB similarity** | **1.0000 (baseline)** | **0.9868 (-0.0132)** | **0.9451 (-0.0549)** |

## Compatibility

- Standard Gemma 4 audio export requires Transformers 5.5 or newer.
- Gemma 4 Unified requires Transformers 5.10 or newer.
- Checkpoints without `audio_config` continue to load without an audio component.
- Audio features are forwarded only during generation prefill; cached decoding does not rerun the audio tower.

## Before submitting

- [ ] This PR fixes a typo or improves the docs (not applicable).


tests/openvino/test_seq2seq.py
```diff
@@ -602,7 +602,7 @@ class OVModelForVisualCausalLMIntegrationTest(OVSeq2SeqTestMixin):
         "qwen3_omni_moe",
     ]
     SUPPORT_VIDEO = ["llava_next_video", "qwen2_vl", "qwen2_5_vl", "qwen3_vl", "videochat_flash_qwen"]
-    SUPPORT_AUDIO = ["qwen3_omni_moe"]
+    SUPPORT_AUDIO = ["gemma4", "qwen3_omni_moe"]
```
rkazants: please double check that here we validate image+text and audio+text separately

rkazants (commented):
please add multi-batch demo support in PR description. I would like to check that multi-batch is supported on real model

tests/openvino/test_seq2seq.py
rkazants: please add gemma4_unified and make sure that it is validated on the precommit

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -5618,6 +5781,15 @@ def preprocess_inputs(
...
             conversation[0]["content"].insert(0, {"type": "image"})
+        if audio is not None:
+            conversation[0]["content"].append({"type": "audio", "audio": audio})
+            return processor.apply_chat_template(
+                conversation,
+                add_generation_prompt=True,
+                tokenize=True,
+                return_dict=True,
+                return_tensors="pt",
+            )
```
popovaan: Is this line needed? We apply chat template and tokenize below (5794-5795 lines), can we do the same with audio?

tests/openvino/test_seq2seq.py
as-suvorov: Double checked, they are

as-suvorov: model type added

as-suvorov:
@rkazants @popovaan 
Current proposal is to export audio embeddings by default with `--task image-text-to-text`.
How do you think, is it ok default scenario or we need some additional task like `omni` when we export every modality available?
As I understand qwen3-omni default task is `image-text-to-text`.

If we go with default exported audio modality, I also need regenerated tiny models with audio embeddings.

as-suvorov:
> please add multi-batch demo support in PR description. I would like to check that multi-batch is supported on real model

multi-batch sample added, supported on a real model

optimum/intel/openvino/modeling_visual_language.py
as-suvorov: not needed, removed, thanks for catch

rkazants (approved):

rkazants:
@as-suvorov, please fix:
FAILED tests/openvino/test_seq2seq.py::OVModelForVisualCausalLMIntegrationTest::test_compare_to_transformers_7_gemma4_unified - ValueError: Audio inputs were provided, but the audio embeddings model is not available.
FAILED tests/openvino/test_seq2seq.py::OVModelForVisualCausalLMIntegrationTest::test_generate_utils_7_gemma4_unified - TypeError: only a single or a list of entries is supported but got type=<class 'tuple'>

lcretan:
Thanks for pushing this forward, @as-suvorov — the coverage for the standard E2B/E4B checkpoints with an audio tower looks good, and having audio gated on `audio_config` is the right shape.

I wanted to add some concrete detail on the `gemma4_unified` case, since I think it explains the two failing tests rather than being separate from them, and since anyone hitting

```
ValueError: Audio inputs were provided, but the audio embeddings model is not available.
```

will probably land on this thread looking for the reason.

## What "Unified" changes

The distinguishing property of Gemma 4 12B **Unified** is the complete elimination of per-modality encoders. There is no audio tower, no vision tower feeding an adapter — every modality is projected straight into the shared latent space. For audio that means `model.embed_audio` is a `Gemma4UnifiedMultimodalEmbedder` — RMSNorm followed by `Linear(640 → 3840)` — consuming raw frames of shape `[batch, n_frames, 640]` directly.

This is visible in the checkpoint. For `google/gemma-4-12B-it`:

- `config.json` → `model_type: gemma4_unified`, and `audio_config` **is** present, so the `if config.audio_config is None` gate passes and audio export is attempted.
- That `audio_config`, however, carries only projection parameters — `audio_embed_dim`, `audio_samples_per_token`, `output_proj_dims`, `rms_norm_eps`, `hidden_size` — with no layer or attention configuration anywhere in it.
- `model.safetensors` contains exactly **one** audio tensor: `model.embed_audio.embedding_projection.weight`. There is no `audio_tower` in the checkpoint at all.

So on Unified checkpoints, this path in `modeling_visual_language.py`

```python
self.audio_tower = model.model.audio_tower
audio_outputs = self.audio_tower(input_features, input_features_mask, return_dict=True)
audio_features = self.embed_audio(inputs_embeds=audio_outputs.last_hidden_state)
```

has nothing to call. `audio_config` is truthy, so export proceeds; the tower it then reaches for does not exist, and the audio embeddings component never materialises — which is what surfaces later as the error above, and I believe what `test_compare_to_transformers_7_gemma4_unified` and `test_generate_utils_7_gemma4_unified` are reporting.

## How #1813 handles it

#1813 was written specifically around this architectural change, and treats Unified as its own case rather than a tower-based model with a shorter chain:

- **`optimum/exporters/openvino/model_patcher.py`** — a `Gemma4UnifiedAudioEmbeddingsModelPatcher` (~27 lines) that unwraps `embed_audio` into a traceable single-input forward: `input_features [batch, n_frames, 640]` → `last_hidden_state [batch, n_frames, 3840]`. There are no data-dependent operations in this embedder, so the patcher stays small — its job is only to expose the signature tracing needs.
- **`tests/openvino/utils_tests.py`** — `audio_embeddings_model` added to the `gemma4_unified` expectations, so the `openvino_audio_embeddings_model.*` artifact is asserted on export and covered by INT8 quantization checks, on the path `test_export.py` and `test_seq2seq.py` already parametrize.

`model_patcher.py` is the one file #1902 does not touch, and on this model it is where the Unified-specific handling has to live.

I mention this because the combination above is what produces a working IR here: I have been running Gemma 4 12B Unified with audio active on an Intel Arc iGPU from an OpenVINO IR built this way, so the approach is exercised beyond the test suite.

## Offer

Happy to help land this however suits you — rebasing #1813's Unified handling onto this branch, or opening it as a follow-up against yours once this merges. Either is completely fine.

@rkazants — on #1813 itself, you asked me to fold the standalone `test_gemma4_audio.py` into the existing suites rather than keep it separate, and I pushed that in `5d59cbc` on 25 June. It has been quiet since, so I wanted to ask whether there is anything further you would like on it, or whether you would prefer the Unified coverage to land through #1902 instead. Genuinely either way — I would just like the Unified path covered before this merges, since it is the case failing today.

Thanks again for the work on this.

optimum/exporters/openvino/model_configs.py
```diff
@@ -5029,6 +5076,9 @@ def __init__(
             behavior=behavior,
         )
         self._behavior = behavior
+        self.SUPPORTED_BEHAVIORS = list(type(self).SUPPORTED_BEHAVIORS)
+        if config.audio_config is None:
+            self.SUPPORTED_BEHAVIORS.remove(Gemma4UnifiedConfigBehavior.AUDIO_EMBEDDINGS.value)
```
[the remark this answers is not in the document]
as-suvorov: This is common pattern for optimum intel. Enums are inherit from str so they are comparable

optimum/exporters/openvino/model_configs.py
```diff
@@ -4701,6 +4704,9 @@ def __init__(
             behavior=behavior,
         )
         self._behavior = behavior
+        self.SUPPORTED_BEHAVIORS = list(type(self).SUPPORTED_BEHAVIORS)
+        if config.audio_config is None:
+            self.SUPPORTED_BEHAVIORS.remove(Gemma4ConfigBehavior.AUDIO_EMBEDDINGS.value)
```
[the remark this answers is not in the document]
as-suvorov: Implemented

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -5480,20 +5524,70 @@ def prepare_inputs_for_generation(
...
+        if processor is None:
+            raise ValueError("Processor is required.")
+        if video is not None:
+            raise ValueError("Video input is not supported")
+
+        conversation = [{"role": "user", "content": [{"type": "text", "text": text}]}]
+        if image is not None:
+            conversation[0]["content"].insert(0, {"type": "image", "image": image})
+        audio_inputs = audio if isinstance(audio, (list, tuple)) else [audio]
```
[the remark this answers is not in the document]
as-suvorov: Fixed

optimum/exporters/openvino/input_generators.py
```diff
@@ -494,6 +494,58 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
         return super().generate(input_name, framework, int_dtype, float_dtype)
 
 
+class DummyGemma4UnifiedAudioInputGenerator(DummyInputGenerator):
+    SUPPORTED_INPUT_NAMES = ("input_features",)
```
popovaan, line 502: Can we reuse `FunASRDummyAudioInputGenerator` instead of creating this class?

popovaan (approved):

optimum/exporters/openvino/input_generators.py
as-suvorov: To my knowledge, we cannot reuse it. Unified gemma has `audio_embed_dim` dimension instead of input features of funasr.

optimum/exporters/openvino/input_generators.py
```diff
@@ -494,6 +494,58 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+
+    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+        return self.random_float_tensor(
+            shape=[self.batch_size, self.sequence_length, self.audio_embed_dim],
+            framework=framework,
+            dtype=float_dtype,
+        )
+
+
+class DummyGemma4AudioInputGenerator(DummyInputGenerator):
```
rkazants, line 526: let us have just this input generatotr for both gemma4 and gemma4_unified. Also, please provide description to this class: what is `input_features` how can we get them in the whole pipeline? Describe a bit audio pre-processing.

optimum/exporters/openvino/model_configs.py
```diff
@@ -5131,11 +5142,38 @@ def with_behavior(self, behavior: Union[str, Gemma4ConfigBehavior]):
...
+                    self.audio_tower = model.model.audio_tower
+                    self.embed_audio = model.model.embed_audio
+
+                def forward(self, input_features: torch.Tensor, input_features_mask: torch.Tensor):
+                    audio_outputs = self.audio_tower(input_features, input_features_mask, return_dict=True)
+                    audio_features = self.embed_audio(inputs_embeds=audio_outputs.last_hidden_state)
+                    audio_mask = getattr(audio_outputs, "attention_mask", None)
+                    if audio_mask is None:
+                        audio_mask = audio_outputs.audio_mel_mask
+                    return audio_features, audio_mask
```
rkazants, line 5294: can you please chech if received IR is optimal. It should contain SDPA, Rope internal after compilation. Is there any operation that can be fused?

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -393,9 +402,19 @@ class OVMultiModalProjector(OVVisionProjection):
...
+            inputs = args[0]
+        elif args:
+            input_names = list(self.input_names.keys())
+            inputs = {input_names[index]: value for index, value in enumerate(args)}
+        else:
+            inputs = {name: value for name, value in kwargs.items() if name in self.input_names}
+        result = self.request(inputs)
+        if len(result) > 1:
+            return result[0], result[1]
+        return result[0]
```
rkazants: why do we need this change?

tests/openvino/utils_tests.py
```diff
@@ -211,7 +211,7 @@ def _create_tiny_kokoro_model():
     "gemma3n_text": "optimum-intel-internal-testing/tiny-random-gemma3n-text",
     "gemma4": "optimum-intel-internal-testing/tiny-random-gemma4",
     "gemma4_moe": "optimum-intel-internal-testing/tiny-random-gemma4-moe",
-    "gemma4_unified": "optimum-intel-internal-testing/tiny-random-gemma4-unified",
+    "gemma4_unified": "optimum-intel-internal-testing/tiny-random-gemma4-unified-it",
```
rkazants: why was it changed?
Let us test both instruct and non-instruct

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -5481,20 +5524,73 @@ def prepare_inputs_for_generation(
         # Map mm_token_type_ids to token_type_ids for the OV language model input
         model_inputs["token_type_ids"] = mm_token_type_ids
         model_inputs["image_position_ids"] = image_position_ids
+        model_inputs["input_features"] = input_features if past_key_values is None else None
+        model_inputs["input_features_mask"] = input_features_mask if past_key_values is None else None
```
rkazants, line 5656: looks not needed change, please check

optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -5570,15 +5711,33 @@ def prepare_inputs_for_generation(
...
+        if input_features is None:
+            input_features = input_audio_embeds
+        if input_features_mask is None:
+            input_features_mask = audio_attention_mask
         if token_type_ids is None and mm_token_type_ids is not None:
             token_type_ids = mm_token_type_ids
         return super().forward(
             input_ids=input_ids,
             pixel_values=pixel_values,
+            input_audio_embeds=input_features,
```
rkazants: why don't we call `input_features` as `input_audio_embeds` initially? Now it is confusing because it looks that these embeds can correspond to other modalities

rkazants:
@as-suvorov, please also rebase. It should help to fix red CI

optimum/exporters/openvino/input_generators.py
as-suvorov: Having 2 dedicated audio inputs generators is aligned with the 2 dedicated visual generators.
Docstrings added

optimum/exporters/openvino/model_configs.py
as-suvorov: IR analysis:
* Audio attention uses chunked learned relative-position logits via relative_k_proj, not rotary embeddings. RoPE belongs to the text/vision configs in config.json:124.
* Both CPU and GPU retain 12 decomposed attention blocks: 36 MatMuls and 12 Softmaxes.
* Standard SDPA fusion is blocked by chunk unfolding, relative-key logits, masking, and tanh logit capping between QKt and Softmax.

Existing CPU fusions:
* 135 FullyConnected nodes, mostly BF16
* 111 fused RMS nodes
* Linear-adjacent clamps folded where possible
* Graph reduced from 4,540 IR operations to 3,359 runtime nodes

For a representative [1, 100, 128] input, fully connected layers consumed about 49.2 ms of 57.5 ms. Attention MatMuls plus Softmax consumed only about 0.76 ms. Fusing attention to custom op unlikely to provide meaningful improvement.

optimum/intel/openvino/modeling_visual_language.py
as-suvorov: We need to pass and return both input features and mask. Implementation simplified

tests/openvino/utils_tests.py
as-suvorov: I keep both `-it` and non `-it` model in the list. Non `-it` model for the moment doesn't have audio config and thus doesn't support audio inputs. Consider re generation aligned with `-it` model.
[`"audio_config": null,`](https://huggingface.co/optimum-intel-internal-testing/tiny-random-gemma4-unified/blob/main/config.json#L5)

optimum/intel/openvino/modeling_visual_language.py
as-suvorov: Double checked, it needed

optimum/intel/openvino/modeling_visual_language.py
as-suvorov: Yeah, it confusing indeed. There are legacy inputs named `input_audio_embeds` which were introduced with some phi model. Renamed to `audio_input_features`

as-suvorov:
@rkazants regarding skipped gemma4 model and CVS-193103. We had same issue with gemma4 unified -it and regeneration with different parameters helped to workaround segfaults. Please consider regenerating gemma4 model model as well. You can use https://huggingface.co/optimum-intel-internal-testing/tiny-random-gemma4-unified-it as a reference.
Could you please also run CI

rkazants (commented):
please make sure that int4 compression still works. Does it work on GPU?

as-suvorov:
> please make sure that int4 compression still works. Does it work on GPU?

I added int4 compression config for e2B e4B audio embedding models. 12B model has only projections in embeddings and there is no point in int4 compression there. Please see updated accuracy results in the description.

I checked e2b int4 model on GPU - works good. I cannot check 12B on GPU on my machine.

optimum/intel/openvino/configuration.py
```diff
@@ -545,6 +555,7 @@ class OVQuantizationMethod(str, Enum):
 model_id_aliases = [
     ("meta-llama/Meta-Llama-3.1-8B-Instruct", "meta-llama/Llama-3.1-8B-Instruct"),
     ("meta-llama/Meta-Llama-3.1-8B", "meta-llama/Llama-3.1-8B"),
+    ("google/gemma-4-E4B-it", "google/gemma-4-e2B-it"),
```
regisss: E4B and E2B are two different models, why aliasing one with the other?

optimum/exporters/openvino/input_generators.py
```diff
@@ -494,6 +494,67 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+
+    SUPPORTED_INPUT_NAMES = ("input_features", "input_features_mask")
+
+    def __init__(
+        self,
+        task: str,
+        normalized_config: NormalizedConfig,
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        # Roughly one second of audio at the feature extractor's 10 ms hop
+        sequence_length: int = 100,
```
echarlaix: why not use DEFAULT_DUMMY_SHAPES["sequence_length"] ? would this have an impact on export ?

regisss:
Some Gemma4-related tests failed, can you take a look @as-suvorov ?

optimum/exporters/openvino/input_generators.py
```diff
@@ -494,6 +494,67 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
...
+        **kwargs,
+    ):
+        self.task = task
+        self.normalized_config = normalized_config
+        self.batch_size = batch_size
+        self.sequence_length = sequence_length
+        self.feature_size = getattr(normalized_config, "feature_size", 128)
```
echarlaix: why 128 as default ? and why a different behavior than DummyAudioInputGenerator https://github.com/huggingface/optimum/blob/52367da7f2d227d82db26d0de10f44c297b8c1eb/optimum/utils/input_generators.py#L891

rkazants (commented):
I think that audio part is currently is not int8 compressed by default unlike other parts. The reason is that quantization config is not updated with audio part. Please check.

as-suvorov:
> I think that audio part is currently is not int8 compressed by default unlike other parts. The reason is that quantization config is not updated with audio part. Please check.

Audio part has special handling for int4 compression only as initially requested by you. No special handling for int8, same behavior like other parts

optimum/intel/openvino/configuration.py
as-suvorov: I misused alias list, reverted it. I added E2B model to the int4 dict.

as-suvorov:
> Some Gemma4-related tests failed, can you take a look @as-suvorov ?

Thanks for review. There was an issue after master update on this branch. Tests fixed now. Except new dflash2 tests which fail on master:
https://github.com/huggingface/optimum-intel/actions/runs/35194620890/job/105114815280
https://github.com/huggingface/optimum-intel/actions/runs/35194620890/job/105114815063
and out of the scope of PR.

regisss (approved):
LGTM

rkazants:
@as-suvorov, please fix [Quality Checks / quality (pull_request)](https://github.com/huggingface/optimum-intel/actions/runs/35216139777/job/105244548164?pr=1902)Failing after 23s
and rebase to latest main

as-suvorov:
> @as-suvorov, please fix [Quality Checks / quality (pull_request)](https://github.com/huggingface/optimum-intel/actions/runs/35216139777/job/105244548164?pr=1902)Failing after 23s and rebase to latest main

@rkazants Black fixed, could you please re-run CI

rkazants (approved):
Also, tested int4 on B70, it looks working.

PR #1911, opened 2026-08-03
feat(openvino): enable Gemma 4 MTP assistant export for speculative decoding

### LINKED ISSUE #1878: [OpenVINO] Support Speculative Decoding in Gemma 4 with model type 'gemma4_assistant' and 'gemma4_unified_assistant'
Hi. I plan to use the assistant model of Gemma 4 for Speculative Decoding (i.e. Multi-token Prediction). Unfortunately, it seems to be not supported here yet. The issue is similar to https://github.com/huggingface/optimum-intel/issues/1764, which has been resolved for 12B model (right?). I found that @openarcbob also mentioned the export for the assistant model there, but the topic seemed to be forgotten or overlooked, and the issue became focusing on 12B model in general instead. That's why I separate the topic to here instead. 

Here is the packages' version:
- openvino 2026.2.1
- openvino-tokenizers 2026.2.1.0
- optimum 2.2.0.dev0
- optimum-intel 2.1.0.dev0+74f530e
- transformers 5.5.4

And I got the same error as in the source issue:
```
optimum-cli export openvino --model google/gemma-4-12B-it-qat-q4_0-unquantized-assistant --task image-text-to-text gemma-4-12B-it-qat-q4_0-unquantized-assistant
```
> ValueError: The checkpoint you are trying to load has model type gemma4_unified_assistant but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.

Then I updated transformers to recognize such model type:
```
# 5.10 is needed for 'gemma4_unified[_asisstant]' model type (or just 5.8.0 for 'gemma4_assistant' model type)
pip install transformers==5.10 
```
> ValueError: Trying to export a gemma4_unified_assistant model, that is a custom or unsupported architecture, but no custom export configuration was passed as custom_export_configs. Please refer to https://huggingface.co/docs/optimum/main/en/exporters/onnx/usage_guides/export_a_model#custom-export-of-transformers-models for an example on how to export custom models. Please open an issue at https://github.com/huggingface/optimum-intel/issues if you would like the model type gemma4_unified_assistant to be supported natively in the OpenVINO export.

abhays02:
### The problem

Try exporting a Gemma 4 *assistant* (drafter) model for speculative decoding:

```bash
optimum-cli export openvino --model google/gemma-4-12B-it-qat-q4_0-unquantized-assistant ...
```

It blows up. The parent models (`gemma4`, `gemma4_unified`) work fine — but the assistant variants were never registered in the tasks manager, even though they're the exact same architecture.

### The fix

**4 decorator registrations** + **2 list entries**. That's it.

The assistant models are architecturally identical to their parents — just smaller drafters for MTP. So we reuse the existing config classes:

- `model_configs.py` — stacked `@register_in_tasks_manager` decorators for `gemma4_assistant`, `gemma4_unified_assistant`, and their `-text` variants
- `utils.py` — added the two assistant types to `SUPPORTED_VLM_MODELS`

### Why it matters

Speculative decoding with a drafter model can **2-3x speed up** Gemma 4 inference in OpenVINO. Right now it's blocked by a simple registration gap.

### Tested on

- `openvino 2026.2.1` · `optimum-intel 2.1.0.dev0` · `transformers 5.10`

Fixes #1878


rkazants:
already opened PR

PR #1913, opened 2026-08-03
Create test_irs.py to detect changes in tiny-model IRs

anatyrova:
# What does this PR do?

This PR implements test_irs.py to detect changes in tiny-model IRs against reference IRs.
It will allow us to accelerate moving to new transformers version by default and check that IRs generated for newer transformers versions stay performant. 

For now it compares the non-trust-remote-code exported models in 5.5.4 transformers. It uses ov branches in optimum-intel-internal-testing.

```
pytest tests/openvino/test_irs.py -v
```

eg. small part of failed traceback:
```
E             Not equal output element types of __module.encoder.block.4.layer.1.layer_norm/aten::mul/Multiply_1 and Constant_35780 on 0 index model_one: <Type: 'float32'>, model_two: <Type: 'int64_t'>.
E             Not equal op partial shapes of self.encoder.block.4.layer.1.DenseReluDense.wi.weight and __module.encoder.final_layer_norm/aten::mean/ReduceMean on 0 index model_one: [37,32], model_two: [?,?,1].
E             Not equal op partial shapes of __module.encoder.block.4.layer.1.DenseReluDense.wi/aten::linear/MatMul and Constant_39730 on 0 index model_one: [?,?,37], model_two: [1,1,1].
E             Not equal op partial shapes of __module.encoder.block.4.layer.1.DenseReluDense.act/aten::relu/Relu and __module.encoder.final_layer_norm/aten::add/Add on 0 index model_one: [?,?,37], model_two: [?,?,1].
E             Not equal op partial shapes of self.encoder.block.4.layer.1.DenseReluDense.wo.weight and __module.encoder.final_layer_norm/aten::rsqrt/Sqrt on 0 index model_one: [32,37], model_two: [?,?,1].
E             Not equal op partial shapes of __module.encoder.block.4.layer.1.DenseReluDense.wo/aten::linear/MatMul and __module.encoder.final_layer_norm/aten::rsqrt/Divide on 0 index model_one: [?,?,32], model_two: [?,?,1].
E             Not equal op partial shapes of Constant_40674 and __module.encoder.final_layer_norm/aten::mul/Multiply_1 on 0 index model_one: [1,1,1], model_two: [?,?,32].
E             Not equal op partial shapes of Constant_40672 and Result_35882 on 0 index model_one: [1,1,1], model_two: [?,?,32].
E           
E           Reference IR was generated with:
E             - transformers: 5.5.4
E             - optimum-intel: 2.1.0.dev0+fd94990
E             - openvino: 2026.2.1-21919-ede283a88e3-releases/2026/2
E             - date: 2026-08-03
E           
E           This may indicate a regression or improvement in IR generation. Review the changes and update reference IRs if the change is intentional.

tests/openvino/test_irs.py:222: Failed
```

## Before submitting


rkazants (commented):
@anatyrova 
please revise changes, something wrong +133,707

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+    ir_files = []
+    for xml_file in directory.rglob("openvino_model.xml"):
+        # Get relative path from directory
+        rel_path = xml_file.relative_to(directory)
+        ir_files.append(rel_path)
+    return sorted(ir_files)
+
+
+def parse_ir_xml(xml_path: Path) -> Dict:
```
rkazants: please use already existing functionality for ov::Model comparation
https://github.com/openvinotoolkit/openvino/blob/master/src/tests/test_utils/common_test_utils/src/graph_comparator.cpp

tests/openvino/reference_irs/tiny-random-flux/text_encoder/config.json
rkazants, line 1: please remove json files

rkazants (commented):
how long will it take if we generate reference IRs during testing time?

optimum/intel/openvino/modeling_diffusion.py
```diff
@@ -2190,6 +2190,9 @@ def __init__(
         self.vae = OVModelVae(decoder=self.vae_decoder, encoder=self.vae_encoder)
         self.scheduler = scheduler
         self.tokenizer = tokenizer
+        self.tokenizer_2 = None
+        self.tokenizer_3 = None
+        self.feature_extractor = None
```
anatyrova, line 2217: these for ovltx2pipeline to work with save_pretrained()

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,568 @@
...
+        if arch in REMOTE_CODE_MODELS:
+            continue
+
+        if arch in {  # seq2seq models
+            "bigbird_pegasus",
+            "blenderbot-small",
+            "longt5",
+            "m2m_100",
+            "mbart",
+            "t5",
```
anatyrova, line 277: These models currently fail the test because they change the order of operations and cannot be compared as the function comapre_models() does it. This will be fixed in the next pr for test_irs

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,568 @@
+"""
+Test IR (Intermediate Representation) stability across transformers versions.
+
+This test compares newly exported OpenVINO IRs against reference IRs to detect
+regressions when upgrading transformers or other dependencies.
+"""
```
rkazants, line 6: please add instructions in the comment: how to add new model in this test suite, how to generate IR and where we should upload it.

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,568 @@
...
+    "OVStableDiffusion3Pipeline": OVStableDiffusion3Pipeline,
+    "OVStableDiffusionPipeline": OVStableDiffusionPipeline,
+    "OVStableDiffusionXLImg2ImgPipeline": OVStableDiffusionXLImg2ImgPipeline,
+    "OVStableDiffusionXLPipeline": OVStableDiffusionXLPipeline,
+}
+
+
+# Additional architecture-to-class mappings from test_export.py, test_decoder.py, and test_quantization.py
+# Merged with ARCH_TO_MODEL_CLASS for more complete coverage
+ADDITIONAL_ARCH_MAPPINGS = {
```
rkazants, line 96: I think we have similar mapping already. Please try to re-use it and extend if needed.

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,568 @@
...
+)
+
+
+# Add tests/openvino to path to import utils_tests
+sys.path.insert(0, str(PathlibPath(__file__).parent))
+from utils_tests import ARCH_TO_MODEL_CLASS, HUB_MODEL_NAMES, REMOTE_CODE_MODELS
+
+
+# Map class name strings to actual class objects
+CLASS_NAME_TO_CLASS = {
```
rkazants, line 60: looks already exists, try to re-use it

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,568 @@
...
+        params.append((model_id, model_class))
+    return params
+
+
+TEST_MODELS = _generate_test_params()
+
+
+# OpenVINO model comparison functions
+# Adapted from: https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/tests/utils/helpers.py
+def _compare_models(model_one: Model, model_two: Model, compare_names: bool = True) -> tuple[bool, str]:
```
rkazants, line 346: let us move it to separate file like test_irs_utils.py And place all utils for graph comparation there.

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,568 @@
+"""
+Test IR (Intermediate Representation) stability across transformers versions.
+
+This test compares newly exported OpenVINO IRs against reference IRs to detect
+regressions when upgrading transformers or other dependencies.
+"""
+
+import json
+
```
echarlaix, line 9: Also would make sense to have a new workflow (that can be triggered manually) to push models directly on the hub (under a specific ov branch) when needed / when changes in the IRs are validated

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,568 @@
...
+            "granitemoehybrid",
+            "mamba",
+        }:
+            continue
+
+        if arch in {  # max transformers 5.4.0 required
+            "lfm2",
+            "lfm2_moe",
+        }:
+            continue
```
echarlaix, line 323: because MIN/MAX_TRANSFORMERS_VERSION can be updated I think it makes more sense to use `is_model_type_transformers_compatible` / `get_supported_model_for_library` to filter instead

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,568 @@
...
+                    repo_id=model_id,
+                    revision="ov",
+                    repo_type="model",
+                )
+            )
+        except Exception as e:
+            pytest.skip(f"Failed to download reference IRs from {model_id} (ov branch): {e}")
+
+        if not ref_ir_dir.exists():
+            pytest.skip(f"Reference IR directory not found for {model_id}")
```
echarlaix, line 477: will be an issue as some tests might silently fail (read token revoked, model is pushed under an unexpected branch,...)

tests/openvino/test_irs.py
```diff
@@ -0,0 +1,568 @@
...
+                f"date={ref_metadata.get('generated_date', 'unknown')}"
+            )
+        else:
+            print(f"[IR-DEBUG] {model_id}: no reference metadata.json found in {ref_ir_dir}")
+
+        # Find all IR files (handles both single models and multi-component pipelines)
+        ref_ir_files = find_ir_files(ref_ir_dir)
+
+        if not ref_ir_files:
+            pytest.skip(f"No openvino_model.xml files found in {ref_ir_dir}")
```
echarlaix, line 497: in which case could this happen ? (no openvino models present in our tests models repos / ov branch)

echarlaix (commented):
thanks for the PR @anatyrova

tests/openvino/test_irs.py
anatyrova: i believe it's easier just to delete them from test_irs, since we run models only on latest transformers. will fix that

tests/openvino/test_irs.py
anatyrova: if model is present in test_irs but it does not have ov branch, the test is skipped. usually model doesn't have the ov branch with exported model, because its transformers version is too new (higher than 5.10.4)

PR #1922, opened 2026-08-10
[WIP] Add INT2 (u2) weight format support for OpenVINO export

Cyberpunk1210:
## What does this PR do?

Adds INT2 (2-bit) weight format support to the OpenVINO export path:
 - Accepts `bits=2` end-to-end through the export CLI/config instead of
   rejecting it.
 - Stops silently overriding `sym=True` for INT2 when the installed NNCF can
   express `CompressWeightsMode.INT2_ASYM` (falls back with a warning only
   when NNCF genuinely lacks it -- see nncf PR
   https://github.com/openvinotoolkit/nncf/pull/4168).
 - Extends default `ignored_scope` matching to also match by local folder
   name, not just hub repo id, so local/renamed checkpoints get the right
   default recipe.
 - Adds a default `ignored_scope` for Qwen/Qwen3.6-35B-A3B (MoE routed
   experts excluded from the non-expert quantization pass).

Tested by exporting a 256-expert Qwen3.6-35B-A3B checkpoint end-to-end with
`bits=2` and running the result on OpenVINO GPU/CPU (companion openvino PR:
https://github.com/openvinotoolkit/openvino/pull/37335); accuracy validated
on BFCL (72% in think mode, symmetric INT2 experts).

No new automated tests were added for this change; validation was done via
the manual export + inference run described above.

## Before submitting
- [ ] This PR fixes a typo or improves the docs.

### DIFF
--- a/optimum/commands/export/openvino.py
+++ b/optimum/commands/export/openvino.py
@@ -67,7 +67,7 @@ def parse_args_openvino(parser: "ArgumentParser"):
     optional_group.add_argument(
         "--weight-format",
         type=str,
-        choices=["fp32", "fp16", "int8", "int4", "mxfp4", "nf4", "cb4"],
+        choices=["fp32", "fp16", "int8", "int4", "int2", "mxfp4", "nf4", "cb4"],
         default=None,
         help=(
             "The weight format of the exported model. Option 'cb4' represents a codebook with 16 fixed fp8 values in E4M3 format."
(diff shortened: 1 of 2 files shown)


rkazants (changes requested):
please add JIRA ticket number and proper PR description: cmd-line to export and inference script

optimum/intel/openvino/configuration.py
rkazants, line 1: tests are needed

optimum/commands/export/openvino.py
```diff
@@ -503,8 +503,9 @@ def run(self):
 
 def prepare_wc_config(args, default_configs):
     is_int8 = args.weight_format == "int8"
+    is_int2 = args.weight_format == "int2"
     return {
-        "bits": 8 if is_int8 else 4,
+        "bits": 8 if is_int8 else 2 if is_int2 else 4,
```
rkazants, line 508: How many models are validated with int2 compression and what are their accuracy, performance on GPU?

PR #1928, opened 2026-08-12
Remove _LTX2AttnProcessorWithEps patch for LTX2

anatyrova:
# What does this PR do?

This PR removes unnecessary patching of attention, keeps only RoPE patch applied. 
```
optimum-cli export openvino --model Lightricks/LTX-2 --task text-to-video output_dir
```

max_diff before change: 
```
video_text_embedding:
    max_abs_diff = 4.3411e-03

  audio_text_embedding:
    max_abs_diff = 5.6419e-03
```
max_diff after change:
```
video_text_embedding:
    max_abs_diff = 4.3411e-03

audio_text_embedding:
    max_abs_diff = 5.6419e-03
```


optimum/exporters/openvino/model_patcher.py
rkazants, line 1: please check if test_transformatons.py needs update.

anatyrova: ltx2 didn't exist before in test_transformations, but i will add with this pr

rkazants (approved):

PR #1930, opened 2026-08-13
Remove flags for fp16

anatyrova:
# What does this PR do?

[CVS-192141] 
This PR removes the flags and the condition that increases the models' sizes in fp16 weight format.


rkazants (changes requested):
Not clear why do we need?

anatyrova:
> Not clear why do we need?

this flag breaks fp16 weight compression for various models now, so their size gets bigger x2 or more with weight-format fp16 that it was before. it was added in https://github.com/huggingface/optimum-intel/pull/1733/changes

popovaan (approved):

rkazants (approved):

PR #1947, opened 2026-08-21
[OpenVINO] Use Grouped MatMul in Granite-Moe-Hybrid

popovaan:
# What does this PR do?

Ticket: CVS-186818

## Before submitting

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -328,16 +328,6 @@ def register_ov_batched_mm(patcher):
     patcher._model.set_experts_implementation("ov_batched_mm")
 
 
-def unregister_ov_grouped_mm(patcher):
-    """Restore grouped_mm_experts_forward to its original implementation."""
-    import transformers.integrations.moe as moe_module
-    from transformers.integrations.moe import ALL_EXPERTS_FUNCTIONS
-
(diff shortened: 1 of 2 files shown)


rkazants (commented):
thanks a lot. Needs tiny model with MoE and tests. I am afraid that the current tiny model does not contain MoE, it has only Mamba2.

Please also measure accuracy and perf.

popovaan:
Measured WWB on 6 samples, as the 7-th samples crushes. Original implementation crushes the same way.

int4 similarity 0.952045
int8 similarity 0.975022
 
LLM bench
int4: First token latency: 1289.59 ms, other tokens latency: 102.26 ms/token
int8: First token latency: 742.17 ms, other tokens latency: 117.58 ms/token

LLM bench on IR-s from master for comparison:
int4: First token latency: 8629.57 ms, other tokens latency: 187.95 ms/token
int8: First token latency: 10777.21 ms, other tokens latency: 206.18 ms/token

popovaan:
> thanks a lot. Needs tiny model with MoE and tests. I am afraid that the current tiny model does not contain MoE, it has only Mamba2.
> 

This tiny model has MoE experts and it is tested already: https://huggingface.co/optimum-intel-internal-testing/tiny-random-granitemoehybrid

I checked that the new patching is applied in this model.

PR #1692, opened 2026-04-16
Use latest OpenVINO 2026.1.0

rkazants:
# What does this PR do?
Install latest OpenVINO. The latest OpenVINO is required for MoE optimizations

## Before submitting
- [N/A] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [N/A] Did you make sure to update the documentation with your changes?
- [N/A] Did you write any new necessary tests?

PR #1953, opened 2026-08-25
Remove skip tests of >=2026.1.0

anatyrova:
# What does this PR do?

revert PR https://github.com/huggingface/optimum-intel/pull/1692


tests/openvino/test_seq2seq.py
rkazants, line 1: let us leave skips here since we have a trouble with seq2seq still

rkazants (approved):

optimum/exporters/openvino/model_configs.py
```diff
@@ -2056,7 +2042,6 @@ def __init__(
         if self._behavior == VLMConfigBehavior.VISION_EMBEDDINGS and hasattr(config, "vision_config"):
             self._config = config.vision_config
             self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
-        _warn_potential_accuracy_issue_ov_2026_1(self._OV_2026_1_MODEL_TYPE, min_transformers_version="5.0")
```
rkazants, line 2066: what is remained in seq2seq for them we shold not remove this warning.

PR #1960, opened 2026-08-28
Add support for `Qwen3Guard-Stream` model with `OVModelForGuard`

p-wysocki:
# What does this PR do?

 - Qwen3Guard-Stream is an Qwen3 decoder with LM head replaced with four token classification heads
 - `feature-extraction` has been extended to serve that
 - The branch runs the actual model correctly for both stateless and stateful pipelines

# Note

- This is my first contribution to optimum-intel, so I'm open to any discussion and feedback

# Ticket
 - CVS-191209


p-wysocki:
@rkazants @popovaan could you please review?

docs/source/openvino/inference.mdx
```diff
@@ -42,6 +42,7 @@ As shown in the table below, each task is associated with a class enabling to au
 | `OVModelForAudioClassification`      | `audio-classification`               |
 | `OVModelForImageClassification`      | `image-classification`               |
 | `OVModelForFeatureExtraction`        | `feature-extraction`                 |
+| `OVModelForGuard`                    | `feature-extraction-with-past`       |
```
rkazants, line 45: can we use existing inference class `OVModelForFeatureExtraction`?

rkazants: check qwen3-vl-embedding model how was it enabled.

p-wysocki: The existing `OVModelForFeatureExtraction` assumes the model only has 1 output: `last_hidden_state`, while `Qwen3Guard-Stream` replaces `Qwen3-0.6b`'s LM head with 4 classification heads, so the wrong tensor is being returned:

https://github.com/huggingface/optimum-intel/blob/dd4ed1a86f9c2abf34d8dce777da5b592765b644/optimum/intel/openvino/modeling.py#L404-L409

From what I see `qwen3-vl-embedding` also is satisfied with returning its hidden state.

Additionally, adding `OVModelForGuard` allows us to run a stateful model, which is useful for streaming, where tokens come in one by one.

We can extend `OVModelForFeatureExtraction` to work with that, but in my opinion it's cleaner to add `OVModelForGuard`. I'm happy to discuss though, as I mentioned in the PR description it's my first PR to optimum and I may not be getting everything right.

rkazants: @p-wysocki,  think it makes sense to extend `OVModelForFeatureExtraction` to support new output format with several tensors of embedding. Also, Qwen3-VL-Embedding is also stateful (with KV-cache) per my knowledge.
ModelForGuard sounds strange a bit, also not known term in domain of generative models, etc.

p-wysocki: I modified `OVModelForFeatureExtraction` to work with the Qwen3Guard, could you please take a look?

PR #1961, opened 2026-08-28
Update Gemma 4 12B Config

anzr299:
# What does this PR do?

Update Gemma 4 12B Config to not compress the vision embeddings model. This is done to resolve accuracy issues with compressed inference of vision embedding model on the GPU.

Fixes # (issue)

191056

## Before submitting


rkazants (changes requested):
provide all wwb accuracy metrics on CPU and GPU

PR #1906, opened 2026-08-01
[OpenVINO] Support MiniCPM-V-4.6 with task image-text-to-text

popovaan:
## Description

Fixed the MiniCPM-V-4.6 (minicpmv4_6) end-to-end HF-vs-OpenVINO greedy divergence that the prior session mis-attributed to an unfixable qwen3_5 OV-conversion decode defect. Root cause was a Python-side patcher bug (reproducible in pure PyTorch, not OpenVINO): Qwen3_5DynamicCacheWrap.get_mask_sizes returned kv_offset=past_length instead of 0, so transformers>=5.6 create_causal_mask shifted the full-attention key positions past the query and masked every cached token during decode (mask [0,-inf,-inf,...]); prefill matched, first decode token onward diverged. Per-layer hooks isolated it to the single full-attention layer (linear layers exact ~1e-6). Fix: kv_offset=0 (canonical DynamicLayer contract). Also fixed a test bug: minicpmv4_6 was missing from test_seq2seq get_transformer_model_class AutoModelForImageTextToText branch, causing an AutoModelForCausalLM load error before inference. After fixes, patched-eager and OV decode match HF exactly, CLI export exit 0, real image-text generate() produces diverse tokens, and all targeted repo tests pass.

## Conversion

```bash
optimum-cli export openvino --model openbmb/MiniCPM-V-4.6 output_dir --task image-text-to-text --trust-remote-code
```

## Reproduce generation

```python
from PIL import Image
from transformers import AutoProcessor
from optimum.intel.openvino import OVModelForVisualCausalLM

model_dir = "output_dir"
processor = AutoProcessor.from_pretrained(model_dir, trust_remote_code=True)
model = OVModelForVisualCausalLM.from_pretrained(model_dir, device="CPU")
image = Image.new("RGB", (64, 64), "white")
inputs = processor(images=image, text="Describe this image.", return_tensors="pt")
output_ids = model.generate(**inputs, max_new_tokens=10)
print(processor.batch_decode(output_ids, skip_special_tokens=True)[0])
```

## Validation

- Validation model: **Tiny model** (`/home/openvino_bot/.local/share/openvino-model-agent/requests/issue-36/repository/workspace/tiny-minicpmv4_6`)
- WWB similarity: **1.0**

## Related model-support PRs

- OpenVINO GenAI: https://github.com/openvinotoolkit/openvino.genai/pull/4249

## Before submitting

- [x] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

PR #1810, opened 2026-06-22
[OpenVINO] Support MiniCPM-V-4.6

openvino-agent:
# What does this PR do?

Installation instructions

```sh
pip install optimum-intel@
pip install transformers==5.7.0
```

Cmd-line for export:

```sh
optimum-cli export openvino -m openbmb/MiniCPM-V-4.6 MiniCPM-V-4.6
```

The script for inference:

```python
from transformers import AutoProcessor
from optimum.intel.openvino import OVModelForVisualCausalLM

model_id = "./MiniCPM-V-4.6"
#model_id = "openbmb/MiniCPM-V-4.6"
model = OVModelForVisualCausalLM.from_pretrained(model_id)

processor = AutoProcessor.from_pretrained(model_id, padding_side="left")

url = "https://media.istockphoto.com/id/1192867753/photo/cow-in-berchida-beach-siniscola.jpg?s=612x612&w=0&k=20&c=v0hjjniwsMNfJSuKWZuIn8pssmD5h5bSN1peBd1CmH4="
messages = [
    {
        "role": "system",
        "content": [
            {"type": "text", "text": "You are a helpful assistant."}
        ]
    },
    {
        "role": "user", "content": [
            {"type": "image", "url": url},
            {"type": "text", "text": "What is shown in this image?"},
        ]
    },
]
inputs = processor.apply_chat_template(
    messages,
    tokenize=True,
    return_dict=True,
    return_tensors="pt",
    add_generation_prompt=True,
)

output = model.generate(**inputs, max_new_tokens=50)
print(processor.decode(output[0, inputs.input_ids.shape[1]:], skip_special_tokens=True))
```

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1962, opened 2026-08-28
[OpenVINO] Enable MiniCPM-V-4.5 3D-Resampler temporal grouping

BiswarupMukherjee1:
## What does this PR do?

MiniCPM-V-4.5's 3D-Resampler jointly compresses several consecutive video frames into a single
`query_num`-token group. The OpenVINO path never consumed `temporal_ids`: `resampling()` did not
take them, and the resampler sub-model is driven with three per-frame inputs. So every frame
produced its own 64-token group, MiniCPM-V-2.6 behaviour, even though the checkpoint config
sets `batch_3d_resampler: true`.

Upstream performs the fold inside `Resampler.batch_attn_forward` using
`torch.nn.utils.rnn.pad_sequence`, which the OpenVINO PyTorch frontend cannot convert
(`aten::pad_sequence` has no translator and is listed under "Do not work on these for now" in
openvinotoolkit/openvino#28584).

**This PR does the fold on the host instead, and changes nothing in the graph.** `kv_proj`
(Linear) and `ln_kv` (LayerNorm over the last dim) act per token, so they commute with
regrouping; the per-frame temporal vector is therefore folded into the per-token `pos_embed`
that the existing graph already adds to `k`. Consequently:

* no new resampler IR input,
* no exporter or model-patcher change,
* no `aten::pad_sequence` support required,
* **already-exported resampler IRs run the grouped path unchanged.**

Temporal position embeddings are computed for the ids actually present rather than cached for
`max_temporal_size` (72000) rows, which at `hidden_size = 4096` would reserve ~1.2 GB for a
handful of rows that are ever indexed.

`temporal_ids=None` reproduces the previous per-frame behaviour bit-for-bit, so MiniCPM-V-2.6
and every image-only path are unaffected.

## Verification

Checked against the **unmodified upstream `Resampler`** with `batch_infer=True` (the path the
real checkpoint takes):

| case | result |
|---|---|
| uniform packing=2 (8 frames → 4 groups) | cosine 1.000000 |
| ragged packing=3 (8 frames → 3 groups) | cosine 1.000000 |
| packing=6 (6 frames → 1 group) | cosine 1.000000 |
| ragged `tgt_sizes` within a group | cosine 1.000000 |
| `temporal_ids=None` | **bit-identical** to current behaviour (`max\|diff\| == 0.0`) |

Also verified end to end through an OpenVINO IR: the three-input resampler graph was converted
once with dynamic shapes, then packings 1/2/3/4/6 were each driven through that single compiled
model, matching the reference every time.

The added unit test (`test_minicpmv_3d_resampler_temporal_grouping`) needs no model download —
it drives `resampling()` with a stub resampler and asserts the group count, the frame-major
fold, the ragged padding/mask, the "grouping must cover every frame" guard, and the
`temporal_ids=None` no-regression case.

I kept the test dependency-free deliberately, but `optimum-intel-internal-testing/tiny-random-minicpm-v-4_5`
already exists in the testing org (`version: 4.5`, `batch_3d_resampler: true`, `query_num: 64`)
and is not yet wired into `MODEL_NAMES`. Happy to add an end-to-end export/generate test on top
of it if you would prefer that, either here or as a follow-up.

## End-to-end demonstration

@rkazants  - here is a self-contained script showing exactly the case this fixes. It uses
MiniCPM-V-4.5's **own** `Resampler` as the reference, and downloads a single ~20 kB source file;
no model weights are required.

```python
# Demonstrates the case fixed by this PR, using MiniCPM-V-4.5's OWN Resampler as the
# reference. Downloads a single ~20 kB source file; no model weights required.
import importlib.util, types, torch, numpy as np
from optimum.intel.openvino.modeling_visual_language import _OVMiniCPMVForCausalLM

# --- MiniCPM-V-4.5's reference Resampler (the behaviour OpenVINO should reproduce) ---
spec = importlib.util.spec_from_file_location(
    "mcpm_resampler", hf_hub_download("openbmb/MiniCPM-V-4_5", "resampler.py"))
mcpm = importlib.util.module_from_spec(spec); spec.loader.exec_module(mcpm)

EMBED, KV, QUERIES = 128, 64, 64          # query_num=64 as in the 4.5 config
torch.manual_seed(0)
ref = mcpm.Resampler(num_queries=QUERIES, embed_dim=EMBED, num_heads=4, kv_dim=KV,
                     adaptive=False, max_size=(8, 8), max_temporal_size=64,
                     batch_infer=True).eval()          # batch_3d_resampler: true

# --- a 6-frame clip whose processor grouped frames into 2 temporal groups of 3 ---
FRAMES, temporal_ids = 6, [[0, 1, 2], [3, 4, 5]]
tgt_sizes = torch.tensor([[2, 3]] * FRAMES, dtype=torch.int32)
torch.manual_seed(1)
vision_out = torch.randn(FRAMES, 6, KV)               # vision-encoder output

with torch.no_grad():
    expected = ref(vision_out, tgt_sizes=tgt_sizes, temporal_ids=temporal_ids)

# --- drive optimum-intel's resampling() with a stand-in for the exported resampler IR ---
class _Model:
    resampling = _OVMiniCPMVForCausalLM.resampling
    _temporal_pos_embed_rows = _OVMiniCPMVForCausalLM._temporal_pos_embed_rows
    embed_dim, _pos_embeds = EMBED, ref.pos_embed.float()
    def _adjust_pos_cache(self, tgt_sizes): pass
    @torch.no_grad()
    def resampler(self, image_feature, pos_embed, key_padding_mask):   # == OVResamplerModel
        x = ref.ln_kv(ref.kv_proj(image_feature)).permute(1, 0, 2)
        q = ref.ln_q(ref.query).unsqueeze(1).repeat(1, x.shape[1], 1)
        o = ref.attn(q, x + pos_embed, x, key_padding_mask=key_padding_mask)[0]
        return (ref.ln_post(o.permute(1, 0, 2)) @ ref.proj).numpy()

m = _Model()
before = m.resampling(vision_out, tgt_sizes)                              # today
after  = m.resampling(vision_out, tgt_sizes, temporal_ids=temporal_ids)   # with this PR

cos = lambda a, b: float(np.dot(np.ravel(a), np.ravel(b)) /
                         (np.linalg.norm(a) * np.linalg.norm(b)))
print(f"frames={FRAMES}  temporal groups={len(temporal_ids)}  query_num={QUERIES}\n")
print(f"MiniCPM-V-4.5 reference : {tuple(expected.shape)}  -> {expected.shape[0]*QUERIES} vision tokens")
print(f"optimum-intel (before)  : {before.shape}  -> {before.shape[0]*QUERIES} vision tokens   <-- one group PER FRAME")
print(f"optimum-intel (after)   : {after.shape}  -> {after.shape[0]*QUERIES} vision tokens")
print(f"\ncosine(after, reference) = {cos(after, expected.numpy()):.6f}")
print(f"max|after - reference|   = {float(np.abs(np.asarray(after) - expected.numpy()).max()):.2e}")
```

Output:

```
frames=6  temporal groups=2  query_num=64

MiniCPM-V-4.5 reference : (2, 64, 128)  -> 128 vision tokens
optimum-intel (before)  : torch.Size([6, 64, 128])  -> 384 vision tokens   <-- one group PER FRAME
optimum-intel (after)   : torch.Size([2, 64, 128])  -> 128 vision tokens

cosine(after, reference) = 1.000000
max|after - reference|   = 1.67e-06
```

**The case fixed:** a 6-frame clip whose processor grouped the frames into 2 temporal groups.
MiniCPM-V-4.5's reference implementation emits **one 64-token group per temporal group** (2
groups -> 128 vision tokens). Before this PR, optimum-intel emitted **one group per frame** (6
-> 384 tokens) because `resampling()` had no way to receive `temporal_ids`. After it, the output
matches the reference exactly.

Beyond the token count, the mismatch is a correctness problem: the number of resampler rows must
equal the number of image placeholder slots the processor put in the prompt, so a grouped video
input produces a different embedding layout than the model was trained for.

## Rebase status

Rebased and re-verified against `main` @ `2039025` (2026-09-08). No conflicts; the new test and
`black`/`ruff` (pinned `black~=23.1`, `ruff==0.4.4`) all pass on the rebased tree. None of the commits merged since this PR opened touch `resampling()`, `get_vision_embeddings()` or any
temporal handling.

## Relationship to the other open MiniCPM PRs

* **#1941** adds `temporal_ids` to `forward()` / `prepare_inputs_for_generation()` so
  `generate()` stops raising `ValueError("model_kwargs not used: ['temporal_ids']")`. Its test
  notes the runtime should "accept and safely ignore" them — correct for still images (all ids
  are `-1`, contributing a zero embedding), but it means grouped **video** is silently
  discarded. This PR is the other half: #1941 makes the argument arrive, this makes it mean
  something. They compose; happy to rebase if #1941 lands first.
* **#1412** adds MiniCPM-V-4/4.5 decoder plumbing plus a 4-input resampler carrying
  `temporal_embed`, doing one attention call with `bs = n_frames`. That applies the temporal
  *embedding* but not the temporal *fold*, so it still emits one vector per frame where upstream
  emits one per group. Folding into `pos_embed` makes the fourth input unnecessary. Happy to
  rebase onto that branch instead if the maintainers prefer.
* **#1945 / #1906 / #1810** target MiniCPM-V-4.6, which replaced the resampler with a spatial
  `merge_kernel_size` merger and has no temporal path, so they do not cover this.

## Before submitting
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests)?

### DIFF
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -16,6 +16,7 @@ import warnings
 from abc import abstractmethod
 from collections import OrderedDict
 from dataclasses import dataclass
+from itertools import chain
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
@@ -2457,6 +2458,8 @@ class _OVMiniCPMVForCausalLM(OVModelForVisualCausalLM):
         if input_ids is not None and input_ids.shape[1] == 1:
             return None
         tgt_sizes = kwargs["tgt_sizes"]
+        # MiniCPM-V-4.5's image processor emits temporal groupings for video inputs
+        temporal_ids = kwargs.get("temporal_ids")
         pixel_values_list = pixel_values
         vision_hidden_states = []
         all_pixel_values = []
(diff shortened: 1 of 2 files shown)


rkazants (commented):
please provide e2e code snippet in PR description to demo what case is exactly fixed.

BiswarupMukherjee1:
> please provide e2e code snippet in PR description to demo what case is exactly fixed.

Thanks @rkazants , added an end-to-end snippet to the PR description. It uses MiniCPM-V-4.5's own Resampler as the reference and needs no model weights (~20 kB source download), so it runs as-is.

Summary: for a 6-frame clip grouped into 2 temporal groups, the reference emits 2 × 64 = 128 vision tokens, while optimum-intel currently emits 6 × 64 = 384, one group per frame instead of per temporal group, because resampling() had no way to receive temporal_ids. With this PR the output matches the reference at cosine 1.000000.

I've also rebased on current main (2039025) and re-run the test and style checks, no conflicts.

PR #1974, opened 2026-09-03
Fix error with timm on test_modeling.py

anatyrova:
# What does this PR do?

This PR changes the test_compare_to_timm reference-model call to pass the Hub repo id with an explicit hf-hub: prefix, which timm 1.0.29 now requires instead of implicitly treating a bare timm/... name as a Hub id.


echarlaix (approved):
Thanks for the fix !

PR #1976, opened 2026-09-04
Patch models with changed IRs for v5.10 rel.

anatyrova:
# What does this PR do?

This PR adds patches for models to make their exported IRs identical to the reference IRs. 

Patched models are:
1) tiny-random-SegformerModel
2) tiny-random-SwinModel (received 7e-7 difference)
3) tiny-random-DonutSwinModel

One patch of CLIPTextModelPatcher for:
4) tiny-stable-diffusion-torch
5) tiny-random-stable-diffusion-with-safety-checker
6) tiny-stable-diffusion-with-textual-inversion
7) tiny-random-stable-diffusion-xl
8) tiny-random-latent-consistency
9) tiny-random-flux
10) tiny-random-flux-fill
11) tiny-stable-diffusion-torch-custom-variant

12) gemma4
13) gemma4-moe (patched + new ref IR created)

14) tiny-random-squeezebert 
15) tiny-random-rembert
16) tiny-random-ConvBertForSequenceClassification
17) tiny-random-roformer 
18) tiny-random-MPNetModel 
19) all-mpnet-base-v2 
20) pix2struct-tiny-random 
21) trocr-small-handwritten 
22) VisionEncoderDecoderModel-vit-gpt2 
23) tiny-random-BeitForImageClassification 
24) tiny-random-DeiTModel 
25) tiny-random-vit 
26) tiny-random-ast

### DIFF
--- a/optimum/exporters/openvino/model_patcher.py
+++ b/optimum/exporters/openvino/model_patcher.py
@@ -16,6 +16,7 @@ import functools
 import inspect
 import logging
 import math
+import sys
 import types
 from dataclasses import dataclass
 from types import SimpleNamespace
@@ -114,6 +115,10 @@ if is_transformers_version(">=", "4.57"):
 
 if is_transformers_version(">=", "5"):
     from transformers.modeling_rope_utils import RotaryEmbeddingConfigMixin
+
+
+if is_transformers_version(">=", "5.6"):
+    from transformers.masking_utils import create_bidirectional_mask
 
 
 if is_diffusers_version(">=", "0.38.0"):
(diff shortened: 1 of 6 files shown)


optimum/exporters/openvino/model_patcher.py
```diff
@@ -9490,6 +9559,26 @@ def patched_gemma4_clippable_linear_forward(self, hidden_states: torch.Tensor) -
     return hidden_states
 
 
+# transformers v5.6 replaced the one-hot @ table formulation of the vision patch position
```
echarlaix: should be
```suggestion
# transformers v5.10.1 replaced the one-hot @ table formulation of the vision patch position
```
https://github.com/huggingface/transformers/blob/v5.10.1/src/transformers/models/gemma4/modeling_gemma4.py#L602

also would be easier to add a link adding the original code it was copied from (using the last version it was from so here v5.10), same comment for all the code that was copied from transformers

optimum/exporters/openvino/model_patcher.py
```diff
@@ -11828,3 +11927,401 @@ def __exit__(self, exc_type, exc_value, traceback):
         if hasattr(self._model, "config") and hasattr(self._model.config, "_orig_ov_attn_impl"):
             self._model.config._attn_implementation = self._model.config._orig_ov_attn_impl
             del self._model.config._orig_ov_attn_impl
+
+
+# Starting from transformers 5.6, `window_partition` / `window_reverse` swap the
+# `permute(0, 1, 3, 2, 4, 5)` call for an equivalent `transpose(2, 3)`. Both are semantically
```
echarlaix, line 11946: thanks for the comments / explanation

optimum/exporters/openvino/model_patcher.py
```diff
@@ -11828,3 +11927,401 @@ def __exit__(self, exc_type, exc_value, traceback):
         if hasattr(self._model, "config") and hasattr(self._model.config, "_orig_ov_attn_impl"):
             self._model.config._attn_implementation = self._model.config._orig_ov_attn_impl
             del self._model.config._orig_ov_attn_impl
+
+
+# Starting from transformers 5.6, `window_partition` / `window_reverse` swap the
```
echarlaix: can you take a look to ensure all given versions are correct ?

the change appeared at v5.9 here is seems  https://github.com/huggingface/transformers/blob/v5.8.1/src/transformers/models/swin/modeling_swin.py

```suggestion
# Starting from transformers 5.9, `window_partition` / `window_reverse` swap the
```

echarlaix (changes requested):
Thanks for your work @anatyrova ! I understand that having invariant IRs across different optimum-intel versions is a requirement for the OV team however I don’t think this makes sense in between two transformers versions as there will always be changes in some models modeling code. To ensure this, we shouldn’t have to patch every part of the modeling code that will change in the future, or we will end up with a copied modeling code from v5.5 for every model in optimum-intel directly (which also will very likely become incompatible with future transformers version as breaking changes are frequently introduced). The only option that makes sense to me if we want to ensure invariance is to fix the transformers version allowed to export our models (could vary depending on the given architecture).

rkazants:
Hi @echarlaix, I see your point that we should anyway expect changes in transformers code that will lead to changes in IR.
However, let us consider trade-off between accepting changes in transformers version and applying patches to preserve IR. If it is easy to patch and preserve some part of IR, let us do it, otherwise, okay let us accept changes,

@echarlaix, the point of Altana's PR is that you will cherry-pick of some patches to preserve invariance of IRs into your PR. Is it possible to accept changes that makes sense for you. Let us see how it goes from one transformers version to another.

Best regards,
Roman

PR #1981, opened 2026-09-08
[OpenVINO] Add support for DeepSpec DFlash export

ofirzaf:
# What does this PR do?

DFlash models produced with [DeepSpec](https://github.com/deepseek-ai/DeepSpec) use a different arch in their config and work slightly different from the z-lab DFlash models. This PR adds support for the export of DFlash models based on DeepSpec.

Commands to export DFlash model with this PR:
```
optimum-cli export openvino \
  --model deepseek-ai/dflash_qwen3_14b_block7 \
  --task text-generation-with-past \
  --weight-format int4 \
  --all-layers \
  --trust-remote-code \
  dflash_qwen3_14b-int4-ov
```

## Before submitting


optimum/exporters/openvino/convert.py
```diff
@@ -151,7 +151,11 @@ def _save_model(
     if getattr(config, "eagle3", False):
         model = _add_eagle3_mode_to_rt_info(model)
     if getattr(config, "dflash", False):
-        model = _add_dflash_mode_to_rt_info(model, config._config)
+        model = _add_dflash_mode_to_rt_info(
+            model,
+            config._config,
+            candidate_position_offset=config.candidate_position_offset,
```
rkazants, line 157: I think candidate_position_offset may be absent. May be better to use `getattr(config, "candidate_position_offset", None)`?

optimum/exporters/openvino/model_configs.py
```diff
@@ -470,14 +480,26 @@ def __init__(
...
         archs = getattr(config, "architectures", None)
-        self.dflash = isinstance(archs, list) and len(archs) > 0 and archs[0] == "DFlashDraftModel"
+        self.dflash = False
+        if isinstance(archs, list) and len(archs) > 0:
+            if "dflash" in archs[0].lower():
+                self.dflash = True
+                self.candidate_position_offset = 1
+            elif archs[0] == "Qwen3DSparkModel":
+                if getattr(config, "markov_rank", 0) != 0:
+                    raise ValueError(
```
rkazants, line 490: just for curiousity, does it make sense to support Markov rank > 0 in future. Or Dflash 2 is better?

optimum/exporters/openvino/model_patcher.py
```diff
@@ -9001,6 +9001,49 @@ def forward(
         )
 
 
+class Qwen3DSparkDraftModel(Qwen3DFlashDraftModel):
```
rkazants, line 9004: can we avoid this class and use `Qwen3DFlashDraftModel` directly?

rkazants (commented):
please also add test into test_genai speculative decoding test suite.

optimum/exporters/openvino/model_configs.py
ofirzaf: Not sure yet. It depends on customer demand and the decisions of labs releasing models. From what I understand there is no clear winner yet

optimum/exporters/openvino/model_patcher.py
ofirzaf: We avoid this class now. If we implement DSpark in the future we might need it. I thought it is more clean to have it and then in `__main__.py`

we will need to change it to
```python
"Qwen3DSparkModel": ("Qwen3DFlashDraftModel", "Qwen3DSparkForCausalLM")
```
which might seem odd using a pair that is not matching.

What do you think?

optimum/exporters/openvino/convert.py
ofirzaf: Right now we set it explicitly when we set `dflash = True` but will add.

ofirzaf:
@rkazants Added the test and need your decision on whether to keep `Qwen3DSparkDraftModel` or not

optimum/exporters/openvino/model_patcher.py
rkazants: Let us move step-by-step by demand. When we need it, we will implement it.

ofirzaf:
Hi @rkazants, please review the changes

rkazants (approved):

IlyasMoutawwakil (approved):
LGTM !

PR #1993, opened 2026-09-11
Improve OpenVINO export errors for unwritable output filesystems

Gunjan00001:
## Problem

Exporting large models (e.g. `optimum-cli export openvino --model ... --weight-format int4 ...`) can fail with the opaque native error:

```
RuntimeError: basic_ios::clear: iostream error
```

This comes from OpenVINO's C++ serialization and only means the model files could not be written to disk (disk full, read-only mount, FAT32 per-file size limit, ...). It reports neither the path nor the cause.

On top of that, exports that apply explicit quantization first write the model to a `TemporaryDirectory()`, which defaults to `$TMPDIR`/`/tmp` (often a size-limited tmpfs). Failures can therefore happen on a filesystem unrelated to the requested output directory, which is why "there is plenty of free space in the output location" does not help.

Related to huggingface/optimum#2487.

## Changes

- Add `optimum/exporters/openvino/disk_utils.py` (stdlib only) with:
  - `check_output_path()`: creates the directory, probes writability, and checks free space, read-only mounts and the FAT32 4 GiB per-file limit.
  - `format_save_error()`: builds an actionable message (path, free space, filesystem type/mount, hints).
  - `get_export_tmpdir()`: default temporary directory selection for the CLI.
- `optimum/exporters/openvino/convert.py`: wrap `save_model(...)` in `_save_model` so failures are re-raised as a `RuntimeError` with the actionable message, chaining the original exception.
- `optimum/commands/export/openvino.py`: run the preflight before exporting, and create the quantization temporary directory next to the output by default (same filesystem, no cross-device move). It can be overridden with the `OPTIMUM_OPENVINO_TMPDIR` environment variable.
- `docs/source/openvino/export.mdx`: troubleshooting section for this error.
- Tests: `tests/openvino/test_disk_utils.py` (14 unit tests) and `tests/openvino/test_export_errors.py` (integration test asserting the wrapped error).

## Testing

```
pytest tests/openvino/test_export_errors.py tests/openvino/test_disk_utils.py -q
# 15 passed
```

Also verified end to end with a tiny model:

```
optimum-cli export openvino --model optimum-intel-internal-testing/tiny-random-LlamaForCausalLM \
  --task text-generation-with-past --weight-format int4 --group-size 16 <out>
```

which exports `openvino_model.xml/.bin` + tokenizer, exercising the preflight, the temp directory next to the output and the merge step.

### DIFF
--- a/optimum/exporters/openvino/disk_utils.py
+++ b/optimum/exporters/openvino/disk_utils.py
@@ -1,0 +1,202 @@
+#  Copyright 2026 The HuggingFace Team. All rights reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Filesystem helpers used to give actionable errors when exporting models to OpenVINO IR.
+
+These helpers only rely on the standard library so they can be used (and tested) without the full OpenVINO stack.
+The OpenVINO serialization is performed by a native ``std::ofstream`` and, when the write fails, only surfaces an
+opaque ``basic_ios::clear: iostream error``. The utilities below translate that into a message describing the most
+likely causes (disk full, read-only mount, FAT32 per-file size limit, ...).
+"""
+
+import os
+import shutil
+from pathlib import Path
+from typing import Optional
+
+
+# A single file cannot exceed 4 GiB - 1 byte on a FAT32 volume.
+FAT32_MAX_FILE_SIZE = 4 * 1024**3 - 1
+
+# Environment variable letting users override where the CLI stores the temporary export/quantization directory.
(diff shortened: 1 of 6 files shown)


Gunjan00001:
Hi, thanks for maintaining Optimum Intel!

This PR addresses the `RuntimeError: basic_ios::clear: iostream error` reported in huggingface/optimum#2487, where exporting large models with explicit quantization (e.g. `--weight-format int4`) fails while writing the model files. The error originates from OpenVINO's native serialization and hides the real cause (disk full, read-only mount, FAT32 per-file limit, ...).

Summary of the change:
- wrap `save_model` in `_save_model` to re-raise an actionable error (path, free space, filesystem type, hints) chained to the original exception;
- add a `check_output_path` preflight (writability probe, free space, read-only and FAT32 detection) before the long export;
- create the quantization temporary directory next to the output by default (same filesystem, avoiding a size-limited `/tmp` and a cross-device move), overridable with `OPTIMUM_OPENVINO_TMPDIR`;
- docs + tests.

All 15 tests pass and I verified a real `optimum-cli export openvino ... --weight-format int4` run end to end.

Would a maintainer be able to take a look when convenient? Thanks!

/cc @echarlaix @IlyasMoutawwakil

optimum/exporters/openvino/disk_utils.py
```diff
@@ -0,0 +1,202 @@
+#  Copyright 2026 The HuggingFace Team. All rights reserved.
```
rkazants, line 1: complicated logic here. seems generated by AI. Needs careful review that takes time.

PR #1994, opened 2026-09-12
Make modeling_visual_language importable on newer transformers (VLLM 0.29 support)

3QL:
This is a PR connected to https://github.com/vllm-project/vllm-openvino/pull/31  (vllm-openvino 0.29 support)

## Problem

`optimum.intel.openvino.modeling_visual_language` fails to import on newer
transformers versions (verified with transformers 5.17.0, required by
vLLM 0.29):

1. `VisionRotaryEmbedding` no longer exists in
   `transformers.models.qwen2_vl.modeling_qwen2_vl` (renamed to
   `Qwen2VLVisionRotaryEmbedding` with a different interface), so the
   unguarded import at module top raises `ImportError`.
2. `Qwen3VLVisionModel.rot_pos_emb` and `Qwen3_5VisionModel.rot_pos_emb`
   no longer exist, so the monkey-patches onto `_OVQwen3VLForCausalLM`
   and `_OVQwen3_5ForCausalLM` raise `AttributeError`.

Because `optimum.intel.openvino.__init__` eagerly imports `quantization`,
which imports this module, **every** consumer of `optimum.intel` is broken
— including text-only model export (e.g. `OVModelForCausalLM`), which never
touches any VLM class.

## Fix

- Import `VisionRotaryEmbedding` defensively; on versions where it is gone,
  bind a placeholder that raises an informative `ImportError` only if a
  Qwen2-VL model is actually instantiated.
- Copy `rot_pos_emb` with `getattr(..., None)` (same convention as the
  existing `get_vision_position_ids` line right below it) so the module
  stays importable.

No behavior change on transformers versions where the names still exist.
.

### DIFF
--- a/optimum/intel/openvino/modeling_visual_language.py
+++ b/optimum/intel/openvino/modeling_visual_language.py
@@ -39,7 +39,20 @@ from transformers import (
 )
 from transformers.modeling_outputs import BaseModelOutputWithPooling
 from transformers.models.qwen2_5_vl.modeling_qwen2_5_vl import Qwen2_5_VLModel
-from transformers.models.qwen2_vl.modeling_qwen2_vl import Qwen2VLModel, VisionRotaryEmbedding
+from transformers.models.qwen2_vl.modeling_qwen2_vl import Qwen2VLModel
+
+try:
+    from transformers.models.qwen2_vl.modeling_qwen2_vl import VisionRotaryEmbedding
+except ImportError:
+    # Newer transformers renamed the Qwen2-VL vision RoPE class and changed
+    # its interface. Qwen2-VL vision support on such versions needs a dedicated
+    # port, so fail loudly at first use instead of at import time.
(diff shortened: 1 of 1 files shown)


optimum/intel/openvino/modeling_visual_language.py
```diff
@@ -39,7 +39,20 @@
 )
 from transformers.modeling_outputs import BaseModelOutputWithPooling
 from transformers.models.qwen2_5_vl.modeling_qwen2_5_vl import Qwen2_5_VLModel
-from transformers.models.qwen2_vl.modeling_qwen2_vl import Qwen2VLModel, VisionRotaryEmbedding
+from transformers.models.qwen2_vl.modeling_qwen2_vl import Qwen2VLModel
+
+try:
```
rkazants, line 44: Please check it in a diffrent way. Check transformers version then based on it import required module and use object if it does not exist in transformers.

PR #2001, opened 2026-09-16
Exclude models that were not correctly saved in the cache

echarlaix:


rkazants (commented):
@echarlaix, am I correct that intent is not to exclude them from tests? So we will not loose test coverage.

